PR #9, opened 2024-02-08
Houdini: Cleanup after publishing

MustafaJafar:
## Changelog Description
This PR does 3 things: 
1. exposes staging directory per creator in Houdini settings
2. organises `filepath`s in all creator plugins
3. adds a collector to add the parent of the exported files to the clean up list.

## Additional info
Cautions, This feature can be dangerous. 

## Testing notes:
1. Publish different product types from Houdini
2. All exported files should be gone after the integration finishes.


MustafaJafar:
I'm converting this to draft as I didn't test it at all. So, I'm not sure if it works properly. 
I'm going to make it active again once I test it.

MustafaJafar:
I think this PR works with local publishing, I'm not sure yet about farm publishing. 
Let me add my thoughts as I 'm not sure if this PR addresses the issue correctly. 
--- 
From digging around discord and the code I found out that there are plugins that do the cleaning up: 
- `ayon-core/plugins/cleanup.py` which remove the staging dir if it were in temp path 
- `ayon_core/plugins/publish/cleanup_explicit.py` which removes in path in this list `context.data["cleanupFullPaths"]` 
- `ayon_core/plugins/publish/cleanup_farm.py` which I'm not sure how to use it.
---
tbh, I find this feature dangerous and requires pipeline admins to be aware of how to use it. 
I think these settings are related, *I don't know if they are working with Houdini *:sweat_smile: : 
- `ayon+settings://core/tools/publish/custom_staging_dir_profiles` 
- `ayon+settings://core/publish/CleanUp` 
- `ayon+settings://core/publish/CleanUpFarm`

MustafaJafar:
Mentioning @BigRoy as I couldn't add him as a reviewer in this PR.

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
```diff
@@ -26,9 +27,14 @@ def create(self, subset_name, instance_data, pre_create_data):
             pre_create_data)  # type: CreatedInstance
 
         instance_node = hou.node(instance.get("instance_node"))
+
+        filepath = "{root}/{subset}/{subset}.abc".format(
+            root=hou.text.expandString(self.staging_dir),
+            subset=subset_name
+        )
```
BigRoy, line 34: This is really much better as channel references to the `subset` parm on the node itself, so that if the user would toggle the target asset/subset etc. down the line that.

```suggestion
        filepath = '$HIP/`chs("subset")`/`chs("subset")`.abc'.format(
            root=hou.text.expandString(self.staging_dir),
            subset=subset_name
        )
```

Also, I really wonder why we've decided to not keep these paths relative to `$HIP` since that's a very normal/regular Houdini workflow but instead making these by expanded strings.

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
```diff
@@ -13,6 +13,7 @@ class CreateAlembicCamera(plugin.HoudiniCreator):
     label = "Camera (Abc)"
     family = "camera"
     icon = "camera"
+    staging_dir = "$HIP/ayon"
```
BigRoy, line 16: I feel like that if we're making this customizable - we're just as well of making the FULL path for the filepaths customizable, with e.g. default template in settings being:

```
$HIP/`chs("subset")`/`chs("subset")`.{ext}
```

Maybe only excluding the extension or putting in `{ext}` as variable.

client/ayon_core/hosts/houdini/plugins/publish/collect_staging_dirs_for_cleaning_up.py
```diff
@@ -0,0 +1,90 @@
...
+    This collector collects staging directories
+    and adds them to file remove list.
+
+    CAUTION:
+        This collector deletes the parent folder of the exported files.
+        It works fine with the default filepaths in the creators.
+        Artist should be aware with that fact so they take care when 
+          changing the file path in the ROP node.
+        Developers should be aware when changing the filepath pattern
+          in creator plugins.
```
BigRoy, line 19: With this huge of a caution this plug-in should NOT be enabled by default and/or it should be very specific about what files it's allowed to clean (e.g. only the expected files of an instance). If for whatever reason other files exist in the folder they should remain on disk.

This is a **huge** caution - and inheritly unsafe I'd say. We're much better of marking the folder to "delete folder only if empty" and then have something else, e.g. the extractors mark the files themselves as temporary.

Anyway, there's been quite a debate on the maya front as well with renders being deleted on publish - then the render had an issue on one frame on the farm and now suddenly you don't have all the renders there anymore. To then just fix that one frame you need to either fully re-render the shot, or manually on your server "revert the files" from backup or whatever. I'd argue that also there, the default should be "OFF" with a clear warning label **in settings** about enabling it.

client/ayon_core/hosts/houdini/plugins/publish/collect_staging_dirs_for_cleaning_up.py
```diff
@@ -0,0 +1,90 @@
...
+                filepath = node.evalParm("ar_ass_file")
+        
+        elif node_type == "vray_renderer":
+            filepath = node.evalParm("SettingsOutput_img_file_path")
+        
+        else:
+            self.log.debug(
+                "ROP node type '{}' is not supported for cleaning up."
+                .format(node_type)
+            )
```
BigRoy, line 83: Reaching this else statement the `filepath` variable is undefined below since `filepath` is only set in the `if` and `elif` branches, and not above it, so this `else` statement should really be setting `filepath = None`

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
MustafaJafar: > Also, I really wonder why we've decided to not keep these paths relative to `$HIP`

Do paths relative to `$HIP` raise a conflict on the farm ?

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
MustafaJafar: > I feel like that if we're making this customizable - we're just as well of making the FULL path for the filepaths customizable

This was my feeling too.
But, I have no idea how to list the available keys because it can be confusing that it doesn't support the same keys as https://ayon.ynput.io/docs/addon_houdini_admin#update-houdini-vars-on-context-change

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
BigRoy: Correct - that would be the case - if that file would be moved over, yes. However, I'd actually resolve that in the area where you're moving things over instead of forcing it in the workfile. So preferably the workfile publish might "bake" it - but leaving the local scene as is.

client/ayon_core/hosts/houdini/plugins/publish/collect_staging_dirs_for_cleaning_up.py
MustafaJafar: > "delete folder only if empty"
> the extractors mark the files themselves as temporary.

I believe this is safer as it doesn't force putting the staged files in separate directories.. 
How can I do that without replicating some of the logic in the integrator ?

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,112 @@
...
+        else:
+            files = [filepath]
+
+        self.log.debug("Add directories to 'cleanupEmptyDir': {}".format(staging_dir))
+        instance.context.data["cleanupEmptyDirs"].append(staging_dir)
+        
+        self.log.debug("Add files to 'cleanupFullPaths': {}".format(files))
+        instance.context.data["cleanupFullPaths"] += files
+
+    def get_filepath(self, node):
```
MustafaJafar, line 60: I think this method works for now but I think it doesn't handle all the cases.
e.g. 
- These Temp folders on Mantra
image

- Render with `split_render` enabled

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
MustafaJafar: >  we're just as well of making the FULL path for the filepaths customizable, with e.g. default template in settings being:

I made that in this commit https://github.com/ynput/ayon-core/pull/9/commits/28ace5313c3826f619fa2f65bae6c2d9ab28a501 but I'm not sure if I made it in a good way. 
please let me know what do you think.

client/ayon_core/hosts/houdini/plugins/create/create_alembic_camera.py
```diff
@@ -26,9 +27,13 @@ def create(self, subset_name, instance_data, pre_create_data):
             pre_create_data)  # type: CreatedInstance
 
         instance_node = hou.node(instance.get("instance_node"))
+
+        filepath = self.staging_dir.format(
+            product_name="`chs(\"subset\")`"  # keep dynamic link to subset
+        )
```
BigRoy, line 33: I would actually do this differently. First off, I'd use similar var naming as anatomy templates, I believe there it's `{product[name]}`.

If that is passed, then we embed the hardcoded value - a studio might want that! (not sure why, but they might)

If they want to instead have the dynamic path, they can use:
```
$HIP/ayon/`chs("subset")`/`chs("subset")`.abc
```

By doing it that way - the studio can decide to use either the dynamic or hardcoded/expanded one?

Also, I can't think of any reason why anyone wanted to adjust the extension - so I'm not entirely sure whether we should expose that in the settings - but hey... whatever rows your both I suppose.

MustafaJafar: > why anyone wanted to adjust the extension 

Some creators/rops allow artists to choose a `ext` therefore I left it with `{ext}`.
---
I'll use `product[name]` in code and an admin can make it dynamic by using `chs("subset")`

MustafaJafar:
Here are the current test results on my side. 
I tested these product types and it deletes files as expected 
```
        "camera",
        "pointcache",
        "imagesequence",
        "mantraifd",
        "review",
        "staticMesh",
        "usd",
        "vdbcache"
```

> FYI, I found out a bug in mantra ifd, I'll fix it in a separate PR.

```
Removed files:
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\S_BigRoboMain\S_BigRoboMain.fbx
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\cameraMain\cameraMain.abc
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\imagesequenceMain\imagesequenceMain.1001.exr
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\imagesequenceMain\imagesequenceMain.1002.exr
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\mantraifdMain\mantraifdMain.1001.ifd
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\mantraifdMain\mantraifdMain.1002.ifd
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\pointcacheABC\pointcacheABC.abc
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\pointcacheBgeo\pointcacheBgeo.1001.bgeo
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\pointcacheBgeo\pointcacheBgeo.1002.bgeo
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\reviewMain\reviewMain.1001.png
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\reviewMain\reviewMain.1002.png
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\reviewMain\reviewMain_h264focal_length_burnin.mp4
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\usdMain\usdMain.usd
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\vdbcacheMain\vdbcacheMain.1001.vdb
\\storage\work\ayon_projects\Robo\Assets\Character\BigRobo\work\cfx\ayon\vdbcacheMain\vdbcacheMain.1002.vdb
```

BigRoy:
The Mantra IFD and .rs files being deleted doesn't seem like what you'd want right? Or are those IFD files actually being published? Aren't those just intermediate rendering files?

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,112 @@
...
+        "ass",
+        "pointcache",
+        "imagesequence",
+        "mantraifd",
+        "redshiftproxy",
+        "review",
+        "staticMesh",
+        "usd",
+        "vdbcache"
+    ]
```
BigRoy, line 33: Do we want to expose these in settings so studio can decide per family whether they want to trigger this or not?

MustafaJafar:
> The Mantra IFD and .rs files being deleted doesn't seem like what you'd want right? Or are those IFD files actually being published? Aren't those just intermediate rendering files?

The deleted Mantra IFD files are part of the mantra IFD publishing not Mantra render publishing.
So these files should exist in the publish folder.
image

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
BigRoy, line 60: Any reason we're implementing this in a "Cleanup files" function. Shouldn't this be part of `lib`? Or be using e.g. [this](https://github.com/pypeclub/OpenPype/blob/4add4e7e46c3fa1e502ade367e0f0fef366c1a79/openpype/hosts/houdini/api/lib.py#L46)

Preferably we have a unified approach that's used throughout.

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: That's fair!

moonyuet:
I tested with redshiftROP and review family, doesn't hit any bug.
I agree with @BigRoy we can implement` cleanup function` in `lib.py` and use it in the collector.

MustafaJafar:
Here's a riddle. 
For some render nodes (Mantra, Arnold, Redshift) .. let's take mantra for example.. 

we allow splitting the render: create IFD files, create render and publish only the render files. 
Also, we support publishing the intermediate IFD files as separated product (without rendering). 
So, using this parameter `soho_diskfile` is not enough to differentiate between these two instances [`mantraifdMain` , `mantra_ropSplit`] in my screenshot.. 

My question is: what do you think about using the `product type` / `family` to differentiate between them ?

image

MustafaJafar:
Any ideas about how to implement function that fetches the intermediate render files.. 
Is it okay to have a hard coded function? which adds some maintenance debt.
image

MustafaJafar:
I gave this PR another go. 
I think there are still missing pieces of the puzzle that need to be figured out before marking this PR Open. 
Anyways, here are the updates and let me know what do you think @BigRoy , @moonyuet , @antirotor 

1. A toggle to delete intermediate render files has been added.
2. A library function `lib.get_output_parameter(node)` is used to retrieve the `filepath` parameter instead of implementing another function.
3. I've updated the logic to get the files to clean up.. So far, I believe we have four categories:
    a. Non Render Product with frames (e.g. Bgeo, Review, IFD)
    b. Render Products (Mantra, Karma, Redshift,..)
    c. Intermediate render files `ifdFile` (we didn't figure out a better generic name it)
    d. Non Render Product with no frames (e.g. alembic)

**Few Notes:** 
1. Intermediate Render file is one file with `$F4` evaluated as `0000` but, in fact it's not one file. they are many files that will be generated with frames.. so, this one needs to be fixed to get the expected `ifdFiles`
2. The current `staging_dir` in the collector is not correct all the time.
    e.g. with `mantra_rop`, it will be the staging directory of the `ifdFile` no the render beauty file. 
3. For some reason I don't know, any rendered products (submitted to farm) are deleted automatically (as well as the json meta data) when the publish job finishes..

client/ayon_core/hosts/houdini/plugins/create/create_mantra_ifd.py
```diff
@@ -25,10 +26,11 @@ def create(self, product_name, instance_data, pre_create_data):
             pre_create_data)  # type: CreatedInstance
 
         instance_node = hou.node(instance.get("instance_node"))
+        
+        filepath = self.staging_dir.format(
+            product={"name": product_name}
+        )
```
BigRoy, line 32: I'd remain consistent across all creators and always format using `{ext}`. That way you can specify the exact same template in settings for all creators if a studio wants to. Otherwise they need to know specifically which needs a hardcoded ext and which doesn't.

client/ayon_core/hosts/houdini/plugins/create/create_review.py
```diff
@@ -31,11 +32,9 @@ def create(self, product_name, instance_data, pre_create_data):
 
         frame_range = hou.playbar.frameRange()
 
-        filepath = "{root}/{product_name}/{product_name}.$F4.{ext}".format(
-            root=hou.text.expandString("$HIP/pyblish"),
-            # keep dynamic link to product name
-            product_name="`chs(\"AYON_productName\")`",
-            ext=pre_create_data.get("image_format") or "png"
+        filepath = self.staging_dir.format(
+            product={"name": "`chs(\"AYON_productName\")`"},
```
BigRoy, line 36: Is there a reason this is the only one (that I could quickly see) that uses channel reference? I suppose this was just missed?

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,85 @@
...
+            return
+
+        files = []
+        # Non Render Products with frames
+        frames = instance.data.get("frames", [])
+        staging_dir, _ = os.path.split(filepath)
+        if isinstance(frames, str):
+            files = ["{}/{}".format(staging_dir, frames)]
+        else:
+            files = ["{}/{}".format(staging_dir, f) for f in frames]
```
BigRoy, line 60: Why format instead of `os.path.join`?

Not a big issue - `os.path.join` might just hint more that we're referring to filepaths here.

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Render Products
+        expectedFiles = instance.data.get("expectedFiles", [])
+        for aov in expectedFiles:
+            for v in aov.values():
+                files += v
```
BigRoy, line 66: Would this work?
```suggestion
        for aov in expectedFiles:
            files.extend(aov.values())
```

client/ayon_core/hosts/houdini/plugins/create/create_review.py
MustafaJafar: yes, it's still the only one. I'll update the rest too.

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: I'll try and let you know.

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # Render Intermediate files.
+        # This doesn't cover all intermediate render products.
+        # E.g. Karma's USD and checkpoint.
+        # For some reason it's one file with $F4 evaluated as 0000
+        # So, we need to get all the frames.
+        ifdFile = instance.data.get("ifdFile")
+        if self.include_intermediate_files and ifdFile:
```
BigRoy, line 74: Intermediate files is quite a generic term whereas ifdFile or "intermediate render files" or whatever is more specific. Should we label this more specific, e.g. 

- include intermediate exported render scenes for cleanup (.idf, .ass, .usd, .rs) for render instances

server_addon/houdini/server/settings/create.py
```diff
@@ -17,6 +18,58 @@ class CreateArnoldAssModel(BaseSettingsModel):
...
+
+class CreateKarmaROPModel(BaseSettingsModel):
+    enabled: bool = SettingsField(title="Enabled")
+    default_variants: list[str] = SettingsField(
+        title="Default Products",
+        default_factory=list,
+    )
+    render_staging_dir: str = SettingsField(title="Render Staging Directory")
+    checkpoint_dir: str = SettingsField(title="Checkpoint Directory")
+    usd_dir: str = SettingsField(title="USD Directory")
```
BigRoy, line 42: Would it be more generalizable if we find a common terminology/name for a Render Scene Description or File so that we don't need unique keys for each renderer but can have just something that describes "exported render scene description" or whatever. Of course the titles can be more specific. Just wonder whether we want to consolidate the logic more for potential re-use?

server_addon/houdini/server/settings/publish.py
```diff
@@ -20,6 +20,39 @@ class CollectChunkSizeModel(BaseSettingsModel):
...
+        {"value": "imagesequence", "label": "Composite (Image Sequence)"},
+        {"value": "review", "review": "Review"},
+        {"value": "ass", "label": "Arnold ASS"},
+        # {"value": "arnold_rop", "label": "Arnold ROP"},
+        {"value": "mantraifd", "label": "Mantra IFD"},
+        # {"value": "mantra_rop", "label": "Mantra ROP"},
+        {"value": "redshiftproxy", "label": "Redshift Proxy"},
+        # {"value": "redshift_rop", "label": "Redshift ROP"},
+        # {"value": "karma_rop", "label": "Karma ROP"},
+        # {"value": "vray_rop", "label": "VRay ROP"},
```
BigRoy, line 39: Are some here still commented on purpsose? If so, should we add a comment as to why?

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,85 @@
...
+        "usd",
+        "vdbcache"
+    ]
+    label = "Collect Files For Cleaning Up"
+
+    def process(self, instance):
+
+        import hou
+
+        node = hou.node(instance.data["instance_node"])
```
BigRoy, line 41: Can we make this ignore instances without an `instance_node`? There may be 'runtime instances' from future USD implementations that are generated from other instances during publishing which themselves do not have Houdini instance node.

(Or I actually think I made the families `usdrop` (geometry) and `usdrender` (rendering) on my end)

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,85 @@
...
+        expectedFiles = instance.data.get("expectedFiles", [])
+        for aov in expectedFiles:
+            for v in aov.values():
+                files += v
+
+        # Render Intermediate files.
+        # This doesn't cover all intermediate render products.
+        # E.g. Karma's USD and checkpoint.
+        # For some reason it's one file with $F4 evaluated as 0000
+        # So, we need to get all the frames.
```
BigRoy, line 72: I feel like this isn't solved  yet and this should be a `# TODO:` comment?

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,85 @@
...
+        ifdFile = instance.data.get("ifdFile")
+        if self.include_intermediate_files and ifdFile:
+            files += [ifdFile]
```
BigRoy, line 75: This is much better as `files.append(idfFile)`. You're not generating a temporary list plus it's clearer what's going on that you're appending to list.

client/ayon_core/hosts/houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: This worked. 
```python 
        for aov in expectedFiles:
            for v in aov.values():
                files.extend(v)
```
or in one line: 
```python
        for aov in expectedFiles:
            files.extend(sum(aov.values(), []))
```

Here's an example `expectedFiles`
```
expectedFiles = [
    {
    
        '_': ['//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.beauty.1001.exr', '//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.beauty.1002.exr'], 
        'Z': ['//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.Z.1001.exr', '//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.Z.1002.exr'], 
        'N': ['//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.N.1001.exr', '//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.N.1002.exr']
    }
]
```

MustafaJafar: I used the one line version with a comment about it.

server_addon/houdini/server/settings/publish.py
MustafaJafar: It was for testing purposes.

server_addon/houdini/server/settings/create.py
MustafaJafar: For these creators we can use the same settings model if we picked up an name for the intermediate exported files.
- `CreateVrayROPModel`
- `CreateRedshiftROPModel`
- `CreateMantraROPModel`
- `CreateArnoldRopModel`

Unlike Karma, it has more options 
- `CreateKarmaROPModel`

---
What could be a good name ? 
- proxy scene
- scene description
- scene definition
- exported file 

I think we can create an issue for it so we can change that name along the plugins to ensure consistency.. what do you think ?

MustafaJafar:
So, the collector tells me that exported RS proxy files are added to `cleanupFullPaths`. 
but, when the export, render and publish job finish the files are not cleaned up. 
tbh, I don't know why it doesn't work on Farm.. 
Any ideas ? 
```
DEBUG: Add directories to 'cleanupEmptyDir': //storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render
DEBUG: Add files to 'cleanupFullPaths': [
'//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.beauty.1001.exr', 
'//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/render/redshift_ropMain.beauty.1002.exr', 
'//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/rs/redshift_ropMain.1001.rs', 
'//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/ayon/redshift_ropMain/rs/redshift_ropMain.1002.rs'
]
```
Also, I have no idea why my render files. render publish json file and  the render directory are deleted.
Any ideas how it works ?

MustafaJafar:
TL;DR 
This PR (mainly) adds cleanup feature after publishing. files to clean up are categorized into 4 categories: 
1. Non Render Products with frames
2. Render Products
3. Intermediate exported render files.
4. Non Render Products with no frames

This is how collector settings look like.
image

---

This PR (secondary) adds custom staging directories in Houdini creator settings. 
Because having organised directories makes it easier to implement this feature.

---

This collector doesn't work on farm but collects the files to cleanup correctly as far as I can tell. 
So, I'd appreciate any help.

antirotor:
We should split it for cleanup of local renders and stuff implemented here https://github.com/ynput/OpenPype/pull/4583/files#

MustafaJafar:
Hey @antirotor 
I have remove the logic related to modifying staging directories.

antirotor (changes requested):
correct me if I am wrong, but this will allow you to set only one family - at least that is how it behaves for me:
image
But what if I want to set multiple product types there? Shouldn't it be list of product types?

Also, I am thinking about the product types enum - since settings are now very static in that regard, this enum must be updated whenever new product type is added (even via studio custom addons). Wouldn't it be better to leave it as a list of strings? I know that it feels TD unfriendly, but versioning up Houdini addon whenever someone somewhere adds new product type is not good. What do you think?

MustafaJafar:
> But what if I want to set multiple product types there? Shouldn't it be list of product types?

we can have multiple selections.
image

>  I know that it feels TD unfriendly, but versioning up Houdini addon whenever someone somewhere adds new product type is not good. What do you think?

That's a total valid point!
My concern behind that enum to make it easier for pipeline admins* who can have some difficult time searching for the right value e.g. `pointcache` or `PointCache`  or `Point Cache`  or  `pointcache(ABC)` or  `pointcache (ABC)` 

_* Because Ayon is amazing. it allows non coders to be pipeline admins. And, they might not be cool with searching the code base for some string values._ 

I think there's a middle ground. what about adding a dedicated table in the admin docs. 
e.g.
| Product Type Label | Product Type Key | Description | 
|--|--|--|
|  `Camera (Abc)` | `camera` |  Single baked camera from Alembic ROP. |
| `PointCache (Abc)` | `pointcache` | Alembic ROP to pointcache |
| `PointCache (Bgeo)` | `pointcache` | BGEO pointcache creator. |

* Product Type Label: The name used in AYON publisher UI
* Product Type Key: The name used in the background, The name used in the Loader tool,  The name you'd need to use in AYON Settings.

MustafaJafar:
Okay, The Addons settings are update in https://github.com/ynput/ayon-core/pull/9/commits/fd819906f9559bc658b0682c9a98d0e022d754f4

And to make it easier for admins and TDs, Here's a PR that adds a list of product types  https://github.com/ynput/ayon-documentation/pull/163

mkolar:
>I know that it feels TD unfriendly, but versioning up Houdini addon whenever someone somewhere adds new product type is not good. What do you think?
Actually I think that's exactly what should happen, how else do you keep track of it. We know that filling in strings is almost a no-go, not even studios that have been using OP and AYON for a while are able to do it correctly. it must be an enum otherwise no-one will use it.

That is tru for most of the settings. We should be looking at keeping the enum up to date, rather than requiring people to jump through the hoops to even know what values are allowed.

MustafaJafar:
@antirotor  should I revert the string list to back to the enum ? 
Also, I forgot that we already have a list of creators in the settings and they should be updated whenever we add a new product type (adding them to settings allows admins to disable them. they are enabled by default.)  

https://github.com/ynput/ayon-core/blob/develop/server_addon/houdini/server/settings/create.py

MustafaJafar:
Settings have been refactored to be an enum list again. 
image

antirotor (approved):

antirotor (dismissed):
actually, is there reason why we are not using `clique` or similar for collecting files?

MustafaJafar:
> actually, is there reason why we are not using clique or similar for collecting files?

I think I've managed to do it in https://github.com/ynput/ayon-core/pull/9/commits/547065d93ac85c3ab9cdd4b5b0db8fbc62d3f5f1 and I ❤️ clique so much. 

Here's a log snippet from some test on my side. 
```
DEBUG: Add directories to 'cleanupEmptyDir': //storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/karma_ropMain
DEBUG: Add files to 'cleanupFullPaths': [
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/karma_ropMain/karma_ropMain.1001.exr', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/karma_ropMain/karma_ropMain.1002.exr', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/karma_ropMain/karma_ropMain.1003.exr'
]
DEBUG: Add directories to 'cleanupEmptyDir': //storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/ifd/mantra_ropMain
DEBUG: Add files to 'cleanupFullPaths': [
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/mantra_ropMain/mantra_ropMain.1001.exr', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/mantra_ropMain/mantra_ropMain.1002.exr', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/renders/mantra_ropMain/mantra_ropMain.1003.exr', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/ifd/mantra_ropMain/mantra_ropMain.1001.ifd', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/ifd/mantra_ropMain/mantra_ropMain.1002.ifd', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/ifd/mantra_ropMain/mantra_ropMain.1003.ifd'
]
DEBUG: Add directories to 'cleanupEmptyDir': //storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish
DEBUG: Add files to 'cleanupFullPaths': [
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheBgeoCache_sho010.1001.bgeo', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheBgeoCache_sho010.1002.bgeo', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheBgeoCache_sho010.1003.bgeo'
]
DEBUG: Add directories to 'cleanupEmptyDir': //storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish
DEBUG: Add files to 'cleanupFullPaths': [
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheAbcCache.abc'
]
DEBUG: Add directories to 'cleanupEmptyDir': //storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish
DEBUG: Add files to 'cleanupFullPaths': [
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheAbcCacheFrames.1001.abc', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheAbcCacheFrames.1002.abc', 
    '//storage/work/ayon_projects/Robo/Assets/Character/moco/work/cfx/pyblish/pointcacheAbcCacheFrames.1003.abc'
]
```

MustafaJafar:
For information: 
I met a bug where files were added twice because they existed in both `frames` and `expectedFiles`.

MustafaJafar:
> Besides that it works as expected, just for clarification of the goals stated in the PR description:
> 
> > exposes staging directory per creator in Houdini settings
> > organises filepaths in all creator plugins
> > adds a collector to add the parent of the exported files to the clean up list.
> 
> How does it achieve the first point? Rest of them are obvious.

My bad, the description should have been updated some time ago. 

For reference, 
at the beginning of working on this PR. I updated the value of filepaths in all creator plugins then dropped that change. 
Also, changing staging directory was reverted, there should be a dedicated PR for it.

moonyuet (approved):
Published successfully in Houdini

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -27,47 +27,35 @@ def process(self, instance):
...
         expected_filepath = output_parm.eval()
         instance.data.setdefault("files", list())
         instance.data.setdefault("expectedFiles", list())
-        if instance.data.get("frames"):
-            files = self.get_files(instance, expected_filepath)
-            # list of files
-            instance.data["files"].extend(files)
-        else:
+
+        frames = instance.data.get("frames", "")
```
BigRoy, line 41: This fallback doesn't seem right? We should either rely on `instance.data["frames"]` or do some explicit skipping like:

```suggestion
        frames = instance.data.get("frames")
        if frames is None:
            self.log.warning(f"No frames data found on instance {instance}. Skipping collection for caching on farm...")
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -27,47 +27,35 @@ def process(self, instance):
...
-        self.log.debug(instance.data["families"])
+            instance.data["files"].append(expected_filepath)
+        else:
+            # list of files
+            staging_dir, _ = os.path.split(expected_filepath)
+            instance.data["files"].extend(
+                ["{}/{}".format(staging_dir, f) for f in frames]
+            )
+
+        cache_files = {"cache": instance.data["files"]}
```
BigRoy, line 52: It feels a bit odd that we need both the `instance.data["files"]` and `instance.data["expectedFiles"]` - they seem to want to do the same thing but for whatever reason we require the duplicated data. Might be good for a follow-up issue to streamline that logic over time?

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -27,47 +27,35 @@ def process(self, instance):
...
-        """
-        directory = os.path.dirname(output_parm)
-
-        files = [
-            os.path.join(directory, frame).replace("\\", "/")
-            for frame in instance.data["frames"]
-        ]
-
-        return files
+        self.log.debug("{}".format(instance.data["expectedFiles"]))
```
BigRoy, line 61: This debug log may benefit from more descriptiveness.
```suggestion
        self.log.debug("Caching on farm expected files: {}".format(instance.data["expectedFiles"]))
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,115 @@
...
+                                AYONPyblishPluginMixin):
+    """Collect Files For Cleaning Up.
+
+    This collector collects output files
+    and adds them to file remove list.
+
+    CAUTION:
+        This collector deletes the exported files and
+          deletes the parent folder if it was empty.
+        Artists are free to change the file path in the ROP node.
```
BigRoy, line 19:
```suggestion
    CAUTION:
        This collector registers exported files and
          the parent folder (if it was empty) for deletion
          in `ExplicitCleanUp` plug-in.
        Artists are free to change the file path in the ROP node.
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,115 @@
...
+
+        expected_files = instance.data.get("expectedFiles", [])
+
+        # 'expectedFiles' are preferred over 'frames'
+        if expected_files:
+            # Products with expected files
+            # This can be Render products or submitted cache to farm.
+            for expected in expected_files:
+                # expected.values() is a list of lists
+                files.extend(sum(expected.values(), []))
```
BigRoy, line 63: I understand what this is doing yet it still looks weird - you seem to want to flatten or concatenate the list into one long list.

But why not just:
```suggestion
                for output_files in expected.values():
                    files.extend(output_files)
```
_Or whatever is a good variable name_. It seems less 'magic' that way?

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,115 @@
...
+        if not filepath:
+            self.log.warning("No filepath value to collect.")
+            return
+
+        files = []
+        staging_dir, _ = os.path.split(filepath)
+
+        expected_files = instance.data.get("expectedFiles", [])
+
+        # 'expectedFiles' are preferred over 'frames'
```
BigRoy, line 57: Can we elaborate on WHY it's preferred? What does `frames` lack? (and why does it lack it?)

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
```diff
@@ -20,82 +19,44 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
...
+
+        # Return as no frame pattern detected.
+        if not frame_collection:
+            return
+
+        # It's always expected to be one collection.
+        frame_collection = frame_collection[0]
+        frame_collection.indexes.clear()
+        frame_collection.indexes.update(list(range(start_frame, (end_frame + 1))))
+        instance.data.update({"frames": list(frame_collection)})
```
BigRoy, line 62:
```suggestion
        instance.data["frames"] = list(frame_collection)
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
```diff
@@ -20,82 +19,44 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
...
+            return
+
+        # Create collection using frame pattern.
+        # e.g. 'pointcacheBgeoCache_AB010.1001.bgeo'
+        # will be <Collection "pointcacheBgeoCache_AB010.%d.bgeo [1001]">
+        frame_collection, _ = clique.assemble(
+            [file_name],
+            patterns=[clique.PATTERNS["frames"]],
+            minimum_items=1
+        )
```
BigRoy, line 52: This was just for my own sanity. But just confirming here that this also works for multi-dot extensions:
```python
import clique

frame_collection, _ = clique.assemble(
    ["test.hello.1001.ass.gz"],
    patterns=[clique.PATTERNS["frames"]],
    minimum_items=1
)
print(frame_collection)
# [<Collection "test.hello.%d.ass.gz [1001]">]

frame_collection, _ = clique.assemble(
    ["test.1001.bgeo.bz2"],
    patterns=[clique.PATTERNS["frames"]],
    minimum_items=1
)
print(frame_collection)
[<Collection "test.%d.bgeo.bz2 [1001]">]


frame_collection, _ = clique.assemble(
    ["test.1001.hello.1001.bgeo.bz2"],
    patterns=[clique.PATTERNS["frames"]],
    minimum_items=1
)
print(frame_collection)
# [<Collection "test.1001.hello.%d.bgeo.bz2 [1001]">]
```

server_addon/houdini/server/settings/publish.py
```diff
@@ -44,6 +44,44 @@ class CollectLocalRenderInstancesModel(BaseSettingsModel):
...
+        {"value": "camera", "label": "Camera (Abc)"},
+        {"value": "pointcache", "label": "PointCache (Abc)/PointCache (Bgeo)"},
+        {"value": "review", "label": "Review"},
+        {"value": "staticMesh", "label": "Static Mesh (FBX)"},
+        {"value": "usd", "label": "USD (experimental)"},
+        {"value": "vdbcache", "label": "VDB Cache"},
+        {"value": "imagesequence", "label": "Composite (Image Sequence)"},
+        {"value": "ass", "label": "Arnold ASS"},
+        {"value": "arnold_rop", "label": "Arnold ROP"},
+        {"value": "mantraifd", "label": "Mantra IFD"},
```
BigRoy, line 58: Mantra IFD publishing should be removed since https://github.com/ynput/ayon-core/pull/681

```suggestion
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
MustafaJafar: thanks for confirming.
As far as I tested this PR, it worked pretty well. 

tbh, I wish if we could use it directly with `$F` or `$F4` tokens. but I couldn't write the proper regex pattern.

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: maybe I should rephrase it. let me try again.
In this plugin, with farm instances, `expectedFiles` exists, therefore we should disregard `frames` to avoid duplication.

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
MustafaJafar: The fall back to empty string is important. 
without it the code will actually look similar to
```
"""
`CollectFrames` doesn't work with all product types in Houdini plugin.
e.g. hda
"""
frames = instance.data.get("frames")
if frames is None: # This means it's either has no files or has a single frame.
    # check file_path
    pass
elif isinstance(frames, str): # single file
    # use file_path directly
    # or combine staging directory and frames
    pass
elif isinstance(frames, list): # list of files
    # combine staging directory and each frame in frames
    pass
```
So, I combined it to 
```
frames = instance.data.get("frames", "")
if isinstance(frames, str): # single file
    # use file_path directly
    pass
else: # list of files
    # combine staging directory and each frame in frames
    pass
```

MustafaJafar: For reference, I used this idea `frames = instance.data.get("frames", "")` 
in 
- collect_files_for_cleaning_up.py
- collect_cache_farm.py

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
BigRoy: I think your explanation here helps a lot - but this should definitely be a "cleanup" issue so that we clean this up over time because as long as there's no reason for both being separate we should towards (once again) streamlining the API to just need a list of files with whatever metadata it may need downstream. Whether that is local, farm, or split renders or whatever.

MustafaJafar: I checked it again. sorry for using this comment to reply to multiple conversations on this PR.

`CollectDataforCache` plugin is simply doing
```
instance.data["expectedFiles"] = {"cache": instance.data["files"]}
```
which follows the same logic as render products i.e. `{"aov": list_of_files}` which in turn allows us to re-use the deadline submitter with small tweaks to make it work with cache.

However, `CollectFilesForCleaningUp` doesn't care whether or not it's render or a cache. doesn't care whether or not it's local or farm. 
It follows a general rule. always check `expectedFiles`  then `frames` then fail to default `filepath`.
`filepath` must exist because, we are using it to compute the staging directory. 
and therefore, such a warning exists

https://github.com/ynput/ayon-core/blob/c2d851a9b578f563b8c4b6eb91923abad5025ace/server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py#L49-L51

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
MustafaJafar: let me know if this commit makes more sense 
https://github.com/ynput/ayon-core/pull/9/commits/37cbcc13527741ad69d1f9c7c2b0145580df5825

MustafaJafar:
Hey, I was giving it more test runs.
I believe this PR doesn't affect farm instances and I don't know how files are cleaned up on farm. check my comment https://github.com/ynput/ayon-core/pull/9#issuecomment-1943772672

Also, any collected locations from farm instances will be cleared before the submission. which is not good ? 

This PR works with local instances and should keep working when implementing render targets with other products than render.

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,112 @@
...
+        if expected_files:
+            # Products with expected files
+            # This can be Render products or submitted cache to farm.
+            for expected in expected_files:
+                # expected.values() is a list of lists
+                for output_files in expected.values():
+                    staging_dir, _ = os.path.split(output_files[0])
+                    if staging_dir not in staging_dirs:
+                        staging_dirs.append(staging_dir)
+                    files.extend(output_files)
```
BigRoy, line 49: I understand why this is needed but I think this makes this go from "oof dangerous" to... "oh boy, that's a lot more potential folders"

If we're trying to detect any generated output file from a rop node - this may potentially be "anywhere" depending on whatever file type we're exporting. This now suddenly goes from "just happening to be the ROP node's output path to a lot of potential folders".

I think what I may be "roughly" ok with is if here we check whether the folder during Collect already exists or not, if not - then set it up for cleanup. (Note however that that may skip folders that may already have been created by another collector or publishing logic before this). Anyway, by only configuring folders that do not exist yet then this may just be 'cleaning up' folders that (likley) got generated due to the publishing logic itself which seems far more reasonable?

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,112 @@
...
+                )
+                return
+            elif isinstance(frames, str):
+                # single file.
+                files.append(frames)
+            else:
+                # list of frame.
+                files.extend(
+                    [f"{staging_dir}/{frame}" for frame in frames]
+                )
```
BigRoy, line 70: We don't ever want `None` in here I suppose? So this `instance.data.get("stagingDir")` maybe needs to be required? 

Also I think you're lacking making the `files` complete for the single file condition?

```suggestion
            frames = instance.data.get("frames")
            if frames is None:
                self.log.warning(
                    f"No frames data found on instance {instance}"
                    ". Skipping collection for caching on farm..."
                )
                return

            staging_dir = instance.data.get("stagingDir")
            staging_dirs.append(staging_dir)

            if isinstance(frames, str):
                # single file.
                files.append(f"{staging_dir}/{frames}")
            else:
                # list of frame.
                files.extend(
                    [f"{staging_dir}/{frame}" for frame in frames]
                )
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,112 @@
...
+
+    hosts = ["houdini"]
+    families = ["*"]
+    label = "Collect Files For Cleaning Up"
+    intermediate_exported_render = False
+
+    def process(self, instance):
+
+        files = []
+        staging_dirs = []
```
BigRoy, line 35: I wonder if we're making our lifes easier no by setting type hints here:
```suggestion
        files: List[str] = []
        staging_dirs: List[str] = []
```
Or even:
```suggestion
        files: List[os.PathLike] = []
        staging_dirs: List[os.PathLike] = []
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
BigRoy:
```suggestion
        # Prefer 'expectedFiles' over 'frames' because it usually contains
        # more output files than just a single file or single sequence of files.
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
```diff
@@ -0,0 +1,112 @@
...
+            for expected in expected_files:
+                # expected.values() is a list of lists
+                for output_files in expected.values():
+                    staging_dir, _ = os.path.split(output_files[0])
+                    if staging_dir not in staging_dirs:
+                        staging_dirs.append(staging_dir)
+                    files.extend(output_files)
+        else:
+            # Products with frames or single file.
+
```
BigRoy, line 52:
```suggestion
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
```diff
@@ -16,86 +15,70 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
...
+        frame_collection.indexes.update(list(range(start_frame, (end_frame + 1))))
+        instance.data["frames"] = list(frame_collection)
 
-        """
 
-        # Get the padding length
-        frame = match.group(1)
-        padding = len(frame)
+class CollectHDAFrames(plugin.HoudiniInstancePlugin):
+    """Collect all frames which would be saved from HDA nodes"""
```
BigRoy, line 67: This seems odd - why do we need this? An HDA doesn't make frames? What are we 'hacking' here?

Is this just so that we can get cleanup for the HDA file? isn't that particular file on disk the whole "workfile"'s actual definition of that HDA that we're trying to publish? By removing that, aren't we basically breaking the ability to go back to that workfile and cointinue on the HDA? (or is that always all encapsulated in the scene).

Also, HDA would be lacking from the settings products enum?

BigRoy:
> This PR doesn't affect farm instances because it works locally only and values collected by it are not propagated to farm. I found out that files on farm are cleaned up by some core plugins, Here is one of them https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/collect_rendered_files.py I think this means that I can wipe out some of the code in the collector. (removing anything related to farm)
> 
> Also, any collected locations from farm instances will be cleared before the submission. which is not good.

Yes, that sounds right!

> This PR works with local instances and should keep working when implementing render targets with other products than render.

👍

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: > I think what I may be "roughly" ok with is if here we check whether the folder during Collect already exists or not, if not - then set it up for cleanup.

Directories added to 'cleanupEmptyDir' will only be cleared if they are empty. 
https://github.com/ynput/ayon-core/blob/10adfbab0e00569e814e4c9867a40f9fd6078910/client/ayon_core/plugins/publish/cleanup_explicit.py#L14-L16

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
MustafaJafar: > Is this just so that we can get cleanup for the HDA file?

yes, and also to have consistent instance data.

---

> isn't that particular file on disk the whole "workfile"'s actual definition of that HDA that we're trying to publish? By removing that, aren't we basically breaking the ability to go back to that workfile and cointinue on the HDA?

Yes, and also, Cleaning up HDA product type will break the scene because Houdini will no longer find the HDA file.

image

I think I should revert changes related to HDA.

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: Lets use `list[str]` because they can include something like 
```
f"{staging_dir}/{frames}"
```

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
MustafaJafar: should be resolved by https://github.com/ynput/ayon-core/pull/9/commits/60ec86206f51cbafdefaa447685e661fc2dc888a

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
MustafaJafar: I added a note about it here https://github.com/ynput/ayon-core/pull/9/commits/d34ccc1495f742a88f783c4b702fb0ec2ca32811

Also, I add a sub folder with some files inside and the plugin was able to detect that didn't delete my directory

from `CollectFilesForCleaningUp`'s log
```
DEBUG: Add directories to 'cleanupEmptyDir': ['//storage/work/ayon_projects/Robo/Assets/Character/roboangry/work/cfx/pyblish/renders/mantra_ropMain']
```
from `ExplicitCleanUp`'s log
```
DEBUG: Skipped directories because they contain files:
\\storage\work\ayon_projects\Robo\Assets\Character\roboangry\work\cfx\pyblish\renders\mantra_ropMain
```

BigRoy: > For reference, Directories added to 'cleanupEmptyDir' will only be cleared if they are empty.

I understand - what I meant is - that if for whatever reason the folder did exist before the publish.. maybe it's part of a studio's template. Then suddenly after publish it is gone because it was empty.

We could e.g. by checking before extraction just whether the folder we are considering to remove is actually not there yet only delete the empty folders we expected to have created. Anyway, maybe it's a non-issue or something for a separate issue.

server_addon/houdini/server/settings/publish.py
```diff
@@ -45,6 +45,39 @@ class CollectLocalRenderInstancesModel(BaseSettingsModel):
...
+        {"value": "imagesequence", "label": "Composite (Image Sequence)"},
+        {"value": "ass", "label": "Arnold ASS"},
+        {"value": "arnold_rop", "label": "Arnold ROP"},
+        {"value": "mantra_rop", "label": "Mantra ROP"},
+        {"value": "redshiftproxy", "label": "Redshift Proxy"},
+        {"value": "redshift_rop", "label": "Redshift ROP"},
+        {"value": "karma_rop", "label": "Karma ROP"},
+        {"value": "vray_rop", "label": "VRay ROP"},
+        {"value": "model", "label": "Model"},
+    ]
```
BigRoy, line 65: @MustafaJafar with your other comment that this 'doesn't work correctly for all options of the render options' is it still valid to include the render rops in this list for CollectFilesForCleanup?

moonyuet:
I encountered this issue during installing addon
```
      File "/backend/ayon_server/addons/definition.py", line 84, in versions
        self.init_addon(version_dir)
      File "/backend/ayon_server/addons/definition.py", line 148, in init_addon
        module = import_module(vname, server_module_path)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/backend/ayon_server/addons/utils.py", line 18, in import_module
        spec.loader.exec_module(module)
      File "<frozen importlib._bootstrap_external>", line 940, in exec_module
      File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
      File "/addons/houdini/0.3.7/server/__init__.py", line 5, in <module>
        from .settings import HoudiniSettings, DEFAULT_VALUES
      File "/addons/houdini/0.3.7/server/settings/__init__.py", line 1, in <module>
        from .main import (
      File "/addons/houdini/0.3.7/server/settings/main.py", line 15, in <module>
        from .publish import (
      File "/addons/houdini/0.3.7/server/settings/publish.py", line 100, in <module>
        class PublishPluginsModel(BaseSettingsModel):
      File "/addons/houdini/0.3.7/server/settings/publish.py", line 111, in PublishPluginsModel
        default_factory=BasicValidateModel,
                        ^^^^^^^^^^^^^^^^^^
    NameError: name 'BasicValidateModel' is not defined
```
We need to change `BasicValidateModel `to `BasicEnabledStatesModel ` after [this](https://github.com/ynput/ayon-core/pull/711/files#diff-8f40276d8c5985f234d410a1f54a071674481d3644882c4108378cf8c5f6ca46)

BigRoy:
Tested out the PR for pointcache and vdb sequences - works for me:

image

And if the folder was empty before publishing (or didn't exist) it also removes it - which is a debug log:

image

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py
BigRoy: It seems my IDE doesn't like it:
image

Let's make it `List[str]` then.

MustafaJafar: meh, another reason to stop using VSCode. Nah I can't cheat on it.

server_addon/houdini/server/settings/publish.py
MustafaJafar: the logic of `CollectFilesForCleaningUp`, it works before `CollectLocalRenderInstances` 
which means instances with `render` doesn't exist yet.
Personally, I think it is easier and it allows to select whether to clean render from mantra or render from karma and etc.

server_addon/houdini/server/settings/publish.py
```diff
@@ -74,6 +107,10 @@ class PublishPluginsModel(BaseSettingsModel):
         default_factory=CollectChunkSizeModel,
         title="Collect Chunk Size"
     )
+    CollectFilesForCleaningUp:CollectFilesForCleaningUpModel = SettingsField(
```
BigRoy, line 110:
```suggestion
    CollectFilesForCleaningUp: CollectFilesForCleaningUpModel = SettingsField(
```

server_addon/houdini/server/settings/publish.py
BigRoy: Not sure I understand. Is there something you wish to change now still? or should it be good as is?

The way I understand, render instances don't exist yet - so it doesn't work? So do the selection entries here still make sense, or?

server_addon/houdini/client/ayon_houdini/plugins/publish/collect_cache_farm.py
BigRoy: since it works I'll leave it be and I guess we'll just track a "cleanup" issue for the future when ayon houdini repo is separated.

server_addon/houdini/server/settings/publish.py
MustafaJafar: I believe it's good as it is.

---

Render instances don't exist yet at that point of execution. 
But, they have nothing to do with the existence of files because extractors works on the original instance. e.g. `mantra_rop`.   

`CollectFilesForCleaningUp` plugins works in a similar fashion to extractors, as it works on the original instance data.

moonyuet (approved):
Tested with Review, Camera and Mantra ROP(both local and farm)
Works smoothly.

server_addon/houdini/server/settings/publish.py
BigRoy: My brain can't cope with the explanation today :D - anyway if it's "good as is" I'll mark this conversation resolved and assume it doesn't break anything for the render products.

moonyuet: > My brain can't cope with the explanation today :D - anyway if it's "good as is" I'll mark this conversation resolved and assume it doesn't break anything for the render products.

It doesn't hit the error although I somehow doubt if it should be really there. (Maybe it should be)

BigRoy (approved):
I pushed a version bump - only thing to double check @MustafaJafar is [Kayla's comment](https://github.com/ynput/ayon-core/pull/9#discussion_r1657002139). If you think that's all resolved / fine as it is feel free to mark resolved and merge the PR.

Did another pointcache + vdb test run just now. Works fine.

moonyuet:
> I pushed a version bump - only thing to double check @MustafaJafar is [Kayla's comment](https://github.com/ynput/ayon-core/pull/9#discussion_r1657002139). If you think that's all resolved / fine as it is feel free to mark resolved and merge the PR.
> 
> Did another pointcache + vdb test run just now. Works fine.

we need to dimiss Ondrej's review to merge it btw.

server_addon/houdini/server/settings/publish.py
MustafaJafar: In contrast, My brain can't explain today :D
I wish if you could explain it furtherly but I can't find any words atm.

PR #13, opened 2024-02-09
Use duration from streams as its more precise

tokejepsen:
## Changelog Description
Port of https://github.com/ynput/OpenPype/pull/6171

Waiting on testing to merge.


jakubjezek001 (approved):
Code seems to be the same as in OpenPype alternative and it was working well based on the testing steps from the original PR

PR #22, opened 2024-02-09
Use AYONAddon base class

iLLiCiTiT:
## Changelog Description
Use `AYONAddon` base class for all possible addons.

## Additional info
Some of the addons still must use `OpenPypeModule` because of settings, but that is minority at this moment. Renamed library loader action to just loader action.

## Testing notes:
1. All addons and hosts should load correctly.


MustafaJafar (approved):
Houdini works as expected. 
It loads fine.

PR #28, opened 2024-02-12
Tray: Connection timeout and update validations

iLLiCiTiT:
## Changelog Description
Tray tool can validate if is required update of AYON bundle and asks to relogin if user credentials are not valid.

## Additional info
The validation happens once per 5 minutes. The credentials check happens withing it. It is still not possible to re-login when inside host and the validation happens only once a 5 minutes, but better than nothing. The update dialog is shown only in production or staging mode.

### Possible enhancements in this PR
Maybe we should show information dialog "Your credentials are not valid anymore. Please close all applications and login again." before the login dialog is showed.

## Testing notes:
Updates detection
1. Start tray in production
2. Change production bundle on server
3. Wait 5 minutes
4. An dialog with restart request should show up

Relogin
1. Start tray
2. Shotdown your server
3. After 5 minutes a login dialog should be shown

ayon-core variant of https://github.com/ynput/OpenPype/pull/6194


LiborBatek (approved):
Works for both cases / occasions... without any issues.

When Ayon Bundle changes during running Ayon client:

Screenshot 2024-02-21 105811

When Ayon server being down:


Screenshot 2024-02-21 110603

PR #32, opened 2024-02-12
Bugfix: fix Houdini publish button bugs

MustafaJafar:
Self publish had few bugs
1. Unnecessary faulty logic in `find_rop_input_dependencies` function which caused karma and HDA publishes to fail.
2. Recursive rop rendering as each extractor was rendering all input nodes! 

Solution:
1. refactor `find_rop_input_dependencies` to `find_rop_inputs_chain` with much simpler logic. 
2. adding `ignore_inputs` flag in `render_rop` function

## Testing notes:
1. use self publish buttons to publish karma and HDA 
2. use different combinations and it should still work


MustafaJafar:
For reference, This PR was made mainly to avoid the nested rendering + fix the problem when using the publish button with karma product type. 

Also, BigRoy mentioned here https://github.com/ynput/OpenPype/pull/5891#issuecomment-1818732845 this PR doesn't solve the problem correctly.

moonyuet:
> For reference, This PR was made mainly to avoid the nested rendering + fix the problem when using the publish button with karma product type.
> 
> Also, BigRoy mentioned here [ynput/OpenPype#5891 (comment)](https://github.com/ynput/OpenPype/pull/5891#issuecomment-1818732845) this PR doesn't "do much" as it doesn't correctly solve the raised problems.. thus I'm converting this to draft and leaving it open for discussion.

I did the simple test run for this PR and it works smoothly. 
@BigRoy @antirotor should we continue to build on the ordering with this PR or should we create issues based upon that to come up with the possible solutions on the chain problem mentioned?

MustafaJafar:
Closing this PR in favor of searching for better solutions.
Also, the best part of this PR (ignore inputs when rendering) was added in a dedicated PR https://github.com/ynput/ayon-core/pull/344

PR #36, opened 2024-02-12
Enhancement: Color management for Houdini workfiles

MustafaJafar:
This PR adds Houdini workfiles color management
- Display: default display, it accepts Colon-separated list of displays, e.g ACES:P3, Read more [Active Displays](https://opencolorio.readthedocs.io/en/latest/guides/authoring/displays_views.html#active-displays)
- View: default view, it accepts Colon-separated list of view names, e.g sRGB:DCDM Read more [Active Views](https://opencolorio.readthedocs.io/en/latest/guides/authoring/displays_views.html#active-views)
- Review Colorspace : Review output transform which exposes OCIO Colorspace parameter in opngl nodes, Read more [Houdini OpenGL](https://www.sidefx.com/docs/houdini/nodes/out/opengl.html#output)

Things to do:
- [x] Set default view and default display: In Houdini this is done via setting `OCIO_ACTIVE_DISPLAYS` `OCIO_ACTIVE_VIEWS`
- [x] make review creator and validator to use `review color space` value specified in the settings
- [x] update color space data while review extraction 
- [x] update color space data while houdini image sequence extraction (P.S. It's  the colorspace name assigned to `scene_linear` role.) 
- [x] fix the conflict mentioned here https://github.com/ynput/OpenPype/pull/5858#discussion_r1387651765

image

## Testing notes:

1. Test the hook functionality, case 1: 
    - test : disable global color management, enable  houdini color management, enable  color management/workfile
    - result : Hook won't work.
2.  Test the hook functionality, case 2: 
    - test : enable global color management, disable houdini color management, enable  color management/workfile
    - result : Hook won't work.
3.  Test the hook functionality, case 3: 
    - test : enable global color management, enable houdini color management, disable color management/workfile
    - result : Hook logs to console that workfile settings are disabled.
4.  Test the hook functionality, case 4: 
    - test : enable global color management, enable houdini color management, enable color management/workfile
    - result : Hook sets the proper environment variables with values specified in the Houdini settings, it logs to console the environment variables and in houdini you are able to see the only color spaces specified in the settings. 
5. Test `Review colospace`, case 1: 
    - test : leave `Review colospace` empty
    - result : default review colorspace will be used when 
        - creating review instances 
        - using `Set Review Color Space` action if `Validate Review Colorspace` failed
6. Test `Review colospace`, case 2: 
    - test : add you preferred `Review colospace` value
    - result : this value will be used when 
        - creating review instances 
        -  `Validate Review Colorspace` will check if `ociocolorspace` parameter has that value
        - using `Set Review Color Space` action if `Validate Review Colorspace` failed


moonyuet (approved):
I have tested all scenarios and all work.
I have one question (but i think it can be some later talk): is the color management working if the houdini color management is enabled and filled with OCIO config filepath? If yes, maybe it is also better to introduce the configuration when only that houdini color management turns on.
image
image

MustafaJafar:
> maybe it is also better to introduce the configuration when only that houdini color management turns on.

I think this already happens... 

I don't know if it covers all the cases. e.g. I don't know what happens if admins turn off ayon ocio management and used env vars. 
but, in general, the hook added in this PR won't work if  `OCIO` env doesn't exist

https://github.com/ynput/ayon-core/blob/601908e2993b4fd7e7d510b69e5ade385b9222b8/client/ayon_core/hosts/houdini/hooks/set_default_display_and_view.py#L19-L25

moonyuet:
> > maybe it is also better to introduce the configuration when only that houdini color management turns on.
> 
> I think this already happens...
> 
> I don't know if it covers all the cases. e.g. I don't know what happens if admins turn off ayon ocio management and used env vars. but, in general, the hook added in this PR won't work if `OCIO` env doesn't exist
> 
> https://github.com/ynput/ayon-core/blob/601908e2993b4fd7e7d510b69e5ade385b9222b8/client/ayon_core/hosts/houdini/hooks/set_default_display_and_view.py#L19-L25


I see.. I recommend this mainly because the testing note says that merely enabling houdini color management means 'hook wont work".

```
Test the hook functionality, case 1:
test : disable global color management, enable houdini color management, enable color management/workfile
result : Hook won't work.
```
But if it doesn't work without turning on global color management, we can just ignore this scenario by now as it seems that it isn't not only  related to the color management issue in Houdini

MustafaJafar:
> But if the hook doesn't work without turning on global color management, we can just ignore this scenario by now as it seems that it isn't not only related to the color management issue in Houdini

I agree, it's not related.. but, I think it worth mentioning that `Enable Color Management` per DCC won't work if the global `Enable Color Management` is disabled. and in this particular case, the hook won't work.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -811,6 +811,48 @@ def get_current_context_template_data_with_folder_attrs():
...
+    the display & view of the current Houdini session will be used.
+
+    Args:
+        opengl_node (hou.Node): ROP node to set its ociocolorspace parm.
+        review_color_space (str): Colorspace value for ociocolorspace parm.
+        log (logging.Logger): Logger to log to.
+    """
+
+    if log is None:
+        log = self.log
```
antirotor, line 829: is there any reason for this? Down the line `self.log` is used nevertheless and `log` in this case will be always `None`

client/ayon_core/hosts/houdini/plugins/publish/extract_composite.py
```diff
@@ -45,8 +46,14 @@ def process(self, instance):
             "frameEnd": instance.data["frameEndHandle"],
         }
 
-        from pprint import pformat
-
-        self.log.info(pformat(representation))
+        if ext == "exr":
```
antirotor, line 49:
```suggestion
        if ext.lower() == "exr":
```
I know we have it elsewhere and that it is total edge case, but ...

antirotor (changes requested):
Just those little things ...

client/ayon_core/hosts/houdini/api/lib.py
BigRoy: Is this function even used anywhere? :) And if not, should it?

client/ayon_core/hosts/houdini/hooks/set_default_display_and_view.py
```diff
@@ -0,0 +1,58 @@
+from ayon_core.lib.applications import PreLaunchHook, LaunchTypes
```
iLLiCiTiT, line 1:
```suggestion
from ayon_applications import PreLaunchHook, LaunchTypes
```

client/ayon_core/hosts/houdini/api/lib.py
MustafaJafar: > Is this function even used anywhere? :) And if not, should it?

I feel so stupid. I forgot to use it. 
added here: https://github.com/ynput/ayon-core/pull/36/commits/e8e1a1a2dffd197e61101ccc910c9d3f69c01a1f

antirotor (approved):

moonyuet (approved):
Tested successfully in Houdini.

PR #42, opened 2024-02-13
Remove multi-context restrictions for publishing image sequences on farm

tokejepsen:
## Changelog Description
Port of https://github.com/ynput/OpenPype/pull/6088


client/ayon_core/plugins/publish/collect_rendered_files.py
```diff
@@ -75,23 +75,13 @@ def _process_path(self, data, anatomy):
...
-        ctx["version"] = ctx.get("version") or data.get("version")
-
-        # basic sanity check to see if we are working in same context
-        # if some other json file has different context, bail out.
-        ctx_err = "inconsistent contexts in json files - %s"
-        assert ctx.get("asset") == data.get("asset"), ctx_err % "asset"
-        assert ctx.get("intent") == data.get("intent"), ctx_err % "intent"
-        assert ctx.get("comment") == data.get("comment"), ctx_err % "comment"
-        assert ctx.get("user") == data.get("user"), ctx_err % "user"
-        assert ctx.get("version") == data.get("version"), ctx_err % "version"
```
BigRoy, line 94: Technically - when is this is valid. If I have two `.json` files that specify a different running `context` to publish in this may be very dangerous because now suddenly that context isn't actually used for ONE of the json files?

I think, yes we should **not** assert that instance context must MATCH the gloibal context - but if we were to load multiple JSON files at once we should still assert the global context of both JSON files is shared between them.

tokejepsen: @BigRoy this is just a port. Could you comment on the original PR; https://github.com/ynput/OpenPype/pull/6088 ?

antirotor:
Implemented by #138

PR #44, opened 2024-02-13
Add Loader OTIO export action - AY-3020

tokejepsen:
## Changelog Description
This adds a Library Loader action to export the representations to an `.otio` editorial file. User is presented with a UI to choose which representations to export:

image

### Features:
- automatically setting otio file path to homedir with project name
- URI file path format on demand
- pick representation to be included in OTIO file
- using method for getting metadata to be used on media resources 

## Additional informations
- added transcoding module function for getting metadata either from ffprobe or oiiotools for getting `timecode` value
- added unittest for the new transcoding method

## Testing notes:
1. Open Library Loader from tray and select some products.
2. Right-click and start `Export OTIO` action.
3. Choose which representations to export.
4. Choose a path to export the editorial file to.
5. Optionally inspect the editorial file with OTIO view.


client/ayon_core/pipeline/editorial.py
```diff
@@ -281,3 +281,32 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
+
+def export_otio(clips_data, output_path):
+    clips = []
+    for name, data in clips_data.items():
+        clips.append(
+            create_clip(data["path"], name, data["frames"], data["framerate"])
+        )
+
+    timeline = otio.schema.timeline_from_clips(clips)
+    otio.adapters.write_to_file(timeline, output_path)
```
BigRoy, line 312: Isn't the entry point here much more sensible to be actual `otio.schema.Clip` entries?

This would make the code that generates the data much more readable.
I'm also a bit unsure why the clips data needs to be a dict by name. It's quite limiting what `export_clips` would support since it cannot ever export the same clip name then.

I think this is totally fine code in the Export OTIO right?

```python
clips = []
for representation in representations:
    version = self._version_by_representation_id[representation["_id"]]
    name = self._version_path_by_id[version["_id"]]

    clip = create_clip(
        path=get_representation_path_with_anatomy(
            representation, anatomy
        ),
        name=
        frames=(
            version["data"]["frameEnd"] -
            version["data"]["frameStart"]
        ),
        framerate=version["data"]["fps"]
    )
    clips.append(clip)

export_clips(clips, output_path)
```

Also, be aware that by adding this global `export_otio` loader that you might be getting import errors on hosts that e.g. do not have OTIO available since it imports `from ayon_core.pipeline.editorial import export_otio`.

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,187 @@
...
+                checkbox.setChecked(False)
+                if name in representations_by_name.keys():
+                    self._representation_checkboxes[name].append(
+                        {
+                            "checkbox": checkbox,
+                            "representation": representations_by_name[name]
+                        }
+                    )
+                else:
+                    checkbox.setEnabled(False)
```
BigRoy, line 120: If you only ever want one of the Checkboxes to be enabled - then I think what you're looking for is making this checkbox group in the Qt UI so Qt itself handles it for you?

tokejepsen: There can be multiple checkboxes enabled. Depends on the representations. Is that what you mean?

client/ayon_core/pipeline/editorial.py
tokejepsen: Yeah, maybe I'll push the lib code to the action. Think eventually we'll need to push some code out of it, but that can come in later PRs.

client/ayon_core/plugins/load/export_otio.py
BigRoy: Ah sorry, ignore me then. :) I thought the idea was that for each version the representations were 'exclusive' checkboxes and only one could be selected.

I'm not sure what to expect from an otio where I pick multiple representations though haha ;) and are you sure that works? Since you're currently storing them by "name" in the clips data `dict` where name is the version path - which would not be unique per its representation, and thus in the clips data only would end up being present.

tokejepsen: Yeah, good point about multiple representations. Not sure about the use case either, so i'll limit it to 1 representation per version.

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,187 @@
...
+
+            for representation in representations:
+                all_representation_names.add(representation["name"])
+                id = representation["_id"]
+                self._version_by_representation_id[id] = version
+
+            self._version_path_by_id[version_id] = "{}/{}/{}/v{:03d}".format(
+                representations[0]["context"]["hierarchy"],
+                representations[0]["context"]["asset"],
+                representations[0]["context"]["subset"],
```
iLLiCiTiT, line 82: You should not expect that representation context has `"asset"`, `"subset"` and `"hierarchy"`.

tokejepsen: Ok, any suggestion how to get the hierarchy path of a version?

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,187 @@
...
+        representations_by_version_id = defaultdict(list)
+        self._version_by_representation_id = {}
+        all_representation_names = set()
+        self._version_path_by_id = {}
+        for context in contexts:
+            version_id = context["version"]["_id"]
+            version = get_version_by_id(self._project_name, version_id)
+            representations = list(get_representations(
+                self._project_name, version_ids=[version_id]
+            ))
```
iLLiCiTiT, line 71: I would avoid querying all the data one by one.

Something like this:
```suggestion
        version_docs_by_id = {
            context["version"]["_id"]: context["version"]
            for context in contexts
        }
        repre_docs = list(get_representations(
            self._project_name, version_ids=set(version_docs_by_id)
        ))
        repre_docs_by_id = {
            repre_doc["_id"]: repre_doc
            for repre_doc in repre_docs
        }
        self._version_by_representation_id = {
            repre_doc["_id"]: version_docs_by_id [repre_doc["parent"]]
            for repre_doc in repre_docs
        }
        self._version_path_by_id = {}
        for context in contexts:
            version_doc = context["version"]
            version_id = version_doc["_id"]
            if version_id in self._version_path_by_id:
                continue
            asset_doc = context["asset"]
            folder_path = get_asset_name_identifier(asset_doc)
            subset_name = context["subset"]["name"]
            self._version_path_by_id[version_id] = "{}/{}/{}/v{:03d}".format(
                folder_path, subset_name, version_doc["name"]
```

tokejepsen:
I've updated the UI with using radio buttons, so the representations chosen per version is exclusive. Also changed the toggle all checkboxes to buttons. Description has updated UI.

jakubjezek001:
Task linked: [OP-8218 Loader OTIO export action](https://app.clickup.com/t/6658547/OP-8218)

jakubjezek001:
Also here is a stupid mistake what could be done by anyone. I had add folder path without a file and extension and this was only printed in terminal and no pop-up warning was shown. 
```
opentimelineio.exceptions.NoKnownAdapterForExtensionError: No adapter for suffix '' on file 'xxxxxx'
```

BigRoy:
> Also here is a stupid mistake what could be done by anyone. I had add folder path without a file and extension and this was only printed in terminal and no pop-up warning was shown.
> 
> ```
> opentimelineio.exceptions.NoKnownAdapterForExtensionError: No adapter for suffix '' on file 'xxxxxx'
> ```

Since I suppose this only works for `.otio` anyway we might just as well force the `.otio` suffix if the user didn't enter it themselves?

jakubjezek001:
Once I had fixed the path for otio file. The file is created, but this Traceback had ocurred:
image

```
  File "C:\CODE\__PYPE\ayon-core\client\ayon_core\plugins\load\export_otio.py", line 202, in export
    run_subprocess(["otioview", output_path])
  File "C:\CODE\__PYPE\ayon-core\client\ayon_core\lib\execute.py", line 128, in run_subprocess
    proc = subprocess.Popen(*args, **kwargs)
  File "C:\Python\pyenv-win\versions\3.9.4\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Python\pyenv-win\versions\3.9.4\lib\subprocess.py", line 1420, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
 ```

**Comment by BigRoy** [comment]
> Once I had fixed the path for otio file. The file is created, but this Traceback had ocurred: image
> 
> ```
>   File "C:\CODE\__PYPE\ayon-core\client\ayon_core\plugins\load\export_otio.py", line 202, in export
>     run_subprocess(["otioview", output_path])
>   File "C:\CODE\__PYPE\ayon-core\client\ayon_core\lib\execute.py", line 128, in run_subprocess
>     proc = subprocess.Popen(*args, **kwargs)
>   File "C:\Python\pyenv-win\versions\3.9.4\lib\subprocess.py", line 951, in __init__
>     self._execute_child(args, executable, preexec_fn, close_fds,
>   File "C:\Python\pyenv-win\versions\3.9.4\lib\subprocess.py", line 1420, in _execute_child
>     hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
> ```

This traceback seems cut off / incomplete. Do you happen to have slightly more?

File: client/ayon_core/plugins/load/export_otio.py [hunk 5]
```diff
@@ -0,0 +1,231 @@
...
+        if check_state == QtCore.Qt.CheckState.Checked:
+            run_subprocess(["otioview", output_path])
+
+    def create_clip(self, path, name, frames, framerate):
+        range = self.otio.opentime.TimeRange(
+            start_time=self.otio.opentime.RationalTime(0, framerate),
+            duration=self.otio.opentime.RationalTime(frames, framerate)
+        )
+
+        media_reference = self.otio.schema.ExternalReference(
```
jakubjezek001, line 210: we should differenciate between sequence media source type and video files. Here is example in their code https://github.com/AcademySoftwareFoundation/OpenTimelineIO/blob/main/src/py-opentimelineio/opentimelineio/schema/image_sequence_reference.py

jakubjezek001: I have tried to import it into Resolve as OTIO and it could not find the clips by urls. Perhaps because those were image sequences.

tokejepsen: Using the `ExternalReference` in Resolve works here with OTIO, but I agree we need to use the `ImageSequence` instead. Will update and we can test Resolve and Hiero again.

tokejepsen: @jakubjezek001 could you maybe post the steps you take to import into Resolve and Hiero?

I've tried using `ImageSequence` in Resolve, but the native OTIO implementation does not seem to support it. Again I got the correct import with `ExternalReference`.
I cant get Hiero to import with either `ExternalReference` or `ImageSequence`.

This does maybe seem like we are better off not relying on native implementations in DCCs and just parse the OTIO files ourselves, if there are going to be different implementations between DCCs.

tokejepsen:
@jakubjezek001

> is there any way we would have Output Path browser button for searching a folder, rather then string input field?

Sure, good idea. The `Output path:` label is now a button for file browsing.

> Also here is a stupid mistake what could be done by anyone. I had add folder path without a file and extension and this was only printed in terminal and no pop-up warning was shown.

Good spot! Have implemented some validations with feedback to the user. Now empty output path is validated and it needs to end with `.otio`.

Also running the OTIO view in detached process so we can close the dialog after showing a successful message to the user.

client/ayon_core/plugins/load/export_otio.py
BigRoy: > This does maybe seem like we are better off not relying on native implementations in DCCs and just parse the OTIO files ourselves, if there are going to be different implementations between DCCs.

FYI, If there are differences I feel time is better spent reporting these as bugs to each individual DCC. We might want a workaround temporarily but we should definitely help push to make sure their implementations start aligning over time if they do not. The developers of those apps are likely unaware of the issues.

tokejepsen: @BigRoy good point!

jakubjezek001:
I am not able to use the `Output path` button in Windows. It is probably set to different property then it should, since it is asking me to open `.otio` file which doesn't exists. Look into the attached gif. 

mutliselectiontest

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def toggle_all(self):
+        representation_name = self.sender().text()
+        for item in self._representation_widgets[representation_name]:
+            item["widget"].setChecked(True)
+
+    def set_output_path(self):
+        file_dialog = QtWidgets.QFileDialog()
+        file_dialog.setNameFilter("OTIO Files (*.otio)")
+        file_dialog.setFileMode(QtWidgets.QFileDialog.ExistingFile)
```
jakubjezek001, line 177:
```suggestion
        file_dialog.setFileMode(QtWidgets.QFileDialog.AnyFile)
```
This had fixed the issue with non existent file from comment.

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,280 @@
...
+                    version["data"]["frameStart"]
+                ),
+                "framerate": version["data"]["fps"]
+            }
+
+        self.export_otio(clips_data, output_path)
+
+        check_state = self.checkbox_inspect_otio_view.checkState()
+        if check_state == QtCore.Qt.CheckState.Checked:
+            run_detached_process(["otioview", output_path])
```
jakubjezek001, line 241: Some sanity check for presence of `otioview` would be desired. In my case it had crashed with:

```
  File "C:\CODE\__PYPE\ayon-core\client\ayon_core\plugins\load\export_otio.py", line 241, in export
    run_detached_process(["otioview", output_path])
  File "C:\CODE\__PYPE\ayon-core\client\ayon_core\lib\execute.py", line 302, in run_detached_process
    process = subprocess.Popen(args, **kwargs)
  File "C:\Python\pyenv-win\versions\3.9.4\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Python\pyenv-win\versions\3.9.4\lib\subprocess.py", line 1420, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

But I am having `otioview` inside of my vendor folder at ayon-launcher. 
image

Perhaps we should direct the path as abs to launcher's `otioview`?

jakubjezek001 (changes requested):

client/ayon_core/plugins/load/export_otio.py
jakubjezek001: But to be honest it still feels bit confusing, since no file exists. I would rather recomend to split the attribute into two: Output Dir and File Name (without extension, since it is quite obvious). Where the File Name can be pre-created by some template in settings and perhaps manually changed if desired.

client/ayon_core/plugins/load/export_otio.py
jakubjezek001: I had tested it with the abs path to my otioview from above comment and it seems we might have some buggy dependency:

```
  File "C:\Python\pyenv-win\versions\3.9.4\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python\pyenv-win\versions\3.9.4\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\CODE\__PYPE\ayon-launcher\vendor\python\bin\otioview.exe\__main__.py", line 4, in <module>
ModuleNotFoundError: No module named 'opentimelineview'
```

jakubjezek001: But @iLLiCiTiT will tell you more about **OpenTimelineIO** dependency redesign.

iLLiCiTiT:
Finally found some time to look into the issue with executable. After some research I didn't find "best" solution, but few possible solutions.

1. Set path to dependency packages on ayon-launcher start and use that path to find the executable. This works, but works in a different way on different platform. Would require new release of ayon-launcher which set the value for future usage, it has many caviads so this is 5/10 solution.

2. Implement custom stript `ayon_core/scripts/otio_scripts.py` with custom handling of the logic.
```python
import sys


def main():
    # Remove first argument - path to this script
    sys.argv.pop(0)
    # Second argument is name of otio script
    script = sys.argv[0]
    if script == "otioview":
        from opentimelineview.console import main
        return main()

    if script == "otioautogen_serialized_schema_docs":
        from opentimelineio.console.autogen_serialized_datamodel import main
        return main()

    from opentimelineio.console import (
        otiocat,
        otioconvert,
        otiopluginfo,
        otiostat,
        otiotool,
    )
    if script == "otiocat":
        return otiocat.main()
    if script == "otioconvert":
        return otioconvert.main()
    if script == "otiopluginfo":
        return otiopluginfo.main()
    if script == "otiostat":
        return otiostat.main()
    if script == "otiotool":
        return otiotool.main()
    raise RuntimeError("Unknown script: {}".format(script))


if __name__ == "__main__":
    main()
```

And use it instead as:
```python
from ayon_core.lib import get_ayon_launcher_args, run_detached_process

script_path = os.path.join(AYON_CORE_ROOT, "scripts", "otio_scripts.py")
args = get_ayon_launcher_args("run", script_path, "otioview", output_path)
run_detached_process(args)
```

tokejepsen:
> I am not able to use the Output path button in Windows. It is probably set to different property then it should, since it is asking me to open .otio file which doesn't exists. Look into the attached gif.

@jakubjezek001 I've fixed the output button, so existing and non-existent file paths can be specified. Also when inputting the file name you dont need the extension.

I've also removed the `otioview` functionality since its complicated to account for the existence of the application and its not critical to the PR.

client/ayon_core/plugins/load/export_otio.py
tokejepsen: Fixed

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def create_clip(self, path, name, frames, framerate):
+        range = self.otio.opentime.TimeRange(
+            start_time=self.otio.opentime.RationalTime(0, framerate),
+            duration=self.otio.opentime.RationalTime(frames, framerate)
+        )
+
+        media_reference = self.otio.schema.ExternalReference(
+            available_range=range,
+            target_url=f"file://{path}"
```
jakubjezek001, line 248:
```suggestion
            target_url=path
```

jakubjezek001: once `file://` is removed then it is loading into OpenRV

jakubjezek001 (changes requested):
please fix the OpenRV issue and it is good to be merged.

tokejepsen:
@jakubjezek001 fixed for latest develop and your OpenRV fix works in Resolve. I dont have access to OpenRV so would appreciate a test on your end.

jakubjezek001 (changes requested):
we need to provide support for image sequences

tokejepsen:
> This is now working only on single frames and video references, not on image sequences. The type of reference is wrong.

@jakubjezek001 as mentioned here; https://github.com/ynput/ayon-core/pull/44/files#r1488373642, using the `ImageSequenceReference` does not work in Resolve or NukeStudio. Do you still want this implemented so you can test in OpenRV?

client/ayon_core/plugins/load/export_otio.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+class ExportOTIO(load.ProductLoaderPlugin):
+    """Export selected versions to OpenTimelineIO."""
+
+    is_multiple_contexts_compatible = True
+    sequence_splitter = "__sequence_splitter__"
+
+    representations = ["*"]
+    families = ["*"]
```
iLLiCiTiT, line 19:
```suggestion
    representations = {"*"}
    product_types = {"*"}
```

client/ayon_core/plugins/load/export_otio.py
jakubjezek001: I was just testing the export/import with use of `ImageSequenceReference` for sequences and all was working as expected. The test had been done with Hiero 15 and DaVinci Resolve 16.5. 

@tokejepsen I am not sure what versions you were testing [here ](https://github.com/ynput/ayon-core/pull/44/files#r1488373642) it in but this is perfectly supported.

Please lets implement it with `ImageSequenceReference`.

jakubjezek001:
> > This is now working only on single frames and video references, not on image sequences. The type of reference is wrong.
> 
> @jakubjezek001 as mentioned here; [#44 (files)](https://github.com/ynput/ayon-core/pull/44/files#r1488373642), using the `ImageSequenceReference` does not work in Resolve or NukeStudio. Do you still want this implemented so you can test in OpenRV?
 
Please read the coment in https://github.com/ynput/ayon-core/pull/44/files#r1590949130. Lets implement it with `ImageSequenceReference`.

jakubjezek001:
My tests have revealed the following:

**Hiero 15** imports OTIO with a file URI, but if sequence files have a timecode value that doesn't match the frame number, then clips are not positioned correctly.

**Resolve 19** can't seem to handle file URIs. I suspect it either can't read `ImageSequenceReference` or the mismatched timecode in the image sequence is causing issues with its conforming process.

**OpenRV** also struggles with file URIs and timecodes.

Based on these findings, here's what we should do for this tool:

*   We need to integrate an oiiotool metadata reader to extract timecode for media references and apply it at the clip source's start frame.
*   We should add a selector widget to choose the type of OTIO file to create, covering all supported DCCs.

tokejepsen:
> We should add a selector widget to choose the type of OTIO file to create, covering all supported DCCs.

Would it make more sense to export all OTIO variants that work for the supported DCCs? Then users do not have to choose the destination DCC at publish, but can decide later which DCC to use.
This would be similar to export MayaAscii and Alembic for a model.

jakubjezek001:
> > We should add a selector widget to choose the type of OTIO file to create, covering all supported DCCs.
> 
> Would it make more sense to export all OTIO variants that work for the supported DCCs? Then users do not have to choose the destination DCC at publish, but can decide later which DCC to use. This would be similar to export MayaAscii and Alembic for a model.

Had been implemented via **URI paths** toggle.

client/ayon_core/lib/transcoding.py
```diff
@@ -834,6 +834,112 @@ def get_ffprobe_streams(path_to_file, logger=None):
     return get_ffprobe_data(path_to_file, logger)["streams"]
 
 
+def get_image_info_metadata(
```
iLLiCiTiT, line 837: I really don't think this is usefull. OIIO might output different data than ffprobe, and whatever is asking for the information should know if the output is from oiio or ffmpeg.

client/ayon_core/lib/transcoding.py
```diff
@@ -834,6 +834,112 @@ def get_ffprobe_streams(path_to_file, logger=None):
...
+    ext = os.path.splitext(path_to_file)[-1].lower()
+    if ext not in IMAGE_EXTENSIONS:
+        logger.info((
+            "File extension \"{}\" is not supported by OpenImageIO."
+            " Trying to get metadata using FFprobe."
+        ).format(ext))
+        ffprobe_stream = get_ffprobe_data(path_to_file, logger)
+        if "streams" in ffprobe_stream and len(ffprobe_stream["streams"]) > 0:
+            metadata_stream = _ffprobe_metadata_conversion(
+                ffprobe_stream["streams"][0])
```
iLLiCiTiT, line 881: What is is first stream audio?

client/ayon_core/lib/transcoding.py
```diff
@@ -834,6 +834,112 @@ def get_ffprobe_streams(path_to_file, logger=None):
...
+    if (
+        "r_frame_rate" in metadata_stream
+        or "framespersecond" in metadata_stream
+    ):
+        rate_info = metadata_stream.get("r_frame_rate")
+        if rate_info is None:
+            rate_info = metadata_stream.get("framespersecond")
+
+        try:
+            metadata_stream["framerate"] = eval(str(rate_info))
```
iLLiCiTiT, line 924: I highly not recommend to use `eval`. Why we can't use `float` conversion?

client/ayon_core/lib/transcoding.py
```diff
@@ -834,6 +834,112 @@ def get_ffprobe_streams(path_to_file, logger=None):
...
+        except Exception as e:
+            logger.warning(
+                "Failed to evaluate '{}' value to framerate. Error: {}".format(
+                    rate_info, e)
+            )
+
+    output = {}
+    for key in keys:
```
iLLiCiTiT, line 932: What should this loop do?

client/ayon_core/lib/__init__.py
```diff
@@ -146,15 +147,11 @@
     "get_local_site_id",
     "get_ayon_username",
     "get_openpype_username",
-
```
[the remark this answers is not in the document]
jakubjezek001: it seems like some auto chenges.. I will reverse those..

client/ayon_core/lib/transcoding.py
jakubjezek001: I am only testing the function in tests. Tests parametrized for both exr and mov. And all it needs to get for the implementation is provided reliably.

client/ayon_core/lib/transcoding.py
jakubjezek001: If I understand the question correctly - are you asking if I am sure that the first stream will always be video?

Is there chance this can happen that steam 0 will be audio?

client/ayon_core/lib/transcoding.py
jakubjezek001: Aggregate all required metadata from prepared metadata stream

client/ayon_core/lib/transcoding.py
iLLiCiTiT: If input contains both video and audio, then audio is usually before video.

client/ayon_core/lib/transcoding.py
iLLiCiTiT: If it's used only in tests then please move it to tests, when it's added to lib it becomes maintained code.

client/ayon_core/lib/transcoding.py
iLLiCiTiT: Why it does combine it with nested keys?

client/ayon_core/lib/transcoding.py
```diff
@@ -834,6 +834,113 @@ def get_ffprobe_streams(path_to_file, logger=None):
...
+    ext = os.path.splitext(path_to_file)[-1].lower()
+    if ext not in IMAGE_EXTENSIONS:
+        logger.info((
+            "File extension \"{}\" is not supported by OpenImageIO."
+            " Trying to get metadata using FFprobe."
+        ).format(ext))
+        ffprobe_stream = get_ffprobe_data(path_to_file, logger)
+        if "streams" in ffprobe_stream and len(ffprobe_stream["streams"]) > 0:
+            metadata_stream = _ffprobe_metadata_conversion(
+                ffprobe_stream["streams"][-1])
```
iLLiCiTiT, line 881: I think this was misunderstood. You shoud look through stream and find type `"video"`.

client/ayon_core/lib/transcoding.py
jakubjezek001: in some cases there could be nested dictionaries where the key should be search in

client/ayon_core/lib/transcoding.py
iLLiCiTiT: I see it's been used in the loader. If that's the case then please add exacmple output to the function docstring. I would also add "base" output of the function, so a developer using the function can know what the output actually should contain.

Right now with the nested dictionary manipulation and combination of OIIO and ffmpeg it is uber confusing to find out what it actually does return.

If it would always return the same data structure, it would help a lot. E.g.
```
{
     "timecode": "00:00:00",
     "frame_start": 1,
     "frame_end": 2,
}
```

client/ayon_core/lib/transcoding.py
jakubjezek001: hey @iLLiCiTiT could we just merge this?

iLLiCiTiT: Linking comment that is the most important https://github.com/ynput/ayon-core/pull/44#discussion_r1695458171

jakubjezek001: have moved it into the loader plugin and improved docstring. [`ede1b17` (#44)](https://github.com/ynput/ayon-core/pull/44/commits/ede1b1737013df3b8f0a38d5d781511e9915d10c)

client/ayon_core/lib/transcoding.py
jakubjezek001: function move into [loader plugin](https://github.com/ynput/ayon-core/pull/44/commits/ede1b1737013df3b8f0a38d5d781511e9915d10c) and [tests file renamed](https://github.com/ynput/ayon-core/pull/44/commits/e485eda6110e89a70c11672f89cb96187b8fc1ad) and in more appropriate stucture.

PR #47, opened 2024-02-13
Max: Add missing workfile creator

moonyuet:
## Changelog Description
Add the missing workfile creator in 3dsMax.
Ported from Openpype: https://github.com/ynput/OpenPype/pull/6203

## Additional info
For some reason, `get_subset_name` function can't set the name with the correct variant, although there is variant.
It is using the old subset name `workfileAsset`. 

## Testing notes:
1. Launch Max via launcher
2. Load Scene
3. Open publisher
4. it should automatically create the workfile instance
5. you can notice the new container (which is hidden and greyed out) named accordingly
6. Publish workfile
7. It should be successfully published.


client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -55,7 +53,7 @@ def process(self, context):
         }]
 
         instance.data.update(data)
-
+        self.log.info('Collected data: {}'.format(data))
```
BigRoy, line 56: This should **definitely** not be an info log - and I'd even say that for debug logging just dumping blobs of data is pretty hard to parse.

client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -23,15 +24,12 @@ def process(self, context):
...
+        ext = os.path.splitext(file)[-1].lstrip(".")
 
         data = {}
 
         # create instance
-        instance = context.create_instance(name=filename)
-        subset = 'workfile' + task.capitalize()
+        subset = instance.data["subset"]
 
         data.update({
```
BigRoy, line 34: A lot of the logic here of data being adjusted can be removed since it SHOULD be handled by the Creator.

The subset, asset, publish, family and families should be keys that should be inherited from the creator plug-in's instance as opposed to set here - as far as I know.

client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -23,15 +24,12 @@ def process(self, context):
 
         context.data['currentFile'] = current_file
 
-        filename, ext = os.path.splitext(file)
-
-        task = context.data["task"]
+        ext = os.path.splitext(file)[-1].lstrip(".")
```
BigRoy, line 27: Note that the usage below with `ext` on lines 49 and 50 there's also `lstrip` doing the same. Can those be removed now?

client/ayon_core/hosts/max/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,123 @@
+# -*- coding: utf-8 -*-
+"""Creator plugin for creating workfiles."""
+from ayon_core import AYON_SERVER_ENABLED
```
iLLiCiTiT, line 3: You can remove this and it's usage in ayon-core.

client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -23,15 +24,12 @@ def process(self, context):
...
 
         data = {}
 
         # create instance
-        instance = context.create_instance(name=filename)
-        subset = 'workfile' + task.capitalize()
+        subset = instance.data["subset"]
 
         data.update({
             "subset": subset,
```
BigRoy, line 35: The issue described:

> For some reason, get_subset_name function can't set the name with the correct variant, although there is variant.
> It is using the old subset name workfileAsset.

Is likely due to this setting `subset`?

Should be removed here.

client/ayon_core/hosts/max/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,114 @@
...
+            self.log.info("Auto-creating workfile instance...")
+            instance_node = self.create_node(subset_name)
+            data["instance_node"] = instance_node.name
+            current_instance = CreatedInstance(
+                self.family, subset_name, data, self
+            )
+            self._add_instance_to_context(current_instance)
+            imprint(instance_node.name, current_instance.data)
+        elif (
+            current_instance_asset != asset_name
```
BigRoy, line 61: Wouldn't this work as well?
```suggestion
        if current_instance is None:
            asset_doc = get_asset_by_name(project_name, asset_name)
            subset_name = self.get_subset_name(
                variant, task_name, asset_doc, project_name, host_name
            )
            data = {
                "folderPath": asset_name,
                "task": task_name,
                "variant": variant
            }

            data.update(
                self.get_dynamic_data(
                    variant, task_name, asset_doc,
                    project_name, host_name, current_instance)
            )
            self.log.info("Auto-creating workfile instance...")
            instance_node = self.create_node(subset_name)
            data["instance_node"] = instance_node.name
            current_instance = CreatedInstance(
                self.family, subset_name, data, self
            )
            self._add_instance_to_context(current_instance)
            imprint(instance_node.name, current_instance.data)
        elif (
            current_instance["folderPath"] != asset_name
```

client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -48,15 +39,14 @@ def process(self, context):
...
         instance.data.update(data)
-
-        self.log.info('Collected instance: {}'.format(file))
-        self.log.info('Scene path: {}'.format(current_file))
-        self.log.info('staging Dir: {}'.format(folder))
-        self.log.info('subset: {}'.format(subset))
+        self.log.debug('Collected data: {}'.format(data))
+        self.log.debug('Collected instance: {}'.format(file))
+        self.log.debug('Scene path: {}'.format(current_file))
+        self.log.debug('staging Dir: {}'.format(folder))
```
BigRoy, line 52:
```suggestion
        self.log.debug('Collected workfile path: {}'.format(current_file))
        self.log.debug('Collected workfile staging dir: {}'.format(folder))
        self.log.debug('Collected workfile data: {}'.format(data))
```
Would this be sufficient debug data to log?

Also, you're using quite a mixture of `'` quotes versus `"` in this file. Should we align them to all use `"` like most of the codebase?

client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -6,15 +6,16 @@
...
-class CollectWorkfile(pyblish.api.ContextPlugin):
+class CollectWorkfile(pyblish.api.InstancePlugin):
     """Inject the current working file into context"""
 
     order = pyblish.api.CollectorOrder - 0.01
     label = "Collect 3dsmax Workfile"
     hosts = ['max']
 
-    def process(self, context):
+    def process(self, instance):
```
BigRoy, line 16: **Warning!** There is a bug here.

This will now run over **all instances** - not only workfile instances.

This will need to be targeted to `families = ["workfile"]`

**However** you might then need to separate the `context.data["currentFile"]` into a separate Collector since it's not related to the workfile instance, but is for the global context and is also used if the workfile instance is disabled during publishing.

client/ayon_core/hosts/max/plugins/publish/collect_workfile.py
```diff
@@ -6,57 +6,43 @@
...
 
         instance.data.update(data)
-
-        self.log.info('Collected instance: {}'.format(file))
-        self.log.info('Scene path: {}'.format(current_file))
-        self.log.info('staging Dir: {}'.format(folder))
-        self.log.info('subset: {}'.format(subset))
+        self.log.debug("Collected data: {}".format(data))
+        self.log.debug("Collected instance: {}".format(file))
+        self.log.debug("Scene path: {}".format(
```
BigRoy, line 46: Would this scene path better be logged from `CollectCurrentFile`? :)

BigRoy (approved):
Didn't test it (don't use Max) but overall code changes look okay on first sight.

client/ayon_core/hosts/max/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,109 @@
...
+from ayon_core.pipeline import CreatedInstance, AutoCreator
+from ayon_core.client import get_asset_by_name, get_asset_name_identifier
+from ayon_core.hosts.max.api import plugin
+from ayon_core.hosts.max.api.lib import read, imprint
+from pymxs import runtime as rt
+
+
+class CreateWorkfile(plugin.MaxCreatorBase, AutoCreator):
+    """Workfile auto-creator."""
+    identifier = "io.openpype.creators.max.workfile"
```
iLLiCiTiT, line 12:
```suggestion
    identifier = "io.ayon.creators.max.workfile"
```

client/ayon_core/hosts/max/plugins/publish/collect_current_file.py
```diff
@@ -0,0 +1,23 @@
+import os
+import pyblish.api
+
+from pymxs import runtime as rt
+
+
+class CollectCurrentFile(pyblish.api.ContextPlugin):
+    """Inject the current working file."""
+
+    order = pyblish.api.CollectorOrder - 0.4
```
iLLiCiTiT, line 10: This is not dependendent on anything before? I would move it to boundaries...
```suggestion
    order = pyblish.api.CollectorOrder - 0.5
```

moonyuet: it isn't, so i would make a change on this

PR #50, opened 2024-02-13
TrayPublisher: Use AYON settings

iLLiCiTiT:
## Changelog Description
Traypublisher does not need AYON settings conversion.

## Additional info
The simple creators did change only single key, the most changes happened in editorial create logic.

## Testing notes:
1. Run at least one simple creator and publish
2. Create editorial and publish


jakubjezek001 (approved):
tested on Editorial simple and Settings creator too and works as before

PR #51, opened 2024-02-13
Tools: Move asset icon functions to assets widget

iLLiCiTiT:
## Changelog Description
The functions should not be used anywhere else than in assets widget at this moment, moving them to assets widget which is marked as deprecated should make it cleaner that they should not be used.

## Testing notes:
Just a technical change of functions and related imports.


BigRoy:
Should we mark the function calls themselves so that any usage would report a deprecated message already?

iLLiCiTiT:
> Should we mark the function calls themselves so that any usage would report a deprecated message already?

They're still used in Publisher. But I will add docstrings to them.

PR #53, opened 2024-02-13
Settings: Host imageio use AYON settings

iLLiCiTiT:
## Changelog Description
Imageio settings in hosts does not require conversion.

## Additional info
The changes should be affecting only settings that are used in `colorspace.py`. Modified `colorspace.py` to use AYON data structure. I'm not sure if there is a host which is using directly imageio settings that are handled by `colorspace.py`? I couldn't find any.

## Testing notes:
I need help with testing notes as I don't know what it in fact affects. Probably validate if imageio is used correctly during publishing?

### DIFF
--- a/client/ayon_core/pipeline/colorspace.py
+++ b/client/ayon_core/pipeline/colorspace.py
@@ -837,13 +837,13 @@ def get_imageio_config(
 
     if override_global_config:
         config_data = _get_config_data(
-            host_ocio_config["ocio_config"], formatting_data, env
+            host_ocio_config["filepath"], formatting_data, env
         )
     else:
         # get config path from global
         config_global = imageio_global["ocio_config"]
         config_data = _get_config_data(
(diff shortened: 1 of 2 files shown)


moonyuet (commented):
It doesn't work at Max when enabled imageio and global ocio config. It gives me an error
image
Maybe it is somehow related to the ocio env hook isn't installed correctly in Max.

jakubjezek001:
> It doesn't work at Max when enabled imageio and global ocio config(not sure if it is related to this particular PR). The script editor shows me an error when I use the Set Colorspace in Max

Have a look if you are having activated `ayon_ocio` in your bundle.

jakubjezek001 (approved):
works perfect in nuke! ;)

moonyuet (approved):
Works now... Thanks @jakubjezek001 for help!
image

PR #54, opened 2024-02-13
Chore: Addons can define their openpype alias name

iLLiCiTiT:
## Changelog Description
Addon can be registered under 2 names in `AddonsManager`.

## Additional info
This is needed for `sitesync` addon which was named `sync_server` in OpenPype. With this change it is possible to do both:
```python
from ayon_core.addon import AddonsManager

manager = AddonsManager()
sync_server = manager["sync_server"]
sitesync = manager["sitesync"]
print(sync_server is sitesync)
>>> True
```

### DIFF
--- a/client/ayon_core/addon/base.py
+++ b/client/ayon_core/addon/base.py
@@ -788,6 +788,7 @@ class AddonsManager:
 
                 addon_classes.append(modules_item)
 
+        aliased_names = []
         for addon_cls in addon_classes:
             name = addon_cls.__name__
             if issubclass(addon_cls, OpenPypeModule):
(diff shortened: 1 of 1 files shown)


client/ayon_core/addon/base.py
```diff
@@ -822,6 +830,17 @@ def initialize_addons(self):
                     exc_info=True
                 )
 
+        for item in aliased_names:
+            name_alias, addon = item
```
BigRoy, line 834:
```suggestion
        for name_alias, addon in aliased_names:
```

iLLiCiTiT: Some pythons have issue with this expanding in for loop.

BigRoy: Which python? 

It works fine for me in Py 2.7, Py 3.7, Py 3.9 and Py 3.10 with the next code snippet:
```python
data = [(1, 2), (1001, 1010)]
for start, end in data:
    print(start, end)
```

iLLiCiTiT: To be honest I don't remember, but I'm scared to use this now as we need to make release today.

kalisp (approved):
Site Sync working.

PR #55, opened 2024-02-14
Multi layout creator publish each layout with frame range instead of clip in

bradenjennings:
## Changelog Description
Paragraphs contain detailed information on the changes made to the product or service, providing an in-depth description of the updates and enhancements. They can be used to explain the reasoning behind the changes, or to highlight the importance of the new features. Paragraphs can often include links to further information or support documentation.

## Additional info

## Testing notes:
1. start with this step
2. follow this step


client/ayon_core/hosts/maya/plugins/publish/extract_layout.py
```diff
@@ -145,6 +145,11 @@ def process(self, instance):
 
             json_data.append(json_element)
 
+        shift_back_to_frame_start = instance.data.get("shift_back_to_frame_start", False)
+        if shift_back_to_frame_start:
+            shot_node = instance.data.get("shot_node")
+            # cmds.setAttr(shot_node + ".sequenceStartFrame", instance.data.get("frameStart"))
```
moonyuet, line 151:
```suggestion
```

client/ayon_core/hosts/maya/plugins/publish/validate_instance_in_context.py
```diff
@@ -34,6 +34,7 @@ class ValidateInstanceInContext(pyblish.api.InstancePlugin,
     ]
 
     def process(self, instance):
+        return
```
moonyuet, line 37:
```suggestion
        return
```
Maybe it's better idea to skip the validator for just layout family? Right now the validator would be skipping for any families for this adjustment.

bradenjennings: so like this. Just add the families field below the Validate instance in context field?
```
    ValidateInstanceInContext: BasicValidateModel = SettingsField(
        default_factory=BasicValidateModel,
        title="Validate Instance In Context",
        section="Validators")
    families: list[str] = SettingsField(
        default_factory=list, title="Families")
```

> (and create proper default values)

How do I create default values?

client/ayon_core/hosts/maya/plugins/publish/extract_layout.py
bradenjennings: @antirotor is it the right approach to set the shot node start frame like this?

I noticed in the code for extract layout it extracts a json, and doesn't seem to concern itself with the frames of the sequence at all. So offsetting the shot node start frame probably will make no difference.

client/ayon_core/hosts/maya/plugins/publish/validate_instance_in_context.py
moonyuet: > so like this. Just add the families field below the Validate instance in context field?
> 
> ```
>     ValidateInstanceInContext: BasicValidateModel = SettingsField(
>         default_factory=BasicValidateModel,
>         title="Validate Instance In Context",
>         section="Validators")
>     families: list[str] = SettingsField(
>         default_factory=list, title="Families")
> ```
> 
> > (and create proper default values)
> 
> How do I create default values?

Consider this pesudo code:
```

class ValidateInstanceInContextModel(BaseSettingsModel):
    enabled: bool = SettingsField(title="Enabled")
    optional: bool = SettingsField(title="Optional")
    active: bool = SettingsField(title="Active")
    families: list[str] = SettingsField(
        default_factory=list, title="Families")


class PublishersModel(BaseSettingsModel):
    ValidateInstanceInContext: BasicValidateModel = SettingsField(
        default_factory=ValidateInstanceInContextModel,
        title="Validate Instance In Context",
        section="Validators"
    )


DEFAULT_PUBLISH_SETTINGS = {
    "ValidateInstanceInContext": {
        "enabled": True,
        "optional": True,
        "active": True,
        "families": ["layout"]
    }
}
```

client/ayon_core/hosts/maya/plugins/publish/extract_layout.py
bradenjennings: @antirotor any comment?

NOTE: This line should actually not be commented out. Its my proposed solution for offseting the clip back to frame start.

client/ayon_core/hosts/maya/plugins/publish/validate_instance_in_context.py
moonyuet: You may need to change `families ` in AYON settings to `excluded_families` so that the system won't get confused on `families` and the families to be excluded.  Consider this pesudo code.
```
from maya import cmds
class ValidateInstanceInContext(pyblish.api.InstancePlugin,
                                OptionalPyblishPluginMixin):
    """Validator to check if instance asset match context asset.
    When working in per-shot style you always publish data in context of
    current asset (shot). This validator checks if this is so. It is optional
    so it can be disabled when needed.
    Action on this validator will select invalid instances in Outliner.
    """
    order = ValidateContentsOrder
    label = "Instance in same Context"
    optional = True
    hosts = ["maya"]
    actions = [
        ayon_core.hosts.maya.api.action.SelectInvalidAction, RepairAction
    ]
    excluded_families = []

    def process(self, instance):
        families = instance.data["families"]
        excluded_product_type = families.intersection(self.excluded_families)
        if excluded_product_type:
            self.log.debug("Skipping Validate Instance In Context")
        
        # Some codes for checking on current context
```

bradenjennings: Validate Instance in Context has no excluded families attribute.
image

Should I just hardcode the excluded families on the class to include layout.

```
class ValidateInstanceInContext(pyblish.api.InstancePlugin, OptionalPyblishPluginMixin):
    ...
    excluded_families = ["layout"]
```

moonyuet: Have you built the sever addons and install maya addon into the AYON server? It seems that your maya addon is not at the latest version with this branch as I didn't notice the `families` as being part of the AYON setting as your screengrab shown. 
Speaking of the `excluded_families`, You can just add `excluded_families` into AYON settings and the `excluded_families` would be registered by this case.

server_addon/maya/server/settings/publishers.py
```diff
@@ -824,7 +832,8 @@ class PublishersModel(BaseSettingsModel):
     "ValidateInstanceInContext": {
         "enabled": True,
         "optional": True,
-        "active": True
+        "active": True,
+        "families": ["layout"]
```
moonyuet, line 836:
```suggestion
        "excluded_families": ["layout"]
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -362,6 +362,14 @@ class ExtractLookModel(BaseSettingsModel):
     )
 
 
+class ValidateInstanceInContextModel(BaseSettingsModel):
+    enabled: bool = SettingsField(title="Enabled")
+    optional: bool = SettingsField(title="Optional")
+    active: bool = SettingsField(title="Active")
+    families: list[str] = SettingsField(
```
moonyuet, line 369:
```suggestion
    excluded_families: list[str] = SettingsField(
```

client/ayon_core/hosts/maya/plugins/publish/extract_layout.py
moonyuet: > @antirotor any comment?
> 
> NOTE: This line should actually not be commented out. Its my proposed solution for offseting the clip back to frame start.

If it is affected the layout workflow, you can just ignore my comment:)

client/ayon_core/hosts/maya/plugins/publish/validate_instance_in_context.py
bradenjennings: > Have you built the sever addons and install maya addon into the AYON server? 
How do I do this?

By the way, I updated my Maya adds bundle, and I still don't see a "families" attribute.
image

server_addon/maya/server/settings/publishers.py
```diff
@@ -397,7 +405,7 @@ class PublishersModel(BaseSettingsModel):
         title="Collect Assets for GLB/GLTF export"
     )
     ValidateInstanceInContext: BasicValidateModel = SettingsField(
```
moonyuet, line 407:
```suggestion
    ValidateInstanceInContext: ValidateInstanceInContextModel= SettingsField(
```

client/ayon_core/hosts/maya/plugins/publish/validate_instance_in_context.py
moonyuet: i have commented the code below.
After the adjustment of the code, it should be like this.
image

server_addon/maya/server/settings/publishers.py
```diff
@@ -362,6 +362,14 @@ class ExtractLookModel(BaseSettingsModel):
     )
 
 
+class ValidateInstanceInContextModel(BaseSettingsModel):
+    enabled: bool = SettingsField(title="Enabled")
+    optional: bool = SettingsField(title="Optional")
+    active: bool = SettingsField(title="Active")
+    excluded_families: list[str] = SettingsField(
+        default_factory=list, title="Families")
```
moonyuet, line 370:
```suggestion
        default_factory=list, title="Excluded Families")
```
It would be more concise to use `Excluded Families` as label than `Families` so that the user won't misunderstand the use of this setting.

LiborBatek:
Would it be possible to update testing steps as these basically being non existent aka default github template... thx

iLLiCiTiT:
Please move to https://github.com/ynput/ayon-maya if is still relevant.

PR #57, opened 2024-02-14
Max: Implementation of Validate Render Passes

moonyuet:
## Changelog Description
Ported from Openpype: https://github.com/ynput/OpenPype/pull/6138
This PR is to enhance the current validator of checking the render output before deadline publish. It does the following:
1. The validator `Render Output for Deadline` would be renamed as `Validate Render Passes`
2. The validator would not only check on the invalid render output folder but the invalid filename of render passes.

## Additional info
n/a

## Testing notes:
1. Launch Max via Launcher
4. Open a Renderable Scene
5. Create Some AOVs in Render Setting
6. Create Render Instance with your preferred camera(s)
7. Publish
8. If the AOVs dont have invalid name or render output folder doesn't match the max scene name, it will error out.
9. You can perform repair action once you encounter 7.


LiborBatek (approved):
Seems working normally and as supposed to. When some invalid file name and/or file path then it correctly kicks in and rise error for user.

Screenshot 2024-03-06 140501

After ```Repair``` action all Validates fine and without any errors...

client/ayon_core/hosts/max/plugins/publish/validate_renderpasses.py
```diff
@@ -0,0 +1,186 @@
...
+            instance (pyblish.api.Instance): instance
+            filename (str): filename of the Max scene
+
+        Returns:
+            list: list of invalid filename which doesn't match
+                with the project name
+        """
+        invalid = []
+        file = rt.maxFileName
+        filename, ext = os.path.splitext(file)
```
BigRoy, line 75: I wonder if `file` or `filename` might be clearer as variable name if it's `workfile_path` or `workfile` or `scene_path` or whatever so it's clear that it's not referring to the render path but the max's scene path.

server_addon/max/server/settings/publishers.py
```diff
@@ -185,6 +189,11 @@ class PublishersModel(BaseSettingsModel):
         "optional": True,
         "active": False,
     },
+    "ValidateRenderPasses": {
+        "enabled": True,
+        "optional": True,
```
BigRoy, line 194: Out of curiosity, ff this plug-in is optional - why do we need to check it in the first place? What does it solve by enforcing e.g. the scene name to be in the output filepaths?

moonyuet: actually it should be set to optional = False instead of being optional validator.

client/ayon_core/hosts/max/plugins/publish/validate_renderpasses.py
```diff
@@ -0,0 +1,185 @@
...
+                    instance, ext)
+                invalid.extend(invalid_image_format)
+        elif renderer == "Arnold":
+            cls.log.debug(
+                "Renderpass validation does not support Arnold yet,"
+                " validation skipped...")
+        else:
+            cls.log.debug(
+                "Skipping render element validation "
+                f"for renderer : {renderer}")
```
BigRoy, line 124:
```suggestion
                f"for renderer: {renderer}")
```

client/ayon_core/hosts/max/plugins/publish/validate_renderpasses.py
```diff
@@ -0,0 +1,185 @@
...
+        invalid = []
+        if instance.name not in file_name:
+            cls.log.error("The renderpass filename should contain the instance name.")
+            invalid.append((f"Invalid instance name",
+                            file_name))
+        if renderpass is not None:
+            if not file_name.rstrip(".").endswith(renderpass):
+                cls.log.error(
+                    f"Filename for {renderpass} should "
+                    f"end with {renderpass}"
```
BigRoy, line 149: Should this also report the current filename that it detected?

LiborBatek (approved):
Working ok and no issues were found.

PR #41, opened 2024-02-13
Max: Implementation of Camera Attributes Validator #6110

moonyuet:
## Changelog Description
Implement Validate Camera Attributes in camera family in Max host
Ported from Openpype: https://github.com/ynput/OpenPype/pull/6110

## Additional info
- This validator is not eligible for the camera created from the external plugins such as Redshift and Arnold etc.
- When the values of the camera attributes in the setting set to 0, the validator would skip checking the attribute.

## Testing notes:
1. Launch Max via launcher
2. Tweak the settings `ayon+settings://max/publish/ValidateCameraAttributes` or `project_settings/max/publish/ValidateCameraAttributes`
image
image
3. Create Camera Instance
4. Publish
5. If the camera attributes dont align with those attributes in the setting, the validator would error out.

PR #58, opened 2024-02-14
Max : Optional validator to check invalid context data

moonyuet:
## Changelog Description
Ported from Openpype: https://github.com/ynput/OpenPype/pull/6198
Add optional validator check on invalid context data for asset and task in 3dsMax

## Additional info
Reference to maya's `Validate Instance in Context`
Related PR for implementing the InvalidAction: https://github.com/ynput/ayon-core/pull/41
## Testing notes:
1. Launch 3dsMax via launcher
2. Create two instances with same families, one of them set different asset and task
3. Publish
4. Error would be raised if you enable the validator which check on invalid context data
5. Repair it 
6. Run publisher again it will pass without error


LiborBatek (changes requested):
Gives me an errror msg on ```Validate same context```

```
    runner(*args)
  File "C:\Work\REPO\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_instance_in_context.py", line 42, in process
    context_asset = self.get_context_asset(instance)
  File "C:\Work\REPO\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_instance_in_context.py", line 103, in get_context_asset
    return instance.context.data["asset"]
KeyError: 'asset'
```

LiborBatek (approved):
Works as expected...

Screenshot 2024-02-26 100821


Screenshot 2024-02-26 100924

moonyuet:
@antirotor can you please take a glance at the code before the merge? Thank you!

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,107 @@
...
+            return
+
+        instance_node = rt.getNodeByName(instance.data.get(
+            "instance_node", ""))
+        if not instance_node:
+            return
+        asset = rt.getUserProp(instance_node, "folderPath")
+        context_asset = self.get_context_asset(instance)
+        task = rt.getUserProp(instance_node, "task")
+        context_task = self.get_context_task(instance)
```
BigRoy, line 44: Why does this need to go over the instance node? Isn't this actually collected data already?

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,107 @@
...
+                )
+            )
+
+    @classmethod
+    def get_invalid(cls, instance):
+        node = rt.getNodeByName(instance.data["instance_node"])
+        asset = rt.getUserProp(node, "folderPath")
+        context_asset = cls.get_context_asset(instance)
+        if asset != context_asset:
+            return instance.data["instance_node"]
```
BigRoy, line 88: The get invalid logic fails to check for differing task. This already shows the issue that the validation logic is scattered - better implement the actual validation once in the plug-in so that it's consistent.

moonyuet: I have rephrased a little bit

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,106 @@
...
+                description=(
+                    "## Publishing to a different asset\n"
+                    "There are publish instances present which are publishing "
+                    "into a different asset than your current context.\n\n"
+                    "Usually this is not what you want but there can be cases "
+                    "where you might want to publish into another asset or "
+                    "shot. If that's the case you can disable the validation "
+                    "on the instance to ignore it."
+                )
+            )
```
BigRoy, line 80: I would always try and report all the issues - instead of only asset mismatch if both asset + task were wrong.
How about this?
```suggestion
        invalid = []
        if asset != context_asset:
            invalid.append(
                "Instance '{}' publishes to different asset than current "
                "context: {}. Current context: {}".format(
                    instance.name, asset, context_asset
                )
            )
        if task != context_task:
            invalid.append(
                "Instance '{}' publishes to different task than current "
                "context: {}. Current context: {}".format(
                    instance.name, task, context_task
                )
            )
        
        if invalid:
            message = "\n".join(invalid)
            raise PublishValidationError(
                message=message,
                description=(
                    "## Publishing to a different asset\n"
                    "There are publish instances present which are publishing "
                    "into a different asset or task than your current context.\n\n"
                    "Usually this is not what you want but there can be cases "
                    "where you might want to publish into another asset or "
                    "shot. If that's the case you can disable the validation "
                    "on the instance to ignore it."
                )
            )
```

Or:
```python
        invalid = False
        if asset != context_asset:
            self.log.error(
                "Instance '{}' publishes to different asset than current "
                "context: {}. Current context: {}".format(
                    instance.name, asset, context_asset
                )
            )
            invalid = True
        if task != context_task:
            self.log.error(
                "Instance '{}' publishes to different task than current "
                "context: {}. Current context: {}".format(
                    instance.name, task, context_task
                )
            )
            invalid = True
        
        if invalid:
            raise PublishValidationError(
                message="Instance '{}' does not publish into the current context".format(instance.name),
                description=(
                    "## Publishing to a different asset\n"
                    "There are publish instances present which are publishing "
                    "into a different asset or task than your current context.\n\n"
                    "Usually this is not what you want but there can be cases "
                    "where you might want to publish into another asset or "
                    "shot. If that's the case you can disable the validation "
                    "on the instance to ignore it."
                )
            )
```

client/ayon_core/settings/defaults/project_settings/max.json
BigRoy, line 1: How comes this adds so much? Were this missing before? Is this file invalid?

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,106 @@
...
+                                OptionalPyblishPluginMixin):
+    """Validator to check if instance asset match context asset.
+
+    When working in per-shot style you always publish data in context of
+    current asset (shot). This validator checks if this is so. It is optional
+    so it can be disabled when needed.
+
+    Action on this validator will select invalid instances.
+    """
+
```
antirotor, line 26: this will trigger **pydocstyle** linter *No blank line after or before docstring*

See https://peps.python.org/pep-0257

antirotor (approved):

antirotor (changes requested):
Actually, yes - `max.json`  shouldn't be here?

antirotor (changes requested):
Actually, yes - `max.json`  shouldn't be here?

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,99 @@
...
+        asset = rt.getUserProp(node, "folderPath")
+        context_asset = instance.context.data["folderPath"]
+        if asset != context_asset:
+            invalid.append(node)
+        task = rt.getUserProp(node, "task")
+        context_task = instance.context.data["task"]
+        if task != context_task:
+            invalid.append(node)
+
+        return invalid
```
BigRoy, line 88: Usually it's the case that `get_invalid` logic is re-used across the plug-in. Currently it isn't. It's not a huge issue, but basically means that `selectinvalidaction` currently relies on different checking logic than `process()` does which could mean more maintenance down the line.

Other than that, overall logic looks good. Can't test myself, not a Max user :)

BigRoy: Oh I did want to add! :)

I think with the NEW publisher we can actually implement this DCC agnostic as a global plug-in.

We can just check `instance.data["folderPath"]` against the `context.data["folderPath"]` and validate that way.
Then to repair, we would just tweak the values in `CreateContext` on the `CreatedInstance` instances, change the values, trigger save on the CreateContext and voila.

BigRoy: And then I remembered. I already did so, but just implemetned it for Houdini: https://github.com/ynput/OpenPype/blob/b4d5a7465e95992e1ae07894f62e189e3559b67d/openpype/hosts/houdini/plugins/publish/validate_instance_in_context.py#L14-L13

Pretty sure that would work for Max too then? (except for the SelectROPAction)

moonyuet: > And then I remembered. I already did so, but just implemetned it for Houdini: https://github.com/ynput/OpenPype/blob/b4d5a7465e95992e1ae07894f62e189e3559b67d/openpype/hosts/houdini/plugins/publish/validate_instance_in_context.py#L14-L13
> 
> Pretty sure that would work for Max too then? (except for the SelectROPAction)

yes, pretty sure it will work. It's just that we need to have separate functions for invalid actions.

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,86 @@
...
+
+        folderPath = instance.data.get("folderPath")
+        task = instance.data.get("task")
+        context = self.get_context(instance)
+        if (folderPath, task) != context:
+            context_label = "{} > {}".format(*context)
+            instance_label = "{} > {}".format(folderPath, task)
+            message = (
+                "Instance '{}' publishes to different folderPath than current"
+                "context: {}. Current context:{}".format(
```
BigRoy, line 44:
```suggestion
                "context: {}. Current context: {}".format(
```

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
+"""Validate if instance asset is the same as context asset."""
+from __future__ import absolute_import
```
BigRoy, line 3: Is this needed?

This should not be needed in Py3+.
Ayon is intended to be Py3+ support only. Should we remove this?

BigRoy (approved):
After the comments are resolved I think this looks good.

LiborBatek (approved):
All working ok. The Validator, ```Select Invalid``` and ```Repair``` actions.

Screenshot 2024-03-01 103525

When correct context being set all Publishes ok.

LiborBatek:
@moonyuet maybe just fix the typo ```currentcontext``` to ``` current context``` would do but this being quite cosmetics only

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,84 @@
...
+            return
+
+        folderPath = instance.data.get("folderPath")
+        task = instance.data.get("task")
+        context = self.get_context(instance)
+        if (folderPath, task) != context:
+            context_label = "{} > {}".format(*context)
+            instance_label = "{} > {}".format(folderPath, task)
+            message = (
+                "Instance '{}' publishes to different folderPath than current "
```
BigRoy, line 41: Folder path might be confusing here since it's also potentially publishing to a different task?
```suggestion
                "Instance '{}' publishes to different context than current "
```

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,84 @@
...
+                "context: {}. Current context: {}".format(
+                    instance.name, instance_label, context_label
+                )
+            )
+            raise PublishValidationError(
+                message=message,
+                description=(
+                    "## Publishing to a different asset\n"
```
BigRoy, line 49: Should we rename `asset` here to `folder` or `context`?

antirotor (approved):

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
```diff
@@ -0,0 +1,84 @@
...
+    PublishValidationError,
+    OptionalPyblishPluginMixin
+)
+from ayon_core.hosts.max.api.action import SelectInvalidAction
+from pymxs import runtime as rt
+
+
+class ValidateInstanceInContext(pyblish.api.InstancePlugin,
+                                OptionalPyblishPluginMixin):
+    """Validator to check if instance context match current context .
```
BigRoy, line 16:
```suggestion
    """Validator to check if instance context match current context.
```

client/ayon_core/hosts/max/plugins/publish/validate_instance_in_context.py
BigRoy: This doesn't seem resolved? :)

BigRoy (approved):
Looks good to me.

PR #63, opened 2024-02-14
Applications: Handle additional applications correctly

iLLiCiTiT:
## Changelog Description
Add additional applications correctly to list of applications.

## Additional info
Additional applications is not a dictionary but list.

## Testing notes:
1. Add any additional application in Applications settings
2. It should not raise error when `ApplicationsManager` is created (e.g. in Launcher tool)


MustafaJafar (approved):
I've just checkedout this PR and the error vanished!

PR #64, opened 2024-02-14
Use new package structure

Minkiu:
This commit will modify the `python` script that allows us to quickly generate packages for `ayon-backend` ingestion, either by copying them to the `rez` repo (not advised) or installing them via the `ayon-frontend` by uploading the resulting `zip`s of the script.

## Changelog Description
This is a PR that will allow us to then test the functionality of this `ayon-backend` PR: https://github.com/ynput/ayon-backend/pull/29

Instead of creating a folder structure at this step, we delegate it to the `rezbuild.py` on the `ayon-backend` which expects the `package.py` and the usual folder structure  of the addon source code (https://github.com/ynput/ayon-addon-template), thus this step is greatly simplified where we only copy the contents of the addon, add a `package.py` and zip them up. 

* 4113ff7bf0 Add `rez` packages generation for `server_addons` Allow to generate `rez` compatible packages for the `server_addons`, will create a `package.py` which then is also included in the `.zip`.

* 0744c27c91 `package.py` Ensure we require the correct `ayon_server` version For more info on `rez` package requests syntax check: https://rez.readthedocs.io/en/latest/basic_concepts.html#package-requests-concept

## Testing notes:
1. Clone this branch
2. Run `python server_addon/create_ayon_addons.py -r`
3.  Confirm `zip` archives have been generated in `ayon-core/server_addon/packages` ending in `-rez.zip` and should have the `package.py` at the root of the zip alongside the `manifest.json`.
4. If you are running the backend with the rez implementation: Install the addon, enable it and make sure it is able to initialize.

### DIFF
--- a/server_addon/create_ayon_addons.py
+++ b/server_addon/create_ayon_addons.py
@@ -144,11 +144,18 @@ def create_addon_zip(
     output_dir: Path,
     addon_name: str,
     addon_version: str,
-    keep_source: bool
+    keep_source: bool,
+    rez_package: bool,
 ):
     zip_filepath = output_dir / f"{addon_name}-{addon_version}.zip"
+
+    if rez_package:
+        zip_filepath = output_dir / f"{addon_name}-{addon_version}-rez.zip"
+
     addon_output_dir = output_dir / addon_name / addon_version
     with ZipFileLongPaths(zip_filepath, "w", zipfile.ZIP_DEFLATED) as zipf:
+        # TODO: Decide if we want to keep this for first pass checking without
+        # executing any python. Otherwiseh we can be removed
         zipf.writestr(
             "manifest.json",
             json.dumps({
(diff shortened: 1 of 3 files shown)


server_addon/create_ayon_addons.py
```diff
@@ -191,22 +205,37 @@ def create_addon_package(
...
+                shutil.copy(src_path, str(dst_root))
+    else:
+        package_py = addon_output_dir / "package.py"
+
+        with open(package_py, "w+") as pkg_py:
+            pkg_py.write(
+                f"""name = "{addon_dir.name}"
+version = "{addon_version}"
+plugin_for = ["ayon_server"]
+build_command = "python {{root}}/rezbuild.py"
```
iLLiCiTiT, line 229: What is this for?

Minkiu: It's for when running the installation of the package on the server side, it will run the [`rezbuild.py` in the backend](https://github.com/ynput/ayon-backend/pull/29/files#diff-3971b40e9cd7b57b0924840af2ddb6af6c487eb73376805823ffe2ec6e517666)

Which as far as I recall, would be the default for simple add-ons.

Initially tried to add this field on the fly if it was missing, but there's a but in rez which won't allow to do such thing.

iLLiCiTiT: In that case I don't think it is necessary
```suggestion
```

iLLiCiTiT:
Modified the code. Now it always create new structure, and removed manifest.json. Also fixed issue in applications addon caused by new structure.

iLLiCiTiT (approved):

PR #66, opened 2024-02-15
Max: Bug fix the "remove items" not working in Scene Inventory Manager

moonyuet:
## Changelog Description
When the users load the objects through loaders and they want to remove them in scene inventory manager, the function doesn't work as expected to remove all data. This PR is to ensure the removal of the objects by using remove items in scene inventory manager. 

## Additional info
n/a

## Testing notes:
1. Launch Max with AYON
2. Load objects in any family
3. Go to Manage... to access Scene Inventory
4. Right Click to click "Remove items"
image

5. The objects and the data would be deleted as expected.


client/ayon_core/hosts/max/api/pipeline.py
```diff
@@ -242,3 +242,20 @@ def get_previous_loaded_object(container: str):
         if str(obj) in sel_list:
             node_list.append(obj)
     return node_list
+
+
+def remove_container_data(container: str):
```
iLLiCiTiT, line 247: I think it is `node_name` instead of `container`, or?
```suggestion
def remove_container_data(node_name: str):
```

moonyuet: maybe `container_node` is better as it is removing the container node which stores the container's data for upversioning and switching assets in scene inventory

iLLiCiTiT (approved):
Code changes look ok, needs testing.

LiborBatek (commented):
It seems to work for max assets (max, abc representations) but fails when loading Abc products comming from different DCC like maya for some reason.

It generally removes the ```asset``` from ```Scene Inventory``` and also the ```asset container``` but not remove the actual ```geometries```.

Not sure if desired aka should work either?

Here is the outcome of ```Remove``` asset action :
image

LiborBatek (changes requested):
Hmm, double checked my bundles and core addon etc but cant make it working...

Neither native max assets works nor outside from other DCC host...assets not being removed after triggering ```Manage``` and ```Remove``` action.

Check my vid for details...

https://github.com/ynput/ayon-core/assets/112623825/b3ddc414-d0d9-4084-a9e7-2f5f55590ec3

moonyuet:
@LiborBatek my bad, I guess I doesn't manage the branch well which leads to the implementation of `remove_container_data` goes missing in each loader. It should be working now.

LiborBatek (approved):
Now all working fine!

Note: would be lovely to perform ```viewport update``` after that ```Remove``` so the asset(s) which being removed wont be present still in the viewport as they are now before triggering any viewport action by the user causing slight confusion right away.

LiborBatek:
All working as expected and without any glitch!

moonyuet:
@iLLiCiTiT @antirotor can you guys take a final glance before merging? Thank you!

PR #69, opened 2024-02-15
Hiero: Use AYON settings

iLLiCiTiT:
## Changelog Description
Hiero can use AYON settings.

## Additional info
Modified pyblish pype to be able to use AYON structure of filters (hiero is last one which has the setting). LoadClip settings are handled in the plugin itself.

## Testing notes:
1. Create hiero and core packages.
2. Upload them to server.
3. Use them in bundle.
4. Start tray with the bundle and launch hiero.
5. Pyblish pype should not crash because of filter presets.
6. LoadClip should be working as before this PR.


jakubjezek001 (approved):
works as expected

PR #70, opened 2024-02-15
Nuke: Use AYON settings

iLLiCiTiT:
## Changelog Description
Nuke is using AYON settings.

## Additional info
Nuke had the most complicated conversions from all hosts. The most notable changes are knobs which completelly changed structure in AYON settings. Another big change was workfile imageio, where I did a little bit more changes.

## Testing notes:
1. Make sure you're using ayon-core from this branch.
2. Launch Nuke
3. Loaders should filter by filters in settings.
4. ImageIO for workfile should be set correctly.
5. Create Writes should create nodes correctly with all knobs.
6. Extract intermediates and slate should apply settings correctly.


jakubjezek001 (approved):
looks good and after my fix it even works ;)

PR #73, opened 2024-02-15
Tools: Simple folders widget

iLLiCiTiT:
## Changelog Description
implemented base of `SimpleFoldersWidget` as base of possible replacement of `AssetsWidget`.

## Additional info
The api of the widget is not as big as `AssetsWidget` had, and does not contain the default widgets around, like filter input and refresh button. But that should be in different widget and different PR. The implementation is reusing `FoldersWidget` but has implemented simplified controller which fills the values. More api methods can be added when needed.

I had this prepared for a while, but there is need to use it now so I've created quick PR.


simonebarbieri (approved):
It works beautifully!

PR #75, opened 2024-02-15
Bugfix: Houdini Mantra IFD local submission

MustafaJafar:
## Changelog Description
Mantra IFD extractor local submission was failing because its extractor didn't include `render_rop`

## Testing notes:
1. publish mantra ifd local and on farm. it should work.


MustafaJafar:
I have a question: 
I'm not sure about the workflow of using published `ifd`s ? 
Also, how `ifd`s are related to these temp paths ? e.g. what would happen if I deleted them after publishing the `ifd`s ? 
image

client/ayon_core/hosts/houdini/plugins/publish/extract_mantra_ifd.py
```diff
@@ -22,8 +22,11 @@ def process(self, instance):
 
         ropnode = hou.node(instance.data.get("instance_node"))
         output = ropnode.evalParm("soho_diskfile")
-        staging_dir = os.path.dirname(output)
+        staging_dir, file_name = os.path.split(output)
         instance.data["stagingDir"] = staging_dir
+        # render rop
+        self.log.debug("Writing IFD '{}' to '{}'".format(file_name, staging_dir))
+        render_rop(ropnode)
```
moonyuet, line 29: You need to clean up the lines from 32 to 41, as it errors out the extractor during publishing

moonyuet:
> I have a question: I'm not sure about the workflow of using published `ifd`s ? Also, how `ifd`s are related to these temp paths ? e.g. what would happen if I deleted them after publishing the `ifd`s ? image

This is just providing the temporary storage for ifd file if there is any. 


We need to double check the integrate.py and resourced_file.py to see if the related family of this extractor is included. It seems that it doesn''t get published correctly and it doesn't show up in the loader.

MustafaJafar:
Note: 
As far as I know the published IFD files will keep the link to the original generated temp data in the workfiles which can error if the temp data are deleted.

Thus, I've enabled `Save Geometry Inline` 
As I have no idea to publish the generated temp data along with the IFD files while updating the link between them.

image

moonyuet:
> Note: As far as I know the published IFD files will keep the link to the original generated temp data in the workfiles which can error if the temp data are deleted.
> 
> Thus, I've enabled `Save Geometry Inline` As I have no idea to publish the generated temp data along with the IFD files while updating the link between them.
> 
> image

maybe we can just add an optional validator to check for the temp to avoid the error with the temp file storage.

moonyuet (approved):

moonyuet:
I tested with houdini, all functions work.
 I think it may be good idea to add optional validator to check on the `Save Geometry Incline` button being turned on, but it can be in different PRs(if anyone needs that)

client/ayon_core/hosts/houdini/plugins/create/create_mantra_ifd.py
```diff
@@ -32,9 +32,13 @@ def create(self, product_name, instance_data, pre_create_data):
         parms = {
             # Render frame range
             "trange": 1,
-            # Arnold ROP settings
+            # Mantra ROP settings
             "soho_diskfile": filepath,
-            "soho_outputmode": 1
+            "soho_outputmode": 1,
+            # Enable `Save Geometery Inline`
```
BigRoy, line 38:
```suggestion
            # Enable `Save Geometry Inline`
```

client/ayon_core/hosts/houdini/plugins/create/create_mantra_ifd.py
```diff
@@ -32,9 +32,13 @@ def create(self, product_name, instance_data, pre_create_data):
...
+            # As it's hard to update the geo reference
+            # inside the IFD file after publishing the IFDs.
+            "vm_inlinestorage": 1
```
BigRoy, line 41: Containing all the geometry into the output file means you're basically bloating the filepath a lot. I really feel like this is the opposite of what a 'render' file should look like - it should be a lightweight file whenever possible.

I understand the comment as to what is hard about changing paths in the output files - but I'm wondering, why would we need changing anyway? Is it because of mixed farms, etc?

Also, if this is a "local" submission - do we even need the IFD file? Or what's this about?

Aren't IFD's intermediate files just used for rendering? They don't need publishing - they are not proxies. They are just files used for the current render. If it's a local render - then technically it can almost live in memory or transparently be in a temp folder on the local machine?

client/ayon_core/hosts/houdini/plugins/publish/extract_mantra_ifd.py
```diff
@@ -22,20 +22,13 @@ def process(self, instance):
...
-        missing_frames = [
-            frame
-            for frame in instance.data["frames"]
-            if not os.path.exists(
-                os.path.normpath(os.path.join(staging_dir, frame)))
-        ]
-        if missing_frames:
-            raise RuntimeError("Failed to complete Mantra ifd extraction. "
-                               "Missing output files: {}".format(
-                                   missing_frames))
```
BigRoy, line 38: Note that a `render_rop` can't always accurately detect whether it failed or whether a user cancelled. A user cancelling the render _may_ just exit the function without an error making it look like it finished. This "check" logic here was once intended to cover some very simple 'check cases' that if the files weren't there we were pretty sure the full render did not succeed.

I think it happened on certain ROPs when cancelling super early or JUST as it went from rendering one frame to another. But still, you wouldn't want the publisher to think it succeeded.

I had reported it to SideFX at the time and I haven't seen any report as of yet that they fixed that.

client/ayon_core/hosts/houdini/plugins/create/create_mantra_ifd.py
MustafaJafar: > why would we need changing anyway? Is it because of mixed farms, etc?

As far as I know it's not only about mixed farms but also about having the IFD files with the correct data. 
and keeping that link may produce IFD files that points to a wrong/non-existing geometry files. 

Also, I doubt this problem is going to occur also with farm IFD publishing. 

> Aren't IFD's intermediate files just used for rendering? They don't need publishing - they are not proxies. They are just files used for the current render.

I'm not sure about the `why we have IFD family` I can obly remember it was added by @moonyuet in https://github.com/ynput/OpenPype/pull/4903 
Could @moonyuet  you shed some light ?

moonyuet: > Containing all the geometry into the output file means you're basically bloating the filepath a lot. I really feel like this is the opposite of what a 'render' file should look like - it should be a lightweight file whenever possible.
> 
> I understand the comment as to what is hard about changing paths in the output files - but I'm wondering, why would we need changing anyway? Is it because of mixed farms, etc?
> 
> Also, if this is a "local" submission - do we even need the IFD file? Or what's this about?
> 
> Aren't IFD's intermediate files just used for rendering? They don't need publishing - they are not proxies. They are just files used for the current render. If it's a local render - then technically it can almost live in memory or transparently be in a temp folder on the local machine?

We can just remove the entire product type instead. i mean same thing has been raised by @fabiaserra before

client/ayon_core/hosts/houdini/plugins/publish/extract_mantra_ifd.py
MustafaJafar: Thanks for informing me! 
I've also furtherly checked it and found 2 things. 
1. If a user interrupts a render, this error will be raised  https://github.com/ynput/ayon-core/blob/7b2dc2676f04cd845669dcad9007b1f1e115728e/client/ayon_core/hosts/houdini/api/lib.py#L253
2. I add a try before `render_rop(ropnode)` in `extract_mantra_ifd` in order to see missing frames logic runs. it works.


Both errors are `RuntimeError` and they don't look good in the UI. 
I tried raising `PublishValidationError` instead but it doesn't work in the same manner as in validations. 

A question may be for @iLLiCiTiT : Do we have a generic `PublishError` that can be used in the same fashion as validation errors ? 

image

BigRoy:
> has there been any progress or plans on separating the extraction/render of nodes from the publish?

As far as I know, there hasn't no. Might be one for Dee Coureau (I dont know his Github handle).

fabiaserra:
> > has there been any progress or plans on separating the extraction/render of nodes from the publish?
> 
> As far as I know, there hasn't no. Might be one for Dee Coureau (I dont know his Github handle).

I'm waiting for the 0.3.0 breaking release (this week?) to update my fork and migrate to `ayon-core` and test any new features before I start being loud in the DCC fronts again, right now I'm busy making the Shotgrid sync stable

client/ayon_core/hosts/houdini/plugins/publish/extract_mantra_ifd.py
MustafaJafar: Ok, I did some experiment here.
It's outside the scope of this PR but this PR was a good example to do the experiment.
This one uses `KnownPublishError` 
image

tbh, I was overwhelmed I expected to have different `report_label`
e.g. 
> I did some ugly modifications in order to have this screenshot. by setting the `ReportsWidget.CrashWidget.report_label` with the `publish_error_msg` 
image

But, Truly I expected some general `PublishError` like `PublishValidationError` I did one modification to create the following screenshot (I removed that line and then used `PublishValidationError` as usual) . 

image

MustafaJafar: I'm going to create a GH issue.. It would be a better place for such a discussion.

BigRoy: Well - it's been requested quite frequently to improve the error view for anything that's not a validation report message. It's "too dumbed down" usually. Also a point @fabiaserra I believe has brought up quite a bit - and some others on discord as well. So thanks for tracking this as an issue. Would be good to include Dee Coureau into the discussion as well.

MustafaJafar:
> > has there been any progress or plans on separating the extraction/render of nodes from the publish?
> 
> As far as I know, there hasn't no. Might be one for Dee Coureau (I dont know his Github handle).

it's @dee-ynput

MustafaJafar:
Marking this PR as a draft in favor of this comment https://github.com/ynput/ayon-core/pull/75#discussion_r1547849161 
waiting till making a decision whether to remove the IFD product type or not.

antirotor:
I don't thing that we want to publish IFDs at all. Let's remove it in another PR.

PR #77, opened 2024-02-16
RoyalRender: Use AYON settings

iLLiCiTiT:
## Changelog Description
RoyalRender is using AYON settings.

## Additional info
RoyalRender is now addon inheriting from `AYONAddon`. Modified how `"rr_paths"` are received from settings. Commented out `api` property which could not work because it was passing in studio settings, but the class expects rr root.

## Testing notes:
Not sure how to test this PR.


kalisp (approved):
I cleaned up a querying of RR root.

There will be need to merge https://github.com/ynput/OpenPype/pull/6160 here eventually and that should make everything working in Ayon.

PR #82, opened 2024-02-16
Chore: Move usdlib to pipeline

iLLiCiTiT:
## Changelog Description
Moved `usdlib.py` from `ayon_core/lib/` to `ayon_core/pipeline/`.

## Additional info
I'm not sure if the lib file is usefull in any way for AYON. It contains `parse_avalon_uri` which probably should not be available at all?
Reason to move is the file is dependent on `ayon_core.pipeline`, but `ayon_core.lib` should not import anything from there.

## Testing notes:
Collectors `CollectInstancesUsdLayered` and `CollectUsdBootstrap` should work as they did.


BigRoy:
So technically this file could move to `ayon-usd` if ALL usd related plug-ins for all hosts would end up relying on that AYON usd addon.

I'm currently using `usdlib.py` a lot, across Maya, Houdini and Blender. Basically it has the unified end point to asset/shot contributions publishing.

It could be that current state of `usdlib` in core is outdated but this PR https://github.com/ynput/OpenPype/pull/5925/files originally implemented a lot of what I'm using - and I have made more tweaks since.

Regarding `parse_avalon_uri` - I need something like that during publishing to investigate the contents of an AYON entity URI so I do need to access the functionality. Where it will reside I'm not sure - what's a good place? Separate it into a `ayon_core.lib.ayon_uri`?

Similarly I also have `construct_ayon_uri` from its components, currently on my end also in `usdlib` which should likely need to move somewhere else then as well.

What's a good location?

> Reason to move is the file is dependent on `ayon_core.pipeline`, but `ayon_core.lib` should not import anything from there.

So, to fully understand:

Anything in **.lib** should **never** import from **.pipeline**? Good to know, thanks.

Note that `ayon_core.lib.applications` seems to also do so, so might need fixing as well? (Although those imports occur from within the functions)

---

In my version of `openpype.lib.usdlib` the imports are:
```python
from openpype.client import (
    get_asset_by_name,
    get_subset_by_name,
    get_representation_by_name,
    get_hero_version_by_subset_id,
    get_version_by_name,
    get_last_version_by_subset_id
)
from openpype.pipeline import (
    get_representation_path
)
```

So only `get_representation_path` would be 'at fault' here? 

The need for it has to do with querying the resulting path for an AYON URI (including paths for publishes that are not yet integrated yet, but by also allowing to query the current publishing context and its instances) so that during publishing you can get the destination filepath for an AYON URI after the publish completed. Like an "assumed destination".

iLLiCiTiT:
I think that none of that would be working in AYON. Asset name is not unique and requires path. I think we have already determined the uri for our resolvers. But I'm not the person to answer.

> In my version of openpype.lib.usdlib the imports are:

Anything in `openpype.lib` should not touch `openpype.pipeline`, including `applications.py` which will be converted to addon.

BigRoy:
> I think that none of that would be working in AYON. Asset name is not unique and requires path. I think we have already determined the uri for our resolvers. And I guess all of that should be in ayon-used, as you mentioned. But I'm not the person to answer @antirotor might know.

- **Asset name is not unique and requires path**

That's totally fine - the AYON URI should still be correct and parseable, etc. I would just need to match with `folderPath` instead of `asset`. The logic transfers, it's just a different attribute.

- **I think we have already determined the uri for our resolvers.**

I personally think we should not consider the AYON entity URIs to be USD-related only. They are just as usable and relevant outside of USD.

> Anything in openpype.lib should not touch openpype.pipeline at all, including applications.py which will be converted to addon.


Fine with me. :) As long I can make it available. So each addon, like `ayon_usd` could expose such a `pipeline` module and would adhere to the same rules as `lib` versus `pipeline` here? Or?

---

Thanks for the quick reply!

iLLiCiTiT:
> I personally think we should not consider the AYON entity URIs to be USD-related only. They are just as usable and relevant outside of USD.

AYON entity uri is a full definition of a context, so you're right. But what is in `usdlib.py` is something that is USD related and I'm not the person to tell how much it will stay untouched as I know nothing about it at this moment. Also the uri will be connected to a schema, the schema will be related to it's usage, and `ayon://` would be too generic.

> Fine with me. :) As long I can make it available. So each addon, like ayon_usd could expose such a pipeline module and would adhere to the same rules as lib versus pipeline here? Or?

I don't think it is doesn't make sense to make such a rule, it always really matters on usage of the addon. Core has `lib`, `settings`, `addon`, `host` and `pipeline`. Whereas `pipeline` and `host` are directly connected to work with the pipeline part so might require some dependency import. And `lib`, `settings` and `addon` should be possible to import with only minimum requirements. And all of that should avoid circular import. But most of addons have usually much simpler structure. e.g. hosts have only `addon.py` and the pipeline part is, currently, in `api` folder. But the `api` subfolder is also not requirement anymore.

iLLiCiTiT:
> Do you mean it should be like ayon+entity://? I believe I heard somewhere from @mkolar that ayon:// would be, as default, resolve as ayon+entity://. Or did you mean something else?

So many questions I don't have answer to 😄

PR #83, opened 2024-02-16
Chore: Remove local settings function

iLLiCiTiT:
## Changelog Description
Removed `get_local_settings` function from `ayon_core.settings`.

## Additional info
Local settings are not implemented in AYON the same way as were in OpenPype. Some of the local settings from OpenPype are at this moment not available in AYON at all, and other ones are stored per addon. It is not possible to easily receive all the values as it was in OpenPype, at least not now, and we have to make full replacement of them in future.

## Testing notes:
There is no visible change.

### DIFF
--- a/client/ayon_core/settings/lib.py
+++ b/client/ayon_core/settings/lib.py
@@ -198,8 +198,8 @@ def merge_overrides(source_dict, override_dict):
 def get_site_local_overrides(project_name, site_name, local_settings=None):
     """Site overrides from local settings for passet project and site name.
 
-    Warning:
-        This function is not implemented for AYON.
+    Deprecated:
+        This function is not implemented for AYON and will be removed.
 
     Args:
         project_name (str): For which project are overrides.
--- a/client/ayon_core/settings/lib.py
+++ b/client/ayon_core/settings/lib.py
@@ -48,11 +48,6 @@ def clear_metadata_from_settings(values):
             clear_metadata_from_settings(item)
 
 
-def get_local_settings():
-    # TODO implement ayon implementation
-    return {}
-
-
 def load_openpype_default_settings():
     """Load openpype default settings."""
     return load_jsons_from_dir(DEFAULTS_DIR)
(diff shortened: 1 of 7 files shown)


client/ayon_core/settings/lib.py
```diff
@@ -203,39 +198,17 @@ def merge_overrides(source_dict, override_dict):
...
-    # First load and use local settings from default project
-    if default_project_locals and site_name in default_project_locals:
-        output.update(default_project_locals[site_name])
 
-    # Apply project specific local settings if there are any
-    if project_locals and site_name in project_locals:
-        output.update(project_locals[site_name])
-
-    return output
+    return {}
```
tokejepsen, line 211: Should we have some warning here about the local settings not being available yet in AYON?

tokejepsen:
@jakubjezek001 any code we can run for testing?

iLLiCiTiT:
> @jakubjezek001 any code we can run for testing?

I guess you meant me. Nothing technically changed, except of removed functions.

client/ayon_core/settings/lib.py
iLLiCiTiT: There are TODOs all over the place, the warning would just scare artists about something they can't affect.

tokejepsen: Ahh, maybe a debug logging for developers then?

iLLiCiTiT: I really don't think so. For last 3 years it was used on 1 place, and I don't think someone would start to use a function with `Warning:` in docstring and empty implementation.

kalisp:
Not sure if Todos are enough, I created issue for dirmap problem, https://github.com/ynput/ayon-core/issues/95. More issues should be created to follow them if necessary.

tokejepsen (approved):
Code looks good to me.

iLLiCiTiT:
> Not sure if Todos are enough, I created issue for dirmap problem, https://github.com/ynput/ayon-core/issues/95. More issues should be created to follow them if necessary.

I've already created tasks for that.

PR #85, opened 2024-02-16
Nuke: openpype expose knobs validator - OP-8166

tokejepsen:
## Changelog Description
Port of https://github.com/ynput/OpenPype/pull/6213

### DIFF
--- a/client/ayon_core/hosts/nuke/plugins/publish/validate_exposed_knobs.py
+++ b/client/ayon_core/hosts/nuke/plugins/publish/validate_exposed_knobs.py
@@ -65,7 +65,7 @@ class ValidateExposedKnobs(
         group_node = instance.data["transientData"]["node"]
         nuke_settings = instance.context.data["project_settings"]["nuke"]
         create_settings = nuke_settings["create"][plugin]
-        exposed_knobs = create_settings["exposed_knobs"]
+        exposed_knobs = create_settings.get("exposed_knobs", [])
         unexposed_knobs = []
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
I don't think this is needed in ayon-core.

tokejepsen:
> I don't think this is needed in ayon-core. Settings overrides do not work the same way as in OpenPype.

So if an existing production is using ayon now without the `Exposed Knobs` setting, what would happen?

iLLiCiTiT:
> So if an existing production is using ayon now without the Exposed Knobs setting, what would happen?

In AYON you'll always get current settings model values. And if the settings model has the attribute, you'll get the value, in other words core version `1.1.1` will always get settings values compatible with model `1.1.1`.

NOTE: That might not be true in case of addons which have settings in `server_addons`, because client code lives in core and does not have to be in sync on server with settings.

tokejepsen:
> NOTE: That might not be true in case of addons which have settings in server_addons, because client code lives in core and does not have to be in sync on server with settings.

So we do have to merge this PR?
Dont mind whether we do or not, just need a decision. If its a problem, then its a quick and easy fix later.

iLLiCiTiT:
> Dont mind whether we do or not, just need a decision. If its a problem, then its a quick and easy fix later.

Huge sorry, I just realized the change is in nuke and not in core settings. Yes we should merge this.

iLLiCiTiT (approved):

PR #89, opened 2024-02-19
Nuke: LoadClip colorspace override - OP-6591

tokejepsen:
## Changelog Description
Port of https://github.com/ynput/OpenPype/pull/6215


client/ayon_core/hosts/nuke/plugins/load/load_clip.py
```diff
@@ -450,25 +473,49 @@ def _get_node_name(self, representation):
...
+        # Get colorspace from representation colorspaceData if colorspace is
+        # not found.
+        colorspace_data = representation_doc["data"].get("colorspaceData", {})
+        colorspace = colorspace or colorspace_data.get("colorspace")
+        self.log.debug(
+            f"Colorspace from representation colorspaceData: {colorspace}"
+        )
+
+        print(f"Colorspace found: {colorspace}")
+
```
tokejepsen, line 506:
```suggestion
```

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -103,33 +103,42 @@ def colorspace_exists_on_node(node, colorspace_name):
...
+    return colorspace_name in get_colorspace_list(colorspace_knob)
 
 
 def get_colorspace_list(colorspace_knob):
     """Get available colorspace profile names
 
+    Because the values returned from colorspace_knob.values() do not correspond
+    to the value returned from colorspace_knob.value(), and the extracted
+    colorspace comes from using colorspace_knob.value(), we need to iterate
+    through all values to get the correct value.
```
BigRoy, line 116: Oh boy - this really sounds like something that needs reporting to Nuke! :)

BigRoy: Or does [`nuke.colorspaces.getColorspaceList(colorspace_knob)` work](https://learn.foundry.com/nuke/developers/100/pythonreference/nuke.colorspaces-module.html#getColorspaceList)?

tokejepsen: > What's the Knob class type? Is it an Array_Knob?
Its an `Enumeration_Knob`

> Does .array() work?
```
node = nuke.selectedNode()
print(node["colorspace"].array())
# Result: [0.0]
```
Does not work unfortunately.

> Or does [nuke.colorspaces.getColorspaceList(colorspace_knob) work](https://learn.foundry.com/nuke/developers/100/pythonreference/nuke.colorspaces-module.html#getColorspaceList)?

It looks like its the roles that are messing up the returned values.
```
import nuke


def get_colorspace_list(colorspace_knob):
    """Get available colorspace profile names

    Because the values returned from colorspace_knob.values() do not correspond
    to the value returned from colorspace_knob.value(), and the extracted
    colorspace comes from using colorspace_knob.value(), we need to iterate
    through all values to get the correct value.

    A code example of the above would be:

    for count, value in enumerate(colorspace_knob.values()):
        colorspace_knob.setValue(count)
        print(colorspace_knob.value() in colorspace_knob.values())

    Args:
        colorspace_knob (nuke.Knob): nuke knob object

    Returns:
        list: list of strings names of profiles
    """
    original_value = colorspace_knob.value()

    colorspaces = []

    try:
        for count, _ in enumerate(colorspace_knob.values()):
            colorspace_knob.setValue(count)
            colorspaces.append(colorspace_knob.value())
    finally:
        colorspace_knob.setValue(original_value)

    return colorspaces


node = nuke.selectedNode()
custom_colorspace_list = get_colorspace_list(node["colorspace"])
nuke_colorspace_list = nuke.getColorspaceList(node["colorspace"])
for count, value in enumerate(custom_colorspace_list):
    nuke_value = nuke_colorspace_list[count]
    if value != nuke_value:
        print(value)
        print(nuke_value)
```
```
get_colorspace_list:  color_picking
nuke.getColorspaceList:  color_picking	color_picking (Output - sRGB)
get_colorspace_list:  color_timing
nuke.getColorspaceList:  color_timing	color_timing (ACES - ACEScc)
get_colorspace_list:  compositing_linear
nuke.getColorspaceList:  compositing_linear	compositing_linear (ACES - ACEScg)
get_colorspace_list:  compositing_log
nuke.getColorspaceList:  compositing_log	compositing_log (Input - ADX - ADX10)
get_colorspace_list:  data
nuke.getColorspaceList:  data	data (Utility - Raw)
get_colorspace_list:  default
nuke.getColorspaceList:  default	default (ACES - ACES2065-1)
get_colorspace_list:  matte_paint
nuke.getColorspaceList:  matte_paint	matte_paint (Utility - sRGB - Texture)
get_colorspace_list:  reference
nuke.getColorspaceList:  reference	reference (Utility - Raw)
get_colorspace_list:  rendering
nuke.getColorspaceList:  rendering	rendering (ACES - ACEScg)
get_colorspace_list:  scene_linear
nuke.getColorspaceList:  scene_linear	scene_linear (ACES - ACEScg)
get_colorspace_list:  texture_paint
nuke.getColorspaceList:  texture_paint	texture_paint (ACES - ACEScc)
```
@jakubjezek001 got any idea why the roles would do this?

tokejepsen: This could be an alternative without the brute forcing value iteration, but there are a few assumptions in this method:
```
def get_colorspace_list_2(colorspace_knob):
    """Get available colorspace profile names

    Args:
        colorspace_knob (nuke.Knob): nuke knob object

    Returns:
        list: list of strings names of profiles
    """
    results = []
    pattern = r".*\t.* \(.*\)"
    for colorspace in nuke.getColorspaceList(colorspace_knob):
        # Regex of pattern matches roles
        match = re.search(pattern, colorspace)
        if match:
            results.append(colorspace.split("\t")[0])
        else:
            results.append(colorspace)

    return results
```
The assumptions being that the value returned from roles such as `texture_paint	texture_paint (ACES - ACEScc)` will be the first string before the tab.

jakubjezek001: > Or does nuke.colorspaces.getColorspaceList(colorspace_knob) work?

I do agree with @BigRoy that it could be the fastest solution, even roles are adding some unnecessary text in items we could still search in its substrings.

So ideally rewrite the `ayon_core.hosts.nuke.api.utils.get_colorspace_list()` so it is rather wrapper of `nuke.colorspaces.getColorspaceList(colorspace_knob)` or even better just ditch it since there might not be any benefit from wrapper. 

Then in the function `ayon_core.hosts.nuke.api.utils.colorspace_exists_on_node()` refactor the check to be checking a substring of in the colorspaces list.  `any("scene_linear" in cl for cl in colorspaces)` will do and it will be much faster.

tokejepsen: Have used the code I proposed.

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -117,19 +118,22 @@ def get_colorspace_list(colorspace_knob):
...
+    # This pattern is to match with roles which uses an indentation and
+    # parentheses with original colorspace. The value returned from the
+    # colorspace is the string before the indentation, so we'll need to
+    # convert the values to match with value returned from the knob,
+    # ei. knob.value().
+    pattern = r".*\t.* \(.*\)"
+    for colorspace in nuke.getColorspaceList(colorspace_knob):
+        match = re.search(pattern, colorspace)
+        if match:
+            results.append(colorspace.split("\t")[0])
```
BigRoy, line 132:
```suggestion
            results.append(colorspace.split("\t", 1)[0])
```
This should be faster since it can stop scanning the string after the first tab found.

I also wonder whether the regex matching is really needed to begin with - not a huge issue but does leave me wondering. Since you're matching anyway you could also do:

```python
import re


names = [
    # Non tabbed
    "mycolorspace",
    "ACES - ACEScg",
    # Tabbed
    "compositing_log	compositing_log (Input - ADX - ADX10)",
    "data	data (Utility - Raw)",
    "default	default (ACES - ACES2065-1)",
    "matte_paint	matte_paint (Utility - sRGB - Texture)",
    "reference	reference (Utility - Raw)",
    "rendering	rendering (ACES - ACEScg)",
    "scene_linear	scene_linear (ACES - ACEScg)",
    "texture_paint	texture_paint (ACES - ACEScc)"
]


for colorspace in names:
    match = re.match(r"(.*)\t.* \(.*\)", colorspace)
    if match:
        colorspace = match.group(1)
    print(colorspace)
```

Also, note that regex is greedy in this case so MIGHT match multiple tabs with the first `.*`. Not a big problem, just wanted to note that.

tokejepsen: > Since you're matching anyway you could also do:

Any benefit over the existing code?

BigRoy: > Any benefit over the existing code?

Speed, the match speed of the regex will be roughly equivalent - and thus you're not doing the extra string parsing for `.split()`

Plus it's clearer that you're matching the regex to ensure you get the right part - now the mixture of matching and then splitting might mean that you're not matching and splitting on the same point of the string.

In this case a string with more than one tab will be matched to the last tab found in the string, whereas the `split` call wouldn't care.

Any not too big of an issue - I guess it was me just trying to micro-optimize and be explicit about the behavior.

jakubjezek001 (approved):
tested and works well

PR #92, opened 2024-02-19
Resolve: avoiding clips without mediapoolitems

jakubjezek001:
## Changelog Description
When encountering a scenario where resolve timelineitems are not connected to any media pool items, it is possible that these items could be either an *Adjustment Clip* or a *Fusion Composition*.

## Testing steps
1. add to your timeline  `Adjustment Clip` (from effect panel)
2. open Loader and try to load anything 
3. no error should happen and clip should be loaded to timeline


jakubjezek001:
Task linked: [OP-8273 Resolve: missing GetClipProperty](https://app.clickup.com/t/6658547/OP-8273)

BigRoy:
Can we get `Resolve:` in the PR title as well - as long as the multiple hosts/applications are in core that can help changelogs and/or e-mail notifications since "labels" don't usually appear there?

BigRoy:
> Asserts creation of media pool item when not found for specified files. This ensures proper handling and error messaging in such cases.

Could you include testing notes?

- When does this occur?
- Is there something the user can do about this? Like, can the error message also describe WHY it's wrong or what the user can do about it?
- As far as I'm aware when using `assert` you don't need to make the value after the `,` a `AssertionError` instance. If it's just a string then Python will raise that string as `AssertionError` automatically. Right?
- Instead of an `assert` should this be `LoadError`?

jakubjezek001:
Thanks to @BigRoy there was found the reason why this error can happen. Thanks man!

client/ayon_core/hosts/resolve/api/lib.py
```diff
@@ -409,6 +409,9 @@ def get_current_timeline_items(
         }
         # get track item object and its color
         for clip_index, ti in enumerate(_clips[track_index]):
+            if not ti.GetMediaPoolItem():
```
BigRoy, line 412: Should we add a comment as to when this can occur?

For example:
```python

            # Skip items that do not have a media pool item, like for example
            # an "Adjustment Clip" or a "Fusion Composition" from the effects
            # toolbox
```

Also, is this the right place to fix this?

From the function name I'd expect this entry to be returned since it **is** a timeline item - it just doesn't have a mediapool item equivalent. I fixed it myself in `get_timeline_item` along the lines of:
```python
def get_timeline_item(media_pool_item: object,
                      timeline: object = None) -> object:
    """ ... """
    clip_name = media_pool_item.GetClipProperty("File Name")
    output_timeline_item = None
    timeline = timeline or get_current_timeline()

    with maintain_current_timeline(timeline):
        # search the timeline for the added clip

        for ti_data in get_current_timeline_items():
            ti_clip_item = ti_data["clip"]["item"]
            ti_media_pool_item = ti_clip_item.GetMediaPoolItem()

            # Skip items that do not have a media pool item, like for example
            # an "Adjustment Clip" or a "Fusion Composition" from the effects
            # toolbox
            if not ti_media_pool_item:
                continue

            if clip_name in ti_media_pool_item.GetClipProperty("File Name"):
                output_timeline_item = ti_clip_item

    return output_timeline_item
```

Since that particular function **does** try to get a matching timeline item to a media pool item.


Also, the logic in that function is much better of directly returning the clip in the if statement instead of storing it in a variable and continuing the loop. Since we'll always be returning one. If we want to preserve the current behavior where we return the LAST one found in the iterations we can do `for ti_data in reversed(get_current_timeline_items()):` just so that the first one we find is the last one returned by that other function.
I'm not sure however if the order things are being returned in there is in any particular order we care about or not?

Also, it seems pretty odd to match a timeline to a media pool item even though it may appear more than once on the timeline. I guess that's the current behavior and there's no bug _yet_ due to that? But it's definitely waiting for odd stuff to occur if we rely on there ever being only one entry present on the timeline of a media clip item. No?

tokejepsen (changes requested):
Lets address @BigRoy comments.

client/ayon_core/hosts/resolve/api/lib.py
jakubjezek001: good point! thanks for suggestion. I had implemented it your way ;)

client/ayon_core/hosts/resolve/api/plugin.py
```diff
@@ -405,6 +406,11 @@ def load(self, files):
             files,
             self.active_bin
         )
+
+        assert media_pool_item, LoadError(
+            "Cannot create media pool item for files: `{}`".format(files)
+        )
+
```
jakubjezek001, line 413:
```suggestion
```

client/ayon_core/hosts/resolve/api/plugin.py
```diff
@@ -11,6 +11,7 @@
     LoaderPlugin,
     Anatomy
 )
+from ayon_core.pipeline.load import LoadError
```
jakubjezek001, line 14:
```suggestion
```

client/ayon_core/hosts/resolve/api/lib.py
```diff
@@ -409,6 +415,9 @@ def get_current_timeline_items(
...
+                continue
+
```
jakubjezek001, line 420:
```suggestion
```

tokejepsen (approved):
Tested successfully in Resolve 18.6.

For completeness, the error in current develop is this when loading in a clip while having an adjustment clip on the timeline;
```
During load error happened on Product: "renderCompositingMain" Representation: "exr" Version: 1

Error message: 'NoneType' object has no attribute 'GetClipProperty'

  File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\loader\models\actions.py", line 763, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\Users\tokejepsen\ayon-core\client\ayon_core\pipeline\load\utils.py", line 323, in load_with_repre_context
    return loader.load(repre_context, name, namespace, options)
  File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\resolve\plugins\load\load_clip.py", line 48, in load
    timeline_item = plugin.ClipLoader(
  File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\resolve\api\plugin.py", line 462, in load
    timeline_item = lib.create_timeline_item(
  File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\resolve\api\lib.py", line 306, in create_timeline_item
    output_timeline_item = get_timeline_item(
  File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\resolve\api\lib.py", line 343, in get_timeline_item
    _ti_clip_property = _ti_clip.GetMediaPoolItem().GetClipProperty
AttributeError: 'NoneType' object has no attribute 'GetClipProperty'
```

PR #93, opened 2024-02-19
Maya: Remove shader definition

iLLiCiTiT:
## Changelog Description
Shader definition editor logic is based on MongoDB and gridfs. Because both are not available anymore, and the logic was not used in last 3 years (as far as I know) we can safely remove it.

## Additional info
I'm not sure if the remaining plugin `ValidateModelName` should stay. It is based on settings, which are not in AYON maya model. Can't tell if the logic makes sense without shader definition.

## Testing notes:
1. Maya publishing of model should work ( I guess )


BigRoy (approved):
Pretty sure this legacy design does not fit with the modern OpenPype and definitely not AYON system. If there's reasons to keep something like this alive I think there are better alternatives with settings, etc. in AYON this should be refactored to.

Since I'm personally not using these for me it's 👍 on just removing this. But it might very well be used by one of Ynput's clients - for which I believe this was originally implemented.

tokejepsen (approved):
Client is not with Ynput anymore.

Lets completely remove the plugin as well.

PR #94, opened 2024-02-19
Maya: Remove remote publish logic

iLLiCiTiT:
## Changelog Description
Removed remote maya publish plugins and scripts.

## Additional info
The remote publish logic for Maya was probably copied from colorbleed fork, but it is missing important pieces to be able to work. The implementation did work in older versions of OpenPype, but now it would not work even in current version OpenPype. The plugins, as far as I can tell, are not used. The script path defined in submit plugin would not work in AYON at all, and the execution of python script would not work either as that would require ayon-launcher executable. The implementation of remote publish script did not implement any install of plugins of targets, nor did the function `remote_publish` in publish utils. The function was removed too as it might be misleading.

## Testing notes:
1. Nothing should change


BigRoy (approved):
This has come up before I believe - and yes, pretty sure it does not work in OpenPype and AYON.

tokejepsen (approved):
Looks good to me.

PR #97, opened 2024-02-20
Fusion: Place footage without start-frame at global start

EmberLightVFX:
If footage doesn't have a start-frame set (eg if you add it from TrayPublisher) the footage now starts at Fusions global start instead of frame 0 (This is the way Fusion would do it)

This is a re-pr of an OpenPype pr: https://github.com/ynput/OpenPype/pull/4932

## Testing notes:
1. Import footage that doesn't have Frames assigned to it. It will now be placed at global start instead of 0


BigRoy:
Yup, sounds like a more sensible default. 👍

BigRoy:
> Maybe stupid question, but how can I produce such seq inside Ayon?! as it always have sort of frame range set even when originating from Tray Publisher.

I believe since a somewhat recent change this toggle was added in Tray Publisher:
image

Unfortunately when disabled it automatically enforces the asset frame range as metadata as opposed to no data. (Which is not clear from that toggle at all.)
I feel like that field would be much clearer if instead of a Toggle it would be a Dropdown, e.g. an option named "Frame Range:" with:
- Use sequence frame range
- Use asset entity frame range
- Use custom frame range
- Do not set frame range metadata (if this should really ever be an option)

I feel like that _when_ using asset or custom frame range it _must_ still be validated that the length of the sequence (if it's a sequence) actually matches the frame ranges that it collected, because otherwise we're basically ingesting invalid entities because the frame ranges are incorrect.

E.g. note how these two frames (sequence frame range) are ingested as if they are five frames (folder entity frame range)
image


Anyway, having said that. I believe thus now it's harder to produce publishes currently that may be a sequence but have no frame range data.

EmberLightVFX:
`Collect Original Sequence Frame Data` is new from when I first created this PR back for OP. I didn't see this new toggle so it kind of fixes the problem for newly published products.

```
I feel like that field would be much clearer if instead of a Toggle it would be a Dropdown, e.g. an option named "Frame Range:" with:

    Use sequence frame range
    Use asset entity frame range
    Use custom frame range
    Do not set frame range metadata (if this should really ever be an option)
```

I fully agree with this. I don't see any reason to be able to publish anything without a set frame range.
`Use custom frame range` could instead be `Use custom first frame number` or something like that so the publishing would automatically calculate the range from set first frame number and you won't get any problems in the validation.

This should be a new PR then and the non-optional sequence validator should be a global validator, so I'm closing this PR with that :)

PR #98, opened 2024-02-20
Chore: AYON metadata ids

iLLiCiTiT:
## Changelog Description
Added AYON variants of `AVALON_CONTAINER_ID` and `AVALON_INSTANCE_ID` as `AYON_CONTAINER_ID` and `AYON_INSTANCE_ID`.

## Additional info
Modified also code in hosts to be able to handle both option. Only where it was "easy" to find out how it's used. The AYON variants are not yet used as default. That should be handled in per host PR (in most of cases I'm not capable of doing it).

Old values:
- `AVALON_CONTAINER_ID` > `"pyblish.avalon.container"`
- `AVALON_INSTANCE_ID` > `"pyblish.avalon.instance"`

New values:
- `AYON_CONTAINER_ID` > `"ayon.load.container"`
- `AYON_INSTANCE_ID` > `"ayon.create.instance"`

## Testing notes:
I don't think it is necessary to test all changed files. But it would be nice to try one publish because of added imports.
1. Go through changes.


client/ayon_core/hosts/maya/api/pipeline.py
```diff
@@ -376,9 +377,12 @@ def _maya_iterate(iterator):
...
             iterator.next()
 
-    ids = {AVALON_CONTAINER_ID,
-           # Backwards compatibility
-           "pyblish.mindbender.container"}
+    ids = {
+        AYON_CONTAINER_ID,
+        # Backwards compatibility
+        AVALON_CONTAINER_ID,
+        "pyblish.mindbender.container"
```
BigRoy, line 384: This backwards compatibility is from the very early days of AVALON itself - not even OpenPype. I think it's time to remove it by now :)
```suggestion
        AVALON_CONTAINER_ID
```

BigRoy:
> Only where it was "easy" to find out how it's used. The AYON variants are not yet used as default. That should be handled in per host (and in most of cases I'm not capable of doing it).

Doesn't this PR only have _any_ value once that's actually occurring? Not necessarily by default, but at least having an env var or whatever to enable or disable that behavior?

There's hardly anything to test whether the new logic is working as intended until we actually have instances with the new ids.

MustafaJafar (commented):
I tested it in Houdini and the publisher only recognised the instances with `ayon.create.instance`. 

image

iLLiCiTiT:
> Doesn't this PR only have any value once that's actually occurring? Not necessarily by default, but at least having an env var or whatever to enable or disable that behavior?

Like is mentioned in description. I don't know for sure if the hosts are ready to change it to new value. The developer who is responsible for the host should do the PR to use only the new value. This PR is adding new constants that should be used instead.

> There's hardly anything to test whether the new logic is working as intended until we actually have instances with the new ids.

It is up to the host integration to decide if is fully ready to use new ids, not sure why env variable should be used? At this moment it is per-host decision.

MustafaJafar (approved):
Houdini works fine with old and new ID values.
Thanks.

---
One more question: 
Is it planned to update the schema ? (I don't know what it's used for tbh)
image

MustafaJafar:
There's a typo in the PR description: 
- `AVALON_CONTAINER_ID` > `"pyblish.avalon.container"`
- `AVALON_INSTANCE_ID` > `"pyblish.avalon.instance"`

moonyuet:
I tested the code with Max, looks good.

PR #102, opened 2024-02-21
Chore: stop removing addons via create package py 

jakubjezek001:
## Changelog Description
If you use `create_package.py` to make a core addon package, it goes straight to the `docker/addons` folder. This way, you don't have to stress about deleting any addons that are already there.

## Testing notes:
1. use following cmd and notice that addons in your ayon-docker/addon folder are not removed.

```
cd [YOUR ayon-core repo dir]
python ./create_package.py --skip-zip --output [ayon-docker repo folder path]/addons
```

### DIFF
--- a/create_package.py
+++ b/create_package.py
@@ -279,7 +279,8 @@ def create_server_package(
 def main(
     output_dir: Optional[str]=None,
     skip_zip: bool=False,
-    keep_sources: bool=False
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
tested several times in my side and works as expected. When I enter the command in the terminal, it will build the core addon into the `docker/addon` directory

image

PR #103, opened 2024-02-21
Chore: Use folder path in current context

iLLiCiTiT:
## Changelog Description
Use `folder_path` instead of `asset_name` in current context and context data.

## Additional info
This is minimalistic PR changing just the keys used when data for context are read or defined.

Breaks backwards compatibility on current context data and launch hooks. Launch hooks don't have access to `asset_name` in data.

## Testing notes:
1. Application can launch on a context.
2. Change of context with workfiles too triggers correct callbacks without issues.

### DIFF
--- a/client/ayon_core/cli_commands.py
+++ b/client/ayon_core/cli_commands.py
@@ -106,7 +106,7 @@ class Commands:
             context = get_global_context()
             env = get_app_environments_for_context(
                 context["project_name"],
-                context["asset_name"],
+                context["folder_path"],
                 context["task_name"],
                 app_full_name,
                 launch_type=LaunchTypes.farm_publish,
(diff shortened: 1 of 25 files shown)


tokejepsen (approved):
Tested in Nuke 15 successfully.

Could we maybe get some checkboxes for the hosts in the description?

MustafaJafar:
I faced two things: 
1) Publisher tab looks like this:
image

2) Sometimes context change works and sometimes it fails.
image

iLLiCiTiT:
> Sometimes context change works and sometimes it fails.

Any traceback? Is the fail related to how the context is changed?

jakubjezek001 (approved):
works in **nuke**

iLLiCiTiT:
For anyone reading this PR in future. The issues with publisher did not happen again, probably was caused by lost connection to server, and the context change issue in houdini is not related to this PR.

PR #112, opened 2024-02-23
Feature/ Zbrush integration

moonyuet:
## Changelog Description
***This is still in a heavy development, will add the testing notes after finishing the TO-DO list
- [x] Apps launched
- [x] GUI installation & environment variable
- [x] Websocket
- [x] Zscript to open AYON Plugins
- [x] Workfile loading implementation
- [x] Instance data
- [x] Context Data
- [x] Container Data (Null data is okay but needs to figure out how to get containers for updating)
- [x] Imprint
- [x] Create Workfile
- [x] Publish Workfile
- [x] Create Mesh 
- [x] Publish Mesh(Obj only)
- [x] Load Mesh
- [x] Publlsh PolyMesh
- [x] validate edit mode for model family
- [x] Zscript GUI loaded with AYON
- [x] context data in AYON
- [ ] documentation (https://github.com/ynput/ayon-documentation/pull/149)

## Additional info
- As Zbrush doesn't have any callback, It is advised to always to use workfile tools to save as or open the file so that the system won't get confused about the current file. 
- The code doesn't get any clean up yet
- The last workfile hook implementation may need some special execute on zscript for "autoloading the last working files" in zbrush
- We can improve the development on communicating with the server by writing some plugins in DLL format, the implementation can take references from the decoupling UI from Unreal and TVPaint
<s>- Enhancement on the loader in terms of upversioning, switching and removing asset</s> Do more research and ask Zbrush community on the possible way to rename and updating object.
- As we can't support use selection due to API limitation, users need to select by their own if they want to export the selected obejcts.

## Testing notes:
1. Launch Zbrush in AYON
2. The AYON plugins would be installed in Zbrush.
3. Testing all the tools
4. They should work mostly fine
5. Please use complex zbrush meshes and simple zbrush meshes to test on the model extractor, as we figure out somehow the files get transfer when the export isn't finished. And we need to figure out the best solution to be exported.

https://github.com/ynput/ayon-core/assets/64118225/99379e61-5527-43d6-9fbe-a3de470ee776

### Use Subdivision Level options
1. By default it is sets to 0, which means exporting the maximum Subdivision level of the mesh.


client/ayon_core/hosts/zbrush/hooks/create_zmenu_script.py
```diff
@@ -0,0 +1,88 @@
...
+[VarSet, arg, "--launcher"]
+[IButton,"Zplugin:AYON:Load","Loader",
+	[VarSet, loader, "loader_tool"]
+    [VarSet, q, [StrFromAsc, 34]]
+	[VarSet, cmd, [StrMerge, start, " ",#q, #q, "  ",#q, "{py}", #q]]
+    [VarSet, cmd, [StrMerge, cmd, " ", #addon, #q]]
+    [VarSet, cmd, [StrMerge, cmd, #q,  " ", #zbrush, #q]]
+    [VarSet, cmd, [StrMerge, cmd, #q,  " ", #zscript, #q]]
+    [VarSet, cmd, [StrMerge, cmd, #q,  " ", #arg, #q]]
+    [VarSet, cmd, [StrMerge, cmd, #q,  " ", #loader, #q]]
```
BigRoy, line 52: No idea what this really does but I wonder if this is just more readable if we do not do all this string merging in Zscript per variable but instead just do e.g.

```
[VarSet, cmd, "{py} addon zbrush run-with-script launcher loader_tool"]
[ShellExecute, cmd]
```

Or even:
```
[ShellExecute, "{py}  loader_tool"]
```

Or if we format with e.g.:
```
launch = f"{python_exe} addon zbrush run-with-script launcher "
```

We can just do:
```python
[ShellExecute, "{launch} loader_tool"]
```
If at the end we format the whole string using `launch` variable.
Again, I know nothing of Zscript so there might be reasons not to do this.
Just felt it was greatly unreadable.

BigRoy: Would make it something like this:

```python
        launch = f"{python_exe} addon zbrush run-with-script launcher "
        ayon_script = ("""
[ISubPalette,"Zplugin:AYON"]
[IButton,"Zplugin:AYON:Load","Loader",
	[ShellExecute, "{launch} loader_tool"], 0, 120
]
[IButton, "Zplugin:AYON:Publish", "Publish Tab for Publisher",
	[ShellExecute, "{launch} publish_tool"], 0, 120
]
[IButton, "Zplugin:AYON:Manage", "Scene Inventory Manager",
	[ShellExecute, "{launch} scene_inventory_tool"], 0, 120
]
[IButton, "Zplugin:AYON:Workfile", "Workfile",
	[ShellExecute, "{launch} workfiles_tool"], 0, 120
]""").format(py=python_exe)
```

No idea whether that's valid Zscript though 🗡️

moonyuet: we can try but i agree it is ugly to read.

client/ayon_core/hosts/zbrush/hooks/create_zmenu_script.py
```diff
@@ -0,0 +1,88 @@
...
+        zscript_txt = os.path.join(zscript_path, "ayon_zbrush_menu.txt")
+        with open(zscript_txt, "w") as zscript:
+            zscript.write(self.ayon_menu())
+            zscript.close()
+        self.launch_context.env["AYON_ZBRUSH_CMD"] = os.environ["AYON_ZBRUSH_CMD"]
+    def ayon_menu(self):
+        zclient_path = os.path.join(ZBRUSH_HOST_DIR, "api", "widgets.py")
+        zclient_path = zclient_path.replace("\\", "/")
+        os.environ["AYON_ZBRUSH_CMD"] = subprocess.list2cmdline(
+            get_ayon_launcher_args()
```
BigRoy, line 35: Please check you linter

```suggestion

        zscript_path = os.path.join(ZBRUSH_HOST_DIR, "api", "zscripts")
        os.makedirs(zscript_path, exist_ok=True)
        zscript_txt = os.path.join(zscript_path, "ayon_zbrush_menu.txt")
        with open(zscript_txt, "w") as zscript:
            zscript.write(self.ayon_menu())
            zscript.close()
        self.launch_context.env["AYON_ZBRUSH_CMD"] = os.environ["AYON_ZBRUSH_CMD"]

    def ayon_menu(self):
        zclient_path = os.path.join(ZBRUSH_HOST_DIR, "api", "widgets.py")
        zclient_path = zclient_path.replace("\\", "/")
        os.environ["AYON_ZBRUSH_CMD"] = subprocess.list2cmdline(
            get_ayon_launcher_args()
```

@jakubjezek001 can we get the hound active? I feel like there might be quite some PRs sneaking in invalid cosmetics currently?

client/ayon_core/hosts/zbrush/hooks/create_zmenu_script.py
```diff
@@ -0,0 +1,88 @@
...
+        os.makedirs(zscript_path, exist_ok=True)
+        zscript_txt = os.path.join(zscript_path, "ayon_zbrush_menu.txt")
+        with open(zscript_txt, "w") as zscript:
+            zscript.write(self.ayon_menu())
+            zscript.close()
+        self.launch_context.env["AYON_ZBRUSH_CMD"] = os.environ["AYON_ZBRUSH_CMD"]
+    def ayon_menu(self):
+        zclient_path = os.path.join(ZBRUSH_HOST_DIR, "api", "widgets.py")
+        zclient_path = zclient_path.replace("\\", "/")
+        os.environ["AYON_ZBRUSH_CMD"] = subprocess.list2cmdline(
```
BigRoy, line 34: Is there a reason we're storing this in `os.environ` in the launch action. I feel like this really is only relevant to the launch context, not to the environment we're launching from. We should really be just returning the value or whatever, not storing it in `os.environ`

client/ayon_core/hosts/zbrush/hooks/create_zmenu_script.py
BigRoy: Looking at the docs you should be able to pass the full command line string to `ShellExecute` directly - since we're building this from Python I'd just to that actually so that this here becomes way more readable.

moonyuet: > Looking at the docs you should be able to pass the full command line string to `ShellExecute` directly - since we're building this from Python I'd just to that actually so that this here becomes way more readable.

it doesn't work as zbrush would see it as str unless you use strMerge.

BigRoy: That seems odd. The `StrMerge` should also just be returning a string so the result should be the same.

BigRoy: See the `ShellExecute` example here: https://docs.pixologic.com/user-guide/customizing-zbrush/zscripting/command-reference/

```
[ShellExecute, "open image.psd"]
```

Should work just fine.

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,112 @@
...
+[MemWriteString, EditMode, #InEditMode, 0]
+[MemSaveToFile, EditMode, "{temp_file}", 0]
+[MemDelete, EditMode]
+]
+""").format(temp_file=temp_file)
+    execute_zscript(in_edit_mode)
+    with open(temp_file) as mode:
+        content = str(mode.read())
+        bool_mode = content.rstrip('\x00')
+        return bool_mode
```
BigRoy, line 112: Would this be possible to write in a way that we're within the context manager when querying the resulting value so that the actual temp file gets deleted?

Or do we need it closed because otherwise zbrush can't write into it? or?

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,112 @@
...
+        mode="w", prefix="a_zb_", suffix=".txt", delete=False
+    )
+    tmp_output_file.close()
+    temp_file = tmp_output_file.name.replace("\\", "/")
+
+    in_edit_mode = ("""
+[IFreeze,
+[MemCreate, EditMode, 20, 0]
+[VarSet, InEditMode, [IsEnabled, Transform:Edit]]
+[Note, InEditMode]
```
BigRoy, line 102: According to the zbrush docs `[Note, x]` is to display a visual note to the user in the UI, as such this should be redundant.

moonyuet: Agree. we should remove this, as it will show up as a print

client/ayon_core/hosts/zbrush/api/lib.py
moonyuet: Maybe we can do not only for edit_mode(validation) but the export model function(extraction).

BigRoy: Yup, I think the logic can basically be shared across all functions where you need to have a return value via a file.

_It would be even better if we could somehow do this without needing the file - but that's optimizations mostly and heavily dependent on what's actually possible from Zbrush scripts - this is likely very limited by their API._

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,112 @@
...
+        return valid_workdir
+
+    data.pop("task", None)
+    workdir = anatomy.templates_obj["work"]["folder"].format(data)
+    valid_workdir = find_first_filled_path(workdir)
+    if valid_workdir:
+        # Normalize
+        valid_workdir = os.path.normpath(valid_workdir)
+        if os.path.exists(valid_workdir):
+            return valid_workdir
```
BigRoy, line 72: Why do we need this in in the zbrush libs? Doesn't seem very zbrush-specific and looks like something that should be covered by all other hosts as well?

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def get_current_file():
+    host = registered_host()
+    return host.get_current_workfile()
```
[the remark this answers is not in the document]
moonyuet: maybe I can just move the `get_current_workfile `function into another function in `pipeline` so that we can reused all the time. so it doesn't need to be there.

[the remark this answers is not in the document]
BigRoy: Honestly `registered_host().get_current_workfile()` isn't a really problematic call right? Don't think we really need a wrapper around it unless it gets in the way?

client/ayon_core/hosts/zbrush/api/lib.py
BigRoy: Why couldn't you use `openpype.pipeline.workfile.get_workdir` or alike
I feel like we'd definitely want to avoid writing our own workdir related logic in other places.

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,113 @@
...
+[IShowActions,0]
+[VarSet, filepath, "{filepath}"]
+[FileNameSetNext, #filepath]
+[IKeyPress, 13, [IPress, Tool:Export]]]
+]
+""").format(filepath=filepath)
+    execute_zscript(save_file_zscript)
+
+
+def has_edit_mode():
```
BigRoy, line 92: I feel like this is much clearer as `is_in_edit_mode`, `is_edit_mode` or `is_edit_mode_active` than `has_edit_mode`.

Also, a docstring would be helpful to describe what it does and also why you might want to call this.

```suggestion
def is_in_edit_mode():
    """Return whether transform edit mode is currently enabled.
    
    Certain actions can't be performed if Zbrush is currently within
    edit mode, like exporting a model.
    
    Returns:
        bool: Whether Edit Mode is enabled.
    """
```

client/ayon_core/hosts/zbrush/api/communication_server.py
```diff
@@ -0,0 +1,746 @@
...
+            mode="w", prefix="a_zb_", suffix=".txt", delete=False
+        )
+        output_file.close()
+        output_filepath = output_file.name.replace("\\", "/")
+        with open(output_filepath, "wt") as tmp:
+            tmp.write(zscript)
+        subprocess.call([zbrush_exe, output_filepath],
+                        shell=True)
+        # remove the temp file
+        os.remove(output_filepath)
```
BigRoy, line 692: Doesn't this work?
```suggestion
        with tempfile.NamedTemporaryFile(
            mode="w", prefix="a_zb_", suffix=".txt"
        ) as f:
            f.write(zscript)
            subprocess.call([zbrush_exe, f.name], shell=True)
```

Why do we need to close it first, then write to it, close it again, etc.

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -77,6 +77,20 @@ def get_current_file():
...
+    save_file_zscript = ("""
+[IFreeze,
+[VarSet, filepath, "{filepath}"]
+[FileNameSetNext, #filepath]
+[IKeyPress, 13, [IPress, Tool:Export]]]
+[Sleep, 2]
+]
+""").format(filepath=filepath)
+    execute_zscript(save_file_zscript)
+    time.sleep(8)
```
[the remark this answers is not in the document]
BigRoy: Just to elaborate on this with the info I also shared on discord, basically this:
```python
import time
start = time.time()
print(f"Start {time.time() - start}")
communicator.execute_zscript("[Sleep, 2]")
print(f"After first sleep {time.time() - start}")
communicator.execute_zscript("[Sleep, 2]")
print(f"After second sleep {time.time() - start}")
communicator.execute_zscript("[Sleep, 5]")
communicator.execute_zscript("[Sleep, 5]")
print(f"After two longer sleeps {time.time() - start}")
```

Should result in:
```
0.0
2.0
4.0
14.0
```
But then slightly higher numbers because of the time taken to execute, etc.
But it may not be lower numbers.

Since we only want to continue our python scripts as soon as the zscript finished completely with its execution.

client/ayon_core/hosts/zbrush/plugins/publish/extract_polymesh.py
```diff
@@ -0,0 +1,60 @@
...
+        if not self.is_active(instance.data):
+            return
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        export_format = attr_values.get("exportFormat")
+        stagingdir = self.staging_dir(instance)
+        filename = f"{instance.name}.{export_format}"
+        filepath = os.path.join(stagingdir, filename)
+
+
```
BigRoy, line 31: Check your linter ;)
```suggestion
```

client/ayon_core/hosts/zbrush/plugins/publish/extract_polymesh.py
```diff
@@ -0,0 +1,60 @@
...
+        instance.data["representations"].append(representation)
+        self.log.info(
+            "Extracted instance '%s' to: %s" % (instance.name, filepath)
+        )
+
+    @classmethod
+    def get_attribute_defs(cls):
+        defs = super(ExtractPolyMesh, cls).get_attribute_defs()
+        export_format_enum = ["abc", "fbx", "ma", "obj"]
```
BigRoy, line 52: You can export `ma` from zbrush, haha - how odd. :) 
Anyway, what if you wanted to export multiple formats?

Also, doesn't zbrush support `usd`  exports? It should [according to this](https://docs.pixologic.com/user-guide/zbrush-other-programs/usd-format/)

client/ayon_core/hosts/zbrush/api/lib.py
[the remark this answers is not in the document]
BigRoy: I feel like someone else at Ynput - who's done work on the After Effects / Photoshop integrations and likely has done some work on these communication servers with other hosts could bring some valuable insights to this as well from that experience.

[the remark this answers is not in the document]
moonyuet: > Just to elaborate on this with the info I also shared on discord, basically this:
> 
> ```python
> import time
> start = time.time()
> print(f"Start {time.time() - start}")
> communicator.execute_zscript("[Sleep, 2]")
> print(f"After first sleep {time.time() - start}")
> communicator.execute_zscript("[Sleep, 2]")
> print(f"After second sleep {time.time() - start}")
> communicator.execute_zscript("[Sleep, 5]")
> communicator.execute_zscript("[Sleep, 5]")
> print(f"After two longer sleeps {time.time() - start}")
> ```
> 
> Should result in:
> 
> ```
> 0.0
> 2.0
> 4.0
> 14.0
> ```
> 
> But then slightly higher numbers because of the time taken to execute, etc. But it may not be lower numbers.
> 
> Since we only want to continue our python scripts as soon as the zscript finished completely with its execution.

Maybe one of the way we can do this is to introduce a while loop to check on the file existence. If you talk about exexute zscript is an issue. this means the issue is everywhere, including get current workfile, save workfile, open workfile, list instance, list container.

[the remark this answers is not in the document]
moonyuet: > I feel like someone else at Ynput - who's done work on the After Effects / Photoshop integrations and likely has done some work on these communication servers with other hosts could bring some valuable insights to this as well from that experience.

The better reference is tvpaint, i have already mentioned that I would write dll plugins in additional info

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,134 @@
...
+    temp_file = tmp_output_file.name.replace("\\", "/")
+
+    in_edit_mode = ("""
+[IFreeze,
+[MemCreate, EditMode, 20, 0]
+[VarSet, InEditMode, [IGet, Transform:Edit]]
+[Note, InEditMode]
+[MemWriteString, EditMode, #InEditMode, 0]
+[MemSaveToFile, EditMode, "{temp_file}", 0]
+[MemDelete, EditMode]
```
BigRoy, line 127: Would `[VarSave]` also be usable here?

> Saves variable value/s to file

moonyuet: It's for .zvr format as far as I know

client/ayon_core/hosts/zbrush/api/lib.py
```diff
@@ -0,0 +1,134 @@
...
+    tmp_output_file.close()
+    temp_file = tmp_output_file.name.replace("\\", "/")
+
+    in_edit_mode = ("""
+[IFreeze,
+[MemCreate, EditMode, 20, 0]
+[VarSet, InEditMode, [IGet, Transform:Edit]]
+[Note, InEditMode]
+[MemWriteString, EditMode, #InEditMode, 0]
+[MemSaveToFile, EditMode, "{temp_file}", 0]
```
BigRoy, line 126: Doesn't the last argument `0` mean "do not overwrite" and since you've pregenerated the file. Doesn't this mean that it never writes? :)

iLLiCiTiT:
Shouldn't this be in separated repository?

client/ayon_core/hosts/zbrush/api/lib.py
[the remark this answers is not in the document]
BigRoy: Here's an example 'workaround' as I described:
```python
import os
import tempfile

def execute_zscript_and_wait(zscript, wait=0.1, timeout=20):
    temp_path = os.path.join(
        tempfile.gettempdir(),
        "{tempfile.gettempprefix()}_{uuid}.txt"
    )
    temp_path = path.replace("\\", "/")
    assert not os.path.exists(temp_path)
    
    # At the end of the zscript write
    # temporary file to confirm the zscript
    # has run to the end of the script
    zscript += (f"""
[MemWriteString, EditMode, "1", 0]
[MemSaveToFile, EditMode, "{temp_path}", 0]
[MemDelete, EditMode]
]
""")
    execute_zscript(zscript)
    
    # Wait around until the zscript finished
    time_taken = 0
    while not os.path.exists(temp_path):
        time.sleep(wait)
        time_taken += wait
        if time_taken > timeout:
            raise RuntimeError(
                f"Timeout. Zscript took longer than "
                "{timeout}s to run."
            )
```

Probably not best performance wise - there may be other ways to get responses from Zbrush than this, but I think this should work at least.

client/ayon_core/hosts/zbrush/api/lib.py
BigRoy: Ah yes, you're right. I think I initially assumed that was an ascii file that we could easily read - but can't find much about the zvr format.

BigRoy:
> Shouldn't this be in separated repository?

I suppose this should be a separate `ayon_zbrush` package in a `ayon-zbrush` repository yes.

client/ayon_core/hosts/zbrush/api/pipeline.py
```diff
@@ -0,0 +1,559 @@
...
+        work_dir = get_workdir(project_name, folder_path, task_name)
+        txt_dir = os.path.join(
+            work_dir, ".zbrush_metadata").replace(
+                "\\", "/"
+        )
+        with open (f"{txt_dir}/current_file.txt", "r") as current_file:
+            content = str(current_file.read())
+            filepath = content.rstrip('\x00')
+            current_file.close()
+            return filepath
```
BigRoy, line 148: Knowing _nothing_ of Zbrush - why can't we rely on:
```
[FileNameGetLastUsed]
```

Or:
```python
def get_current_workfile():
    temp_file = get_tempfile_path()
    script = """
// Get FileNameGetNext if any defined, else return FileNameGetLastUsed
[If, [FileNameHasNext], [VarSet, filename, [FileNameGetNext]], [VarSet, filename, FileNameGetLastUsed]]
// Return filename variable value file temp file
[MemWriteString, filename, #filename, 0]
[MemSaveToFile, filename, "{temp_file}", 0]
[MemDelete, filename]
"""
    execute_zscript(script)

    # Get returned value as written from the zscript
    with open(temp_file, "r") as f:
        return f.read()
```
Of course consider this psuedocode.

BigRoy: Ah it seems the first returns ANY filename that was used - including e.g. file exports or reading scripts.

And even though the docs state that `[IGetTitle, Tool:Current Tool]` returns the full path to the tool - it doesn't seem like it.

Anyway, Zbrush itself seems to "remember" the current project just fine. If you use File > Save As the window title updates accordingly, then you can change tools, export objects, and click File > Save Next and it'll increment your project file just fine to a new version.

As such, the current zbrush project path _should_ be available somewhere?

BigRoy: Ok, so Zbrush is a pain in the ass apparently in this area. However, I'd still argue that we should avoid trying to store the current file in a file in the working directory and instead store the current filepath on save in a scripting memory block of ZBrush, e.g. using something like:
```
// On save and on open to set current filepath
// Create a memory block to hold the filepath
[MemDelete, AYON_CurrentFilePath] // Delete if it already existed
[MemCreate, AYON_CurrentFilePath, [StrLength, "{workfile_path}"]]
[MemWriteString, AYON_CurrentFilePath, "{workfile_path}"]

// Get current workfile
// Note: File will not be written if no current file memory block exists
// Write the memory block to file (to get output from zscript to python)
[MemSaveToFile, AYON_CurrentFilePath, "{tmpfile}", 1]
```

BigRoy: Here are some of the options I've tried: https://www.zbrushcentral.com/t/scripting-get-current-file-project-path/458261

moonyuet: > Ok, so Zbrush is a pain in the ass apparently in this area. However, I'd still argue that we should avoid trying to store the current file in a file in the working directory and instead store the current filepath on save in a scripting memory block of ZBrush, e.g. using something like:
> 
> ```
> // On save and on open to set current filepath
> // Create a memory block to hold the filepath
> [MemDelete, AYON_CurrentFilePath] // Delete if it already existed
> [MemCreate, AYON_CurrentFilePath, [StrLength, "{workfile_path}"]]
> [MemWriteString, AYON_CurrentFilePath, "{workfile_path}"]
> 
> // Get current workfile
> // Note: File will not be written if no current file memory block exists
> // Write the memory block to file (to get output from zscript to python)
> [MemSaveToFile, AYON_CurrentFilePath, "{tmpfile}", 1]
> ```

I can try with this again but i tried this similar method in my fork repo(you can find the old commit), but for some reason it doesnt work when saving and open other workfiles.

moonyuet:
> > Shouldn't this be in separated repository?
> 
> I suppose this should be a separate `ayon_zbrush` package in a `ayon-zbrush` repository yes.

Maybe we can move this after finishing the code review here,(or we can do it now)

client/ayon_core/hosts/zbrush/api/pipeline.py
BigRoy: > but for some reason it doesnt work after saving and open other workfiles.

Interesting - did you try with memory blocks or vars? Because vars should be lost indeed with other script executions - and memory blocks should stick around.

Also, you might have hit the bug that the script execution actually didn't finish and you checked the 'written' file before the zscript finished?

moonyuet: > > but for some reason it doesnt work after saving and open other workfiles.
> 
> Interesting - did you try with memory blocks or vars? Because vars should be lost indeed with other script executions - and memory blocks should stick around.
> 
> Also, you might have hit the bug that the script execution actually didn't finish and you checked the 'written' file before the zscript finished?

it doesnt hit bug it is just that  it doesnt save the data correctly in the temp file.

client/ayon_core/hosts/zbrush/hooks/create_zmenu_script.py
BigRoy:
```
// Set a variable to " so we can quote the command line arguments for ShellExecute
[VarSet, q, [StrFromAsc, 34]]

[IPalette, "AYON", 1]

[ISubPalette, "AYON:Tools", 2]
[IButton, "AYON:Tools:Create...", "Open AYON Publisher on Create Tab",
	[ShellExecute, [StrMerge, #q, "{launch}", #q, "addon zbrush run-with-zscript --launcher create_tool"]], 0, 180
]

// Load
[IButton, "AYON:Tools:Load...", "Open AYON Loader",
	[ShellExecute, [StrMerge, #q, "{launch}", #q, "addon zbrush run-with-zscript --launcher loader_tool"]], 0, 180
]

// Publish
[IButton, "AYON:Tools:Publish...", "Open AYON Publisher",
	[ShellExecute, [StrMerge, #q, "{launch}", #q, "addon zbrush run-with-zscript --launcher publish_tool"]], 0, 180
]

// Manage
[IButton, "AYON:Tools:Manage...", "Open AYON Scene Inventory UI",
	[ShellExecute, [StrMerge, #q, "{launch}", #q, "addon zbrush run-with-zscript --launcher scene_inventory_tool"]], 0, 180
]

[ISubPalette, "AYON:Project", 2]
// Workfile
[IButton, "AYON:Project:Work Files...", "Open AYON Work Files UI",
	[ShellExecute, [StrMerge, #q, "{launch}", #q, "addon zbrush run-with-zscript --launcher workfiles_tool"]], 0, 180
]
```

image


This should work @moonyuet

client/ayon_core/hosts/zbrush/api/pipeline.py
BigRoy: It worked fine for me from a single session. However, the memory block must exist - otherwise it won't create/write the file. If the memory block is also the exact size fo the filepath there are no null bytes in the written file and the contents is _exactly_ the filepath so you can just read the file in full.

client/ayon_core/hosts/zbrush/hooks/create_zmenu_script.py
moonyuet: but the shellExecute code doesn't work, knowing that zscript only allows certain characters in each line

moonyuet: the ui is successfully installed but the button cant trigger any tool

client/ayon_core/hosts/zbrush/api/communication_server.py
moonyuet: > with tempfile.NamedTemporaryFile(
>             mode="wt", prefix="a_zb_", suffix=".txt"
>         ) as f:
>             f.write(zscript)
>             subprocess.call([zbrush_exe, f.name], shell=True)

it will error out when i tested with loading model, so no.

BigRoy: Could you provide the error?

moonyuet: > with tempfile.NamedTemporaryFile(
>             mode="wt", prefix="a_zb_", suffix=".txt"
>         ) as f:
>             f.write(zscript)
>             subprocess.call([zbrush_exe, f.name], shell=True)

image

moonyuet:
@BigRoy I get this error when i was testing with AYON, but the same code works fine in OP legacy.
image

BigRoy:
> @BigRoy I get this error when i was testing with AYON, but the same code works fine in OP legacy. Not sure if something happened to the workfile. image

I have seen something like this in the past when the `current_file` would return None and it was unable to detect the extension it would end up needing to set for the new file due to that.

Anyway, since this error is in the UI I would expect to at least see a traceback in the server's console?

moonyuet:
> > @BigRoy I get this error when i was testing with AYON, but the same code works fine in OP legacy. Not sure if something happened to the workfile. image
> 
> I have seen something like this in the past when the `current_file` would return None and it was unable to detect the extension it would end up needing to set for the new file due to that.
> 
> Anyway, since this error is in the UI I would expect to at least see a traceback in the server's console?

I will show the comparison between OP legacy and AYON.
It runs perfectly fine in OP while it errors out in AYON, when I comment out line 124 and not saving global context, it doesn't fail anymore in AYON. But it doesn't save any context data, of which needed to have when switching different context. 

OP Legacy:
https://github.com/ynput/ayon-core/assets/64118225/fd4e2df9-6e18-4c63-a814-72506c221bca

AYON:
https://github.com/ynput/ayon-core/assets/64118225/dd2c05e9-b2b5-4997-ac97-aa86595d69db

LiborBatek:
...so after some tweaking of the ```Publish``` model instance and switching off ```Extract Polymesh``` I was able to produce my ```modelHires``` publish and succesfully load it into ```maya``` host or more like ```import``` it via Action in the ```Loader```

Screenshot 2024-03-08 143420

and in maya:

image

As it was imported it cant be managed as an ordinary ayon asset atm tho

image

LiborBatek:
I would like to add a few of my observations when testing the very first ```ZBrush``` addon version here:

1. The Ayon tools windows like ```Publisher``` ```Workfiles``` etc is constantly dissapearing during user interaction forcing user to find any of those floating wins via Alt+TAB keys to see it again
2. Im not sure whats the purpose of ```Extract Polymesh``` option on Publish instance of ```model```
3. We should add option whether user wants to output ztool (aka geometry) in active ```SDiv Level``` or the ```Highest``` no matter what being the displayed/active level he is working in
4. Publish does not produce the selected ```representation``` atm as mine been ```ABC``` but ```OBJ``` been produced instead

Sligthly confusing Publish Instance setting as seen here:
Screenshot 2024-03-08 144825

SDiv of the active model (ztool...you name it:)

Screenshot 2024-03-08 145008

BigRoy:
> 1. The Ayon tools windows like `Publisher` `Workfiles` etc is constantly dissapearing during user interaction forcing user to find any of those floating wins via Alt+TAB keys to see it again

I had the same - I think we're best of setting the Always on Top flag on the Qt UIs. I'm not sure if that also makes it float on top of zbrush in full screen mode, but I hope so!

> Im not sure whats the purpose of Extract Polymesh option on Publish instance of model

Confused me too - not sure what the idea was. Also not too happy with the dropdown myself. I see myself wanting to export one or more formats, not `.obj` and one other. I might want to export, three, four formats, etc.

> We should add option whether user wants to output ztool (aka geometry) in active SDiv Level or the Highest no matter what being the displayed/active level he is working in

Agree. It should likely be a dropdown or integer number which defaults to the "Use current subdiv" or sets a particular subdiv during extraction.

I'd rather not have it 'validate' it but rather have the option to pick the subdiv level for export so that I can export both low-res and high-res in one publish session.

> Publish does not produce the selected representation atm as mine been ABC but OBJ been produced instead

I believe the errors you had were due to the Alembic export failing. It currently ALWAYS exports the Obj format - which I think is odd as well.

BigRoy:
> and in maya:
> image
> 
> As **it was imported it cant be managed as an ordinary ayon asset atm tho**

Correct, the `.obj` file format I believe can't be referenced in Maya and thus will always be loaded unmanaged via the Import loader.

client/ayon_core/hosts/zbrush/api/pipeline.py
```diff
@@ -147,12 +147,20 @@ def list_instances(self):
...
+        """Get the data of the containers
+
+        Returns:
+            list: the list which stores the data of the containers
+        """
         return get_containers()
 
     def initial_app_launch(self):
+        """The actions should be done when Zbrush launches.
+        """
```
BigRoy, line 163:
```suggestion
        """Triggers on launch of the communication server for Zbrush.
        
        Usually this aligns roughly with the start of Zbrush.
        """
```
I wonder if this _may_ be before Zbrush actually fully launched. If so, we may want to include that in the docstring

client/ayon_core/hosts/zbrush/api/pipeline.py
```diff
@@ -203,10 +211,24 @@ def containerise(
...
+        context (dict): context data
+
+    """
     return write_context_metadata(ZBRUSH_SECTION_NAME_CONTEXT, context)
 
 
 def write_context_metadata(metadata_key, context):
+    """Write context data into the related json
+    which stores in zbrush_metadata folder
+    in the project work directory.
```
BigRoy, line 226: As stated on Discord - the biggest issue I had with these functions was that some seemed to write into:

```
.zbrush_metadata/{workfile}/...
```

Whereas others seemed to read/write directly to
```
.zbrush_metadata/...
```

It seemed like the first was intended to be:
- Used when saving/opening the workfile (so like persisting and reading files)

And the second was intended to be:
- Store the 'in-memory' context, instances and other metadata to the current zbrush process. 
- Then on workfile save this is the data that would get copied/saved into the `.zbrush_metadata/{workfile}/...` folder
- Then on workfile open this is the data that would be replaced by any `.zbrush_metadata/{workfile}/...` of the workfile being opened.

If there is indeed this difference between writing to either of these two locations - then it's best to clarify in the docstring what the intended use case is of the function. The fact that `save_current_workfile_context` "saves current workfile context data" and that it is a function for it - tells me nothing more from the docstring than then function name. The docstring should instead be describing what it does and why.

E.g.
```python
def save_current_workfile_context(context):
    """Save current workfile context data to `.zbrush_metadata/{workfile}/key`

    This persists the current in-memory context to be set for a specific
    workfile on disk. Usually used on save to persist the local sessions'
    workfile context on save.

    The context data includes things like the project name, asset name,
    etc.

    Args:
        context (dict): context data
    """
```
This description might be inaccurate - I'm just writing out what I _think_ may be the case. In a way I hope I'm wrong in my description because that would actually showcase how limitied information I get from the function alone.



I see a `write_workfile_metadata` and `write_context_metadata` which I feel are the differences between writing to one of the two folders. They should really explain why they differ, what their intent is, etc.

BigRoy: Just to clarify:

- I feel like all data should be related to a workfile, there's nothing else than persisting data with the workfile.
- The only other data is the data in-memory for the current session - that is either data that would not need to persist (since it's not related to a workfile) or is just 'in-memory' up until it gets persisted to disk on workfile save.

How the functions used here differ between these two - and whether that's the case - I have no idea.

moonyuet: I agree with you that we need to further elaborate the docstring and we should rename the function for better scope on what the function does in the integration. 
I guess we can also mention why the zbrush_metadata folder is there and how it works in the integration (in admin docs) in the documentatio.

moonyuet:
@LiborBatek I have introduced the fix on ma format. It is possible that you may find integrate hero version errors out even if it is successfully published.
And I think we can try on different scenarios of which creating instances with different formats, to see if it is best options for users to set the export format in just the creator.

LiborBatek (changes requested):
I gave it a spin again and this time the publishing went without any errors when choosing ```ma``` format but unfortunatelly the file being empty and no model data present inside of it .

image

It need some resolution then, either by fixing or just disabling the publishing in ```ma``` format... which could do imho (as OBJ, Abc, FBX being sufficient)

moonyuet:
Closed for https://github.com/ynput/ayon-zbrush and https://github.com/ynput/ayon-core/pull/198

PR #114, opened 2024-02-23
Unreal: Fix Render Instance Collector to use folderPath

simonebarbieri:
## Changelog Description
Fix Render Instance Collector to use folderPath instead of just the asset name.

## Additional info
This was causing the `ValidateSequenceFrames` validator to fail because couldn't find the asset data in the database.

PR linked to https://github.com/ynput/OpenPype/pull/6233


LiborBatek (approved):
All working fine. All rendered frames got published and integrated to Ayon.

PR #113, opened 2024-02-23
Chore: Product name and type

iLLiCiTiT:
## Changelog Description
Goal of this PR is to start using product naming in code base. Which means to use `productName` and `productType` instead of `subset` and `family` during create and publish phase.

## Additional info
First of all I'm 100% sure that I broke something, if you can, please do code review first before you try to run. If you find any issues please report as soon as possible, you can DM me with quick questions.

All changes should support compatibility with workfiles before this PR in all hosts.

All the time when a `family` did represent product type, it should be changed to `product type`. e.g. create plugin does have `product_type` instead of `family` and created instance has `productType` in data instead of `family`.

One of the goals was to also use `productType` instead of `family` during publishing and use only `families` for pyblish filtering. The key `family` key is filled by pyblish automaticlly to `"default"`, so it should not be an issue. **But** when we planned it, we didn't realize that sometimes `instance.data` are filled with raw data from scene, and in that case there might be `family` with a value that is not managed by us, speaking only about logic or workfiles with instances before this PR. I've changed that idea and all the time `productType` is changed during publishing, `family` is changed too.

## What I couldn't resolve
1. Json files in **Blender** and **Unreal**
Sometimes it looks like the json is storing data that are unused, and sometimes I couldn't find source of their data so I couldn't tell if/how should I modify them. Sometimes the data are stored unclearly under key that says nothing about the value, e.g. json has `"_id"` which in fact represents representation id. Some of the json files are storing `family` which should be probably changed to `product_type` but I didn't find where the value is used. etc.

2. Json files in **Maya**
This is just about looks publishing, needs validation that I didn't break anything.

3. **Resolve** and **Hiero**
I probably failed at backwards compatibility.

## Testing notes:
- For host testing go one by one.
- Go through code if you can and scan if you find invalid changes.
- Open host and try all create plugin and publish. Validated output if can be loaded.
- Open the host you're testing with workfile that does have instances before this PR and publish them.
- [x] 3dsMax
- [x] AfterEffects
- [x] Blender
- [x] Fusion
- [x] Hiero
- [x] Houdini
- [x] Maya
- [x] Nuke
- [x] Photoshop
- [x] Resolve
- [x] SubstancePainter
- [x] TrayPublisher - simple creators
- [x] TrayPublisher - editorial
- [x] TVpaint
- [x] Unreal
- [ ] ~~Celaction~~
- [ ] ~~Flame~~
- [ ] ~~Harmony~~

[...]

PR #115, opened 2024-02-23
Houdini: use product name n product type (follow up PR)

MustafaJafar:
## Changelog Description
A separate PR for dealing with Houdini for https://github.com/ynput/ayon-core/pull/113 

## Additional info 1
The new `productName` Key conflicts with `Karma ROP`. 
So, instead we are going to use `AYONProductName` 
The caveat is using  `AYONProductName`  when creating instances but reading its value as `productName` when collecting instances.

## Additional info 2
I don't sure if we should provide a convert legacy code in this PR.

## Testing notes:
From the main PR:
- Open Houdini and try all create plugin and publish. Validated output if can be loaded.
- Open Houdini with workfile that does have instances before this PR and publish them.

### DIFF
--- a/client/ayon_core/hosts/houdini/api/plugin.py
+++ b/client/ayon_core/hosts/houdini/api/plugin.py
@@ -271,8 +271,8 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
     def imprint(self, node, values, update=False):
         # Never store instance node and instance id since that data comes
         # from the node's path
-        if "AYONProductName" in values:
-            values["productName"] = values.pop("AYONProductName")
+        if "productName" in values:
+            values["AYONProductName"] = values.pop("productName")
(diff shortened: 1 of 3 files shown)


iLLiCiTiT (approved):

BigRoy:
This issue actually shows me that we should've prefixed the attributes, with e.g. `ayon_` or `AYON_` or along the lines of this PR prefixed with `AYON`. Preferably all, not just this one. The fact that we just happened to find a particular node where there's a conflict now just makes it seems like there'll be other conflicts that are bound to happen in the future? Having them namespaced would make things much clearer as well maybe?

iLLiCiTiT:
I'm not capable of doing it. I f you can to do it, in a way that it can read data from scenes before https://github.com/ynput/ayon-core/pull/113 , then I don't see a reason why not.

PR #116, opened 2024-02-26
Max: Validate No Animation

moonyuet:
## Changelog Description
This PR is to add the optional validator to ensure no keyframe set when the objects being published in model family

## Additional info
n/a
## Testing notes:
1. Launch Max via AYON
2. Create Model instances with some models(some with no animation while some have)
3. Publish
4. If the publishing model has animation, it will error out
5. If the publishing model has no animation, it will pass the validation.


client/ayon_core/hosts/max/plugins/publish/validate_no_animation.py
```diff
@@ -0,0 +1,67 @@
...
+            if ctl_count > 0:
+                invalid.append(
+                    ((sel), f"Object Rotation(s) has {ctl_count} keyframe(s)"))
+            sel_scale_ctl = rt.getPropertyController(
+                sel.controller, "Scale"
+            )
+            ctl_count = (sel_scale_ctl.keys).count
+            if ctl_count > 0:
+                invalid.append(
+                    ((sel), f"Object Scale(s) has {ctl_count} keyframe(s)"))
```
BigRoy, line 65: Shouldn't we somehow just check whether it has ANY animation instead of only checking pos, rot, scale? Isn't there an equivalent "check if it has any keys?" type logic?

Also, why is `sel` between brackets like `(sel)` that seems redundant?

client/ayon_core/hosts/max/plugins/publish/validate_no_animation.py
```diff
@@ -0,0 +1,67 @@
...
+    def get_invalid(instance):
+        invalid = []
+        selected_objects = instance.data["members"]
+        for sel in selected_objects:
+            sel_pos_ctl = rt.getPropertyController(
+                sel.controller, 'Position')
+            ctl_count = (sel_pos_ctl.keys).count
+            if ctl_count > 0:
+                invalid.append(
+                   ( (sel), f"Object Position(s) has {ctl_count} keyframe(s)"))
```
BigRoy, line 51: Are we lacking the Hound on this repo? I feel like this `( (sel)` would be something the houdin in OpenPype repo would've triggered over.

client/ayon_core/hosts/max/plugins/publish/validate_no_animation.py
```diff
@@ -0,0 +1,67 @@
...
+    OptionalPyblishPluginMixin
+)
+from ayon_core.hosts.max.api.action import SelectInvalidAction
+
+
+class ValidateNoAnimation(pyblish.api.InstancePlugin,
+                          OptionalPyblishPluginMixin):
+    """Validates No Animation
+
+    Ensure no keyframes on nodes in the Instance
```
BigRoy, line 15: If the current logic remains where it checks only the controller's pos, rotate and scale I'd change this docstring to explain it only validates that.

However, as said below - I think it'd be better to just check for ANY animation on the node whatsoever instead of only transforms.

client/ayon_core/hosts/max/plugins/publish/validate_no_animation.py
moonyuet: this is perhaps one of the ways check whether there is keyframe in Max. But from my current experience I can say they are not Maya alike.

BigRoy: You should at least be able to do:
```
numKeys controller
```

According to:
 https://help.autodesk.com/view/MAXDEV/2023/ENU/?guid=GUID-B1700B1D-B1EA-4A6C-B4A3-A29DB26C8C02

LiborBatek (commented):
The Validator works when ```animation``` keys present on the geo and correctly kicks in these circumstances.

Screenshot 2024-03-06 110330

Also ```Select Invalid``` works ok and select the geo with ```animation``` for the user so he can treat it.

The ```Validator``` could benefit from more readible messaging tho as now being somewhat unclear and containing unnecessary info about some values of keys or transforms, not very helpful/ usable in the end.

Also one thing to consider: 

When Validation happens it does not error out if just single ```keyframe``` set on the ```geometry```, in theory it is also sort of animation in a way as keyframe exists and could prevent some action later on when asset loaded elswhere. 
So I propose to Validate such occasion and error out too.

moonyuet:
@LiborBatek i have adjusted the code regarding to your suggestion.

antirotor (approved):
👍🏻

PR #117, opened 2024-02-26
Bugfix/houdini default renderproduct name

MustafaJafar:
## Changelog Description
This is fixing key name for default render products in VRay. Original name `RGB Color` caused issues during job submission.

## Additional info
Fixing error like:

```
ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Name 'render_ropMainRGB Color' does not match regex '^[a-zA-Z0-9]([a-zA-Z0-9.-]*[a-zA-Z0-9])?$' on item 'project/products' (Line 3 Column 5)
```

## Testing notes:
1. Run Houdini with Vray for Houdini
2. Create VRay ROP and publish

[Ported From OpenPype [6083](https://github.com/ynput/OpenPype/pull/6083) ]


moonyuet:
I currently dont have the vray license as I accidentally deleted my xml config for that.
But I tested it previously it should be good.

antirotor (approved):
looks good, the same as with OP that was now merged

PR #119, opened 2024-02-26
Don't use "modules" subkey in system settings

iLLiCiTiT:
## Changelog Description
Fixed places where subkey `"modules"` from system settings was used.

## Additional info
I'm not sure if maya should know how to get royal render root, but that's question for a different PR.

## Testing notes:
1. Launch of application should not crash because of missing `"modules"` key in system settings


tokejepsen (approved):
Tested in Nuke 15 successfully.

PR #121, opened 2024-02-27
Max: Validate Mesh Has UVs

moonyuet:
## Changelog Description
This PR is to add the validator to check mesh has UVs in model product type.

## Additional info
n/a

## Testing notes:
1. Install the latest Max Addon
2. Launch Max via AYON
3. Create Model Instance with some meshes
4. Turn on Validate Mesh Has UVs in Publish Tab
5. If the mesh does not have UVs, the validator errors out.


client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,44 @@
...
+from pymxs import runtime as rt
+
+
+class ValidateMeshHasUVs(pyblish.api.InstancePlugin,
+                         OptionalPyblishPluginMixin):
+
+    """Validate the current mesh has UVs.
+
+    It validates whether the current mesh has texture vertex(s).
+    If the mesh does not have texture vertex(s), it does not
```
BigRoy, line 18: The added (s) here after the word is weird because the plural of vertex is vertices. I'd just write it as:
```suggestion
    It validates whether the current mesh has texture vertices.
    If the mesh does not have texture vertices, it does not
```

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,44 @@
...
+    families = ['model']
+    label = 'Validate Mesh Has UVs'
+    actions = [SelectInvalidAction]
+    optional = True
+
+
+    @classmethod
+    def get_invalid(cls, instance):
+        invalid = [member for member in instance.data["members"]
+                   if not member.mesh.numTVerts > 0]
```
BigRoy, line 33: Having no knowledge of Max - but in Maya it's possible to have a mesh that has only partial UVs, e.g. a few faces missing UVs while the rest does have UVs. Is that possible in Max as well?

If so, we might want to clarify in the wording that it currently only checks if an object has ANY uvs and not whether all faces have UVs.

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def process(self, instance):
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                title="Mesh has missing UVs",
+                message="Model meshes are required to have UVs.<br><br>"
+                        "Meshes detected with invalid or missing UVs:<br>"
+                        "{0}".format([err.name for err in invalid])
+            )
```
BigRoy, line 44: Is there a reason you're using html over regular line breaks?

I think the `<br>` might display ok in the UI but in logs will not be resolved to a new line and thus look weird there.
```suggestion
                message="Model meshes are required to have UVs.\n\n"
                        "Meshes detected with invalid or missing UVs:\n"
                        "{0}".format([err.name for err in invalid])
            )
```

Also, with this being an artist facing message we might want to turn it into bullet points OR at least `", ".join(names)` so they aren't printed as a Python list.

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,44 @@
...
+    def get_invalid(cls, instance):
+        invalid = [member for member in instance.data["members"]
+                   if not member.mesh.numTVerts > 0]
+        return invalid
+
+    def process(self, instance):
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                title="Mesh has missing UVs",
```
BigRoy, line 40: The docstring tells a bit about what this plug-in validates and does. However, that doesn't show by default in the Publisher UI.

I'd recommend adding a `description` argument here that shows in a nice super artist-friendly manner what the validation is doing, why it's doing it and how the artist can best solve it.

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
moonyuet: it doesn't really have the faces missing UVs check. there is a function to get the coordinates of texture faces but there is no exact function to get number of texture faces(but it does have the function to get number of face)
https://help.autodesk.com/view/MAXDEV/2022/ENU/?guid=GUID-2E5C8194-F165-4D09-8A19-069BFCF83104

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,56 @@
...
+from pymxs import runtime as rt
+
+
+class ValidateMeshHasUVs(pyblish.api.InstancePlugin,
+                         OptionalPyblishPluginMixin):
+
+    """Validate the current mesh has UVs.
+
+    This validator only checks if the mesh has UVs but not
+    whether the faces of the mesh have UVs.
```
BigRoy, line 18:
```suggestion
    This validator only checks if the mesh has UVs but not
    whether all the individual faces of the mesh have UVs.
```

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,57 @@
...
+    families = ['model']
+    label = 'Validate Mesh Has UVs'
+    actions = [SelectInvalidAction]
+    optional = True
+
+
+    @classmethod
+    def get_invalid(cls, instance):
+        invalid = [member for member in instance.data["members"]
+                   if not member.mesh.numTVerts > 0]
```
BigRoy, line 37: What happens here if there's a member in the instance that is not a mesh? does it error on the `member.mesh` access?

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,57 @@
...
+        invalid = self.get_invalid(instance)
+        if invalid:
+            bullet_point_invalid_statement = "\n".join(
+                "- {}".format(invalid.name) for invalid
+                in invalid
+            )
+            report = (
+                "Model meshes are required to have UVs.\n\n"
+                "Meshes detected with invalid or missing UVs:\n"
+                f"{bullet_point_invalid_statement}\n\n"
```
BigRoy, line 50: Why do you need the extra line breaks at the END of the report?
```suggestion
                f"{bullet_point_invalid_statement}"
```

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    order = ValidateMeshOrder
+    hosts = ['max']
+    families = ['model']
+    label = 'Validate Mesh Has UVs'
+    actions = [SelectInvalidAction]
+    optional = True
+
+
```
BigRoy, line 33:
```suggestion
```
Still think the hound is disabled on this repo? @jakubjezek001 
This should've been caught?

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
moonyuet: if there is instance which is not mesh, yes it will error out.

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,57 @@
...
+import pyblish.api
+from ayon_core.hosts.max.api.action import SelectInvalidAction
+from ayon_core.pipeline.publish import (
+    ValidateMeshOrder,
+    OptionalPyblishPluginMixin,
+    PublishValidationError
+)
+from pymxs import runtime as rt
```
BigRoy, line 9:
```suggestion
```
This is unused?
Hound should've caught this?

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
BigRoy: Do we want that? It will likely be quite an unclear error to the user as well?

moonyuet: > Do we want that? It will likely be quite an unclear error to the user as well?

we need to also build a check on this, i will make an update later

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,66 @@
...
+    optional = True
+
+
+    @classmethod
+    def get_invalid(cls, instance):
+        invalid_mesh_type = [member for member in instance.data["members"]
+                             if not rt.isProperty(member, "mesh")]
+        if invalid_mesh_type:
+            cls.log.error("Non-mesh type objects detected")
+            return invalid_mesh_type
```
BigRoy, line 40: Honestly - it's not up to this validator to check for only mesh types in model instances. It clutters the validator and doesn't sound like what it actually should be validating. It checks for uvs on meshes. Not whether everything is a mesh.

I'd say this validator should only validate meshes, ignore the rest silently.
Then if model instances may only have meshes - then that should be a separate validator.

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
BigRoy: There's still a enter too many here. Right?

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @classmethod
+    def get_invalid(cls, instance):
+        invalid_mesh_type = [member for member in instance.data["members"]
+                             if rt.isProperty(member, "mesh")]
+        if invalid_mesh_type:
+            invalid_uv = [member for member in instance.data["members"]
+                        if not member.mesh.numTVerts > 0]
+            if invalid_uv:
+                cls.log.error("Meshes detected with invalid UVs")
```
BigRoy, line 41:
```suggestion
        meshes = [member for member in instance.data["members"]
                  if rt.isProperty(member, "mesh")]
        invalid = [member for member in meshes
                   if not member.mesh.numTVerts > 0]
        return invalid
```

BigRoy (approved):
After the last comment I think this should be good.

client/ayon_core/hosts/max/plugins/publish/validate_mesh_has_uv.py
```diff
@@ -32,15 +32,11 @@ class ValidateMeshHasUVs(pyblish.api.InstancePlugin,
...
-            invalid_uv = [member for member in instance.data["members"]
-                        if not member.mesh.numTVerts > 0]
-            if invalid_uv:
-                cls.log.error("Meshes detected with invalid UVs")
-
-            return invalid_uv
+        meshes = [member for member in instance.data["members"]
+                  if rt.isProperty(member, "mesh")]
+        invalid = [member for member in meshes
+                   if not member.mesh.numTVerts > 0]
```
BigRoy, line 38: I wonder if this would read better?
```suggestion
                   if member.mesh.numTVerts == 0]
```

moonyuet: yes it is

LiborBatek (approved):
When setting the Validator for UVs it correctly kicks in during Publishing of a model
Screenshot 2024-03-01 095949

when UVs are missing:

Screenshot 2024-03-01 095236

And if any sort of UVs applied (also just on selected only faces of the object) then it Validates ok and publish will succeed correctly...
Screenshot 2024-03-01 095732

BigRoy:
Screenshot 2024-03-01 095732

Just wanted to say that the Validating model content for modelMain should not be a `info` level log. The artist doesn't care that the validation is happening - they only need to know about data that might not be obvious or they want to double check OR any issues. This should thus be a debug log.

The project context at the top however is useful for artists however since it confirms for them that they were in the correct context, etc.

PR #123, opened 2024-02-27
Houdini: Add 'AYON_' prefix to all instance keys in metadata

iLLiCiTiT:
## Changelog Description
Add `AYON_` prefix to instance metadata in houdini instances.

## Additional info
This would be done to differentiate ayon parameters from others, even built-in liek `productName`. This change does not affect Workfile create plugin which has different logic. The prefixed keys will always override other parameters so if there is `productName` and `AYON_productName` the value from `AYON_productName` will be used.

The PR does not handle removement of old paramaters for case when scene already has existing instances without the prefix e.g. `subset` and `family`, they won't cause issues technically but may be confusing to have the same key multiple times. I don't know how to resolve that, changes related to that are welcomed.

## Testing notes:
1. It should be possible to open scene from OpenPype and Publisher should discover existing instances.
2. Save in publisher should store prefixed parameters to the metadata.
3. On reset the data should be propagated correctly from the prefixed metadata.

### DIFF
--- a/client/ayon_core/hosts/houdini/api/plugin.py
+++ b/client/ayon_core/hosts/houdini/api/plugin.py
@@ -18,6 +18,7 @@ from ayon_core.pipeline import (
 from ayon_core.lib import BoolDef
 from .lib import imprint, read, lsattr, add_self_publish_button
 
+KEY_PREFIX = "AYON_"
 
 class OpenPypeCreatorError(CreatorError):
     pass
@@ -237,14 +238,18 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
 
             node_data = read(instance)
 
+            for key, value in tuple(node_data.items()):
+                if not key.startswith(KEY_PREFIX):
+                    continue
+                new_key = key[len(KEY_PREFIX):]
+                node_data[new_key] = node_data.pop(key)
+
             # Node paths are always the full node path since that is unique
             # Because it's the node's path it's not written into attributes
             # but explicitly collected
             node_path = instance.path()
             node_data["instance_id"] = node_path
             node_data["instance_node"] = node_path
-            if "AYONProductName" in node_data:
-                node_data["productName"] = node_data.pop("AYONProductName")
 
             created_instance = CreatedInstance.from_existing(
                 node_data, self
@@ -255,7 +260,7 @@ class HoudiniCreator(NewCreator, HoudiniCreatorBase):
         for created_inst, changes in update_list:
             instance_node = hou.node(created_inst.get("instance_node"))
             new_values = {
-                key: changes[key].new_value
+                "{}_{}".format(KEY_PREFIX, key): changes[key].new_value
                 for key in changes.changed_keys
             }
             # Update parm templates and values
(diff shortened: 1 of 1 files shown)


client/ayon_core/hosts/houdini/api/plugin.py
```diff
@@ -237,14 +238,18 @@ def collect_instances(self):
 
             node_data = read(instance)
 
+            for key, value in tuple(node_data.items()):
+                if not key.startswith(KEY_PREFIX):
+                    continue
+                new_key = key[len(KEY_PREFIX):]
+                node_data[new_key] = node_data.pop(key)
```
BigRoy, line 245: So to be clear - just for backwards compatibility we're still reading the full `node_data` but just allow `AYON_` prefixed keys to override the collected data`.

This does help with backwards compatibility to load them - but will from that moment after the first 'update' have duplicated data where the old will remain lingering around. Yes?

### Setting a decent label?

Another note is that with this change now this in the Houdini UI will now all get `AYON_` prefixes on the labels, making them quite long and not as nice visually:

image

I'm wondering, if instead we should maybe update the `imprint` function to support setting labels for the attributes where it e.g. strips of `AYON_` for the nice human readable label but keeps the `AYON_` prefix for the attribute name itself.


```python
# pseudocode
def imprint(node, data, labeler=None):
    for attr_name, attr_value in data.items():
        attr_label = None
        if labeler is not None:
            attr_label = labeler(attr_name)
        set_attribute(node, attr_name, value, label=attr_label)


def labeler(name):
    """Label 'AYON_creator_attributes' as 'Creator Attributes'"""

    def uppercase_first(word):
        if not word:
            return word    
        return word[0].upper() + word[1:]

    name = name.removeprefix(KEY_PREFIX)
    
    return " ".join(uppercase_first(part) for part in split("_"))


# For example
imprint(node, {
    "task": "modeling",
    "creator_attributes": "hello",
    "publish_attributes": "boo"
}, labeler=labeler)
```

MustafaJafar: > I'm wondering, if instead we should maybe update the imprint function to support setting labels for the attributes where it e.g. strips of AYON_ for the nice human readable label but keeps the AYON_ prefix for the attribute name itself.

I tried to implement it here https://github.com/ynput/ayon-core/pull/124

BigRoy: @MustafaJafar can you try and change the target branch to this one - just so it shows just the diff to this one.

MustafaJafar: I will but, I couldn't see any difference 🤔

BigRoy: That's odd - does it contain that much more changes than this PR?

Ah wait - it does. Why does it?

iLLiCiTiT:
I hoped that we'll be able to finish the enhancement quickly before merging the base PR, but they're not production ready, and are not close to the state.
As it might block the base PR I'm closing both additional PRs. For now, the existing code works, and this was/is enhancement. I've also changed `AYONProductName` in base branch to `AYON_productName` to match changes in the PRs.

The changes can be implemented in future PRs.

PR #124, opened 2024-02-27
Houdini: Add 'AYON_' prefix when using `'imprint'` and `'read'`

MustafaJafar:
## Changelog Description
This PR adds two things: 
1. Use `AYON` name instead of `Extra` (the created tab name on ROP nodes)
2. Add 'AYON_' prefix when using `'imprint'` and `'read'`

## Additional info
This PR is my attempt to follow BigRoy's recommendation in his [comment](https://github.com/ynput/ayon-core/pull/123/files#r1504564660)
**I don't know how this integrates in the publish code.** 
**And, for some reason the publisher doesn't detect my new instances (with `AYON_` prefix)**
Many thanks to @BigRoy as his comment encouraged me to make this PR.

## Demo
imprint and read  from python code. 
each node produces the same results although one of them contain a prefix
```python
import hou
from ayon_core.hosts.houdini.api.lib import imprint, read

data = {
    "variant": "LLLL"
}

nodes = [
    hou.node('/out/cameraMain'),     # with Extra Tab, parms without prefix
    hou.node('/out/cameraAyonMain'), # with AYON Tab, parms without prefix
    hou.node('/out/cameraNewMain')   # with AYON Tab, parms with prefix
]

for i, node in enumerate(nodes):
    imprint(node, data, update=1, labeler=1)

 
for node in nodes:
    print(read(node))
```

Animation_36

## Testing notes:
I assume the same testing notes from https://github.com/ynput/ayon-core/pull/123
- It should be possible to open scene from OpenPype and Publisher should discover existing instances.
- Save in publisher should store prefixed parameters to the metadata.
- On reset the data should be propagated correctly from the prefixed metadata.


MustafaJafar:
So, I assumed that we deal with nodes via functions implemented in `ayon_core.hosts.houdini.api.lib` to deal with nodes' parameters.  
But, I found out that sometimes native calls are used e.g. (node.evalParm("...")) which doesn't know about our prefix.
----
Also, I think I found a bug in `lsattrs` when `attrs` is a string.
So, this doesn't work. 
```
>> lsattr("id")
["myNode", "myOtherNode"]
```

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -370,6 +371,16 @@ def imprint(node, data, update=False):
...
 
+        if labeler is not None:
+            if not key.startswith(KEY_PREFIX):
+                key = KEY_PREFIX + key
+
+        # Use the key with the prefix by default.
+        # But, if that key doesn't exist, use the old key.
+        key_without_prefix = key.replace(KEY_PREFIX, "", 1)
+        if node.parm(key_without_prefix):
+            key = key_without_prefix
```
BigRoy, line 382: So the way I proposed it in the comments was to make the `labeler` argument be a `function` so that you can imprint with any custom labeling making the imprint function much more re-usable for custom things.

Note that this function is inside `lib` - it should be as generic as possible if we can do so. As such this is also not the place to do any magic parm remapping.

Also, to swap the prefix you do not want to use `str.replace()` because it's not the same and may result in unexpected behavior, e.g. `my_ayon_parm` would suddenly become `my_parm` even though you wouldn't expect it.

From Py3.9 there is `str.removeprefix`. @iLLiCiTiT might know whether we still need to support older?
Otherwise do:
```python
if text.startswith(prefix):
    text = text[len(prefix):]
```

---

Side note, don't forget to update the docstring as well.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -391,12 +402,14 @@ def imprint(node, data, update=False):
 
     # Add new parm templates
     if new_parm_templates:
-        parm_folder = parm_group.findFolder("Extra")
+        parm_folder = (
+            parm_group.findFolder("AYON") or parm_group.findFolder("Extra")
+        )
```
BigRoy, line 407: Interesting. I had originally thought of making the folder name an argument to the `imprint` function because it can be quite useful for other usages of the `imprint` library function.

Anyway, I can see how this might be worth it from a backwards compatibility standpoint to do it like this.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -496,7 +509,8 @@ def read(node):
             except json.JSONDecodeError:
                 # not a json
                 pass
-        data[parameter.name()] = value
+        # Use parameter name without prefix if prefix exists.
+        data[parameter.name().replace(KEY_PREFIX, "", 1)] = value
```
BigRoy, line 513: This should not be done here - much better to prepare the data as needed inside the Creator plug-ins. Again, this is `lib` so I'd expect the functions to be somewhat generic and re-usable.

client/ayon_core/hosts/tvpaint/plugins/publish/collect_workfile_data.py
```diff
@@ -65,7 +65,7 @@ def process(self, context):
         # Collect and store current context to have reference
         current_context = {
             "project_name": context.data["projectName"],
-            "asset_name": context.data["folderPath"],
+            "folder_path": context.data["folderPath"],
```
BigRoy, line 68: This might not be due to this PR. But **why** is it sometimes `folder_path` key and sometimes `folderPath`? This is bound for confusion. Let's stick to one, no?

Or is there a reason to deviate the two?

client/ayon_core/modules/timers_manager/rest_api.py
```diff
@@ -71,11 +71,11 @@ async def get_task_time(self, request):
         data = await request.json()
         try:
             project_name = data['project_name']
-            asset_name = data['asset_name']
+            asset_name = data['folder_path']
```
BigRoy, line 74: I've said this before elsewhere - but is there a reason we're not refactoring the variable names?

iLLiCiTiT: There are multiple opened big PRs changing the same lines. It's PR conflicts hell why it's not changed.

client/ayon_core/hosts/houdini/api/lib.py
iLLiCiTiT: I'll make it quick.
- I don't think `imprint` should use the prefix by default, it should be possible to use imprint for other data that are not AYON related.
- Ideally this PR should affect only instances from create plugins, if possible.
- Don't use `replace`, use `text = text[len(prefix):]`.
- Keep in mind that some places in code might access the parameters directly without using read functions and the function should know what the parameter name is.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -391,12 +402,14 @@ def imprint(node, data, update=False):
...
-        parm_folder = parm_group.findFolder("Extra")
+        parm_folder = (
+            parm_group.findFolder("AYON") or parm_group.findFolder("Extra")
+        )
 
         # if folder doesn't exist yet, create one and append to it,
         # else append to existing one
         if not parm_folder:
-            parm_folder = hou.FolderParmTemplate("folder", "Extra")
+            parm_folder = hou.FolderParmTemplate("folder", "AYON")
```
iLLiCiTiT, line 412: Wouldn't this break backwards compatibility with older scenes?

BigRoy: As far as I know - it shouldn't. The folder isn't part of the unique identifier of the attribute - the attribute name itself is. As such, no matter what folder it is in or we move it to... it'll behave the same and is accessible the same - it's just in a visually different tab for the user after this PR.

Of course, something to test.

iLLiCiTiT:
I hoped that we'll be able to finish the enhancement quickly before merging the base PR, but they're not production ready, and are not close to the state.
As it might block the base PR I'm closing both additional PRs. For now, the existing code works, and this was/is enhancement. I've also changed `AYONProductName` in base branch to `AYON_productName` to match changes in the PRs.

The changes can be implemented in future PRs.

PR #130, opened 2024-02-28
Chore: Loader plugin expect representation context in methods

iLLiCiTiT:
## Changelog Description
LoaderPlugin does expect representation context instead of representation document in arguments or `update` and `switch`.

## Additional info
There are many reasons why this change should happen. First reason is inconsistency to `load` method where representation context is already expected. Another reason is that crutial data for update or switch are taken from representation context, which in fact are just template data to fill representation path template. We should not be dependent on this dictionary to receive crutial information like product name/type or folder name/path. We might also need to add more data to context for USD workflow.

What is representation context? In fact just a dictionary with `"representation"`, `"version"`, `"subset"`, `"asset"` and `"project"` that contain representation document with all parents.

All existing loader plugins in repository were modified to expect representation context.

## Testing notes:
- Launch host.
- Open loader tool.
- Choose product and representation to load and load it -> loader plugin was available on right click.
- Open scene inventory tool.
- Change version of loaded representation.
- Switch representation.

### Hosts
- [x] 3dsMax
- [x] Aftereffect
- [x] Blender
- [x] Fusion
- [x] Hiero
- [x] Houdini
- [x] Maya
- [x] Nuke
- [x] Photoshop
- [x] Resolve
- [x] SubstancePainter
- [x] TVpaint
- [x] Unreal
- [ ] ~~Flame~~
- [ ] ~~Harmony~~


antirotor (approved):
I just quickly went through the code changes, not tested it but it seems reasonable. What we need to clearly communicate (but I am not sure exactly where) is that all custom loaders will be broken?

iLLiCiTiT:
> What we need to clearly communicate (but I am not sure exactly where) is that all custom loaders will be broken?

I think we'll need to communicate more breaking changes that happened in ayon-core.

MustafaJafar (approved):
Works in Houdini. 
Tested Loading multiple products, updating their version and switching their folders.

dee-ynput:
> I just quickly went through the code changes, not tested it but it seems reasonable. What we need to clearly communicate (but I am not sure exactly where) is that all custom loaders will be broken?

Maybe we could inspect the signature of `plugin.update()` in `ayon_core.pipeline.load.plugins.discover_loader_plugins()` and issue a warning if there's an kwarg named `representation`?

LiborBatek:
Been tested in Blender host and works ok.

LiborBatek:
Been tested in ```3dsmax``` seems working essentially but

@moonyuet seems we having some issues with multiplying asset in the work file when performing Manage, probably different issue and not related to this PR?! Please inspect attached vid for more insight


https://github.com/ynput/ayon-core/assets/112623825/1378e19c-f7f0-44a7-86bd-81d31802543f

LiborBatek:
Works normally in ```UE``` too

LiborBatek (changes requested):
```Photoshop``` seems to having issue with ```Manage``` and ```Set Version``` and gives error msg:

```Unable to find layer metadata for [number]```

**Review by LiborBatek — COMMENTED** [review]
Been tested in ```Fusion``` and works ok.

**Comment by LiborBatek** [comment]
Been also tested in ```Substance Painter``` all working ok!

**Review by kalisp — APPROVED** [review]
Tested in AE

File: client/ayon_core/pipeline/load/plugins.py [hunk 1]
```diff
@@ -192,13 +192,13 @@ def load(self, context, name=None, namespace=None, options=None):
...
-    def update(self, container, representation):
+    def update(self, container, context):
         """Update `container` to `representation`
 
-        Arguments:
+        Args:
             container (avalon-core:container-1.0): Container to update,
                 from `host.ls()`.
-            representation (dict): Update the container to this representation.
+            context (dict): Update the container to this representation.
```
BigRoy, line 201: I really feel like we need _some_ documentation on what `context` means in this case and preferably specify what keys we expect to exist, like e.g. `folder`, `version`, `representation` etc.

Even better if we can also specify more specific type hints for these, apparently you can specify a Type hint for a dict for keys of their own types known up-front, like [described here](https://stackoverflow.com/a/54198204)

Then we could do:

```python
from typing import TypedDict

class Project(TypedDict):
    name: str
    data: dict

class Folder(TypedDict):
    name: str
    data: dict

class Version(TypedDict):
    name: int
    data: dict

class Representation(TypedDict):
    name: str
    data: dict

class RepresentationContext(TypedDict):
    project: Project
    folder: Folder
    version: Version
    representation: Representation
```

And then instead of `dict` as type here set `Representation` and then your IDE knows what it'll be a dict and if a certain key is accessed that it's of a particular type, etc.

Then, by doing so if we have a function that takes a `RepresentationContext` as input we can use these same type hints.

iLLiCiTiT: Out of scope of this PR, and current state.

64qam:
Loader in Hiero doesn't show any product: 
```Traceback (most recent call last):
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\ui\window.py", line 515, in _on_folders_refresh
    self._products_widget.refresh()
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\ui\products_widget.py", line 230, in refresh
    self._refresh_model()
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\ui\products_widget.py", line 274, in _refresh_model
    self._selected_folder_ids
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\ui\products_model.py", line 384, in refresh
    sender=PRODUCTS_MODEL_SENDER_NAME
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\control.py", line 188, in get_product_items
    project_name, folder_ids, sender)
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\models\products.py", line 229, in get_product_items
    project_name, folder_ids_to_update, sender)
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\models\products.py", line 616, in _refresh_product_items
    folder_items=folder_items
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\models\products.py", line 534, in _query_product_items_by_ids
    project_name, products, versions, folder_items=folder_items
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\models\products.py", line 462, in _create_product_items
    loaded_product_ids = self._controller.get_loaded_product_ids()
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\tools\loader\control.py", line 351, in get_loaded_product_ids
    for container in containers:
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\hosts\hiero\api\pipeline.py", line 128, in ls
    all_items = lib.get_track_items()
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\core_0.3.0-dev.1\ayon_core\hosts\hiero\api\lib.py", line 247, in get_track_items
    tracks = list(sequence.audioTracks()) + list(sequence.videoTracks())
AttributeError: 'NoneType' object has no attribute 'audioTracks'```

64qam (approved):

PR #137, opened 2024-02-29
TrayPublisher: Use AYON projects model

iLLiCiTiT:
## Changelog Description
TrayPublisher tool is using projects model from ayon tools.

## Additional info
The model does differentiate standard and library projects. Removed `ProjectModel` and `ProjectSortFilterProxy` from utils because they're not used anymore. Changed traypublisher registry to use `AYONSettingsRegistry` as base.

## Testing notes:
1. Make sure you're using ayon-core from this branch.
2. Open TrayPublisher.
3. You should see projects.
4. You should be able to select project.
5. You should be able to change project.


LiborBatek (approved):
All works ok and without any issues, tried switching projects and all good.

image

PR #138, opened 2024-02-29
Remove multi-context restrictions for publishing image sequences on farm

antirotor:
## Changelog Description
Removing assert that context data must be same as instance data will allow publishing image data to different context in one go.
That sanity check was there because we didn't support publishing to different context before (like multishot workflow), but this should be possible now. 

⚰️ This is also removing support for multiple jsons files as it was just complicating the code and it wasn't used anywhere.

## Additional info
More work on different DCCs needs to be done to seamlessly support multishot workflow - like Maya restrict only one rendering instance, Nuke enforce unique instance name (so you can't publish renderMain to multiple shots), etc.

## Testing notes:
1) Open Nuke
2) Create 3 write instances (you need to change variant name to make them unique).
3) Direct them in the Publisher to different shots.
4) Render and publish on farm.

> [!NOTE]
> This is `ayon-core` version of https://github.com/ynput/OpenPype/pull/6088

[cuID:[AY-1003](https://app.clickup.com/t/6658547/AY-1003)]


client/ayon_core/plugins/publish/collect_rendered_files.py
```diff
@@ -73,30 +73,23 @@ def _process_path(self, data, anatomy):
         data_err = "invalid json file - missing data"
         required = ["user", "comment",
                     "job", "instances", "version"]
-        assert all(elem in data.keys() for elem in required), data_err
+
+        if any(elem not in data.keys() for elem in required):
```
BigRoy, line 77: If `data` is a `dict` then this should work fine.
```suggestion
        if any(elem not in data for elem in required):
```

client/ayon_core/cli_commands.py
```diff
@@ -122,7 +125,7 @@ def publish(paths, targets=None, gui=False):
         else:
             pyblish.api.register_target("farm")
 
-        os.environ["AYON_PUBLISH_DATA"] = os.pathsep.join(paths)
+        os.environ["AYON_PUBLISH_DATA"] = os.pathsep.join(path)
```
BigRoy, line 128: Hehe, pretty sure this is not what you want.

```python
path = "hello"
print(os.pathsep.join(path))
# h;e;l;l;o
```

```suggestion
        os.environ["AYON_PUBLISH_DATA"] = path
```

client/ayon_core/cli_commands.py
```diff
@@ -111,7 +114,7 @@ def publish(paths, targets=None, gui=False):
                 app_full_name,
                 launch_type=LaunchTypes.farm_publish,
             )
-            os.environ.update(env)
+            os.environ |= env
```
BigRoy, line 117: Do we really feel this is more readable? I prefer the old style myself.

My first thought here was - what does this do?

client/ayon_core/cli_commands.py
```diff
@@ -73,6 +74,11 @@ def publish(paths, targets=None, gui=False):
         import pyblish.api
         import pyblish.util
 
+        if not isinstance(path, str):
+            warnings.warn(
+                "Passing list of paths is deprecated.",
+                DeprecationWarning)
```
BigRoy, line 80: It's deprecated, but with the logic changes will also just break completely. This should just be an error at this point right?

client/ayon_core/cli.py
```diff
@@ -102,19 +102,18 @@ def extractenvironments(output_json_path, project, asset, task, app, envgroup):
 
 
 @main_cli.command()
-@click.argument("paths", nargs=-1)
-@click.option("-t", "--targets", help="Targets module", default=None,
+@click.argument("path", nargs=1)
```
BigRoy, line 105: I believe `nargs=1` still makes the `path` variable a `list` but just containing the one entry - which means in effect we're now passing the wrong argument type to `Commands.publish()` because the docstring there is now stating it should be `str` type.

client/ayon_core/cli_commands.py
antirotor: This happens when you rush out and do only half of the thing you have in your mind. Thanks

client/ayon_core/plugins/publish/collect_rendered_files.py
antirotor: true, but I think with `.keys()` it is much more explicit that you are working on dict by my opinion. But without keys it seems it is slightly faster :)

client/ayon_core/cli_commands.py
antirotor: union operators are the future! 🚀

BigRoy: Hah, they have really been there for a long time of course ;) I'm just debating there readability.

tokejepsen (approved):
Tested successfully in Nuke 15.

PR #140, opened 2024-02-29
Resolve: mediapool loader and inventory clearing plugins

jakubjezek001:
## Changelog Description
Adding plugin for loading into mediapool with option to also add into timeline. Inventory plugin for clearing timeline item container included unused media's from mediapool. Clips in mediapool are colormanaged via remapping. 

## Additional info
Following changes are adaptation from Colorbleed's code https://github.com/BigRoy/OpenPype/commit/37da9c6f785a645d11a6274fa6ef738450ce967c#diff-3ec58869aa5ff3f62216d417256778fd5b44ea0132b953e67d7880f5e7b53df5. Thanks @bigroy
The fix dealing with Adjustment Clips on timeline which was originally here were moved into https://github.com/ynput/ayon-core/pull/92.

## Next steps

-  [x] Adding settings for metadata loading

image

### Load media (loader)
1. Open Resolve menu and chose Loader
2. select any product type version and RMB click
3. Chose `Load media` plugin and load anything
4. see that the media is added to mediapool and also into timeline
5. load again but this case use loader option box and disable `add to Timeline`
6. This time it will load only into mediapool

### Remove Unused Media (inventory)
1. make sure you are having loaded some clip into timeline
2. duplicate the mediapool clip so there is something to be erased
3. open Inventory manager
4. select any item in manager and RMB click 
5. in submenu go to actions and select `Remove Unused Media`
6. it should remove all duplicates and only one clip should remain


client/ayon_core/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,365 @@
...
+        #     # Assume no timecode was detected from the source media
+        #
+        #     fps = float(media_pool_item.GetClipProperty("FPS"))
+        #     handle_start = context["version"]["data"].get("handleStart", 0)
+        #     frame_start = context["version"]["data"].get("frameStart", 0)
+        #     frame_start_handle = frame_start - handle_start
+        #     timecode = frames_to_timecode(frame_start_handle, fps)
+        #
+        #     if timecode != start_tc:
+        #         media_pool_item.SetClipProperty("Start TC", timecode)
```
BigRoy, line 318: The workaround we discussed on discord, to in `ExtractReview` write along the `-timecode` output argument for FFMPEG works nicely for Resolve to pick up the timecode of the media, which doesn't require us to at least for those files do a "dance" like this. Which makes this much less important; and because it's currently not possible in Resolve it seems to fix this anyway via the code (as seen in [the linked thread](https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=197327) in the comment) I think we should remove this commented code.

client/ayon_core/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,365 @@
...
+        metadata = {
+            "Clip Name": "{asset[name]} {subset[name]} "
+                         "v{version[name]:03d} ({representation[name]})",
+            "Shot": "{asset[name]}",
+            "Take": "{subset[name]} v{version[name]:03d}",
+            "Comments": "{version[data][comment]}"
+        }
+        for clip_property, value in metadata.items():
+            media_pool_item.SetClipProperty(clip_property,
+                                            value.format_map(context))
```
BigRoy, line 330: We really like this on our end - but this MIGHT be studio specific.

This metadata is required for the load media stuff to work. We might want to expose these in project settings with these as defaults but allowing others to customize/remove things they don't need.

So we can have a Set metadata on load `dict` type of settings entry so once could set e.g. "Shot" metadata to "{asset[name]}", etc.

client/ayon_core/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,365 @@
...
+        version = context["version"]
+
+        # Get the start and end frame of the image sequence, incl. handles
+        frame_start = version["data"].get("frameStart", 0)
+        frame_end = version["data"].get("frameEnd", 0)
+        handle_start = version["data"].get("handleStart", 0)
+        handle_end = version["data"].get("handleEnd", 0)
+        frame_start_handle = frame_start - handle_start
+        frame_end_handle = frame_end + handle_end
+        padding = len(representation["context"].get("frame"))
```
BigRoy, line 348: I've heard of cases where a particular representation might have differing frame ranges than the version data (e.g. slates added, etc.) and as such this MIGHT be incorrect for some edge cases. We might need to at some point store those ranges in `representation["data"]` so we can reliably get frame ranges per representation instead.

BigRoy (commented):
Thanks for transferring the logic into a PR. Here are some notes to keep in mind.

client/ayon_core/hosts/resolve/plugins/load/load_media.py
BigRoy: Here's that workaround logic I implemented for outputs from reviews: https://github.com/BigRoy/OpenPype/commit/4be1d58a5c53b6ac2640f161a6d47b77540ee44b

Then you can add it as output argument in settings (e.g. here in OpenPype):
image

Then resolve loads that natively for the clip, like seen here:
image

client/ayon_core/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,363 @@
...
+            "value": "{subset[name]} v{version[name]:03d}"
+        },
+        {
+            "name": "Clip Name",
+            "value": (
+                "{asset[name]} {subset[name]} "
+                "v{version[name]:03d} ({representation[name]})"
+            )
+        }
+    ]
```
BigRoy, line 135: Why is this suddenly a list instead of a dict? This seems less efficient and less sensible since we shouldn't ever need duplicate keys either. Can we keep this a `dict`?

jakubjezek001: This is just a prep for settings in AYON. There is not the same editable dictionary objects as in OpenPype.

BigRoy: Oh boy, that's awful design really. I understand that maybe the backend is set up to be ordered for nice front-end but I'd expect the settings Pydantic stuff could auto-convert to a Python `dict` just fine, no?

It's really extra confusing here already that it's `name` instead of `key` (and should admittedly be just a dict)

tokejepsen:
> load again but this case use loader option box and disable add to Timeline
This time it will load only into mediapool

This wont load the same product if its already in the timeline. Do you mean we should remove it first?

tokejepsen (changes requested):
> Remove Unused Media (inventory)

This does not work atm. The duplicated clip is not removed.

BigRoy:
> This does not work atm. The duplicated clip is not removed.

Could you show what happens for you?
Does Resolve itself still list usages for that clip in the media pool? Right click it > Usages.
If no usages, only then should it remove it.

BigRoy:
> This wont load the same product if its already in the timeline. Do you mean we should remove it first?

There is an option box on the Load Media loader - in there you can disable "re-use existing" - when disabled this should load the media pool item another extra time into the media pool instead of continuing to use the existing media pool item of this version that you had loaded previously.

Note however: there are certain file types that I loaded where Resolve just seemed incapable of loading it again (I believe exr sequence?) even though I'm explicitly requesting there API to load it again. In those cases, the logic to 'not re-use, load again' will likely fail and actually continue to re-use since Resolve API just doesn't want to load it again into the same Media Pool Folder.

client/ayon_core/hosts/resolve/api/lib.py
```diff
@@ -330,19 +341,25 @@ def get_timeline_item(media_pool_item: object,
...
+            ti_media_pool_item = ti_clip_item.GetMediaPoolItem()
+
+            # Skip items that do not have a media pool item, like for example
+            # an "Adjustment Clip" or a "Fusion Composition" from the effects
+            # toolbox
+            if not ti_media_pool_item:
+                continue
+
+            if clip_name in ti_media_pool_item.GetClipProperty("File Name"):
+                output_timeline_item = ti_clip_item
```
jakubjezek001, line 362:
```suggestion
        for _ti_data in get_current_timeline_items():
            _ti_clip = _ti_data["clip"]["item"]
            _ti_clip_property = _ti_clip.GetMediaPoolItem().GetClipProperty
            if clip_name in _ti_clip_property("File Name"):
                output_timeline_item = _ti_clip
```

client/ayon_core/hosts/resolve/api/lib.py
```diff
@@ -330,19 +341,25 @@ def get_timeline_item(media_pool_item: object,
     Returns:
         object: resolve.TimelineItem
     """
-    _clip_property = media_pool_item.GetClipProperty
-    clip_name = _clip_property("File Name")
+    clip_name = media_pool_item.GetClipProperty("File Name")
```
jakubjezek001, line 344:
```suggestion
    _clip_property = media_pool_item.GetClipProperty
    clip_name = _clip_property("File Name")
```

client/ayon_core/hosts/resolve/plugins/load/load_media.py
jakubjezek001: This is good point for the new Representation model @antirotor is working on.

client/ayon_core/hosts/resolve/plugins/load/load_media.py
jakubjezek001: I really like this as well. It had been moved to settings.

client/ayon_core/hosts/resolve/plugins/load/load_media.py
BigRoy: This is marked resolved but I believe hasn't changed in data structure itself. Is there anything we can do to actually improve this becoming an actual `dict` datatype as opposed to the odd list structure for what is essentially intended to be a dict on the python side.

It's _likely_ slower, uses more memory but above all it's harder to read and an unnecessary bloated data structure for something that is essentially a `dict`.

jakubjezek001:
> This does not work atm. The duplicated clip is not removed.

Could you confirm there are no logs in Console in regards of removing?

Here is a video on how it works on myside and where to look for logging.

Uploading removing_mediapoolitem.mp4…

BigRoy:
> Uploading removing_mediapoolitem.mp4…

@jakubjezek001 I think you just submitted your comment too early before the file finished uploading. 🌪️

client/ayon_core/hosts/resolve/plugins/load/load_media.py
```diff
@@ -0,0 +1,346 @@
...
+        # Update MediaPoolItem filepath and metadata
+        item = container["_item"]
+
+        # Get the existing metadata before we update because the
+        # metadata gets removed
+        data = json.loads(item.GetMetadata(lib.pype_tag_name))
+
+        # Update path
+        path = self._get_filepath(representation)
+        item.ReplaceClip(path)
```
BigRoy, line 206: 🔴 

Just found a bug in this code in our production. This doesn't actually work.

Even though the documentation states this can take a filepath, it actually doesn't - it takes a folder path and will remap the MediaPoolItem with a clip in that new folder. And it looks like it will just take the FIRST media file found in that folder, instead of the one we actually want.

I'm looking into other options.

BigRoy: This fixed it for us: https://github.com/BigRoy/OpenPype/commit/e22d5dbb9fe5b9006a6b81bb156e875c47ab3638

The issue was only with sequences.

tokejepsen (changes requested):
> load again but this case use loader option box and disable add to Timeline

I cant get the `media` to load a second time. I've tried disabling the `Re-use existing` option, but it does not bring in any extra clip.

> select any item in manager and RMB click

Looks like you need to select at least the group of containers with unused media in it. Maybe this action should either be fixed so you can select any item in the manager or entry to the tool should be somewhere else than the manager?

BigRoy:
> I cant get the media to load a second time. I've tried disabling the Re-use existing option, but it does not bring in any extra clip.

What type of media are you loading? A sequence or video file? What extension?

I've had this a few times where Resolve just didn't want to load it again but it only occurred with certain files. That's the [note](https://github.com/ynput/ayon-core/pull/140#issuecomment-1988062565) I mentioned there. I didn't investigate too much as to why that happens exactly.

> Looks like you need to select at least the group of containers with unused media in it. Maybe this action should either be fixed so you can select any item in the manager or entry to the tool should be somewhere else than the manager?

Any inventory action only runs over the entries you've selected in the manager - so yes. It would only clear from your selection.
It is possible to ignore the selection in the action and remove ALL regardless of selection - but usually these actions apply to the selection.

tokejepsen:
> What type of media are you loading? A sequence or video file? What extension?

EXR sequence, so maybe this is the issue you are referring to. Maybe we can warn the user at least about the non-existent functionality of EXRs?

> Any inventory action only runs over the entries you've selected in the manager - so yes. It would only clear from your selection.
It is possible to ignore the selection in the action and remove ALL regardless of selection - but usually these actions apply to the selection.

Yeah, question is what makes sense to the user?
I would have thought an action like `Remove unused...` would remove all, but maybe we just need to change the label of the action to `Remove unused... from selection`?

BigRoy:
> Yeah, question is what makes sense to the user?
> I would have thought an action like `Remove unused...` would remove all, but maybe we just need to change the label of the action to `Remove unused... from selection`?

We have the same in [Maya "Remove unsused looks"](https://github.com/BigRoy/OpenPype/blob/colorbleed/openpype/hosts/maya/plugins/inventory/remove_unused_looks.py) which also removes only from the selected containers so might be that we were just familiar with that 'workflow'.

However there if you were to run `remove_unused_looks` without containers then it does remove all. I wonder if that's actually possible from the manager UI, to trigger it without selection. If so, then I'd also do that here - remove all if none selected, otherwise operate only on selection?

tokejepsen:
> If so, then I'd also do that here - remove all if none selected, otherwise operate only on selection?

Sounds good to me.

BigRoy:
Would be good to:

- [ ] [Fix the updating](https://github.com/ynput/ayon-core/pull/140#discussion_r1523031147)
- [ ] [Add setting the colorspace](https://github.com/BigRoy/OpenPype/commit/a403669e8389681f4041438ecf443365c0f972b9)
- [ ] [Tweak the inventory action](https://github.com/ynput/ayon-core/pull/140#issuecomment-1999304420)

BigRoy:
Might be nice to include [this code](https://github.com/BigRoy/ayon-core/commit/80e277d4ee325c95078601e85a3c551dbdee4c11) as well so that one can pick which loader should be enabled.

client/ayon_core/hosts/unreal/integration
BigRoy, line 1: Something iffy is going on here with this showing in the diff?

BigRoy (approved):
Changes look good to me. :)

BigRoy:
@jakubjezek001 could you let me know what the conditions are to test [this fix](https://github.com/ynput/ayon-core/pull/140/commits/7b1c4a8a080114e299dfe26ff52ccf63776edfcb) since I had tested both sequences and video files just fine before that.

I want to make sure I run through what's needed there. The code seems quite "involved" for something that should be trivial - so I'm hoping to find a way better than this madness.

jakubjezek001:
> I want to make sure I run through what's needed there. The code seems quite "involved" for something that should be trivial - so I'm hoping to find a way better than this madness.

There was bug in the code, for some reason sequences were not loading at all for me until the fix. Also I wanted to make sure frame range is taken from real frames rather then from version data.

BigRoy:
For what it's worth - seemed to work fine on my end also with the changes. Thanks!

PR #142, opened 2024-03-01
Publisher: Add plugin details widget

bradenjennings:
## Changelog Description
Show basic plugin details in report.

## Additional information
Took over the PR and changed it a little. Each tab in publish report viewer does refresh content only if is set as active. The plugins details widget does show details for each plugin or for selected plugins.

At this moment shows only basic information like documentation, families, order, class name, filepath and process time. There is no visual indication if was processed, skipped, succeded or failed and there are no logs visible in UI Those features can be added in future PRs (logs might make it uber slow).

The UI can get quite slow with bigger list of plugins. I was thinking that most of the information would be not created until user clicks on expand button or change to advanced mode.

### Screenshots
image

## Testing notes:
1. Open publisher in host of your choice (traypublisher is easiest).
2. Go to `Details` tab.
3. Change tab on right side to `Plugins Details`
4. You should see cards with details about plugins.


BigRoy:
Could you fix the PR title + enter a PR description? In this case, for the feature without context it would also be nice to include screenshots of the implementation or maybe reference designs (if not implemented completely yet) of what the PR intends to implement.

Keep in mind this is an active and public repository, seeing notifications about this PR I literally have no idea what this intends to solve, wants to implement and/or what I should comment on.

The current code appears to implement an extra widget in the publisher that lists only the families and filepath of a particular plug-in. Is this based on a public issue or design reference?

I feel like at this stage it's at most a draft PR or not even at a stage it maybe warrants a PR (at least not without a description?)

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,44 @@ def _set_logs(self, logs):
...
+        try:
+            plugin_filter = list(plugin_filter)[0]
+        except Exception:
+            self._lineedit_file_path.setText(str())
+            self._lineedit_families.setText(str())
+            return
+        self._plugin_filter = plugin_filter
+        plugin_item = self._report_item.plugins_items_by_id[plugin_filter]
+        self._lineedit_file_path.setText(plugin_item.filepath)
+        self._lineedit_families.setText(",".join(plugin_item.families or list()))
```
BigRoy, line 326:
```suggestion
        self._lineedit_families.setText(", ".join(plugin_item.families or list()))
```

bradenjennings:
Here is a screenshot of the new widget showing Class Info of plugin
image

bradenjennings:
> Could you fix the PR title + enter a PR description?

Sorry added now. I probably should of held off making a PR, until I provided more details

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
...
+        self._lineedit_families.setReadOnly(True)
+        form_layout.addRow("Families", self._lineedit_families)
+
+        vertical_layout.addStretch(100)
+
+        self._report_item = None
+        self._instance_filter = set()
+        self._plugin_filter = set()
+
+    def set_plugin_filter(self, plugin_filter):
```
iLLiCiTiT, line 335: Please use variable name that tells what it actually holds.
```suggestion
    def set_plugin_filter(self, plugin_ids):
```

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
...
+        try:
+            plugin_filter = list(plugin_filter)[0]
+        except Exception:
+            self._lineedit_file_path.setText(str())
+            self._lineedit_families.setText(str())
+            return
```
iLLiCiTiT, line 341: This can be handled explicitly.
```suggestion
        if not plugin_ids:
            self._lineedit_file_path.setText("")
            self._lineedit_families.setText("")
            return

        plugin_id = list(plugin_ids)[0]
```

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
...
+        # instance_labels = list()
+        # for instance_item in self._report_item.instance_items_by_id.values():
+        #     instance_labels.append(instance_item.label)
+        self._plugin_filter = plugin_filter
+        plugin_item = self._report_item.plugins_items_by_id[plugin_filter]
+        self._textedit_doc.setText(plugin_item.doc or str())
+        self._lineedit_file_path.setText(plugin_item.filepath or str())
+        families_str = ", ".join(plugin_item.families or list())
+        self._lineedit_plugin_type.setText(plugin_item.plugin_type or str())
+        # self._lineedit_instances_labels.setText(", ".join(instance_labels or list()))
```
iLLiCiTiT, line 351:
```suggestion
```

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
...
+    def set_plugin_filter(self, plugin_filter):
+        try:
+            plugin_filter = list(plugin_filter)[0]
+        except Exception:
+            self._lineedit_file_path.setText(str())
+            self._lineedit_families.setText(str())
+            return
+        # instance_labels = list()
+        # for instance_item in self._report_item.instance_items_by_id.values():
+        #     instance_labels.append(instance_item.label)
```
iLLiCiTiT, line 344:
```suggestion
```

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
...
+        self._plugin_filter = plugin_filter
+        plugin_item = self._report_item.plugins_items_by_id[plugin_filter]
+        self._textedit_doc.setText(plugin_item.doc or str())
+        self._lineedit_file_path.setText(plugin_item.filepath or str())
+        families_str = ", ".join(plugin_item.families or list())
+        self._lineedit_plugin_type.setText(plugin_item.plugin_type or str())
```
iLLiCiTiT, line 350:
```suggestion
        self._textedit_doc.setText(plugin_item.doc or "")
        self._lineedit_file_path.setText(plugin_item.filepath or "")
        families_str = ", ".join(plugin_item.families or []))
        self._lineedit_plugin_type.setText(plugin_item.plugin_type or "")
```

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
         self._output_widget.setPlainText(text)
 
 
+class ClassInfoWidget(QtWidgets.QWidget):
```
iLLiCiTiT, line 297: Rather use `PluginInfoWidget`, also please change the tab to `"Plugin info"`.

client/ayon_core/tools/publisher/control.py
```diff
@@ -150,10 +150,15 @@ def _create_plugin_data_item(self, plugin):
         label = None
         if hasattr(plugin, "label"):
             label = plugin.label
-
+        result = inspect.getmro(plugin)
+        plugin_type = result[1].__name__
```
iLLiCiTiT, line 154: What is `plugin_type`?

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -294,6 +294,70 @@ def _set_logs(self, logs):
...
+
+        # self._lineedit_instances_labels = QtWidgets.QLineEdit()
+        # self._lineedit_instances_labels.setReadOnly(True)
+        # form_layout.addRow("Runs on instances", self._lineedit_instances_labels)
+
+        self._lineedit_families = QtWidgets.QLineEdit()
+        self._lineedit_families.setReadOnly(True)
+        form_layout.addRow("Families", self._lineedit_families)
+
+        vertical_layout.addStretch(100)
```
iLLiCiTiT, line 329: Few notes about widgets in host tools and tools:
- Use parenting if possible, some hosts have different behavior of style propagation when parent is not passed on `__init__`. It also applies to adding layout to layout, in that case the styles are not sometimes propagated to children widgets.
- If possible, store all widgets at the end of `__init__` so you can easily see what the widget actually has. When someone else will look in future he should look at one place, going through each line of `__init__` is hell. Also define layout at one place so you can see what is set and what is added to the layout.
```suggestion
        info_widget = QtWidgets.QWidget(self)

        plugin_doc_widget = QtWidgets.QTextEdit(info_widget)
        plugin_doc_widget.setReadOnly(True)
        plugin_doc_widget.setMaximumHeight(300)

        plugin_path_widget = QtWidgets.QLineEdit(info_widget)
        plugin_path_widget.setReadOnly(True)

        plugin_type_widget = QtWidgets.QLineEdit(info_widget)
        plugin_type_widget.setReadOnly(True)

        plugin_families_widget = QtWidgets.QLineEdit(info_widget)
        plugin_families_widget.setReadOnly(True)

        info_layout = QtWidgets.QFormLayout(info_widget)
        info_layout.setContentsMargins(0, 0, 0, 0)
        info_layout.addRow("Documentation", plugin_doc_widget)
        info_layout.addRow("File Path", plugin_path_widget)
        info_layout.addRow("Plugin type", plugin_type_widget)
        info_layout.addRow("Families", plugin_families_widget)

        main_layout = QtWidgets.QVBoxLayout(self)
        main_layout.setContentsMargins(8, 8, 8, 8)
        main_layout.addWidget(info_widget, 0)
        main_layout.addStretch(1)

        self._plugin_doc_widget = plugin_doc_widget
        self._plugin_path_widget = plugin_path_widget
        self._plugin_type_widget = plugin_type_widget
        self._plugin_families_widget = plugin_families_widget
```

client/ayon_core/tools/publisher/control.py
BigRoy: I guess the idea was for it to be `InstancePlugin` or `ContextPlugin` according to this screenshot: https://github.com/ynput/ayon-core/pull/142#issuecomment-1975322813

However, it always being the second entry in the MRO order I think isn't what we should rely on.

I feel like we're much better of here doing something like:

```python
if isinstance(plugin, pyblish.api.InstancePlugin):
    plugin_type = "InstancePlugin"
elif isinstance(plugin, pyblish.api.ContextPlugin):
    plugin_type = "ContextPlugin"
else:
    plugin_type = "unknown"
```

Or `issubclass` if the `plugin` variable is the class and not an instance.

BigRoy:
> Here is a screenshot of the new widget showing Class Info of plugin image

Very nice seeing some work being done in this area - UI design wise I think there's still lots to improve to make it prettier, but having these starting to become accessible from here is a great start.

client/ayon_core/tools/publisher/control.py
```diff
@@ -150,10 +150,15 @@ def _create_plugin_data_item(self, plugin):
...
         return {
             "id": plugin.id,
             "name": plugin.__name__,
+            "filepath": inspect.getfile(plugin),
+            "doc": plugin.__doc__,
```
BigRoy, line 159: Should this maybe still go through [`inspect.cleandoc`](https://docs.python.org/3/library/inspect.html#inspect.cleandoc)?

If so, we might be better of directly going through [`inspect.getdoc(plugin)`](https://docs.python.org/3/library/inspect.html#inspect.getdoc)?

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
```diff
@@ -538,6 +606,7 @@ def _on_plugin_change(self, *_args):
                 plugin_ids.add(index.data(ITEM_ID_ROLE))
 
         self._logs_text_widget.set_plugin_filter(plugin_ids)
+        self._class_info_widget.set_plugin_filter(plugin_ids)
```
BigRoy, line 609: Preferably these would only end up triggering 'their logic' if they are currently visible. In these cases both these widgets maybe aren't doing too much complicated work yet and thus it might be quite fast - but as we expand on these UIs we might want to, not necessarily in this PR, just pass on 'marking the widgets dirty for a refresh' and only actually trigger the refresh code if the widget is visible?

client/ayon_core/tools/publisher/control.py
iLLiCiTiT: In that case we have to face fact, that in pyblish you can't rely on class inheritance.

```python
if plugin.__instanceEnabled__:
    plugin_type = "Instance plugin"
else:
    plugin_type = "Context plugin"
```

BigRoy: Yes, thanks! :)

client/ayon_core/tools/publisher/publish_report_viewer/widgets.py
iLLiCiTiT: Agree, but not in this PR please.

client/ayon_core/tools/publisher/control.py
bradenjennings: Is this how I should distinguish between plugin types
```
if plugin.__instanceEnabled__:
    plugin_type = "Instance plugin"
else:
    plugin_type = "Context plugin"
```
Does this really work reliably?

bradenjennings: thanks this seems to work reliably
```
if plugin.__instanceEnabled__:
    plugin_type = "Instance plugin"
else:
    plugin_type = "Context plugin"
```

client/ayon_core/tools/publisher/control.py
bradenjennings: I now use 
`inspect.getdoc(plugin)`

jakubjezek001:
It is looking great and definitely good direction to go. Suggestion from me would be to make the text look like it is not editable. At the moment it is evoking to change the text as in form inputs. 
image

BigRoy:
> Suggestion from me would be to make the text look like it is not editable. At the moment it is evoking to change the text as in form inputs.

It is set to Read Only so I guess this is an AYON stylesheet issue maybe? Or do you mean you want specific stylesheet overrides for these?

BigRoy:
The documentation does look off for plug-ins that do not specify any docstring due to than inheriting from Pyblish plug-ins classes?

See:
image

---

Can we make the documentation field extend more in height if the UI provides the empty space?

image

iLLiCiTiT:
I would change the `File Path`, `Plugint type` and maybe even `Families` fields to `QLabel` with `TextBrowserInteraction` flag. They should be probably also moved before the documentation. I would maybe skip the `Documentation` label or move it on row before so the docs input is full width.

client/ayon_core/tools/publisher/publish_report_viewer/report_items.py
```diff
@@ -13,6 +13,11 @@ def __init__(self, plugin_data):
         self.skipped = plugin_data["skipped"]
         self.passed = plugin_data["passed"]
 
+        self.doc = plugin_data["doc"]
+        self.filepath = plugin_data["filepath"]
+        self.plugin_type = plugin_data["plugin_type"]
+        self.families = plugin_data["families"]
```
iLLiCiTiT, line 19: This is also used as standalone tool where you can drop report files, but the old files won't have these new keys.
```suggestion
        self.doc = plugin_data.get("doc", "N/A")
        self.filepath = plugin_data.get("filepath", "N/A")
        self.plugin_type = plugin_data.get("plugin_type", "N/A")
        self.families = plugin_data.get("families", "N/A")
```

bradenjennings:
> I would change the `File Path`, `Plugint type` and maybe even `Families` fields to `QLabel` with `TextBrowserInteraction` flag. They should be probably also moved before the documentation. I would maybe skip the `Documentation` label or move it on row before so the docs input is full width (that would require `QGridLayout`).
> 
> EDITED: And yeah, probable remove the `addStretch(1)` and add stretch to the documentation input field so it fills the height.

I've done all that and pushed. 

Here is a screenshot. NOTE: label text has been selected as suggested.
image

bradenjennings:
So do we need to override the stylsheet if read only items, or should the core stylesheet already be doing this for us?

iLLiCiTiT:
>  NOTE: label text has been selected as suggested.

They're still `QLineEdit` widgets and not moved above the docstring.

mkolar:
@iLLiCiTiT I'm going to assume that this will need to be redone, correct? No point merging it in semi done state

iLLiCiTiT:
> @iLLiCiTiT I'm going to assume that this will need to be redone, correct? No point merging it in semi done state

Not ready to be merged, but can be used as base.

MilaKudr (approved):

BigRoy (commented):
So it works - but... 

image

1. The Plugin Details tab is really slow to open, unless you have a plugin or a few selected. But displaying all takes a long time (about a second at best? but the majority of the time it takes about 4-5 seconds for me? and clicking just reset in Maya sometimes now takes 15 seconds for me, in an empty scene.)
    - I also have the feeling the publisher UI is a bit slower in general, even on another tab on reset it took a long time for me.
    - If the plugin details tab is open on Details.. but you go to the Create tab (and hence Plugins Details is not visible anymore) it's still extremely slow on reset because it still seems to refresh the widgets?

2. Some docstrings represent some characters as 'special characters' e.g. see `cleanupEmptyDirs` here:
image

3. When no processing has occurred the "Time" displays: 0.000s but it may be nicer for it to state that it hasn't run yet.

4. I wonder if families are better listed as `", ".join(families)` instead of as the Python list being printed?

When on the details pane with Plugin Details open and you publish, it deselects your current selection hence it's realllly slow to get to the publish result. (Just because for whatever reason that list is really slow to generate?)

5. I wonder if it makes sense to also print which instances it processed. E.g. Processed: instance1, instance2 or Processed: context. Bonus points if somehow we could track the time spent per instance, but that's a nice to have.

Overall  my biggest concern is **the performance of the tab.** If that wasn't there I would approve this now and leave maybe the other notes for improvements.

iLLiCiTiT:
> The Plugin Details tab is really slow to open, unless you have a plugin or a few selected. But displaying all takes a long time (about a second at best? but the majority of the time it takes about 4-5 seconds for me? and clicking just reset in Maya sometimes now takes 15 seconds for me, in an empty scene.)

It is slow only if the tab is activated. With my testing it doesn't really matter if all plugins are shown, or just one of them (didn't test in nuke or maya so I might add it another try). I'm not sure how much artists would go there? -> Meaning how much is it important for this PR?

> If the plugin details tab is open on Details.. but you go to the Create tab (and hence Plugins Details is not visible anymore) it's still extremely slow on reset because it still seems to refresh the widgets?

It might, only if you selected `Plugins Details` tab, but I'm 50:50 sure it should not be slower even then (will re-validate).

> When no processing has occurred the "Time" displays: 0.000s but it may be nicer for it to state that it hasn't run yet.

Makes sense.

> I wonder if families are better listed as ", ".join(families) instead of as the Python list being printed?

Fine for me.

> I wonder if it makes sense to also print which instances it processed. E.g. Processed: instance1, instance2 or Processed: context. Bonus points if somehow we could track the time spent per instance, but that's a nice to have.

Agree, feature for future PR.

BigRoy:
> It might, only if you selected `Plugins Details` tab, but I'm 50:50 sure it should not be slower even then (will re-validate).

Sure - I was just wondering whether it made sense to just not default to listing all but only ever display for an active selection to avoid the common case where by default the tab is really slow to open. And by default it would just show "Select one or multiple plugins to display their details." Also because the FULL list of data is hardly usable anyway because it's so much text? That'd only be really useful if you could e.g. reorder them by e.g. Time taken. So that e.g. you would have the slowest plug-ins displayed at the top, or the entries that generated the most errors or whatever.

iLLiCiTiT:
NOTE: When I try to re-create the details widget on selection change, it becomes really laggy. I found out that filling all plugins and hiding/showing them is much faster thant recreating the widgets and layout.

iLLiCiTiT:
Made small modifications.
- It is required to select plugin/s to show their details.
- Small enhancements of the speed if Plugins Details tab is not visible.
- Families are joined by `,`.
- Time does not show spent time if plugin was skipped/did not stary yet.

BigRoy (approved):
Thanks for the changes - this is a much better first version since it feels less slow in usage. It's much better.

For whatever reason @iLLiCiTiT the first switch to Plugin Details tab, even with nothing selected, is still slow.

Other than that, it works as intended!

iLLiCiTiT:
> For whatever reason @iLLiCiTiT the first switch to Plugin Details tab, even with nothing selected, is still slow.

It is because of this https://github.com/ynput/ayon-core/pull/142#issuecomment-2298996333 . Changing it to create only widgets for selected plugins makes is more laggy, would require more optimization, future enhancement 🙂  .

PR #143, opened 2024-03-03
Publisher: Double click at instance switch to publish tab

bradenjennings:
## Description
Double click on instances under `Create` tab changes tab to `Publish`.

### Additional information
The double click is ignored if happens on checkbox or button. Double click on instances group does not change the tab either.

**Additional context**
image

Source issue description https://github.com/ynput/OpenPype/issues/4567

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -534,6 +534,8 @@ class InstanceCardView(AbstractInstanceView):
     Wrapper of all widgets in card view.
     """
 
+    double_clicked = QtCore.Signal()
+
     def __init__(self, controller, parent):
         super(InstanceCardView, self).__init__(parent)
 
@@ -577,6 +579,9 @@ class InstanceCardView(AbstractInstanceView):
             QtWidgets.QSizePolicy.Minimum,
             self.sizePolicy().verticalPolicy()
         )
+
+    def mouseDoubleClickEvent(self, event):
+        self.double_clicked.emit()
 
     def sizeHint(self):
         """Modify sizeHint based on visibility of scroll bars."""
@@ -715,6 +720,7 @@ class InstanceCardView(AbstractInstanceView):
                 )
                 group_widget.active_changed.connect(self._on_active_changed)
                 group_widget.selected.connect(self._on_widget_selection)
+                # group_widget.double_clicked.connect(self._on_widget_double_clicked)
                 self._content_layout.insertWidget(widget_idx, group_widget)
                 self._widgets_by_group[group_name] = group_widget
 
@@ -824,6 +830,11 @@ class InstanceCardView(AbstractInstanceView):
             self._select_item_extend_to(instance_id, group_name, new_widget)
 
         self.selection_changed.emit()
+
+    # def _on_widget_double_clicked(self):
+    #     print("_on_widget_double_clicked")
+    #     widgets = self._get_selected_widgets()
+    #     print(widgets)
 
     def _select_item_clear(self, instance_id, group_name, new_widget):
         """Select specific item by instance id and clear previous selection.
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -581,7 +581,8 @@ class InstanceCardView(AbstractInstanceView):
         )
 
     def mouseDoubleClickEvent(self, event):
-        self.double_clicked.emit()
+        if event.button() == QtCore.Qt.LeftButton:
+            self.double_clicked.emit()
 
     def sizeHint(self):
         """Modify sizeHint based on visibility of scroll bars."""
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -720,7 +720,6 @@ class InstanceCardView(AbstractInstanceView):
                 )
                 group_widget.active_changed.connect(self._on_active_changed)
                 group_widget.selected.connect(self._on_widget_selection)
-                # group_widget.double_clicked.connect(self._on_widget_double_clicked)
                 self._content_layout.insertWidget(widget_idx, group_widget)
                 self._widgets_by_group[group_name] = group_widget
 
@@ -830,11 +829,6 @@ class InstanceCardView(AbstractInstanceView):
             self._select_item_extend_to(instance_id, group_name, new_widget)
 
         self.selection_changed.emit()
-
-    # def _on_widget_double_clicked(self):
-    #     print("_on_widget_double_clicked")
-    #     widgets = self._get_selected_widgets()
-    #     print(widgets)
 
     def _select_item_clear(self, instance_id, group_name, new_widget):
         """Select specific item by instance id and clear previous selection.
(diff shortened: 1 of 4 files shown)


client/ayon_core/tools/publisher/widgets/overview_widget.py
```diff
@@ -293,6 +299,10 @@ def _on_product_change(self, *_args):
             instances, context_selected, convertor_identifiers
         )
 
+    def _on_double_clicked(self):
+        from ayon_core.tools.utils import host_tools
+        host_tools.show_publisher(tab="publish")
```
iLLiCiTiT, line 304: Please do not do it this way. The publisher tool does not have to shown using `host_tools`. Resubmit signal to `PublisherWindow`.

client/ayon_core/tools/publisher/widgets/list_view_widgets.py
```diff
@@ -454,6 +457,7 @@ def __init__(self, controller, parent):
         instance_view.collapsed.connect(self._on_collapse)
         instance_view.expanded.connect(self._on_expand)
         instance_view.toggle_requested.connect(self._on_toggle_request)
+        instance_view.doubleClicked.connect(self._on_double_clicked)
```
BigRoy, line 460: Is there a particular reason this isn't just like this and don't need the `_on_double_clicked` in-between?
```suggestion
        instance_view.doubleClicked.connect(self.double_clicked)
```

bradenjennings: no there is no particular reason why.
I implemented the suggestion.

client/ayon_core/tools/publisher/widgets/overview_widget.py
bradenjennings: I am now emitting a publish_tab_requested signal.

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -578,6 +580,9 @@ def __init__(self, controller, parent):
             self.sizePolicy().verticalPolicy()
         )
 
+    def mouseDoubleClickEvent(self, event):
+        self.double_clicked.emit()
```
BigRoy, line 584: Does this now always go to the publish tab if double clicking - even if not double clicking on an instance? I assume it might but I'm not sure if that's the intended behavior. Is it what we want?

jakubjezek001: the approach is great! thanks

jakubjezek001 (approved):
LGTM! tested in Traypublisher

BigRoy:
My notes:
- The double click behavior also occurs if one e.g. double clicks the "instance active" state (the slider). Should that be avoided since the click is then technically handled already?
- It also happens on double right mouse click and middle mouse clicks. Maybe odd? (Not sure whether we care)
- On the list view (not the card view) double clicking the empty area below instances does nothing. Maybe inconsistent between the two?
- Clicking the header (not the dropdown arrow) in the List view collapses/expands the instances section. Now if you take a quick peek at the contents if collapsed currently, have seen it and rather click it again to close it'll get registered as the same double click. Again, since I feel the "click" is handled by the UI already it should not trigger this middle click to go to the other view. (Same like the slider bullet point).

Other than that, I tested Maya, Houdini and Tray Publisher. Behaved the same in all 👍

iLLiCiTiT:
I have question. Why we need this double click when you can actually just click the big button which appears when you hover to instance view?

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
iLLiCiTiT: > It switches to the publish tab if you double click anywhere within the card view widget.

I would not do that. It should happen only if you double click the card itself. There might be other widgets that want to capture double click on the card itself.

BigRoy:
> Why we need this double click when you can actually just click the big button which appears when you hover to instance view?

Because user testing had shown that users expected the double clicking to work to 'edit it' but it didn't. Plus it's much easier to click the instance twice if you want to edit a specific one instead of clicking instance, then arrow (or arrow, then instance).

So, mostly came from user testing I suppose.

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
BigRoy: > I would not do that. It should happen only if you double click the card itself.

I must admit, I kind of agree - similarly if the "enable/disable" slider on the instance was double clicked (and thus the click was 'handled' already it should not trigger the double click behavior either) - kind of like my [notes here](https://github.com/ynput/ayon-core/pull/143#issuecomment-1984724600).

jakubjezek001: > similarly if the "enable/disable" slider on the instance was double clicked (and thus the click was 'handled' already it should not trigger the double click behavior either)

I have just tested the double click on the slider and nothing happened. It was treated as two single click in sequence, so it disabled and then immediately enabled back again.

bradenjennings:
> * It also happens on double right mouse click and middle mouse clicks. Maybe odd? (Not sure whether we care)

I've fixed this behaviour for both views

bradenjennings:
> * On the list view (not the card view) double clicking the empty area below instances does nothing. Maybe inconsistent between the two?

I've corrected this behavior. double clicking the empty area in both the card view and tree view, both switch tabs to publish.

bradenjennings:
> The double click behavior also occurs if one e.g. double clicks the "instance active" state (the slider). Should that be avoided since the click is then technically handled already?

I'm not sure how to differentiate between this click state. The card view isn't a Qt view so its difficult to hit test the children.
Personally I don't think it's likely the user is going to double click the instance active state with the intention of toggling on and off quickly.
Is it possible someone can point me to some code that would help?

bradenjennings:
> * Clicking the header (not the dropdown arrow) in the List view collapses/expands the instances section. Now if you take a quick peek at the contents if collapsed currently, have seen it and rather click it again to close it'll get registered as the same double click. Again, since I feel the "click" is handled by the UI already it should not trigger this middle click to go to the other view. (Same like the slider bullet point).

The activation time for double click requires such a rapid speed, it's unlikely the user would toggle expand on and off that quickly. 

I'm not sure how to fix this particular behavior. It's a core part of how tree view work. Can someone point to some code that would help?

bradenjennings:
this ticket is in approved state. Does that mean I can merge it in its current state?

tokejepsen:
> this ticket is in approved state. Does that mean I can merge it in its current state?

We usually operate with 2 approvals for a merge. Second approver would merge as well. Post on discord (dev and review channel) that it ready to merge for better visibility.

BigRoy:
> I've corrected this behavior. double clicking the empty area in both the card view and tree view, both switch tabs to publish.

This sounds like the opposite behavior of what @iLLiCiTiT [described here](https://github.com/ynput/ayon-core/pull/143#discussion_r1517474897). I understood he said: Only switch to publish tab when double clicking on a card or list view instance entry - do not switch when double clicking the empty area.

So just tagging @iLLiCiTiT to see whether he's ok with the current behavior or not.

iLLiCiTiT:
> The activation time for double click requires such a rapid speed, it's unlikely the user would toggle expand on and off that quickly.

Oh, you don't know artists.

> Only switch to publish tab when double clicking on a card or list view instance entry - do not switch when double clicking the empty area.

Yes, that's correct.

mkolar:
@iLLiCiTiT can I get a thumbs up or down here? We want the feature, just not sure if it's mergable

iLLiCiTiT:
This needs some changes.

jakubjezek001 (approved):
works as expected

PR #144, opened 2024-03-05
Updating tools defaults

m-u-r-p-h-y:
## Changelog Description
Updating the tools defaults for Ayon

- updating version of plugins to recent versions
- setting up plugins per DCC
- adding (example) to labels so it is clear every tool needs to be set again to accommodate each studio setup
- changing most env definitions to additive
- adding all three platforms to the setup
- adding new tools (mGear, Yeti)

## Testing notes:
1. build a new application addon using create_ayon_addons.py in server_addon
2. upload the new addon to the server
3. make the new addon active in your bundle
4. test the new defaults


server_addon/applications/server/tools.json
```diff
@@ -1,55 +1,149 @@
...
+                    "name": "5-4-2-7",
+                    "label": "",
+                    "host_names": [
+                        "houdini"
+                    ],
+                    "environment": "{\n    \"HTOA_VERSION\": \"5.4.2.7\"\n}",
+                    "app_variants": []
+                }
+            ],
+            "environment": "{\n    \"_comment_\": \"{STUDIO_SW} points to software repository. Can be defined in Core addon globally\",\n\n    \"HOUDINI_PATH\": [\n        \"{STUDIO_SW}/APP/HTOA/{HTOA_VERSION}/HOUDINI{HOUDINI_VERSION}/WINDOWS/htoa-6.1.3.3_rdb15014_houdini-{HTOA_VERSION}\",\n        \"{HOUDINI_PATH}\"\n    ],\n    \"PATH\": {\n        \"windows\": [\n            \"{STUDIO_SW}/APP/HTOA/{HTOA_VERSION}/HOUDINI{HOUDINI_VERSION}/WINDOWS/htoa-6.1.3.3_rdb15014_houdini-{HTOA_VERSION}/scripts/bin\",\n            \"{PATH}\"\n        ]\n    }\n}"
```
BigRoy, line 17: Is this comment a literal key/value in settings - or is it somehow used by the front end to turn this into a literal comment? I understand the thinking - I just feel it's weird that we're actually abusing the defaults for this.

I almost feel like we might be better of setting the "Placeholder text" for the settings field instead if we really just want to provide hints?

m-u-r-p-h-y: as there is no possibility to have comments in json this way was the simplest to point out the need of setting up STUDIO_SW variable to make the example work

m-u-r-p-h-y: also the advantage here is that you can easily remove it, while adjusting it to your studio environment. Any hardcoded hint will stay there, even though it is not valid to your setup anymore.
I would definitely not use the word `abuse` but creative use instead ;-)

BigRoy: Yup - thanks. I mean, it's not a huge issue - just wondered if we could have better tools at our disposal for this. E.g. maybe a button that may apply a 'template' preset or whatever based on some built-in examples as opposed to having them predefined completely.

Anyway, nice work on cleanup!

LiborBatek (approved):
LGTM

PR #147, opened 2024-03-06
Publisher: context menu option to revert each attribute to default

bradenjennings:
## Changelog Description
Publisher - now has a "Revert to default" context menu option for each menu of various widgets ( Int, Float, String and Enum)

## Additional info

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_core/tools/attribute_defs/widgets.py
+++ b/client/ayon_core/tools/attribute_defs/widgets.py
@@ -313,13 +313,28 @@ class NumberAttrWidget(_BaseAttrDefWidget):
         self.main_layout.addWidget(multisel_widget, 0)
 
     def eventFilter(self, obj, event):
-        if (
+        if event.type() == QtCore.QEvent.Type.ContextMenu:
+            QtCore.QTimer.singleShot(0, self._on_menu_timeout)
+            super(NumberAttrWidget, self).contextMenuEvent(event)
+        elif (
             self._multivalue
             and obj is self._input_widget
             and event.type() == QtCore.QEvent.FocusOut
         ):
             self._set_multiselection_visible(True)
         return False
+
+    def _on_menu_timeout(self):
+        menu = self.findChild(QtWidgets.QMenu, "qt_edit_menu")
+        if menu is not None:
+            first_action = menu.actionAt(QtCore.QPoint())
+            action = QtWidgets.QAction(menu)
+            action.setText("Reset to default")
+            action.triggered.connect(self.reset_to_default_requested)
+            menu.insertAction(first_action, action)
+
+    def reset_to_default_requested(self):
+        self.set_value(self.attr_def.default)
 
     def current_value(self):
         return self._input_widget.value()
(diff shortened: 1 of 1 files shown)


bradenjennings:
Revert to default menu options now appears for Int, Float, String and Enum menus. 

It adds itself to existing menus, instead of replacing them.

image

jakubjezek001:
image

Not sure how much more complexity it would add, but this approach seems to be best UX. Reset button on right of each row.

tokejepsen:
Maybe @Innders could chip in with ideas or whether there is some convention in AYON already.

BigRoy:
> image BlackMagic Resolve
> 
> Not sure how much more complexity it would add, but this approach seems to be best UX. Reset button on right of each row.

`qargparse` actually does that and thus the OpenPype Loader Option box fields also do that, e.g. see the button to right of Remove Publish Folder here:

image

There does seem to be a bug though that if you click that reset button it doesn't disappear in that particular UI - but it's a decent reference.

iLLiCiTiT:
I do agree that right click is not good option to reset value, consideting I want to reset all attributes to default state.

BigRoy:
> I do agree that right click is not good option to reset value, **consideting I want to reset all attributes to default state.**

Well, not _always_. I usually want to revert an attribute or two - not the full instance. But if I'd want to revert 10, then yes. This will be very slow. However, the context menu could show a "shortcut" label like e.g. `CTRL+LMB` and when clicking that on an attribute that it reverts to default. Works perfectly fine like that in Houdini.

bradenjennings:
There is a separate ticket for a reset all attributes button. 
So in this case you wouldn't need to click every attribute directly.
https://github.com/ynput/ayon-core/pull/153

What's the consensus should we keep the context menu, or add a button next to each attribute. 
Personally I find its a lot of visual noise having a button next to each attribute.

bradenjennings:
I find the Houdini context menu option to be a clean non distacting implementation.

iLLiCiTiT:
> I find the Houdini context menu option to be a clean non distacting implementation.

You just described hidden feature, that requires 2 clicks. But I'm not known for UX, so hard to tell.

client/ayon_core/tools/attribute_defs/widgets.py
```diff
@@ -313,14 +313,29 @@ def _ui_init(self):
...
+    def _on_menu_timeout(self):
+        menu = self.findChild(QtWidgets.QMenu, "qt_edit_menu")
+        if menu is not None:
+            first_action = menu.actionAt(QtCore.QPoint())
+            action = QtWidgets.QAction(menu)
+            action.setText("Reset to default")
+            action.triggered.connect(self.reset_to_default_requested)
+            menu.insertAction(first_action, action)
+
+    def reset_to_default_requested(self):
```
iLLiCiTiT, line 336: The method name is not ideal. It should be something like `set_to_default_value`, or private method for signal handling and in that case `_on_default_value_requested`.

client/ayon_core/tools/attribute_defs/widgets.py
```diff
@@ -459,6 +494,23 @@ def _ui_init(self):
         self.main_layout.addWidget(input_widget, 0)
         self.main_layout.addStretch(1)
 
+        self.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
+        self.customContextMenuRequested.connect(self._on_context_menu)
+
+    def _on_context_menu(self, pos):
```
iLLiCiTiT, line 500: Menu and action does not have to be stored under `self` they live and die during stack of method.

client/ayon_core/tools/attribute_defs/widgets.py
```diff
@@ -313,14 +313,29 @@ def _ui_init(self):
...
+        elif (
             self._multivalue
             and obj is self._input_widget
             and event.type() == QtCore.QEvent.FocusOut
         ):
             self._set_multiselection_visible(True)
         return False
 
+    def _on_menu_timeout(self):
+        menu = self.findChild(QtWidgets.QMenu, "qt_edit_menu")
```
iLLiCiTiT, line 328: I wonder if all DCCs are using this name of menu, considering Houdini, Nuke, Maya and SubstancePainter etc. Could someone validate those DCCs if the right click does work for text fields there?

iLLiCiTiT:
> That way you can also perform the revert in one click per attribute.

So, let it in context menu with hint to use Ctrl + MMB instead (oh, Mac users will be pissed 😈 ).

BigRoy:
> (oh, Mac users will be pissed 😈 ).

Would be Command + MMB then I suppose. No idea how that works. Didn't Qt have something built-in for that and also to labelize menu entries, etc?

iLLiCiTiT:
> Would be Command + MMB then I suppose.

It was not about CTRL, you would have to find MMB :)

BigRoy:
> It was not about CTRL, you would have to find MMB :)

🤦‍♂️ Of course. Anyway, not sure what Houdini does there.

mkolar:
Seriously folks. Context menu is more than enough here. We're talking about something that is probably used very occasionally and it's super common to add these kinds of things to right click. It's a few attributes, so resetting all at once is really on a fringe of usefulness compared to the effort.

iLLiCiTiT:
Closed in favor of https://github.com/ynput/ayon-core/pull/986

PR #149, opened 2024-03-06
Resolve: Adjust clip start if slate is present

jakubjezek001:
## Changelog Description
Starting of clip should be adjusted if slate frame is detected.

## Additional info
Originally submitted here https://github.com/ynput/OpenPype/pull/6126

## Testing notes:
1. in Resolve load into the timeline representation with slate
2. Slate should be trimmed out


jakubjezek001:
@jrsndl here is the alternative on ayon-core.

jakubjezek001 (commented):
This was adapted originally from OpenPype Repo and I was testing it in both and it is working, so strangely although I am author of this PR I am approving it and merging it as well ;)

PR #154, opened 2024-03-07
Batch Folder Publisher tool

bradenjennings:
Port of https://github.com/ynput/OpenPype/pull/6144

## Changelog Description
Batch Publisher can be opened from AYON launcher tray. 
It is a tool for batch publishing multiple files each as a separate product.

## Testing notes:
1. git clone the OpenPype repo
2. checkout this branch "Feature/op 7692 folder batch publishing tool"
3. Setup dev environment on AYON server and set OpenPype directory for bundle to directory of OpenPype
4. Ask @bradenjennings for ayon internal testing data
5. Make sure your AYON server has a project available. For example make sure you setup AYON server on AY_CG_demo project .
6. Open Batch publisher for AYON launcher tray
7. Choose the browse button and pick the root directory of test data, here
C:\errolDotting
8. The table view should show some fbx file paths
9. Choose your folder (asset) to make the line "defined"

### DIFF
--- a/client/ayon_core/hosts/batchpublisher/__init__.py
+++ b/client/ayon_core/hosts/batchpublisher/__init__.py
@@ -1,0 +1,5 @@
+from .addon import BatchPublishAddon
+
+__all__ = (
+    "BatchPublishAddon",
+)
--- a/client/ayon_core/hosts/batchpublisher/addon.py
+++ b/client/ayon_core/hosts/batchpublisher/addon.py
@@ -52,13 +52,13 @@ class BatchPublishAddon(AYONAddon, IHostAddon, ITrayAction):
         import ayon_core.hosts.batchpublisher.ui.batch_publisher_view
         import ayon_core.hosts.batchpublisher.ui.window
 
-        # # TODO: These lines are only for testing current branch
-        # reload(ayon_core.hosts.batchpublisher.controller)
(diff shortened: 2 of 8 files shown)


iLLiCiTiT:
I think this should be new addon instead of adding it to ayon-core.

bradenjennings:
> I think this should be new addon instead of adding it to ayon-core.

Is there any documentation about how to make a new add on?

moonyuet:
> > I think this should be new addon instead of adding it to ayon-core.
> 
> Is there any documentation about how to make a new add on?

@mkolar @m-u-r-p-h-y can you help Barden on this? Thank you!

mkolar:
That is a pretty good point actually. Let's do it. We'll create a repo today so it could be ported.

jakubjezek001:
Probably unrelated of this particular PR but here are some ideas for enhancements:

- Copying /pasting cell values triggered by `CTRL+C/V`
- multiple cells selection and then CTRL+V would past to all selected cells
- disabling validators optional box - Its failing now since folder attributes are not the same as frame ranges of representations
- deal with multiple versions of the same product - since the situation can happen that productions will need to ingest multiple different versions. At the moment this will fail on Validators.

kalisp:
Replaced by https://github.com/ynput/ayon-batch-publisher/pull/1

PR #164, opened 2024-03-08
Max: Allows user to select material duplicates option during loading max scene object

moonyuet:
## Changelog Description
This PR is to add the options/dialogs to allow user to select material duplicates options when loading max scene via Loader.
(More Info to be updated)
## Additional info
The material duplicate option would be at the default setting `promptMtlDups` when 3dsmax is at headless mode
## Testing notes:
Load...
1. Launch 3dsMax 
2. Go to AYON -> Load...
3. Right Click load Max Scene and you will find memo icon 
image
4. Click that memo icon, you can select your preferred material duplicate options.
image
5. Click Ok and it would load the max scene objects with your choice on material duplicate option.
Updating/Switching Version...
1. Go to AYON -> Manage...
2. Set Version with the max scene product type
3. After finishing to choose your version, it will pop up a dialog to ask to choose material duplicate types.
4. Click Ok button
5. The scene will update and load the objects with the chosen material duplicate type.
6. If clicking cancel button, the scene will still update and load the objects with the default material duplicate type.


client/ayon_core/hosts/max/plugins/load/load_max_scene.py
```diff
@@ -67,7 +129,9 @@ def update(self, container, context):
         for prev_max_obj in prev_max_objects:
             if rt.isValidNode(prev_max_obj):  # noqa
                 rt.Delete(prev_max_obj)
-        rt.MergeMaxFile(path, quiet=True)
+        window = MaterialDupOptionsWindow(self.mtl_dup_enum)
+        window.exec_()
```
BigRoy, line 133: Note: Updating should work in headless/command line mode as well without UI interference - make sure to not always force a dialog.

BigRoy (commented):
It's hard for me to really understand whether this is the best design approach. You might have 10 max scenes loaded. I'd personally not want to click through 10 popups on update.

So can't really decide on what the best approach is for this use case but it does feel like an anti-pattern from the outside.

moonyuet:
> It's hard for me to really understand whether this is the best design approach. You might have 10 max scenes loaded. I'd personally not want to click through 10 popups on update.
> 
> So can't really decide on what the best approach is for this use case but it does feel like an anti-pattern from the outside.

I am thinking about adding as loader setting into 3dsmax addon, when talking about the best approach.

LiborBatek (approved):
All seems working fine in multiple occasions and circumstances...did test different scenarios with ```material``` and all options seems working as expected.

LGTM

image

BigRoy:
> All seems working fine in multiple occasions and circumstances...did test different scenarios with `material` and all options seems working as expected.
> 
> LGTM
> image

Can we please improve the labels on those drop down menu entries? :) What is `promptMtlDups`? Is there no way to improve readability on those labels?

moonyuet:
> > All seems working fine in multiple occasions and circumstances...did test different scenarios with `material` and all options seems working as expected.
> > LGTM
> > image
> 
> Can we please improve the labels on those drop down menu entries? :) What is `promptMtlDups`? Is there no way to improve readability on those labels?
> 
> Also @LiborBatek have you also tested the prompt/dialog on Update instead of Load?

I edited the code to be more descriptives and artist friendly in terms of the choosing options.
image
image

BigRoy:
> > > All seems working fine in multiple occasions and circumstances...did test different scenarios with `material` and all options seems working as expected.
> > > LGTM
> > > image
> > 
> > 
> > Can we please improve the labels on those drop down menu entries? :) What is `promptMtlDups`? Is there no way to improve readability on those labels?
> > Also @LiborBatek have you also tested the prompt/dialog on Update instead of Load?
> 
> I edited the code to be more descriptives and artist friendly in terms of the choosing options. image image

Much better - so what is **Prompt Material**? Is that like a Pick & Choose dialog?

moonyuet:
> > > > All seems working fine in multiple occasions and circumstances...did test different scenarios with `material` and all options seems working as expected.
> > > > LGTM
> > > > image
> > > 
> > > 
> > > Can we please improve the labels on those drop down menu entries? :) What is `promptMtlDups`? Is there no way to improve readability on those labels?
> > > Also @LiborBatek have you also tested the prompt/dialog on Update instead of Load?
> > 
> > 
> > I edited the code to be more descriptives and artist friendly in terms of the choosing options. image image
> 
> Much better - so what is **Prompt Material**? Is that like a Pick & Choose dialog?

According to the documentation, it means `prompting the user when duplicated materials are found.` But I think it's too long as the option's name.

BigRoy:
I see.

How about "Prompt on duplicate materials" - and I'm not sure what the "prompt" offers for the user. What happens with the prompt itself - is that like a 'decide what to include or skip' dialog?

client/ayon_core/hosts/max/plugins/load/load_max_scene.py
```diff
@@ -14,6 +16,56 @@
 from ayon_core.pipeline import get_representation_path, load
 
 
+class MaterialDupOptionsWindow(QtWidgets.QDialog):
+
```
BigRoy, line 20: Can we add a simple rudimentary docstring for this?

Future readers of the code will love you.

client/ayon_core/hosts/max/plugins/load/load_max_scene.py
```diff
@@ -14,6 +16,56 @@
...
+        layout.addWidget(self.widgets["label"])
+        layout.addWidget(self.widgets["material_options_list"])
+        layout.addWidget(self.widgets["buttons"])
+
+        self.widgets["okButton"].pressed.connect(self.on_ok_pressed)
+        self.widgets["cancelButton"].pressed.connect(self.on_cancel_pressed)
+        self.widgets["material_options_list"].itemPressed.connect(
+            self.on_material_optionsPressed)
+
+    def on_material_optionsPressed(self, item):
```
BigRoy, line 54: I think code style wise this should be:

```suggestion
    def on_material_options_pressed(self, item):
```

client/ayon_core/hosts/max/plugins/load/load_max_scene.py
```diff
@@ -14,6 +16,56 @@
...
+        self.material_option = next((key for key in self.material_options.keys()
+                                     if self.material_options[key]== item.text()),
+                                     None)
```
BigRoy, line 57: This would be so much nicer if we could match the item by key instead of the value. Why does the option match the value instead of the key which is what we should be matching against instead?
```suggestion
        self.material_option = self.material_options.get(item.text(), None)
```

If we would append QListWidgetItem instead  of just the text

```python
item = QtWidgets.QListWidgetItem(text)
item.setData(QtCore.Qt.UserRole, key)
```

Then here we could just do:
```suggestion
        self.material_option = item.data(QtCore.Qt.UserRole)
```

BigRoy (approved):
Can't speak to the 3DsMax side of things nor from an angle of artists/usability (don't use max myself) but code seems good.

LiborBatek (approved):
All works fine and under all circumstances...

Also now being the options when using load much clearer for the user

image

PR #166, opened 2024-03-08
Blender: Implement 'compress' settings extract blend

### LINKED ISSUE #156: Blender: Extract Blend allow to "compress" the output file

When we publish something from Blender, I just realized that the resulting `.blend` file is a lot bigger than the workfile from which it originates.
It's because the extractor uses default arguments from `bpy.data.libraries.write`, which means the file is written as uncompressed.

Could you please add an option to enable compression ?

See:
- [Blender API `write` function, notably the `compress` argument](https://docs.blender.org/api/current/bpy.types.BlendDataLibraries.html)
- [Ayon Blender: Blend Extractor](https://github.com/ynput/ayon-core/blob/7dc8505194b50eaa1cfd7ade7a9d400be74796c4/client/ayon_core/hosts/blender/plugins/publish/extract_blend.py#L56)

So basically exposing the `compress` argument to be either True or False. 

> Does that influence anything else?

Both compressed and uncompressed blend files are non-ascii (not human readable) but one is (likely) slower to read/write whereas the other has a better filesize.

So companies may prefer the current uncompressed way. So the perfect solution would be to have an option in `Studio Settings-->Blender` so that we can enable or disable that compression. 

As Alain Xerri states on Discord:

> Without compression it's still a binary, so we cannot read it anyway in case there is a problem.
So I don't know why Blender decided to have uncompressed as a default (but on some forums people argue that compressed files take longer to be written by Blender, so maybe speed is the reason for uncompressed as default). 

---

Add a Project Setting to enable or disable the "compress" argument, likely keeping the default off for backwards compatibility.

_No response_

### Additional context:

[Discord conversation where Alain Xerri requested this](https://discord.com/channels/517362899170230292/517382145552154634/1215302041094586489)

BigRoy:
## Changelog Description

Implement a toggle in AYON settings to "compress" the `.blend` exports.
It defaults to False, to preserve the old behavior prior to this PR.

## Additional info

Fix #156

## Testing notes:

1. Enable the settings on Extract Blend
2. Export a `.blend` file.
3. The file should be smaller than when the toggle is off.
4. Loading the file should continue to work fine.


simonebarbieri (approved):
It works very well! Thanks for this, I actually didn't know that there was an option to export blend files as compressed!

PR #167, opened 2024-03-11
Publisher: Promote information to be shown on instance label

bradenjennings:
## Changelog Description
We need to be able to display more information on instance label in publisher than just it's name.

image
In this case for example renderCompositingMain  could show there also frame range or some badge that review will be created.

What information shown there could be decided by plugins themselves. But how to solve conflicting interests, available ways to show things, etc. is up to the implementation.

## Additional info

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/create/create_render.py
+++ b/client/ayon_core/hosts/maya/plugins/create/create_render.py
@@ -106,3 +106,9 @@ class CreateRenderlayer(plugin.RenderlayerCreator):
                     default=self.render_settings.get("enable_all_lights",
                                                      False))
         ]
+
+    def get_extra_label(self):
+        """A label that can be made custom for a creator"""
+        print("get_extra_label() dir(self)", dir(self))
+        # TODO: How do you get to the instance attributes from here, such as frameStart and frameEnd?
+        return "{}-{}".format(1001, 1010)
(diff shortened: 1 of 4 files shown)


bradenjennings:
I wasn't sure how to approach this ticket.

So for now I've added an extra label attribute which can be implemented in the creator plugin.
The problem is I can't return anything useful from the creator plugin method get_extra_label, such as frame range.

Any guidance would be appreciated.

iLLiCiTiT:
It should be done via a method on a create plugin, it does not make much sense, you want to show the information per instance. But there already is a way how to approach it, just fill `"label"` key in data. That is not enough?

BigRoy:
> It should not be done via a method on a create plugin, it does not make much sense, you want to show the information per instance and get the information from instnace. But there already is a way how to approach it, just fill `"label"` key in data. That is not enough?

I believe the original new publisher designs had a special layout for the cards to e.g. show the frame range below the instance's name, and potentially make room for other info:

- frame ranges
- asset + task names
- enabled output types
- etc.

Anyway, what's suitable info for there I guess isn't really defined. Maybe @mkolar has a good link to some of those designs (and can debate whether they are still relevant or not).

Just wanted to remark that even if frame range would be in the label - it'd still need to update when e.g. frame range attributes would be "changed" on e.g. creator attributes or publish attributes if those frame ranges are defined there - and not collect or whatever? (Again coming back to the "update on attribute changes"?)

BigRoy:
I know this has been lingering for a long time. But reading this again, isn't this just the same as setting the instance's `label` attribute? :)

iLLiCiTiT:
NOTE: This modifications are not the way, if we'll add "additional information" to the instance label it should be instance specific -> stored on `CreatedInstance` and handled by UI based on the availabilities in the UI. So e.g. Card view can show in under the instance label, but list view can show it next to it.

PR #171, opened 2024-03-12
Max: Use instance.data["productName"] instead of instance.context.data["subset"] in validate model name

moonyuet:
## Changelog Description
In AYON, there is no `instance.context.data["subset"]`. This PR is to replace the use of this with instance name on validate model name
## Additional info
n/a

## Testing notes:
1. Launch Max
2. Create Model Instance
3. Enable `Validate Model Name`
4. It should validate subset correctly if the validation regex sets to subset-related param.

### DIFF
--- a/client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
+++ b/client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
@@ -103,7 +103,7 @@ class ValidateModelName(pyblish.api.InstancePlugin,
         invalid = False
         compare = {
             "project": instance.context.data["projectName"],
-            "asset": instance.context.data["folderPath"],
+            "asset": instance.data["folderPath"],
             "subset": instance.data["productName"]
         }
         for key, required_value in compare.items():
--- a/client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
+++ b/client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
@@ -100,11 +100,12 @@ class ValidateModelName(pyblish.api.InstancePlugin,
             return obj
 
         # Validate regex groups
+        cls.log.debug(instance.data["productName"])
         invalid = False
         compare = {
             "project": instance.context.data["projectName"],
             "asset": instance.context.data["folderPath"],
-            "subset": instance.name,
+            "subset": instance.data["productName"],
         }
         for key, required_value in compare.items():
             if key in regex.groupindex:
--- a/client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
+++ b/client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
@@ -100,12 +100,11 @@ class ValidateModelName(pyblish.api.InstancePlugin,
             return obj
 
         # Validate regex groups
-        cls.log.debug(instance.data["productName"])
         invalid = False
         compare = {
             "project": instance.context.data["projectName"],
             "asset": instance.context.data["folderPath"],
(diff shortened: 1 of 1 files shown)


client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
```diff
@@ -104,7 +104,7 @@ def invalid_name(cls, instance, obj):
         compare = {
             "project": instance.context.data["projectName"],
             "asset": instance.context.data["folderPath"],
-            "subset": instance.context.data["subset"],
+            "subset": instance.name,
```
BigRoy, line 107: Wait a second - this taking from `context` is wrong to begin with. It should match the `instance` - not the `context`

Also, if `instance.data["subset"]` does not exist then likely in AYON it is `instance.data["product"]` or `instance.data["productName"]` that you should be matching against, not `instance.name` - there is no requirement for the instance name to be or match the product (== subset in OpenPype).

moonyuet: fixed

client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
```diff
@@ -100,11 +100,12 @@ def invalid_name(cls, instance, obj):
             return obj
 
         # Validate regex groups
+        cls.log.debug(instance.data["productName"])
```
BigRoy:
```suggestion
```

client/ayon_core/hosts/max/plugins/publish/validate_model_name.py
```diff
@@ -100,11 +100,12 @@ def invalid_name(cls, instance, obj):
...
         invalid = False
         compare = {
             "project": instance.context.data["projectName"],
             "asset": instance.context.data["folderPath"],
```
BigRoy, line 106: I suspect these should actually also be using `instance.data` as opposed to `context.data`? Depends on the use case of the validator.

moonyuet: I dont think instance.data include ["projectName"] as I tested:
```
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\dependency_packages\ayon_2402141620_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_model_name.py", line 51, in process
    invalid = self.get_invalid(instance)
  File "D:\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_model_name.py", line 77, in get_invalid
    if cls.invalid_name(instance, obj):
  File "D:\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_model_name.py", line 103, in invalid_name
    cls.log.debug(instance.data["projectName"])
KeyError: 'projectName'

```

BigRoy: Ah yes, project name is always on the context since it can't be customized on the instance - true. It should likely be `folderPath` only then. My bad.

LiborBatek (approved):
All working as expected...

My ```Max addon``` settings

Screenshot 2024-03-12 173059

if different name present:


Screenshot 2024-03-12 172856

and when the name adjusted according to the ```regex``` defined in the settings:

Screenshot 2024-03-12 172825

And last but not least succesful ```Publish``` in the end...


Screenshot 2024-03-12 172938

PR #177, opened 2024-03-13
Publisher: Change scrollbar stylesheet to avoid issues on resize

iLLiCiTiT:
## Changelog Description
Fix issue with infinite expanding of publisher UI.

## Detailed issue description
There is scroll area showing error labels with instance view, when this scroll area height gets bigger that it shows scroll bar the widget is forced to resize to width of the content widget and scroll bar width. The issue is that with specific stylesheets may some DCCs misuse widths of the widgets. The handle may have backend width 14px but the real width in UI is 10px, in that exact case the window will start to exponentially grow because we're resizing it.

### How to replicate
It happens in Substance painter, where I faked 3 instance validators to raise `ValidationError` (you can somehow make them fail). The idea is that you have to make the list of validation errors height enough so scroll bar can be shown.
image
When the scrollbar is visible, the window should not start to expand. 

## Change
The scroll bar for that wiget was modified to have different stylesheet which does not affect the width, but has almost none margins. In this case it is probably not a big issue as it is on left side of the view.

## Testing notes:
1. Launch substance painter.
2. Make sure there are instnaces that will cause a lot of validation errors so it is possible to see scroll bar when window is resized.
3. The window does not expand.

### DIFF
--- a/client/ayon_core/style/style.css
+++ b/client/ayon_core/style/style.css
@@ -1067,6 +1067,38 @@ PixmapButton:disabled {
     font-size: 13pt;
 }
 
+#PublisherVerticalScrollArea QScrollBar {
+    background: transparent;
+    margin: 0;
+    border: none;
+}
+
+#PublisherVerticalScrollArea QScrollBar:horizontal {
+    height: 10px;
+    margin: 0;
+}
+
+#PublisherVerticalScrollArea QScrollBar:vertical {
+    width: 10px;
+    margin: 0;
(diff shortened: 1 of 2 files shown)


LiborBatek (approved):
Prepared some weird output template with multiple outputs so it didnt fit into default Publisher UI (check the issue outside of this PR - malfunctioning) 

and when check within this PR no issue happens, all working as supposed to and OK.

image

PR #178, opened 2024-03-13
bug fix on getting aov expected files when global aov mode disabled in Redshift

moonyuet:
## Changelog Description
This PR is to fix the bug on getting AOV render products as expected files when global aov mode is turned off in Redshift renderer

## Additional info
n/a
## Testing notes:
1. Launch Maya via AYON
2. Create Render Instance with Redshift as current renderer
3. Disabled global aov mode 
image
4. Publish

### DIFF
--- a/client/ayon_core/hosts/maya/api/lib_renderproducts.py
+++ b/client/ayon_core/hosts/maya/api/lib_renderproducts.py
@@ -1136,68 +1136,72 @@ class RenderProductsRedshift(ARenderProducts):
         light_groups_enabled = False
         has_beauty_aov = False
         colorspace = lib.get_color_management_output_transform()
-        for aov in aovs:
-            enabled = self._get_attr(aov, "enabled")
-            if not enabled:
-                continue
+        global_aov_enabled = bool(
+            self._get_attr("redshiftOptions.aovGlobalEnableMode")
+        )
+        if global_aov_enabled:
+            for aov in aovs:
+                enabled = self._get_attr(aov, "enabled")
+                if not enabled:
+                    continue
 
-            aov_type = self._get_attr(aov, "aovType")
-            if self.multipart and aov_type not in self.unmerged_aovs:
-                continue
+                aov_type = self._get_attr(aov, "aovType")
+                if self.multipart and aov_type not in self.unmerged_aovs:
+                    continue
 
-            # Any AOVs that still get processed, like Cryptomatte
-            # by themselves are not multipart files.
+                # Any AOVs that still get processed, like Cryptomatte
+                # by themselves are not multipart files.
 
-            # Redshift skips rendering of masterlayer without AOV suffix
-            # when a Beauty AOV is rendered. It overrides the main layer.
-            if aov_type == "Beauty":
-                has_beauty_aov = True
+                # Redshift skips rendering of masterlayer without AOV suffix
+                # when a Beauty AOV is rendered. It overrides the main layer.
+                if aov_type == "Beauty":
+                    has_beauty_aov = True
 
-            aov_name = self._get_attr(aov, "name")
+                aov_name = self._get_attr(aov, "name")
 
-            # Support light Groups
-            light_groups = []
-            if self._get_attr(aov, "supportsLightGroups"):
-                all_light_groups = self._get_attr(aov, "allLightGroups")
-                if all_light_groups:
-                    # All light groups is enabled
-                    light_groups = self._get_redshift_light_groups()
-                else:
-                    value = self._get_attr(aov, "lightGroupList")
-                    # note: string value can return None when never set
-                    if value:
-                        selected_light_groups = value.strip().split()
-                        light_groups = selected_light_groups
+                # Support light Groups
+                light_groups = []
+                if self._get_attr(aov, "supportsLightGroups"):
+                    all_light_groups = self._get_attr(aov, "allLightGroups")
+                    if all_light_groups:
+                        # All light groups is enabled
+                        light_groups = self._get_redshift_light_groups()
+                    else:
+                        value = self._get_attr(aov, "lightGroupList")
+                        # note: string value can return None when never set
+                        if value:
+                            selected_light_groups = value.strip().split()
+                            light_groups = selected_light_groups
 
-                for light_group in light_groups:
-                    aov_light_group_name = "{}_{}".format(aov_name,
-                                                          light_group)
-                    for camera in cameras:
-                        product = RenderProduct(
-                            productName=aov_light_group_name,
-                            aov=aov_name,
-                            ext=ext,
-                            multipart=False,
-                            camera=camera,
-                            driver=aov,
-                            colorspace=colorspace)
-                        products.append(product)
+                    for light_group in light_groups:
+                        aov_light_group_name = "{}_{}".format(aov_name,
+                                                            light_group)
+                        for camera in cameras:
+                            product = RenderProduct(
+                                productName=aov_light_group_name,
+                                aov=aov_name,
+                                ext=ext,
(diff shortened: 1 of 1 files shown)


client/ayon_core/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1136,68 +1136,72 @@ def get_render_products(self):
...
-                            ext=ext,
-                            multipart=False,
-                            camera=camera,
-                            driver=aov,
-                            colorspace=colorspace)
-                        products.append(product)
+        global_aov_enabled = bool(
+            self._get_attr("redshiftOptions.aovGlobalEnableMode")
+        )
+        if global_aov_enabled:
```
BigRoy, line 1142:
```suggestion
        global_aov_enabled = bool(
            self._get_attr("redshiftOptions.aovGlobalEnableMode")
        )
        if not global_aov_enabled:
            # Nothing else to collect, because all aovs are disabled
            return products
```

Can we avoid the extra nesting by doing this instead?

moonyuet: Not sure if this works but we can try.

client/ayon_core/hosts/maya/api/lib_renderproducts.py
```diff
@@ -1136,6 +1136,21 @@ def get_render_products(self):
...
+        )
+        if not global_aov_enabled:
+            beauty_name = "BeautyAux" if has_beauty_aov else ""
+            for camera in cameras:
+                products.insert(0,
+                                RenderProduct(productName=beauty_name,
+                                            ext=ext,
+                                            multipart=self.multipart,
+                                            camera=camera,
+                                            colorspace=colorspace))
```
BigRoy, line 1150:
```suggestion
            # only beauty output
            for camera in cameras:
                products.insert(0,
                                RenderProduct(productName="",
                                            ext=ext,
                                            multipart=self.multipart,
                                            camera=camera,
                                            colorspace=colorspace))
```
If there are no AOVs then `has_beauty_aov`  may never be True?

Also, I'd move `light_groups_enabled` and `has_beauty_aovs` to below this to make it clear that those vars are unrelated to this part of the code.

BigRoy (approved):
I didn't test the functionality - but code makes sense. On to @LiborBatek for testing.

LiborBatek (commented):
I have run this test, first adding some AOVs for render aka enabling those...

```Cryptomatte```
```Z```
```N```

Run Validate action via Publisher so I have proper AOVs naming via ```Repair``` action so validations goes well.

Then cancel the publishing and finally ```Disable Aovs``` via that RS option we are refering to in this PR.

After submitting it to DL my job fails on ```Publish Job``` due to expected ```Cryptomatte``` aov...

image

obviously when inspecting my JSON, it still expect those files...

So my question is, does this PR solves this scenario too? Speaking of Cryptomattes as optional AOVs with Global AOVs switch Disabled...

Right now it does not solve this issue.

**Comment by LiborBatek** [comment]
Im enclosing the JSON for rendered frames and scrn grab of the error from DL log console... just as additional info for you

[renderLightingMain_metadata.json.txt](https://github.com/ynput/ayon-core/files/14872003/renderLightingMain_metadata.json.txt)


Screenshot 2024-04-04 152300

**Comment by moonyuet** [comment]
> Im enclosing the JSON for rendered frames and scrn grab of the error from DL log console... just as additional info for you
> 
> [renderLightingMain_metadata.json.txt](https://github.com/ynput/ayon-core/files/14872003/renderLightingMain_metadata.json.txt)
> 
> Screenshot 2024-04-04 152300

Which means by this case we need to also exclude the cryptomatte.

**Comment by BigRoy** [comment]
> > Im enclosing the JSON for rendered frames and scrn grab of the error from DL log console... just as additional info for you
> > [renderLightingMain_metadata.json.txt](https://github.com/ynput/ayon-core/files/14872003/renderLightingMain_metadata.json.txt)
> > Screenshot 2024-04-04 152300
> 
> Which means by this case we need to also exclude the cryptomatte.

But.. doesn't this code already exclude that? It excludes all AOVs - also the Cryptomattes, right?

**Comment by moonyuet** [comment]
> > > Im enclosing the JSON for rendered frames and scrn grab of the error from DL log console... just as additional info for you
> > > [renderLightingMain_metadata.json.txt](https://github.com/ynput/ayon-core/files/14872003/renderLightingMain_metadata.json.txt)
> > > Screenshot 2024-04-04 152300
> > 
> > 
> > Which means by this case we need to also exclude the cryptomatte.
> 
> But.. doesn't this code already exclude that? It excludes all AOVs - also the Cryptomattes, right?

for some reason, it doesn't work in my side too.

File: client/ayon_core/hosts/maya/api/lib_renderproducts.py [hunk 3]
```diff
@@ -1133,9 +1133,26 @@ def get_render_products(self):
             aovs = list(set(aovs) - set(ref_aovs))
 
         products = []
+        # global_aov_enabled = bool(
+        #     self._get_attr("redshiftOptions.aovGlobalEnableMode")
+        # )
+        global_aov_enabled = bool(
+            cmds.getAttr("redshiftOptions.aovGlobalEnableMode"))
```
moonyuet, line 1140: @BigRoy, pretty strange that `self._get_attr` doesn't work for this case. 
Not sure if the bool convert is the issue.

BigRoy: I understand why! :) `self._get_attr` likely calls the get attribute with `asString=True` and thus returns the 'string' result of the attribute and not the numeric one?

Anyway, please do still use `_get_attr` since it actually considers renderlayer overrides which it must (since it could be disabled just in one layer (we do that all the time).

We might want to expose the `as_string` argument to `_get_attr` and for this call have it be `_get_attr(attr, as_string=False)`?

LiborBatek:
@BigRoy @moonyuet  could be that RS always keep ```Cryptomatte``` as separate files always? (to my surprise, as Arnold dont do that by nature when multilayer exr enabled)

moonyuet:
> @BigRoy @moonyuet could be that RS always keep `Cryptomatte` as separate files always? (to my surprise, as Arnold dont do that by nature when multilayer exr enabled)

If AOV mode is globally turned off, Redshift won't render Cryptomatte no matter what. It's just beauty rendering out.

LiborBatek (approved):
All working now without any errors shown...

My job successfully finished on DL.

image

Note: For some reason my rendered frames when integrated didnt produce any ```reviewable mp4``` and ```thumbnail```. Is it related to this PR or not? For this reason, Im enclosing DL publish job LOG too

[Job_2024-04-05_09-48-47_660facdfc97c3c722d19f16d.txt](https://github.com/ynput/ayon-core/files/14881859/Job_2024-04-05_09-48-47_660facdfc97c3c722d19f16d.txt)

moonyuet:
> All working now without any errors shown... there werent other files expected besides the `renderLightingMain` one, so all good.
> 
> My job successfully finished on DL.
> 
> image
> 
> Note: For some reason my rendered frames when integrated didnt produce any `reviewable mp4` and `thumbnail`. Is it related to this PR or not? For this reason, Im enclosing DL publish job LOG too
> 
> [Job_2024-04-05_09-48-47_660facdfc97c3c722d19f16d.txt](https://github.com/ynput/ayon-core/files/14881859/Job_2024-04-05_09-48-47_660facdfc97c3c722d19f16d.txt)

I dont think it is related to this PR as this PR is just about redshift aov being enabled globally.
This is more likely about submit publish job have encountered some issues. Can you try with different renderers to see the issue still exists?

PR #179, opened 2024-03-13
3dsMax: Retaining Viewport Layout during Preview

r42-chun:
Retaining 3dsMax Viewport layout and checking for non-extended viewport for preview

## Changelog Problem
Previously, when doing a preview in 3ds max, the top left viewport would always be changed to the review camera view.
Even if it is a non camera view. Like a top or a right view.
It also doesn't like it if the top left viewport is an extended viewport (like a redshift renderview)
If another viewport (not the top left) is an extended viewport, the process of switching layouts remove all extended viewports.

## Changelog Description
Instead of switching layouts and switching back, we maximize the first non-extended viewport to do the review.
We store the camera as well as the type and switch it back to it's original camera or type when the preview is done.
We then un-maximize it if we had to maximize it before.

Added a validator to also make sure that there is a valid viewport that can be used for review and not have extended viewports occupying every available view.

## Testing notes for retaining layout:
1. Have a viewport layout, it could be a camera, orthographic, or extended viewport
2. Create a review
3. Make sure the layout before and after the review are the same.

## Testing notes for validator:
1) Switch every available viewport into an extended viewport
2) Validate and see if it errors.

### DIFF
--- a/client/ayon_core/hosts/max/api/preview_animation.py
+++ b/client/ayon_core/hosts/max/api/preview_animation.py
@@ -39,7 +39,7 @@ def viewport_layout_and_camera(camera, layout="layout_1"):
     review_camera = rt.getNodeByName(camera)
 
     try:
-        if rt.viewport.getLayout()!=rt.name(layout):
+        if rt.viewport.getLayout() != rt.name(layout):
             rt.execute("max tool maximize")
             needs_maximise = 1
         rt.viewport.setCamera(review_camera)
--- a/client/ayon_core/hosts/max/api/preview_animation.py
+++ b/client/ayon_core/hosts/max/api/preview_animation.py
@@ -31,23 +31,26 @@ def viewport_layout_and_camera(camera, layout="layout_1"):
         layout (str): layout to use in viewport, defaults to `layout_1`
             Use None to not change viewport layout during context.
     """
+    needs_maximise = 0
+    # Set to first active non extended viewport
+    rt.viewport.activeViewportEx(1)
     original_camera = rt.viewport.getCamera()
-    original_layout = rt.viewport.getLayout()
-    if not original_camera:
-        # if there is no original camera
-        # use the current camera as original
-        original_camera = rt.getNodeByName(camera)
+    original_type = rt.viewport.getType()
     review_camera = rt.getNodeByName(camera)
+
     try:
-        if layout is not None:
-            layout = rt.Name(layout)
-            if rt.viewport.getLayout() != layout:
-                rt.viewport.setLayout(layout)
+        if rt.viewport.getLayout()!=rt.name(layout):
+            rt.execute("max tool maximize")
+            needs_maximise = 1
         rt.viewport.setCamera(review_camera)
         yield
     finally:
-        rt.viewport.setLayout(original_layout)
-        rt.viewport.setCamera(original_camera)
+        if needs_maximise == 1:
+            rt.execute("max tool maximize")
+        if original_type == rt.Name("view_camera"):
+            rt.viewport.setCamera(original_camera)
+        else:
+            rt.viewport.setType(original_type)
 
 
 @contextlib.contextmanager
--- a/client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
+++ b/client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+import pyblish.api
+from ayon_core.pipeline import PublishValidationError
+from pymxs import runtime as rt
+
+
+class ValidateExtendedViewport(pyblish.api.InstancePlugin):
+    """Validate if the first viewport is an extended viewport."""
+
+    order = pyblish.api.ValidatorOrder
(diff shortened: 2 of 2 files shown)


client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
```diff
@@ -0,0 +1,21 @@
...
+    families = ["review"]
+    hosts = ["max"]
+    label = "Validate Extended Viewport"
+
+    def process(self, instance):
+        try:
+            rt.viewport.activeViewportEx(1)
+        except RuntimeError:
+            raise PublishValidationError(
+                "Please make sure one viewport is not an extended viewport", title=self.label)
```
BigRoy, line 20: Preferably this provides the `description` argument explaining what is wrong with it being an extended viewport (like WHY must it not be an extended viewport)

Even better if this could explain how the user can make it a **not** extended viewport. I don't know 3DsMax so do not know what an extended viewport is.

Also, is there any way to disable extended viewport - and re-enable it. If so, what's holding us from just disabling it in a context manager so the review still works fine? (Again, might be stupid question; I don't use max)

r42-chun: Understand. Thank you for your feedback! I'll update the PublishValidationError to include the way people could resolve it in 3dsMax.

So an extended viewport are special viewports that are not all supported by 3dsmax. This could be for example if you use redshift, you could have redshift renderview docked as one of the viewports. 
image
image

The above example is alright and the validator won't see that as a problem as there is still one available viewport that is a 3dsmax viewport that programmatically we have access to it.

So the problem comes about if all available viewports are extended viewports. Like the image below.
image
Cause there isn't any functions exposed to maxscript or python that allows setting or unsetting the viewport. You'll have to do it manually by using the UI.

image
image

client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
```diff
@@ -0,0 +1,24 @@
+# -*- coding: utf-8 -*-
+import pyblish.api
+from ayon_core.pipeline import PublishValidationError
+from pymxs import runtime as rt
+
+
+class ValidateExtendedViewport(pyblish.api.InstancePlugin):
```
moonyuet, line 7:
```suggestion
class ValidateExtendedViewport(pyblish.api.ContextPlugin):
```

client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
```diff
@@ -0,0 +1,24 @@
...
+    """Validate if the first viewport is an extended viewport."""
+
+    order = pyblish.api.ValidatorOrder
+    families = ["review"]
+    hosts = ["max"]
+    label = "Validate Extended Viewport"
+
+    def process(self, instance):
```
moonyuet, line 15:
```suggestion
    def process(self, context):
```

moonyuet: we can use context plugin instead for this case as we dont need to grab any instance data for the validator

moonyuet (commented):
Thank you @r42-chun for coming up on this. I have commented on two spots and already tested by my own. Will be good for merging after resolving these.

client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
r42-chun: Thank you. I have done that :)

client/ayon_core/hosts/max/api/preview_animation.py
```diff
@@ -31,23 +31,26 @@ def viewport_layout_and_camera(camera, layout="layout_1"):
...
-        original_camera = rt.getNodeByName(camera)
+    original_type = rt.viewport.getType()
     review_camera = rt.getNodeByName(camera)
+
     try:
-        if layout is not None:
-            layout = rt.Name(layout)
-            if rt.viewport.getLayout() != layout:
-                rt.viewport.setLayout(layout)
+        if rt.viewport.getLayout()!=rt.name(layout):
```
BigRoy, line 42:
```suggestion
        if rt.viewport.getLayout() != rt.name(layout):
```
Cosmetics - should've been caught by the hound.

client/ayon_core/hosts/max/plugins/publish/validate_extended_viewport.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def process(self, context):
+        try:
+            rt.viewport.activeViewportEx(1)
+        except RuntimeError:
+            raise PublishValidationError(
+                "Please make sure at least one viewport is not an extended viewport but a 3dsmax supported viewport "
+                "i.e camera/persp/orthographic view. To rectify it, please go to view in the top menubar, "
+                "go to Views -> Viewports Configuration -> layout and right click on one of the panels to change "
+                "it.", title=self.label)
```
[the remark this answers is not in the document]
r42-chun: Thank you for the suggestions :)

PR #182, opened 2024-03-13
Chore: Remove requests get and post from lib functions

iLLiCiTiT:
## Changelog Description
Removed deadline specific wrapper functions for `requests` python module.

## Additional info
The removed functions are wrappers for `requests.get` and `requests.post` functions, the only difference is that that enable/disable ssl verification based on environment variable `"OPENPYPE_DONT_VERIFY_SSL"`. The functions were added because of deadline module, which already have the same functions implemend in the module itself.
The functions are removed because the logic based on environment variable is just "wild" guess to be safely used in a generic way, even the deadline logic should be modified in future to define the ssl verification per url.

## Testing notes:
1. Logic did not change so validate the code changes.


BigRoy (approved):
Looks good to me. Only other difference appears to be that the ones in the deadline module also specify `timeout` of `10` as kwargs - so that logic did change. Not sure what that means in practice for the `get_deadline_pools` call.

kalisp (approved):
Seems that it doesnt break anything, but I dont use SSL on Deadline

PR #165, opened 2024-03-08
Chore: Use AYON entities

iLLiCiTiT:
## Changelog Description
This PR is removing `ayon_core.client` and replaces it's usage with `ayon_api` calls. That changed structure of all entity types from OpenPype compatible to AYON.

## Additional info
I did what I could to change every line of code which are working with entities to work properly, but I've probably missed something so please report anything you hit. And I didn't test most of the changes.

Entity structure changes sometimes required change of logic and function arguments. Some of those changes happened on the spot, so if you hit a change that is questionable then please comment. I'll mention few questionable changes myself. In some cases I did more changes than just change of query and used keys because it helper me with other changes, or I couldn't resist to refactor few lines.

## Major changes
- Project, asset, subset, version, representation and workfile documents are now project, folder, product, version, representation and workfile entities.
- Removed `ayon_core.client`. Its functions were replaced with function from `ayon_api`.
- Implemented `initialize_ayon_connection` in `ayon_core.lib.ayon_connection` to initialize ayon connection. Initialization of the connection happened automatically when a call in `ayon_core.client` happened. It is automatically called when `cli.main` is called and on host install or ayon plugin install, so all important places should be covered. This affects local root overrides so it is important to happen as soon as possible in any process that might work with Anatomy.
- Added usage of task entity. Before this PR task information was on asset document, but AYON has task entity separated from folder entity so many places did have to change. Almost all places did expect asset document and task name and it was enough, now is usually needed only task entity, which affected arguments of functions (e.g. `get_product_name`).
- Renamed multiple funtions which contained asset, e.g. `get_current_asset_name` > `get_current_folder_path`, `get_current_project_task` > `get_current_project_folder`.
- Context in load plugins does not have `"asset"` and `"subset"` keys, but `"folder"` and `"product"`.

### Notes
I was very struggling with changes in unreal host. There is a lot of "asset" naming that is not related to OpenPype asset, and also there was a lot of unusuall way of working with representations and what is stored to container data. If it is working as expected now then it's fine, details can be changed when unreal will be in separated addon.

### Questions
- ~~Function `get_product_name` now expects task entity, but from api point of view it might be better to expect `task_name` and `task_type` as that's what is really needed for the function.~~ Changed the function to expect task name and task type instead of task entity.
- Function `initialize_ayon_connection` is really important and I'm not sure if should stay in `ayon_core.lib`, I couldn't find more suitable place and name.

## Testing notes:
[...]

PR #185, opened 2024-03-14
General: fix wrong assignment

kalisp:
## Changelog Description
Without it additional values from data (as author) couldnt be passed through.

## Additional info
It seems as obvious typo/wrong assignment.

## Testing notes:
This would be a bit difficult as even if this is fixed, there is probably additional bug in server backend.  I noticed this when I was trying to push through different `author` than currently logged in in Webpublisher (that would be for services some `service` bogus account). 
I want to store the original author (and I set it into `context.data["user"]`), but it was not getting through. Now I think it is getting through, but server doesn't recognize it either.

But this seems like a required step to figure out server issue.

### DIFF
--- a/client/ayon_core/client/conversion_utils.py
+++ b/client/ayon_core/client/conversion_utils.py
@@ -908,7 +908,7 @@ def convert_create_version_to_v4(version, con):
         converted_version["attrib"] = attribs
 
     if data:
-        converted_version["data"] = attribs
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Closing as https://github.com/ynput/ayon-core/pull/165 is removing the whole conversion logic.

PR #189, opened 2024-03-15
Max: Always save max scene file before deadline submission

moonyuet:
## Changelog Description
When users submit the render to deadline with different versions, some specific versions don't get rendered out correctly.  The reason behind this strange bug contributes to the scene is being saved always after validation fixes on the output folder.
This PR is to fix this bug and ensure the scene is being incremented and saved before deadline submission.

## Additional info
n/a

## Testing notes:
1. Create Render Instance in Max
2. Publish multiple times to farm
3. The deadline submission shouldn't be errored out anymore.


LiborBatek (approved):
Did produced multiple renders from single workfile and all smooth and working well. The workfile got incremented and saved automatically each time publishing happens.

All good!

Screenshot 2024-03-18 101949

PR #191, opened 2024-03-18
Load plugins: Use product types for filtering

iLLiCiTiT:
## Changelog Description
Rename filtering attribute `families` to `product_types` on loader plugins to match what is actually used for filtering.

## Testing notes:
All loader plugins should be filtering correctly.
1. Open loader tool in a DCC or from tray.
2. Show availabl loaders for selection and validate that correct loaders are shown for the selection.

Load plugins which had settings for `families/product_types` should be working based on the settings. As far as I can tell, loaders with the settings are only in hiero and flame.


BigRoy:
- Would this also be the perfect moment to make `product_types` a `set`? Or do we have good reasons not to?
- Also, I'd recommend doing some "extra testing" for any loader that has customizable product types in settings. (I believe there are some but I don't see any settings changes?)

iLLiCiTiT:
> Would this also be the perfect moment to make product_types a set? Or do we have good reasons not to?

Very good question, I guess we should change them to set, but not enforce it. Will do.

> Also, I'd recommend doing some "extra testing" for any loader that has customizable product types in settings. (I believe there are some but I don't see any settings changes

Forgot to mention that in testing note, will add. I found only flame and hiero having settings for product types, and hiero settings hanling was removed.

LiborBatek (commented):
Been successfully tested in ```3dsmax``` ```blender``` and ```PS```

...```maya``` for some reason starts without ```Ayon``` menu present tho ...cant be tested atm

jakubjezek001 (approved):
Nuke, Resolve, Fusion, Hiero, Flame is OK

PR #192, opened 2024-03-18
Enhancement: add ruff workflow and codespell pre-commit hooks

antirotor:
## Changelog Description
This is adding [ruff](https://github.com/astral-sh/ruff) as Github action on PR, also as pre-commit hook along with [codespell](https://github.com/codespell-project/codespell) - spelling checker.

## Additional info
To setup your dev python environment, this is adding "regular" `pyproject.toml` file (apart of the one inside *client* folder that is used for dependency package). To help setting everything up, there are shell scripts for both linux and windows under `./scripts/` directory that will do the same stuff as in OpenPype or AYON Launcher - they will install local poetry and local .venv with all prerequisities. Because this is not synced to dependency package you need to be careful you have everything you need - this needs to be solved somehow in the future.

To install pre-commit hooks manually (or update them):

```sh
./.poetry/bin/poetry run pre-commit install
```

Run codespell on repository manually:
```sh
./.poetry/bin/poetry run codespell
```

Run ruff on repository manually:
```sh
./.poetry/bin/poetry run ruff check
```

You can configure both codespell and ruff in `pyproject.toml` file in their respective sections `[tool.ruff]` and `[tool.codespell]`

## Github Action

Ruff is running as Github Action - it won't work as Hound is - so it doesn't comment on the code itself (I couldn't find GH action that does that unfortunately). But results can be found on reports of the failed (if) action.


pyproject.toml
```diff
@@ -0,0 +1,107 @@
+[tool.poetry]
+name = "ayon-core"
+version = "0.3.0"
+description = ""
+authors = ["Ynput Team <team@ynput.io>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
```
iLLiCiTiT, line 8: Why are these here?

antirotor: to run at least some of the unit tests, but maybe this is for another PR in a future solving development environment with stuff coming from dev bundle. Otherwise just `pre-commit` and `codespell` is needed.

scripts/setup_env.ps1
```diff
@@ -0,0 +1,67 @@
+$current_dir = Get-Location
```
iLLiCiTiT, line 1: Shouldn't we rename `scripts` to `tools` and `setup_env` to `manage` to match all other repositories?

pyproject.toml
```diff
@@ -0,0 +1,102 @@
+# WARNING: This file is used only for development done on this addon.
+# Be aware that dependencies used here might not match the ones used by
+# the specific addon bundle set up on the AYON server. This file should
+# be used only for local development and CI/CD purposes.
+
+[tool.poetry]
+name = "ayon-core"
+version = "0.3.0"
```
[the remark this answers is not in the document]
antirotor: I think new poetry will bitch about it because it doesn't follow https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers

scripts/setup_env.ps1
antirotor: well, I am not a friend of `manage` script because it is hiding functionality. Also any changes in one place of the script might affect the rest of it? `scirpts` -> `tools` I can live with, but mange script I would avoid. Even the difference between windows platform "manage" and linux "make" we have is bothering me...

iLLiCiTiT: > Also any changes in one place of the script might affect the rest of it?

Well but you can change it in single file, which can have help to provide more information than just filename. Also helps to avoid duplicated code, that has to be changed in every separate script if you need to change it.

jakubjezek001 (approved):
lets merge this, it is looking good

BigRoy:
> Ruff is running as Github Action - it won't work as Hound is - so it doesn't comment on the code itself (I couldn't find GH action that does that unfortunately). But results can be found on reports of the failed (if) action.

Oh boy - this might be waiting for 'disaster'. Haha, looking forward to what will come out of this!

antirotor:
> > Ruff is running as Github Action - it won't work as Hound is - so it doesn't comment on the code itself (I couldn't find GH action that does that unfortunately). But results can be found on reports of the failed (if) action.
> 
> Oh boy - this might be waiting for 'disaster'. Haha, looking forward to what will come out of this!

It is not disabling Hound right now, but I am also somewhat nervous about it :)

BigRoy:
image

Haha. Love it, here we go!

PR #193, opened 2024-03-18
Editorial: Data in hierarchy context cleanup

iLLiCiTiT:
## Changelog Description
Cleanup of data that are used to **create** or **update** entities in a project, based on editorial data.

## Additional info
The data structure was created in Pype 2 for ftrack workflow and did not change since that time. This PR tries to match naming in AYON and fixes some grammar. Most of changes are related to `hierarchyContext` key and the source data which are used to create it. Key `"custom_attributes"` was renamed to `"attributes"`, `"childs"` renamed to `"children"`. Key `"entity_type"` does not hold folder type, but when it's value is `"folder"` it is expected there is another key `"folder_type"` with the specific value.

Changes should be auto-fixed by collect plugins if scene metadata contains old structure. In creators was changed just the key `"entity_type"` > `"folder_type"` the rest is related to global plugins.

Some smaller changes, like renaming variable `type` > `src_type` (`type` is python builtin function/class).

## Testing notes:
1. Validate that all editorial DCCs can be used for publishing new hierarchy and update hierarchy data.
- [x] Hiero
- [x] Resolve
- [x] Traypublisher
- [x] ~Flame~


client/ayon_core/hosts/traypublisher/api/editorial.py
```diff
@@ -186,14 +186,15 @@ def _create_parents_from_settings(self, parents, data):
             # in case first parent is project then start parents from start
             if (
                 _index == 0
-                and parent_token_type == "project"
+                and parent_token_type == "Project"
```
BigRoy, line 189: Where does this capitalized P come from in this `Project`?

jakubjezek001 (approved):
LGTM, tested and works well

client/ayon_core/plugins/publish/validate_editorial_asset_name.py
```diff
@@ -35,7 +35,7 @@ def process(self, context):
 
         existing_folder_paths = {
             folder_entity["path"]: (
-                folder_entity["path"].lstrip("/").rsplit("/")[0]
+                folder_entity["path"].lstrip("/").split("/")[:-1]
```
BigRoy, line 38: This doesn't seem right. This will now return a list but I wouldn't  expact `folder_entity["path"]` value to be a list.

What are you describing to do here?

This might also benefit from better variable naming. What is `existing_folder_paths` if it's a dictionary? What is the key, and what is the value supposed to be?

client/ayon_core/hosts/traypublisher/api/editorial.py
jakubjezek001: from settings https://github.com/ynput/ayon-core/blob/enhancement/AY-1112_Editorial-data-in-hierarchyContext/server_addon/traypublisher/server/settings/editorial_creators.py#L32

iLLiCiTiT:
```suggestion
                and parent_token_type.lower() == "project"
```

client/ayon_core/plugins/publish/validate_editorial_asset_name.py
jakubjezek001: I have added some comments in code to describe required value

iLLiCiTiT: We just realized that the plugin is not needed in AYON at all. There is no need to validate unique folder name across project in AYON.

BigRoy: If the values are just the list of parent folder **names** (not their paths) then please update the `existing_folder_paths` variable name to describe what it contains, e.g. `folder_parent_names` or alike.

PR #199, opened 2024-03-19
Create: Add missing variable 'task_name'

iLLiCiTiT:
## Changelog Description
Handled case when `task_entity` in `create` method is `None`.

## Testing notes:
1. Create object of `CreateContext` and run `create` method without task entity.
2. It should not crash because of missing `task_name` variable.


jakubjezek001:
could we also update the docstring for the Create function.

iLLiCiTiT:
> could we also update the docstring for the Create function.

Done

jakubjezek001 (approved):

PR #201, opened 2024-03-19
Max & Maya: Make sure validators being shown in the Publisher UI when they set to be optional in AYON setting  

moonyuet:
## Changelog Description
This PR is to make sure validators being shown correctly in the Publisher UI when they are being set to be optional in AYON setting.

## Additional info
n/a

## Testing notes:
1. Launch Max/Maya
2. Create Instances
3. Toggle the validators optional switch in Max/Maya settings
4. Check to see whether the validators are being shown in the Publisher UI

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_current_renderlayer_renderable.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_current_renderlayer_renderable.py
@@ -1,10 +1,13 @@
 import pyblish.api
 
 from maya import cmds
-from ayon_core.pipeline.publish import context_plugin_should_run
+from ayon_core.pipeline.publish import (
+    context_plugin_should_run,
+    OptionalPyblishPluginMixin
+)
 
-
-class ValidateCurrentRenderLayerIsRenderable(pyblish.api.ContextPlugin):
+class ValidateCurrentRenderLayerIsRenderable(pyblish.api.ContextPlugin,
+                                             OptionalPyblishPluginMixin):
     """Validate if current render layer has a renderable camera
 
     There is a bug in Redshift which occurs when the current render layer
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_current_renderlayer_renderable.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_current_renderlayer_renderable.py
@@ -20,9 +20,11 @@ class ValidateCurrentRenderLayerIsRenderable(pyblish.api.ContextPlugin):
     order = pyblish.api.ValidatorOrder
     hosts = ["maya"]
     families = ["renderlayer"]
+    optional = False
 
     def process(self, context):
-
+        if not self.is_active(context.data):
+            return
         # Workaround bug pyblish-base#250
         if not context_plugin_should_run(self, context):
             return
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_glsl_material.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_glsl_material.py
@@ -6,10 +6,11 @@ from ayon_core.pipeline.publish import (
     RepairAction,
     ValidateContentsOrder
 )
-from ayon_core.pipeline import PublishValidationError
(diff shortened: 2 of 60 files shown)


client/ayon_core/hosts/maya/plugins/publish/validate_current_renderlayer_renderable.py
```diff
@@ -20,9 +20,11 @@ class ValidateCurrentRenderLayerIsRenderable(pyblish.api.ContextPlugin):
     order = pyblish.api.ValidatorOrder
     hosts = ["maya"]
     families = ["renderlayer"]
+    optional = False
 
     def process(self, context):
-
+        if not self.is_active(context.data):
```
BigRoy, line 26: This plugin must still inherit from `OptionalPyblishPluginMixin`

client/ayon_core/hosts/maya/plugins/publish/validate_glsl_material.py
```diff
@@ -23,6 +24,8 @@ class ValidateGLSLMaterial(pyblish.api.InstancePlugin):
     active = True
 
     def process(self, instance):
+        if not self.is_active(instance.data):
```
BigRoy, line 27: This plugin must still inherit from `OptionalPyblishPluginMixin`

client/ayon_core/hosts/maya/plugins/publish/validate_loaded_plugin.py
```diff
@@ -35,7 +36,8 @@ def get_invalid(cls, context):
         return invalid
 
     def process(self, context):
-
+        if not self.is_active(context.data):
```
BigRoy, line 39: This plugin must still inherit from `OptionalPyblishPluginMixin`

client/ayon_core/hosts/maya/plugins/publish/validate_look_shading_group.py
```diff
@@ -14,17 +14,20 @@ class ValidateShadingEngine(pyblish.api.InstancePlugin):
...
     hosts = ["maya"]
     label = "Look Shading Engine Naming"
     actions = [
         ayon_core.hosts.maya.api.action.SelectInvalidAction, RepairAction
     ]
+    optional = True
 
     # The default connections to check
     def process(self, instance):
+        if not self.is_active(instance.data):
```
BigRoy, line 29: This plugin must still inherit from `OptionalPyblishPluginMixin`

BigRoy (commented):
I'm personally not a fan of just throwing this into **ALL** validator plug-ins. Some just really aren't meant to ever be optional plus it's more code to maintain - and in the end does more 'logic' per plug-in and thus will be (maybe measurably?) slower to run.

Anyway, I can see how having the consistency might make it 'less error prone' on changes in the future if someone did start exposing one to settings. But half of these plug-ins don't have matching "optional" settings I believe, and thus this logic being implemented now "does nothing meaningful" because the plug-ins are never actually optional.

So I'd recommend just 'fixing' the ones that need fixing - instead of cluttering it across all plug-ins where it doesn't currently do anything useful. **OR** add them all to the customizable settings?

m-u-r-p-h-y:
> So I'd recommend just 'fixing' the ones that need fixing - instead of cluttering it across all plug-ins where it doesn't currently do anything useful. **OR** add them all to the customizable settings?

We are not making any new validator optional here, I suppose. We are just fixing the ones that already have optional tag to propagate it to the publishing window so they are really optional.

image

BigRoy:
> > So I'd recommend just 'fixing' the ones that need fixing - instead of cluttering it across all plug-ins where it doesn't currently do anything useful. **OR** add them all to the customizable settings?
> 
> We are not making any new validator optional here, I suppose. We are just fixing the ones that already have optional tag in settings to propagate it to the publishing window so they are really optional.
> 
> image

Thanks, yes. Kayla just hit me up on discord in a PM as well mentioning these were already in the settings - in that case. Great fixing! 🧑‍🔧

mkolar (approved):

PR #202, opened 2024-03-19
Bugfix: Consider Hero versions correctly for checking outdated

BigRoy:
## Changelog Description

- Fix outdated check in scene inventory - it now considers hero and latest version to both correctly be latest
- Optimize the querying of highest versions for all products in the scene inventory.

## Additional info

~~**Requires #211 to be merged**~~ _This has since been merged_

See [discord](https://discord.com/channels/517362899170230292/517382145552154634/1219695832295800943).

image
image
image

## Testing notes:

1. Latest version should not appear "red" in scene inventory
2. The "Filter to outdated" (filter button top right) should correctly show only outdated entries
3. The "Update to latest" button on an outdated entry should set latest version not the hero version.


client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -490,27 +505,26 @@ def set_hierarchy_view(self, state):
...
+            if version is None:
                 return True
 
             # If either a version or highest is present but not the other
             # consider the item invalid.
             if not self._hierarchy_view:
                 # Skip this check if in hierarchy view, or the child item
                 # node will be hidden even it's actually outdated.
-                if version is None or highest is None:
+                if version is None:
```
BigRoy, line 524: This will now technically never occur because `version is None` is already checked and returned for above.

BigRoy:
This should now include #211 to function correctly.

@iLLiCiTiT would you like me to merge that branch into this one to ease testing? Then we should just hold off merging this one until that PR is merged - but at least this PR could be tested accordingly.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -321,7 +321,17 @@ def _add_containers(self, containers, parent=None):
...
             group_node["representation"] = repre_id
             group_node["version"] = version_entity["version"]
-            group_node["highest_version"] = highest_version["version"]
+
+            # We check against `abs(version)` because we allow a hero version
+            # which is represented by a negative number to also count as
+            # latest version
+            # If a hero version for whatever reason does not match the latest
+            # positive version number, we also consider it outdated
+            group_node["isOutdated"] = (
```
iLLiCiTiT, line 330: I would say that hero versions should not be considered as outdated if are not from latest version...
```python
if version_entity["version"] < 0:
    is_outdated = False
else:
    last_version = highest_version_by_product_id.get(version_entity["productId"])
    is_outdated = version_entity["version"] != last_version
group_node["isOutdated"] = is_outdated
```

MustafaJafar:
I'm confused a little bit (_Note that: I don't know much about Hero versions in AYON_) ...

I tested this PR and found out that update to latest always fall to the Hero version (which appears in red as my Hero version is equivalent to `v001`) 

Anyway, here's my test.

https://github.com/ynput/ayon-core/assets/20871534/3385944a-c641-4f89-a3a2-73e025ca917c

BigRoy:
> I'm confused a little bit (_Note that: I don't know much about Hero versions in AYON_) ...
> 
> I tested this PR and found out that update to latest always fall to the Hero version (which appears in red as my Hero version is equivalent to `v001`)
> 
> Anyway, here's my test.
> 2024-03-25.13-49-25.mp4

Thanks @MustafaJafar - as the PR described it required #211 to be merged - which was merged but I believe this PR did not yet have the upstream updates merged and thus was still lacking that PR. Could you try again now?

client/ayon_core/tools/sceneinventory/model.py
BigRoy: If the hero version is v001 and the latest version is v006 (so for whatever reason somehow no hero versions got created or updated anymore?) you'd still want to the v001 hero version to be counted as "latest"? Seems somewhat counter-intuitive?

I'm fine with it to make the change though, since we don't usually use hero versions on our end. But it does make me doubt whether it really makes sense that way.

iLLiCiTiT: >  you'd still want to the v001 hero version to be counted as "latest"?

Not as latest, but not as outdated.

BigRoy: I suspect that in OpenPype it would not be considered outdated - no. I'll refactor the logic here as you requested to remain in sync with that. 👍

MustafaJafar:
> Thanks @MustafaJafar - as the PR described it required #211 to be merged - which was merged but I believe this PR did not yet have the upstream updates merged and thus was still lacking that PR. Could you try again now?

I gave it a test and 
- [x] Latest version should not appear "red" in scene inventory
- [ ] The "Filter to outdated" (filter button top right) should correctly show only outdated entries..  (Hero is not considered latest. and it appears in red.)
- [x] The "Update to latest" button on an outdated entry should set latest version not the hero version.


---

On triggering `Switch to versioned`, should scene inventory switch to the equivalent version of the hero version ? 
because my hero version `[v001]` is being switched to `v001` which is not logically correct ?

BigRoy:
> * (Hero is not considered latest. and it appears in red.)

This should now be fixed with https://github.com/ynput/ayon-core/pull/202/commits/b8d1377904d4f830145b53dc8ad4d8ef1b3a1c3a

To me that actually was correct behavior - since you hero version seemed to be last updated with v001 and does not match the latest v006. Likely hero version publishing was disabled all of a sudden for you and thus future versions did not have a hero version.

Anyway, I've refactored it to always consider hero version to be up-to-date no matter what as @iLLiCiTiT requested [here](https://github.com/ynput/ayon-core/pull/202#discussion_r1537598792).

iLLiCiTiT:
NOTE: Title of PR is "Bugfix: Consider Hero versions correctly for checking outdated". The conversation about changes in the tool are relevant, but not for this PR.

MustafaJafar:
> > because my hero version [v001] is being switched to v001 which is not logically correct ?
> 
> Sorry, why isn't that logically correct - that does sound correct. No?

I mean my hero version [v001] is not the same as version v001.. 
Maybe because hero version publishing was disabled during that publish (I randomly select some published product. so it might be the case.)

iLLiCiTiT:
> Not sure I understand - this PR fixes how "hero" versions outdated state actually gets validated. Without this PR hero versions would ALWAYS be considered outdated, now with the latest commit they will never be considered outdated.

I'm lost in the PRs, I though the origin of the PR was created because of the bug in ayon api, and then transformed to something else. But if that's the case then we should use `HeroVersionType` for hero versions. The object is important as `-1` integer would update to latest instead of hero version 1.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -321,7 +321,17 @@ def _add_containers(self, containers, parent=None):
...
+            # Check if the version is outdated. If the version is below 0 it
+            # is a hero version and will never be considered outdated
+            is_outdated = False
+            if version_entity["version"] >= 0:
+                last_version = highest_version_by_product_id.get(
+                    version_entity["productId"])
+                if last_version is not None:
+                    is_outdated = version_entity["version"] != last_version
+            group_node["isOutdated"] = is_outdated
+
```
iLLiCiTiT, line 334:
```
            version = version_entity["version"]
            if version < 0:
                version = HeroVersionType(version)
            group_node["version"] = version

            # Check if the version is outdated. If the version is below 0 it
            # is a hero version and will never be considered outdated
            is_outdated = False
            if not isintance(version, HeroVersionType):
                last_version = highest_version_by_product_id.get(
                    version_entity["productId"])
                if last_version is not None:
                    is_outdated = version_entity["version"] != last_version
            group_node["isOutdated"] = is_outdated
```

client/ayon_core/tools/sceneinventory/model.py
iLLiCiTiT: Then remove it...

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -297,6 +288,21 @@ def _add_containers(self, containers, parent=None):
...
         sites_info = self._controller.get_sites_information()
 
+        # Query the highest available version so the model can know
+        # whether current version is currently up-to-date.
+        highest_version_by_product_id = ayon_api.get_last_versions(
+            project_name,
+            product_ids={
+                group["version"]["productId"] for group in grouped.values()
+            },
+            fields=["productId", "version"]
```
iLLiCiTiT, line 298:
```suggestion
            fields={"productId", "version"}
```

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -8,10 +8,7 @@
 from qtpy import QtCore, QtGui
 import qtawesome
 
-from ayon_core.pipeline import (
-    get_current_project_name,
-    HeroVersionType,
-)
+from ayon_core.pipeline import get_current_project_name
```
iLLiCiTiT, line 11:
```suggestion
from ayon_core.pipeline import (
    get_current_project_name,
    HeroVersionType,
)
```

client/ayon_core/tools/sceneinventory/model.py
iLLiCiTiT: I guess this can be just
```
def outdated(node):
    return node.get("isOutdated", True)
```

client/ayon_core/tools/sceneinventory/model.py
iLLiCiTiT: Feel free to change lines and make it "nicer".

BigRoy: Thanks - seemed good to me - just changed comments a bit.

client/ayon_core/tools/sceneinventory/model.py
BigRoy: Done with https://github.com/ynput/ayon-core/pull/202/commits/910c79f886274d66e616229ad88afd90d3f5aa90 and https://github.com/ynput/ayon-core/pull/202/commits/e735c5dc723bc127eba840a03569e7e0c3db05e9

BigRoy:
> > I feel like we're over complicating things and wonder why this wasn't an issue in earlier versions of OpenPype?
> 
> Because hero versions were not separated entities and did not have negative version number, and because sceneinventory is using `-1` to identify "Update to latest" then it must be handled.
> 
> > Would you like me to update the delegate?
> 
> Give me a minute.
> 
> EDITED: Function `format_version` should be modified to:
> 
> ```python
> def format_version(value):
>     """Formats integer to displayable version name"""
>     if isinstance(value, HeroVersionType):
>         value = int(value)  # 'value = value.version' would work too
>     label = "v{0:03d}".format(abs(value))
>     if value < 0:
>         return "[{}]".format(label)
>     return label
> ```
> 
> EDITED 2: Or implement `__abs__` on `HeroVersionType`?
> 
> ```python
> def __abs__(self):
>     return abs(self.version)
> ```

Thanks, it'd have to be the first one so we can still also include the `[]` brackets displayed around a hero version. I'll just do `int(value)` at the beginning of the delegate and it'd already work.

iLLiCiTiT (approved):

iLLiCiTiT:
@MustafaJafar can you re-validate please?

MustafaJafar (approved):
Looking Good.

- [x] Latest version should not appear "red" in scene inventory
- [x] The "Filter to outdated" (filter button top right) should correctly show only outdated entries..
- [x] The "Update to latest" button on an outdated entry should set latest version not the hero version.

PR #10, opened 2024-02-08
Bugfix: Make houdini submissions respect pools groups

MustafaJafar:
## Changelog Description
Make houdini submissions respect pools groups.
This is done by:
1. Make collect pools works with some Houdini families/product types.
2. Make Ayon Houdini submitters get group names from Houdini deadline settings.

## Additional info
I made three Houdini deadline setting categories. each for a particular case.
image

## Testing notes:
1. Set pools and groups in ayon deadline addon settings
3. Publish render and cache to deadline.

PR #206, opened 2024-03-19
Bugfix: Houdini Karma/Mantra collectors wrong indentation

BigRoy:
## Changelog Description

Fix indendations

## Additional info

These would've broken in the case the "chunkSize" parm did not exist because `render_products` and `files_by_aov` would have been undefined variables.

## Testing notes:

1. Publish Karma ROP instance
2. Publish Mantra ROP instance

(Also test instances without the `chunkSize` parm, likely from older legacy instances - but maybe best to reproduce by removing the parms manually?)

### DIFF
--- a/client/ayon_core/hosts/houdini/plugins/publish/collect_karma_rop.py
+++ b/client/ayon_core/hosts/houdini/plugins/publish/collect_karma_rop.py
@@ -41,23 +41,23 @@ class CollectKarmaROPRenderProducts(pyblish.api.InstancePlugin):
             instance.data["chunkSize"] = chunk_size
             self.log.debug("Chunk Size: %s" % chunk_size)
 
-            default_prefix = evalParmNoFrame(rop, "picture")
-            render_products = []
+        default_prefix = evalParmNoFrame(rop, "picture")
+        render_products = []
 
-            # Default beauty AOV
-            beauty_product = self.get_render_product_name(
-                prefix=default_prefix, suffix=None
-            )
-            render_products.append(beauty_product)
+        # Default beauty AOV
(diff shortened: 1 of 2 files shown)


MustafaJafar (approved):
Looking good so far. 
I tested the following variations for both Mantra and Karma: 
1. Use smaller value than my scene frame range .. (cool, separated in tasks)
2. Use bigger value than my scene frame range .. (cool, appeared as one task)
3. Remove chunk size from nodes .. (cool too.)

---

I have some thought.. Let me think out loud.
[**Not Related to this PR, I guess.**]
The life cycle of fetching chunk size is:
1. `collect_karma_rop.py`.. (read `chunkSize` from node, save it in `instance.data`)
2. `submit_houdini_render_deadline.py` .. (which adds attribute defs in publisher UI)
    i. in `get_attribute_defs` default of `cls.chunk_size` is  fetched from class attributes that can be overridden by ayon houdini deadline settings. 
    ii. set `job_info.ChunkSize` to `self.chunk_size` 

That life cycle made me question: Is `chunk_size` parm being ignored ? 
I checked `abstract_submit_deadline.py` but I found `chunkSize` has a hard coded default.
While I'm not sure about it because I might have skipped something. 
But, this test is some how confirm my concern.  
image

Also, Another question which default value should `chunkSize` fall to ?
e.g. `chunkSize` parm in Houdini node or `chunkSize` in [Ayon Houdini deadline settings](https://github.com/ynput/ayon-core/pull/10) ?

mkolar:
Looking good and ready to merge. I'd say we can change the chunksize default in a different PR

PR #207, opened 2024-03-19
Bugfix: Maya validate rig contents geometry actually return invalid nodes

BigRoy:
## Changelog Description

Fix geometry validation for rig validator

## Additional info

It was not returning the invalid nodes, so returned None always.
Also - fixed the docstrings.

## Testing notes:

1. Rig validation of geometry node types should work as intended.


moonyuet (approved):
Happens to be more concise with the edited docstrings

LiborBatek (approved):
It works for me without any issues. At least when speaking of correct validation and successful publish as there hasnt been provided any further testing steps to get invalid content / validation error occasion aka when particular circumstances happening and that it has been fixed for such occasion by this PR.

so, works normally imho

PR #209, opened 2024-03-20
Chore: Cleanup redundant code

BigRoy:
## Changelog Description

Cleanup some cosmetic / redundant / dead code

## Additional info

See commit messages for details.

## Testing notes:

1. Shouldn't do much
2. Maybe double check load workfile template for harmony?

### DIFF
--- a/client/ayon_core/hosts/houdini/api/plugin.py
+++ b/client/ayon_core/hosts/houdini/api/plugin.py
@@ -147,7 +147,6 @@ class HoudiniCreatorBase(object):
     def create_instance_node(
         folder_path, node_name, parent, node_type="geometry"
     ):
-        # type: (str, str, str) -> hou.Node
         """Create node representing instance.
 
         Arguments:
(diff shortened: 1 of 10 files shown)


client/ayon_core/hosts/houdini/api/plugin.py
```diff
@@ -147,7 +147,6 @@ def cache_instance_data(shared_data):
     def create_instance_node(
         folder_path, node_name, parent, node_type="geometry"
     ):
-        # type: (str, str, str) -> hou.Node
```
iLLiCiTiT, line 150: This is valid type-hinting.

BigRoy: Because the docstring is **also** defining type hints but with more details - so they would essentially be redundant.
Plus this type-hint was actually invalid, because it missed one input argument.

iLLiCiTiT (approved):

PR #214, opened 2024-03-20
Blender: Extract pointcache alembic with subdiv schema

BigRoy:
## Changelog Description

Add toggle to export Alembic with subdiv schema.

## Additional info

image

## Testing notes:

1. Test pointcache export with and without subdiv schema enabled.


LiborBatek:
@BigRoy Just want to ask about these... 

I can see that ```Alembic Subdiv Schema``` knob but I couldnt make it work when  my ```model``` which outputs ```alembic``` repre too. 

So my question is, is it also valid when outputting anything else in ```Alembic``` or just for ```pointcache``` product?? As Im can see that option present when publishing ```model``` product type.

And as I have inspected the code, this should propagate ```crease``` edges into ```Alembic``` cache but this wasnt the case - simply not.

Am I missing something, or?

BigRoy:
> I can see that Alembic Subdiv Schema knob but I couldnt make it work when my model which outputs alembic repre too.

If it displays for the model product - then it should also work on that product's alembic export. (It shouldn't influence other output formats). It should work for anything that goes through this Alembic Export plug-in. Unless @iLLiCiTiT sees anything why it might not work for the model product type?

> but I couldnt make it work 

What in particular made you think it did not work? The output file might not differ much based on how your mesh is build. It just enables the "Use Subdivision Schema" checkbox on export. You can maybe try exporting manually in Blender with that toggle enabled/disabled to see if that does give very different outputs (if so, then indeed there must be something wrong with the logic if Blender behaves differently there)

> And as I have inspected the code, this should propagate `crease` edges into `Alembic` cache but this wasnt the case - simply not.

What it should do is just enable [the _Use Subdivision Schema_](https://docs.blender.org/manual/en/2.80/files/import_export/alembic.html#object-options) option in Blender. What that means in practice and what it _really_ does in the background I'm not entirely sure of. We needed it for a particular scene an artist made on our end to make the output come out as we intended/expected it to be. Why particularly this solved it I'm not entirely sure now in hindsight.

iLLiCiTiT:
> Unless @iLLiCiTiT sees anything why it might not work for the model product type?

Nope.

LiborBatek:
@BigRoy seems like ```Subdiv Schema``` means this (as copilot told me :)

"When exporting geometry using Alembic, you can specify the subdivision scheme (such as Catmull-Clark, Loop, or Bilinear) to ensure that the subdivided surfaces are correctly represented in the Alembic file."

So its generally a flag deciding which type of subdivs method to use...

LiborBatek (dismissed):
So I have done another test round with ```Subdiv Schema``` enabled and come up with following findings:

1) If done outside of ```Ayon``` and check this option during native ABC export

It propagates edge creasing correctly when imported the ABC back into blender.

Here is my scene before export/publish

Screenshot 2024-04-16 104854

And when imported by native ABC in blender:


Screenshot 2024-04-16 104814


2) When Publishing within Ayon, those wont propagate the same way (note the zero values in Crease value)

Screenshot 2024-04-16 105022

And when loaded via Ayon:


Screenshot 2024-04-16 105501

And it also doesnt have those when importing via native blender toolset.

So there must be some issue with that...```subdiv schema```

LiborBatek (approved):
Ok one more update...I have applied the ```subdiv modifier``` and now it all correctly propagate...

Finally seems working ok! Sorry one has to apply subdiv modifier or it wont produce desired data...

Now all ok!

image

Note the crease values are present when Loaded with Ayon load ABC

PR #215, opened 2024-03-20
Blender: Validate Transform Zero report + repair

BigRoy:
## Changelog Description

Validate Transform Zero report + repair improve validation report + add repair actionpair action

## Additional info

n/a

## Testing notes:

1. Publish model with transforms on objects
2. Validation report should be clear
3. Repair action should work

### DIFF
--- a/client/ayon_core/hosts/blender/plugins/publish/validate_transform_zero.py
+++ b/client/ayon_core/hosts/blender/plugins/publish/validate_transform_zero.py
@@ -6,7 +6,7 @@ import bpy
 
 import pyblish.api
 
-from ayon_core.hosts.blender.api import plugin
+from ayon_core.hosts.blender.api import plugin, lib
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
It correctly validates when any sort of transform present but it fails on ```Repair``` action which doesnt repair aka apply the transforms to deltas.

I have found that if  I use first ```Select Invalid``` and then ```Repair``` action then it works well.
I guess we should adjust so it works in both cases when the model being first selected (as it works in this condition) and also if not selected....which do not work atm.
image

BigRoy:
@LiborBatek thanks - should be fixed now with https://github.com/ynput/ayon-core/pull/215/commits/5b0826a16da2678ca26581dc6df933593550bed4

LiborBatek (approved):
Now all works in both cases...I havent spot any issues this time.

Good to go!

PR #218, opened 2024-03-21
Nuke: Publish and load slate frame explicitly - AY-4055

tokejepsen:
## Changelog Description
Explicitly publish representations with `slateFrame` data member, so loading can query whether there is slate frame.

## Testing notes:
1. Publish a slate render/review from Nuke.
2. Load the render/review into Nuke.
3. Validate slate is at the Nuke workfiles first frame minus 1.


client/ayon_core/hosts/nuke/plugins/load/load_clip.py
```diff
@@ -130,6 +130,19 @@ def load(self, context, name, namespace, options):
...
             last = first + duration
 
+        # If a slate is present, the frame range is 1 frame longer for movies,
+        # but file sequences its the first frame that is 1 frame lower.
+        slate_frame = repre_entity["data"].get("slateFrame", False)
+        if slate_frame:
+            extension = "." + repre_entity["context"]["ext"]
+
+            if extension in VIDEO_EXTENSIONS:
+                last += 1
```
jakubjezek001, line 140: This does not look right. How is it possible that video file is shorter one frame?

BigRoy: It's not shorter? It's adding to the `last` (which is the `end`, right?)

So there's an extra frame at the end.
However, as the comment says - for image sequences instead there's an extra frame at the start.

Which admittedly the video example sounds wrong. The timecode for the slate frame should still be 'before' the start of the video.

BigRoy (commented):
Even though I understand what it is trying to solve I feel things here are making this way more complicated than it really should be.

The assumption that a `slate` is only ever 1 frame seems wrong - I have also historically seen 'animated slates' that last up to a second or two, like almost a bumper with color checks, etc. I feel like the `slateFrame` true/false just doesn't account to that?

tokejepsen:
> The assumption that a slate is only ever 1 frame seems wrong - I have also historically seen 'animated slates' that last up to a second or two, like almost a bumper with color checks, etc. I feel like the slateFrame true/false just doesn't account to that?

Yeah, I was thinking whether to account for that, but I've personally never encountered slate longer than 1 frame.

@jakubjezek001 got any experience with this?

jakubjezek001:
> yeah, I was thinking whether to account for that, but I've personally never encountered slate longer than 1 frame.

Never had but we could be more safe if we do `slateFrames` as **Int** instead. This way in future we would be covered. But the refractory of the code for publishing plugins should happen in future on demand (those are only counting with single slate frame at the moment).

Thanks @BigRoy for mentioning this!

jakubjezek001 (approved):
perfect! thanks @tokejepsen

PR #220, opened 2024-03-24
Maya Settings: Extract Camera bake attributes title fix + add description

BigRoy:
## Changelog Description

Settings for Maya Extract Camera "bake attributes" should now be labeled correctly and have a description.
## Additional info

I did not bump the server addon version because it's cosmetics only for the settings. So it's fine it only starts showing in future releases.

## Testing notes:
1. Settings for Maya Extract Camera "bake attributes" should now be labeled correctly and have a description.

### DIFF
--- a/server_addon/maya/server/settings/publishers.py
+++ b/server_addon/maya/server/settings/publishers.py
@@ -322,7 +322,9 @@ class ExtractCameraAlembicModel(BaseSettingsModel):
     optional: bool = SettingsField(title="Optional")
     active: bool = SettingsField(title="Active")
     bake_attributes: str = SettingsField(
-        "[]", title="Base Attributes", widget="textarea"
+        "[]", title="Bake Attributes", widget="textarea",
+        description="List of attributes that will be included in the alembic "
+                    "export.",
     )
 
     @validator("bake_attributes")
--- a/server_addon/maya/server/settings/publishers.py
+++ b/server_addon/maya/server/settings/publishers.py
@@ -315,16 +315,13 @@ class ExtractMayaSceneRawModel(BaseSettingsModel):
 
 
 class ExtractCameraAlembicModel(BaseSettingsModel):
-    """
-    List of attributes that will be added to the baked alembic camera. Needs to be written in python list syntax.
-    """
     enabled: bool = SettingsField(title="ExtractCameraAlembic")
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (approved):

BigRoy:
The settings field is currently a string that should be a `json` list. 
Would it make sense to actually transition that to a list of strings as input fields?
Similar to #221

Innders:
I'm wondering if a "?" button that can either be clicked or hovered to reveal a tooltip would be better.

Otherwise it's not obvious there's extra info to see.

BigRoy:
> I'm wondering if a "?" button that can either be clicked or hovered to reveal a tooltip would be better.

Having a question mark icon for now while there aren't too many attributes with tooltips I think could be nice. However, I think we should just try and work towards almost everything getting tooltip descriptions (and then at some point it's everywhere and the question mark might become redundant.)

BigRoy:
> My first question would be: "Why it was json at first place?". Conversion on server is possible, but client code conversion might be complicated, with combination of new maya settings and old ayon-core (or openpype). Could that wait until maya client code is bound to the server addon?

Sure can - this PR doesn't touch the fact that it's JSON - it just fixes the label "Base" to "Bake" and adds a description. 
We'd just need to make an issue to track the fact that we want to convert the JSON to a list down the line.

BigRoy:
> Shall we also remove the docstring?

I've gone ahead and done this with https://github.com/ynput/ayon-core/pull/220/commits/fe8613b55e14d529ad76d9cd8a1e25e13631f6af

iLLiCiTiT:
I was about to write that you can probably keep `"Attributes that will be added to the baked alembic camera."`, if it's true statement, but it is already gone.

BigRoy:
> I was about to write that you can probably keep `"Attributes that will be added to the baked alembic camera."`, if it's true statement, but it is already gone.

The issues was that it showed in the frontend as "global" header description for the full plug-in, instead of for that particular settings field. I've now moved whatever 'info' it contained that was lacking in the description into the description.

Happy to revert if you prefer the old.

mkolar (approved):

mkolar:
I'd merge it, a lot of talk over a small pr :D

PR #222, opened 2024-03-24
Houdini: Add generic filepath loader

BigRoy:
## Changelog Description

Add generic filepath loader that can be loaded and managed via AYON.

## Additional info

This is useful if for a particular workflow there is no existing loader yet. A Houdini artists can load as the generic filepath loader and then reference the relevant Houdini parm to use the exact value. The benefit is that this filepath will be managed and can be updated as usual.

## Testing notes:

Each of the below should work:

1. Load some paths via loader
2. Update / switch folders via scene inventory (manage)
3. Remove via scene inventory  (manage)


BigRoy:
> I think Houdini addon version should be bumped to indicate the change.

@iLLiCiTiT @antirotor thoughts on this? Should I bump the Houdini server addon version?

iLLiCiTiT:
> Should I bump the Houdini server addon version?

Server version bump at this moment makes sense only if you change server code (only settings at this moment).

MustafaJafar:
> Server version bump at this moment makes sense only if you change server code (only settings at this moment). 

Fair!

BigRoy:
> > Should I bump the Houdini server addon version?
> 
> Server version bump at this moment makes sense only if you change server code (only settings at this moment). The code change is in ayon-core and change of server addon version has no effect.

In that case, all should be good as is.

PR #224, opened 2024-03-24
Blender: Model publish uv map name validator

BigRoy:
## Changelog Description

Add model publish validator to ensure uv set is named `map1`.

## Additional info

Blender defaults to UV sets named `UVMap` which can be an issue for a Maya-centric pipeline.
Note that the plug-in is disabled in settings by default.

## Testing notes:

1. Enable validator in settings
2. Publish a blender cube
3. It should invalidate and have a working select and repair action


client/ayon_core/hosts/blender/plugins/publish/validate_model_uv_map1.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+
+    order = ValidateContentsOrder
+    hosts = ["blender"]
+    families = ["model"]
+    label = "Mesh UVs named map1"
+    actions = [ayon_core.hosts.blender.api.action.SelectInvalidAction,
+               RepairAction]
+    optional = True
+    enabled = False
```
moonyuet, line 33:
```suggestion
```
I dont think this is needed to be there, as I tested on the code it won't show up with or without this line when it is being disabled in AYON setting.

moonyuet:
I have done simple test with the codes(also glanced the code a little bit), the validators blocks the publishing as expected when the uv names are not being named to `map1`.  But somehow the font of the error report looks a bit strange where the bold text somehow appear to be much bigger than the normal text.  
image

client/ayon_core/hosts/blender/plugins/publish/validate_model_uv_map1.py
BigRoy: Correct - I added this, and the optional line, just to make it clear that those are the default values in settings - I can add both under a comment like `# overridden by settings`. Would that help?

client/ayon_core/hosts/blender/plugins/publish/validate_model_uv_map1.py
```diff
@@ -0,0 +1,94 @@
...
+        invalid = self.get_invalid(instance)
+        if invalid:
+            raise PublishValidationError(
+                f"Meshes found in instance without valid UV's: {invalid}",
+                description=self.get_description()
+            )
+
+    def get_description(self):
+        return inspect.cleandoc(
+            """## Meshes must have `map1` uv set
```
BigRoy, line 87:
```suggestion
            """## Meshes must have map1 uv set
```

BigRoy:
> I have done simple test with the codes(also glanced the code a little bit), the validators blocks the publishing as expected when the uv names are not being named to `map1`. image But somehow the font of the error report looks a bit strange where the bold text somehow appear to be much bigger than the normal text. If the debug message and the comment above can be resolved, I will approve the ticket. image

Thank you, I expect this to be fixed with https://github.com/ynput/ayon-core/pull/224/commits/83fa71982a109cc41460b8ca510538a5dfa67c5e

LiborBatek (approved):
Works fine for me. Also tested by loading tha ```blender``` made model with ```map1``` uvset into ```maya``` host to check if all propagated correctly and all good.
image

note: only small glitch being those non bold typings when messages shown during validation in the publisher

Screenshot 2024-03-25 103145

LiborBatek:
Can be merged imho...

moonyuet (approved):

Issue #134, opened 2024-02-28
Enhancement: Filtering shown creator plugins based on preset / settings profile


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

For certain workflows and pipeline steps, it would be easier to limit number of creators presented in Publisher UI. For example, modeler willl probably never need to publish VDBs, etc. This is highly workflow specific so it needs to be customizable by the studio itself.

### How would you imagine the implementation of the feature?

There are two ways (or more):

**Preset**
Define sets of creators in lists and show those list as preset selector in the Publisher

**Profiles**
Define list of creators per host on project level with set of task type/task name/asset or even user role?

All of it needs button in publisher to ignore and show all anyway.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[AY-2539](https://app.clickup.com/t/86bxp3p06)]

BigRoy:
I believe this was implemented by @kalisp with https://github.com/ynput/ayon-core/pull/560

The setting is: `ayon+settings://core/tools/creator/filter_creator_profiles`

PR #225, opened 2024-03-24
Fusion: Allow render publishes to have custom frame range per saver

BigRoy:
## Changelog Description

Allow render publishes to have custom frame range per saver

## Additional info

Screenshot:
image

There is currently no way to hide the 'custom frame range' attributes in the UI if the frame range source was not set to Custom Frame Range so it will always (confusingly) be shown. I know that is an issue @antirotor and @iLLiCiTiT are aware of but is currently in the backlog. I believe https://github.com/ynput/ayon-core/issues/134 is part of that as a tracked issue.

---

Note: It may potentially be that this does not work correctly for farm submitted renders - but if that's the case then I suspect that's also originally an issue with mixing the other "frame range sources" if they were different frame ranges. If that's indeed not working as intended, then that'd be a separate issue from this PR's functionality so would be great if you could report that as a separate issue if that's the case.

## Testing notes:

1. Create a few saver instances, say five different ones.
2. Set them to different Frame Range Source
3. Set two of them to "Custom frame range"
4. Set different custom frame ranges 

Each of the instances should render and publish only the specified frame ranges.


BigRoy:
@EmberLightVFX I vaguely recall you were interested in this in the past, but I can't link you as reviewer.

EmberLightVFX:
Yes! I would actually need it for the project I'm currently working on 😄
Will try and test it out when I arrive to the office in two days!

EmberLightVFX:
Tried it out and works great.
The only problem I'm seeing is that the comp render ones per saver and not everything together. To solve this, when setting the `Frame Start` and `Frame End` we can change the in/out point of the node in question.
This would also be a great addition to with frame range set to `From render in/out` or `From composition timeline` that would set a custom render in/out point but that might be for another PR?

As on how the fusion integration currently works I would say this pr LGTM

EmberLightVFX:
> Actually - wait a second. **How** do you set a custom in/out point for a saver node so that _that_ works? I never got that to work. But if we can render different frame ranges in one API call to "render" - yes please. Separate PR maybe, but yes - I'd like that!

You just simply drag the in/out point of the saver node in the keyframes window. Fusion will then try and render the comp from your set in/out point as as with any node, if it's not active on specific frames Fusion skips its processing, thus resulting in those frames not rendering.
image

The only problem is that Fusion reports the render as failed if you try to render more frames than exists for an image-stream.
Eg you have a loaded sequence that goes from 1001 to 1011 that you try to render but your comp goes from 1001 to 1021.
Fusion will try to render but there is no image to render, thus the error.
I'll take a look at this for another PR and see if I can get it working as expected :)

EmberLightVFX:
> I mean - that technically is a correct error then, yes? Or do you mean that if you use the saver's output downstream? (not sure if anyone does that?)

Yes it is technically correct but at the same time it rendered correctly.
The problem ends up with fusion reporting that the render failed to the publisher even if it kind of didn't.

It's a bit complicated 😅

BigRoy:
> It's a bit complicated 😅

Looking forward to the reported issue or even PR trying to target that! :)

I will try to help where I can.

LiborBatek (approved):
Did setup three different ```Render Saver``` instances with custom frame ranges and all published ok and with review mp4. All frame ranges matched those set on publish instances.

Works nicely!

image

mkolar (approved):

mkolar:
happy to merge this. Regarding the in and out points, that's probably best for another PR

PR #227, opened 2024-03-24
Houdini: Allow loading any alembic file by .abc extension

BigRoy:
## Changelog Description

Allow loading any alembic file by .abc extension, regardless of representation name

## Additional info

I'm not entirely sure anymore why we needed [this in OpenPype](https://github.com/BigRoy/OpenPype/commit/3419281f6a273849c0ae3788c7f4d60c90269e1e) and what family of `{"model", "animation", "pointcache", "gpuCache"}` actually exported to a different representation name than `abc`. Anyway it shouldn't change existing behavior and the change does make sense as to "on what" it should show.

I wonder if we should remove the product types filter as well.

## Testing notes:

1. Load an Alembic in Houdini should work (also as 'archive')


MustafaJafar (approved):
Looking good to me.

> I'm not entirely sure anymore why we needed [this in OpenPype](https://github.com/BigRoy/OpenPype/commit/3419281f6a273849c0ae3788c7f4d60c90269e1e) and what family of {"model", "animation", "pointcache", "gpuCache"} actually exported to a different representation name than abc.

I think it might have been added for future considerations.

client/ayon_core/hosts/houdini/plugins/load/load_alembic.py
```diff
@@ -11,7 +11,8 @@ class AbcLoader(load.LoaderPlugin):
 
     product_types = {"model", "animation", "pointcache", "gpuCache"}
     label = "Load Alembic"
-    representations = ["abc"]
+    representations = ["*"]
+    extensions = {"abc"}
```
fabiaserra, line 15: unrelated to this PR but why are we defaulting unpacking the alembic on load?

BigRoy: Mostly legacy logic from our pipeline because artists preferred it that way - note that it does not 'display' that part of the node network by default and you can object merge any of the nodes you want. So it was more 'easy access' which maybe from a Houdini artist POV should maybe not be done?

I feel like for Houdini artists they prefer "automate less" so they can do whatever they want.

fabiaserra: yeah that's not great from a render POV as then that's not treated as a procedural and a lot of artists then end up not working efficiently. Arnold reads alembics as procedurals only if they are packed, otherwise it's like Houdini geo

BigRoy: The way it's currently set up shouldn't make it use the unpacked one by default. It's only accessible there for convenience and shouldn't influence the use case you're describing. If it does, then we should hotfix this asap. Otherwise I'd say, we might consider removing this in a backwards compatible manner over time.

Thanks for the details!

fabiaserra: anyway, I didn't mean to go in a tangent as this PR sorts out a very specific case and it should be merged but it caught my attention. In the future I think that should be an exposed option on the loader that artists can easily control during the load action, and since I'm already going in a big tangent... not sure if we ended up creating a discussion to refactor the loader UX? I think it should be much easier to visualize the control of how things are loaded in a friendlier way than the current options dialog that's hidden on the action menu and you need to do extra clicks. Thinking for example on the ftrack loader https://www.ftrack.com/wp-content/uploads/2021/04/ftrack-Studio-Houdini-%E2%80%93-import-alembic-jpg-1536x840.jpg (even though there's no exposed options in this example) but that's where you'd control whether the representation gets loaded as archive, packed, unpacked, assign materials.... (and you can leave it by default if you don't care) and not the current approach of having an action menu for each permutation of load plugin and representation

antirotor (approved):
I think pointcache can also produce bgeo

PR #229, opened 2024-03-25
Chore: Simple OpenPype to AYON changes

iLLiCiTiT:
## Changelog Description
Another PR changing Openpype/Pype to AYON in codebase. All changes are affecting only prints, comments or UI elements and should not affect real logic.

## Additional info
Use AYON naming in UIs, prints, class & function names, readme or button identifiers.
Removed unused `OpenPypeCreatorError` in 3dsMax integration.
Removed `OpenPypeCreatorError` from houdini which just inherited from `CreatorError`, I didn't find any reason to keep it.

## Testing notes:
- [x] Changed comments/logs make sense.
- [x] 3dsMax menu works.
- [x] Fusion menu works.
- [x] Resolve menu works.
- [x] Houdini does raise correct Create error (don't know how to).

EDITED:
Also renamed avalon to AYON.

### DIFF
--- a/client/ayon_core/hosts/houdini/startup/MainMenuCommon.xml
+++ b/client/ayon_core/hosts/houdini/startup/MainMenuCommon.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <mainMenu>
     <menuBar>
-        <subMenu id="openpype_menu">
+        <subMenu id="ayon_menu">
             <labelExpression><![CDATA[
 import os
 return os.environ.get("AYON_MENU_LABEL") or "AYON"
--- a/client/ayon_core/hosts/max/api/pipeline.py
+++ b/client/ayon_core/hosts/max/api/pipeline.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""Pipeline tools for AYON Houdini integration."""
+"""Pipeline tools for AYON 3ds max integration."""
 import os
 import logging
 from operator import attrgetter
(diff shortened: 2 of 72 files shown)


client/ayon_core/hosts/houdini/startup/MainMenuCommon.xml
antirotor, line 1: This will cause that any third party tool injecting itself there will stop working. Just noting...for the release notes?

client/ayon_core/hosts/max/api/pipeline.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""Pipeline tools for OpenPype Houdini integration."""
+"""Pipeline tools for AYON Houdini integration."""
```
antirotor, line 2: I know that this is changing OP->AYON, but this is so obviously wrong...

```suggestion
"""Pipeline tools for AYON 3ds max integration."""
```

antirotor (approved):

client/ayon_core/pipeline/schema/__init__.py
```diff
@@ -44,11 +44,6 @@ def validate(data, schema=None):
         _precache()
 
     root, schema = data["schema"].rsplit(":", 1)
-    # assert root in (
-    #     "mindbender-core",  # Backwards compatiblity
```
BigRoy, line 48: Should we also take this PR to remove any other remaining `mindbender` reference in the codebase. That would all be from the early days of Avalon. Same goes for "polly" or "polly config" which is still in a few README files (after effects, photoshop, harmony).

iLLiCiTiT: Removed `mindbender` only from places I know I can safely change, but I would say the README files will have to be modified anyway when they'll be separated to addons.

BigRoy: Fine with me - just wanted to make you aware those were legacy things as well.

PR #231, opened 2024-03-25
Fusion: Also update saver tool path on only task or asset change

BigRoy:
## Changelog Description

Also update saver tool path on only task or asset change.

Without this the path would change to the other asset/task if you _combined that_ with a change to subset or image format but not if you only changed asset or task. Which made it quite confusing and could potentially have been overwriting files on the wrong location (not particular to that asset or task it switched to). This changes makes it way more reliable, switching asset or task will result in the new path.

## Additional info

n/a

## Testing notes:

1. Fusion publishing of render and image should work.
2. Changing asset or task (or both) should update the saver's output path.


EmberLightVFX:
Maybe I'm doing something wrong but there is no difference for me with or without this pr.
For me if I simply change the task type from comp to animation, save the change and validate, publish or close the publisher the path in my saver changes to the correct path.

BigRoy:
> Maybe I'm doing something wrong but there is no difference for me with or without this pr. For me if I simply change the task type from comp to animation, save the change and validate, publish or close the publisher the path in my saver changes to the correct path.

Ok, so the reason for that is because the `subset` includes the `task` name by default in the subset naming template. However, if you were to ONLY switch the folder to another with the same task then without this PR the filepath would not update.

Also, if you were to change the templates for your project so that the subset does not include the task name then also on task change it would not change the save path to the other folder.

Anyway, in short, you got 'lucky' with the fact that the task name change influences the subset name and thus would be captured as a change already in the original design.

kalisp:
It is failing for me when changing asset:
image

BigRoy:
> It is failing for me when changing asset: image

Should be fixed with https://github.com/ynput/ayon-core/pull/231/commits/98bed9f5c610deb2f947d643573ce18bd85054da

kalisp:
Should it change folder path when I change asset in the Publisher? It does not in Fusion, change in product name propagates, but character stays the same in the path.

image

EmberLightVFX:
> Should it change folder path when I change asset in the Publisher? It does not in Fusion, change in product name propagates, but character stays the same in the path.

Did you save the change?

kalisp:
YY, I did.

BigRoy:
> YY, I did.

Do you get any errors in the console? Are you able to debug up to where the code triggers? Does your instance somehow not have `productName` data due to it being a legacy older instance? (No idea why it might not trigger for you)

Does it work for a newly created instance?

kalisp:
I think that the issue is that `workdir` in `formatting_data` doesn't get refreshed, it gets pulled only from `AYON_WORKDIR` env var, which will not change if you change asset in Publisher.
image

BigRoy:
> I think that the issue is that `workdir` in `formatting_data` doesn't get refreshed, it gets pulled only from `AYON_WORKDIR` env var, which will not change if you change asset in Publisher. image

Ok, so - I have implemented a change that should work for this. But admittedly, I'm not sure if I really like that having that code there. It feels cluttered and likely slow for many instances. I wonder if we have better options than this.

@EmberLightVFX @iLLiCiTiT thoughts?

kalisp (approved):
This works.

Do we think that this would have such a massive occurrence (changing asset from current context) that it would have any noticeable impact?

BigRoy:
> Do we think that this would have such a massive occurrence (changing asset from current context) that it would have any noticeable impact?

I don't suspect day-to-day usage should see performance issues due to this. I'd have just preferred this to be much simpler and cleaner than this.

PR #234, opened 2024-03-25
Houdini: Load usd to SOPs directly

BigRoy:
## Changelog Description

Implement a SOPs level `usdimport` loader.

## Additional info

This allows artists to directly load managed data into SOPs instead of having load a LOPs state first and then have a `lopimport` inside SOPs. Without that extra bunny-hopping the load logic here may be slightly faster for certain cases.

## Testing notes:

1. Load a USD file to SOPs
2. Manage versions should also work (update)


MustafaJafar (approved):
This is cool! 
It works on my side.

mkolar (approved):

PR #235, opened 2024-03-25
Copy non-python launch script in each integration

iLLiCiTiT:
## Changelog Description
Preparation for split of Photoshop, AfterEffects and Harmony integrations into their addon repositories which requires to not use global script and prelaunch hook.

## Additional info
Each integration has own launch script and prelaunch hook which cares about the same logic, and removed the global script and prelaunch hook. Removed `get_non_python_host_kwargs` function from applications logic. Implemented new helper function `is_using_ui_executable` to determine if UI executable is used on windows.

Renamed constants in each of the host to have `_ADDON_ROOT` suffix instead of `_HOST_DIR` to match where they point (will point).

## Testing notes:
- [ ] Photoshop should start ok, when running from code, using `ayon_console.exe` or `ayon.exe`.
- [ ] AfterEffects should start ok, when running from code, using `ayon_console.exe` or `ayon.exe`.
- [ ] Harmonyshould start ok, when running from code, using `ayon_console.exe` or `ayon.exe`. (if we have a way how to test?)


client/ayon_core/lib/ayon_info.py
```diff
@@ -24,15 +30,41 @@ def is_running_from_build():
...
 
+    """
     executable_path = os.environ["AYON_EXECUTABLE"]
     executable_filename = os.path.basename(executable_path)
     if "python" in executable_filename.lower():
         return False
     return True
 
 
+def is_using_ui_executable():
```
BigRoy, line 42:
```suggestion
def is_using_windows_ui_executable():
```

Should we make it clear that this is windows-specific from the function name as well?

iLLiCiTiT: You're mind reader. We were just discussing it, and came to conclusion that `is_using_ayon_console` might be the best.

iLLiCiTiT: Now I realized I also have to switch the logic...

kalisp (approved):
Small fix, but then it works.

Btw could we change that it would behave when started from code same way as with `ayon_console`, eg. separate console with logs. 
Now it from code it behaves as with `ayon.exe`, eg no separate window.

Or is it for separate PR?

kalisp:
And actually this https://community.ynput.io/t/dcc-communication-over-websocket-code-duplication/1328 is somehow connected to this PR. Could you please comment here (or there) reasoning/future plans?

PR #239, opened 2024-03-25
Chore: Move workfile utils functions to workfile pipeline code

iLLiCiTiT:
## Changelog Description
Moved `should_start_last_workfile` and `should_workfile_tool_start` from applications logic to workfile pipeline code.

## Additional info
Both functions are pipeline related, are using core settings and should be available from anywhere. Because applications logic will be separate into addon this is the ideal change.

## Testing notes:
1. Last workfiles should be opened correctly.
2. Showing of workfiles tool on startup should work too (not sure for which host integrations it really works?).

### DIFF
--- a/client/ayon_core/lib/applications.py
+++ b/client/ayon_core/lib/applications.py
@@ -1789,6 +1789,10 @@ def _prepare_last_workfile(data, workdir, addons_manager):
 
     from ayon_core.addon import AddonsManager
     from ayon_core.pipeline import HOST_WORKFILE_EXTENSIONS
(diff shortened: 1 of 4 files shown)


kalisp (approved):
Works.

Out of curiosity, what parts will stay in `ayon-core` and what parts will go to `ayon-applications` (assuming `applications.py` would go)

iLLiCiTiT:
> Out of curiosity, what parts will stay in ayon-core and what parts will go to ayon-applications (assuming applications.py would go)

Only `applications.py` and some cli (extractenvironments).

PR #242, opened 2024-03-25
Maya: Add setting to disable `cbId` workflow

BigRoy:
## Changelog Description

Add a single setting to disable the `cbId` workflow **validations** in Maya.

## Additional info

This is an equivalent to [this commit](https://github.com/BigRoy/OpenPype/commit/9a543e9fffc1bb3ea58a22b95efaccfb2f6b8632) that we did to disable `cbId` validations in Maya in our USD project.

Implemented because you might want to disable `cbId` completely when moving to USD since it's layering can replace the full `cbId` workflow for looks, etc. Also relevant to this [Ynput discord discussion](https://discord.com/channels/517362899170230292/1219996539146076160/1220057378808922173) with Lucas Bruchhage at BCN Visuals.

This can benefit from a follow up PR that will, using the same toggle, disable the `cbId` creation on file save completely. However - removing that completely might be hard with certain other areas of the code relying on them existing, e.g. certain Arnold exports or Yeti cache workflows. It is entirely true that even with the validations disabled you can hardly rely on them in production - but it's a start to remove them. If you still need them, then you're better of keeping this setting enabled.

Let me know if you want me to remove the creation of the ids on scene save as well.

Note that this may be considered a **major change** since the `cbId` functionality was a key component of lots of logic in the Maya implementation. With e.g. `cbId` workflow disabled the `look` product type would have no real value anymore.

## Testing notes:

1. When the setting is disabled, publishing should never error on `cbId` related logic.
2. When the setting is enabled, all `cbId` validators should function as usually.

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_animation_out_set_related_node_ids.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_animation_out_set_related_node_ids.py
@@ -32,6 +32,13 @@ class ValidateOutRelatedNodeIds(pyblish.api.InstancePlugin,
     ]
     optional = False
 
+    @classmethod
+    def apply_settings(cls, project_settings):
+        # Disable plug-in if cbId workflow is disabled
+        if not project_settings["maya"].get("use_cbid_workflow", True):
+            cls.enabled = False
+            return
+
     def process(self, instance):
         """Process all meshes"""
         if not self.is_active(instance.data):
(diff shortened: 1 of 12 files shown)


LiborBatek (approved):
When ```cbId``` being ```ON``` validators being used and trigger error when multiple Ids being the same
Screenshot 2024-03-26 154226

All works as expected and I was able to produce publish with the same ```cbId``` when turned off as those werent taken into account obviously and the validator was not triggered...

Screenshot 2024-03-26 154542


Screenshot 2024-03-26 154616

LiborBatek:
PS: @BigRoy  I do like very very much your hover over ```hint``` ...we need those pretty much on every settings!! very very cool

Screenshot 2024-03-26 154542

moonyuet (approved):
Looks good for me. It's interesting to know some way to put more info to describe what the option does.

antirotor (approved):
I would leave `cbID` creation there as disabling it optionally would destroy traditional look workflows. When we are ready for that step, we need to provide clear documentation for it and migration steps.

PR #241, opened 2024-03-25
Maya: Fix Validate Node No Ghosting

BigRoy:
## Changelog Description

Fix validator for no enabled 'ghosting' on models.

## Additional info

Fixes:
```python
Traceback:
Traceback (most recent call last):
File "E:\dev\ayon-core\client\ayon_core\pipeline\publish\lib.py", line 257, in publish_plugins_discover
module = import_filepath(abspath, mod_name)
File "E:\dev\ayon-core\client\ayon_core\lib\python_module_tools.py", line 38, in import_filepath
module_loader.exec_module(module)
File "", line 883, in exec_module
File "", line 241, in _call_with_frames_removed
File "E:\dev\ayon-core\client\ayon_core\hosts\maya\plugins\publish\validate_node_no_ghosting.py", line 12, in 
OptionalPyblishPluginMixin):
AttributeError: type object 'InstancePlugin' has no attribute 'OptionalPyblishPluginMixin'
```

## Testing notes:

1. Validator should not be listed under crashed plug-ins.
2. Validate should correctly invalidate ghosting on models.

PR #247, opened 2024-03-26
Maya: Raise PublishValidationError in validators

BigRoy:
## Changelog Description

Improve validation errors to generate an artist-facing report.

## Additional info

The reports are very minor/simple and can be improved over time, but at least now they don't crash the publisher UI.

Also includes the fix of #241 

## Testing notes:

1. Test the relevant validators, they should work.


client/ayon_core/hosts/maya/plugins/publish/validate_vrayproxy.py
```diff
@@ -17,18 +20,18 @@ def process(self, instance):
         if not self.is_active(data):
             return
         if not data["setMembers"]:
-            raise KnownPublishError(
+            raise PublishValidationError(
```
iLLiCiTiT, line 23: Should all these be PublishValidationError? User can't do anything about it...

BigRoy: I wasn't entirely sure - it felt like something the artists did have control over. 

- If there's no set members, then they just added no objects to the instance and they should add it.
- If frame end is below frame start they should fix it on the instance
- If both vrmesh and alembic are disabled for extraction - that'd also be a toggle for the artist to fix.

Not saying that the validation message is perfect the way they are, but at least I feel like these are something _for the artist_ that they can handle?

iLLiCiTiT: I don't know the logic behind, but the message should be modified, at least remove the "This is a bug". If it is a bug it is not artists issue...

antirotor: I agree that this is something artist has power to change, but I would remove the "bug" sentence also.

antirotor (changes requested):

antirotor (approved):
Looks good - it is adding some f-strings but I guess we don't support Maya with python 2.x in AYON anymore.

BigRoy:
> it is adding some f-strings but I guess we don't support Maya with python 2.x in AYON anymore.

Indeed, kill with fire! :)

PR #248, opened 2024-03-26
Maya: optimize validate no null transform

BigRoy:
## Changelog Description

Tiny optimization to validate no null transforms for models with many transforms

## Additional info

n/a

## Testing notes:

1. Empty transforms should still be detected by the validator


mkolar (approved):
Code makes sense. Need to test in app

LiborBatek (approved):
LGTM ...get triggered when the occasion happens

image

PR #256, opened 2024-03-26
Hiero: Collect tracks only if sequence by name exists

iLLiCiTiT:
## Changelog Description
Do not crash when no sequence is selected.

## Additional info
Function `get_current_sequence` may return `None` and in that case the function would crash with `AttributeError`.

This was reported by a client in past, I don't know how it did happen and if this change is "correct" way how to handle it.


tokejepsen:
Found here as well; https://github.com/ynput/OpenPype/pull/6249

iLLiCiTiT:
> Found here as well; https://github.com/ynput/OpenPype/pull/6249

Oh, does it have ayon-core PR?

tokejepsen:
> Oh, does it have ayon-core PR?

Nope, not yet so just get this merged.

jakubjezek001 (approved):

PR #258, opened 2024-03-26
Maya: Load plugins use correct source for product type

iLLiCiTiT:
## Changelog Description
Load plugins are looking for product type on product entity instead of context data.

## Additional info
Key `"family"` may not available on representation context as we've changed it to product type.

## Testing notes:
1. Load plugins can be used on newest published instances from current develop.

EDITED:
The same fix was also done for Unreal plugins.

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/load/load_redshift_proxy.py
+++ b/client/ayon_core/hosts/maya/plugins/load/load_redshift_proxy.py
@@ -32,10 +32,7 @@ class RedshiftProxyLoader(load.LoaderPlugin):
 
     def load(self, context, name=None, namespace=None, options=None):
         """Plugin entry point."""
-        try:
-            product_type = context["representation"]["context"]["family"]
-        except ValueError:
-            product_type = "redshiftproxy"
+        product_type = context["product"]["productType"]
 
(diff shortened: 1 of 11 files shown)


LiborBatek (approved):
I did test on pretty much any type of product and loader....LGTM

BigRoy (approved):
Thanks @LiborBatek for testing - unable to test this quickly myself but the code changes make sense.

Note that the same issue seems to exist in Unreal:
- https://github.com/ynput/ayon-core/blob/576817b8eaa832da111cf2326dabf1460191b80a/client/ayon_core/hosts/unreal/plugins/load/load_alembic_animation.py#L75
- https://github.com/ynput/ayon-core/blob/576817b8eaa832da111cf2326dabf1460191b80a/client/ayon_core/hosts/unreal/plugins/load/load_layout.py#L662
- https://github.com/ynput/ayon-core/blob/576817b8eaa832da111cf2326dabf1460191b80a/client/ayon_core/hosts/unreal/plugins/load/load_layout_existing.py#L396

But maybe that's for @simonebarbieri ?

iLLiCiTiT:
> Note that the same issue seems to exist in Unreal:

It is the same. Fixed in this PR.

PR #259, opened 2024-03-26
Fusion: Prompt reset scene context on saving to another task

BigRoy:
## Changelog Description

Add a prompt to reset comp or content in scene values to the folder or task being saved to.

## Additional info

image

I'll be preparing a similar matching feature for Houdini and Maya.

## Testing notes:

1. Open a comp
2. With workfiles tool, save it to another folder or task


BigRoy:
@EmberLightVFX is this something you might be interested in testing?

EmberLightVFX:
> @EmberLightVFX is this something you might be interested in testing?

When I get some time over I'll do!
Just a question. What does "Reset" mean in this context? Would it mean the checked items would get their info from the current folder/task or the folder/task it's being copied over to?

BigRoy:
> Just a question. What does "Reset" mean in this context? Would it mean the checked items would get their info from the current folder/task or the folder/task it's being copied over to?

It will set it to the folder/task you're saving into. So that if you save your comp to e.g. another folder you directly have it set up to match e.g. that shots' frame range (if you picked that option).

client/ayon_core/hosts/fusion/api/lib.py
```diff
@@ -285,3 +308,99 @@ def comp_lock_and_undo_chunk(
...
+    from ayon_core.lib import BoolDef, UILabelDef
+    from qtpy import QtWidgets, QtCore
+
+    definitions = [
+        UILabelDef(
+            label=(
+                "You are saving your scene into a different task."
+                "\n\n"
+                "Would you like to reset some settings for the "
+                "for the new context?\n"
```
EmberLightVFX, line 353: You have two "for the" here.

EmberLightVFX:
> It will set it to the folder/task you're saving into. So that if you save your comp to e.g. another folder you directly have it set up to match e.g. that shots' frame range (if you picked that option).

Maybe it's just me that finds the sentence `Would you like to reset some settings for the new context?` as it's not my first language but what do you think about `Would you like to change some settings with the new context?`

BigRoy:
> > It will set it to the folder/task you're saving into. So that if you save your comp to e.g. another folder you directly have it set up to match e.g. that shots' frame range (if you picked that option).
> 
> Maybe it's just me that finds it a bit hard to understand `Would you like to reset some settings for the new context?` as it's not my first language but what do you think about `Would you like to change some settings with the new context?`

Thanks, I've updated it and tweaked the wording.

image

If you think that looks great and is clear - then I'll match it in the Houdini and Maya PRs that match this feature.

client/ayon_core/hosts/fusion/api/lib.py
BigRoy: Fixed with https://github.com/ynput/ayon-core/pull/259/commits/a94081e820f562dd7574607a4ce4e24bada666a5

EmberLightVFX:
The only thing I see is the title of the window says "saving to different context" while the first row says "...saving your scene to a different task".

Ifk exactly how the terminaligy is for Ayon but should "scene" be "product" and "task" be "context"?

BigRoy:
> Ifk exactly how the terminaligy is for Ayon but should "scene" be "product" and "task" be "context"?

I think scene should then be `workfile` - and `context` is basically the "folder path + task".

I'll change:
- `scene` -> `workfile`
- `task` -> `context`

BigRoy:
image

Actually - I tried one more different thing.

By mentioning that you are saving to a different "folder" or "task" I'm hoping this clarifies a bit what the 'context' is the pop-up refers to. What do you think of this?

EmberLightVFX:
> By mentioning that you are saving to a different "folder" or "task" I'm hoping this clarifies a bit what the 'context' is the pop-up refers to. What do you think of this?

Love it!

BigRoy:
I fixed the merge conflicts - should be ready to merge.

PR #260, opened 2024-03-26
Fusion: Support nested usage of the `comp_lock_and_undo_chunk` context manager

BigRoy:
## Changelog Description

Support nested usage of the `comp_lock_and_undo_chunk` context manager

## Additional info

As far as I know there's no current pressing issue that should be 'tested' against whether it's resolved. Consider this more of a 'safe improvement' to allow nesting the comp locking.

## Testing notes:

1. Publishing should work as usual.
2. Nothing special to test really.

### DIFF
--- a/client/ayon_core/hosts/fusion/api/lib.py
+++ b/client/ayon_core/hosts/fusion/api/lib.py
@@ -278,6 +278,13 @@ def comp_lock_and_undo_chunk(
     keep_undo=True,
 ):
     """Lock comp and open an undo chunk during the context"""
+
+    # If comp is already locked do nothing and consider us to
+    # be inside another undo chunk currently. This allows nesting the chunks.
+    if comp.IsLocked():
+        yield
+        return
+
     try:
         comp.Lock()
         comp.StartUndo(undo_queue_name)


EmberLightVFX:
Fusion already support nested lock and undo chunks. If you run comp.Lock() 3 times you need to run comp.Unlock() 3 times before you can start using the UI again.

I actually have this script in case I locked the UI x amount of times but didn't unlock 😄 
```
if comp:IsLocked() then
	while comp:IsLocked() do
		comp:EndUndo(true)
		comp:Unlock()
	end
	print("Comp unlocked!")
else
	print("Comp not locked")
end
```

BigRoy:
> Fusion already support nested lock and undo chunks. If you run comp.Lock() 3 times you need to run comp.Unlock() 3 times before you can start using the UI again.
> 
> I actually have this script in case I locked the UI x amount of times but didn't unlock 😄
> 
> ```
> if comp:IsLocked() then
> 	while comp:IsLocked() do
> 		comp:EndUndo(true)
> 		comp:Unlock()
> 	end
> 	print("Comp unlocked!")
> else
> 	print("Comp not locked")
> end
> ```

Thanks! Looking back at the comment in the code this brought me back to why I did this at some point:

> This allows nesting the chunks.

Without this, using the context manager would basically mean (I think?) that any change inside of it would always turn into its own undo - instead of the outer chunk. Anyway, I just tried to reproduce that issue - but I can't seem to reproduce that.

I'll close this PR.

PR #261, opened 2024-03-26
Maya: Prompt reset scene context on saving to another task

BigRoy:
## Changelog Description

Add a prompt to reset workfile data (frame_range, fps, resolution, colorspace) or publish instances (target folder+task) to the folder or task being saved to.

## Additional info

image

I'll be preparing a similar matching feature for Houdini

## Additional info

This is the Maya equivalent of #259 

## Testing notes:

1. Open a workfile
2. Save it to another folder or task via workfile tool


MustafaJafar (approved):
Tested as user, it works on my side.

tokejepsen (approved):

client/ayon_core/hosts/maya/api/pipeline.py
```diff
@@ -67,6 +67,9 @@
 
 AVALON_CONTAINERS = ":AVALON_CONTAINERS"
 
+# Track whether the workfile tool is about to save
+ABOUT_TO_SAVE = False
```
iLLiCiTiT, line 71: Already merged, but this looks like public constant. But the value is changing. Can this be changed to private, not capitalized variable `_about_to_save`?

BigRoy: Will make a PR in a bit that does this for all hosts - same logic is also in Fusion and Houdini

PR #262, opened 2024-03-26
Houdini: Prompt reset scene context on saving to another task

BigRoy:
## Changelog Description

Add a prompt to reset workfile data (frame_range, fps) or publish instances (target folder+task) to the folder or task being saved to.

## Additional info

image

This is the Houdini equivalent of #259 and #261 

## Testing notes:

1. Open a workfile
2. Save it to another folder or task via workfile tool


MustafaJafar (approved):
It works on my side. 
I think we may replace the current popups (update FPS , etc ) with this cool prompt. 

I was going to suggest adding `update_houdini_vars_context` but this could be another PR. ( we might like to show the value before and after)

MustafaJafar:
For sake of curiosity: 
Do we have a list events ? and when they are emitted ? @iLLiCiTiT 
The logic behind callbacks in this PR is tested and working, but I find it a little confusing.

BigRoy:
> I was going to suggest adding `update_houdini_vars_context` but this could be another PR. ( we might like to show the value before and after)

Not entirely sure whether this is the perfect 'merge' since that should also run on file _open_ in another context I believe, not just on save - anyway. Definitely something for another PR.

antirotor (approved):
works nice, I think it can always be enhanced in the future

BigRoy:
Thanks so much for the quick testing and merge! Looking forward to seeing the Fusion equivalent #259 and Maya equivalent #261 to get merged!

PR #263, opened 2024-03-26
Houdini: Remove legacy unused id logic

BigRoy:
## Changelog Description

Remove generation of id logic in Houdini.

## Additional info

A long time ago this was once implemented to try and match the `cbId` workflow in Maya - but from Houdini's perspective that never got traction and had no use. The code remained unused for any production use cases so can be safely removed.

## Testing notes:

1. Saving scene should work
2. Publishing and loading should work as usual

You should notice no differences.


MustafaJafar (approved):
Tested. everything works as usual.

antirotor (approved):

PR #264, opened 2024-03-26
Houdini: Fix correctly updating context data on saving in publisher

BigRoy:
## Changelog Description

Fix correctly updating context data on saving in publisher

## Additional info

Without this fix it would not work to save and persist changes to e.g. ignoring outdated containers.

## Testing notes:

1. Publishing from Houdini should work
2. Enabling/disabling attributes on the Context should persist correctly and remaind applied on "reset" of the publisher and scene re-opens, etc.


MustafaJafar:
> 2. Enabling/disabling attributes on the Context should persist correctly and remaind applied on "reset" of the publisher and scene re-opens, etc.

Just to make sure I got it. is my screenshot an example of `attributes on the Context`  ? 
image

BigRoy:
> Just to make sure I got it. is my screenshot an example of `attributes on the Context` ?

No. Top left of the publisher click the **Context** item. The attributes on that one are the Context attributes.

image

antirotor (approved):

PR #267, opened 2024-03-26
Maya: Validate Model Content improve validation message

BigRoy:
## Changelog Description

Validator for model product type content should provide a better error message.

## Additional info

Also fix `get_invalid` actually returning the instance node.

## Testing notes:

1. Publish model with invalid geometry, e.g. a proxy shape inside - it should be captured by this validator.

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_model_content.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_model_content.py
@@ -28,13 +28,15 @@ class ValidateModelContent(pyblish.api.InstancePlugin,
     validate_top_group = True
     optional = False
 
+    allowed = ('mesh', 'transform', 'nurbsCurve', 'nurbsSurface', 'locator')
+
     @classmethod
     def get_invalid(cls, instance):
 
         content_instance = instance.data.get("setMembers", None)
         if not content_instance:
             cls.log.error("Instance has no nodes!")
-            return [instance.data["name"]]
+            return [instance.data["instance_node"]]
 
         # All children will be included in the extracted export so we also
         # validate *all* descendents of the set members and we skip any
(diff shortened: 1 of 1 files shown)


MustafaJafar (approved):
Tested as user.
It works on my side.

image

image

BigRoy:
Actually - let me actually improve the "description" for this validator. Because this still doesn't look very nice.

BigRoy:
Report should be improved now.

MustafaJafar (approved):
This is helpful! 

image

image

LiborBatek (approved):
I have ran several different test occasions with following results (in terms of Validation message) for your consideration @BigRoy 

Multiple Roots present:


Screenshot 2024-04-05 101343

Empty GRP with no geometries at all:


Screenshot 2024-04-05 101428

And if some Shape not being visible (actually I have turned on ```Intermediate Shape``` in the ShapeNode attribs) ...sort of works for this occasion too even tho its not 100% accurate info for user whats causing it and also when triggering ```Repair``` it simply removes the object which is a bit destructive imho...but works somewhat

Screenshot 2024-04-05 101639

Would be even cooler if it could detect the Intermediate shape and switch it off...and notify this more accurately....however its a bit of edge case aka synthetic circumstance anyway

LiborBatek:
@BigRoy  maybe for the last occasion with ```Empty Transform``` we could enhance the detailed message on the right with something like:

```Empty transforms found without Shapes or with "Intermediate" shape enabled. Warning: by hitting "Repair" action these will be removed!```

or similar. What u think?

LiborBatek (changes requested):
Would be nice to adress the occasion when Shape not visible and informing user that ```Repair``` will actually delete those invalid entities...

BigRoy:
> @BigRoy maybe for the last occasion with `Empty Transform` we could enhance the detailed message on the right with something like:
> 
> `Empty transforms found without Shapes or with "Intermediate" shape enabled. Warning: by hitting "Repair" action these will be removed!`
> 
> or similar. What u think?

Thanks for the thorough testing. I'd like to argue that the current behavior is actually correct.

The model family exports without construction history and thus should ignore the intermediate objects. As such, to what would be published that is just an "empty mesh". In day to day production I've also never seen a modeler count a Transform with an intermediate object shape under it (which Maya doesn't show in outliner by default, etc.) as a "mesh" - they just see the empty transform and often don't even know that it was containing an intermediate object - which usually happens due to modeling tools and construction history they make. As such, the input node that had the intermediate mesh usually also counts as "history" and thus should be removed.

I'd say, if there's more debate about it we can do so in another PR? @LiborBatek

LiborBatek (approved):

LiborBatek:
@BigRoy yeah, I get your point...no need to take such occasion into account

in terms of "improving validation message" I can see some benefits to include that "removal warning" as its not repairing but deleting stuff from the workfile altering it. 

And it does not affect the current correct behaviour in any way just improving the message for a user.

Lets merge it anyway...

PR #269, opened 2024-03-26
Maya: Tweak validate transform zero formatting

BigRoy:
## Changelog Description

Tweak validate transform zero artist report formatting.

- Add line break before first object name
- Add static description instead of listing objects right-hand side.

## Additional info

image

## Testing notes:

1. Enable `ValidateTransformZero` plugin in settings
2. Publish model with transforms
2. Validation should be clear.


MustafaJafar (approved):
Tested as a user, Looks cool.

LiborBatek (approved):
LGTM

image

PR #271, opened 2024-03-26
Fix: Report the duplicated creator class

BigRoy:
## Changelog Description

Fix warning message: Actually report the duplicated creator class.

## Additional info

Before it wouldn't actually list the duplicated class.

## Testing notes:

1. Register two creators with same identifier.
2. The log should show the conflicting classes.

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -1790,10 +1790,10 @@ class CreateContext:
 
             creator_identifier = creator_class.identifier
             if creator_identifier in creators:
-                self.log.warning((
-                    "Duplicated Creator identifier. "
-                    "Using first and skipping following"
-                ))
+                self.log.warning(
+                    "Duplicated Creator identifier. Using first and "
+                    "skipping following: {}".format(str(creator_class))
+                )
                 continue
 
             # Filter by host name


client/ayon_core/pipeline/create/context.py
```diff
@@ -1790,10 +1790,10 @@ def _reset_creator_plugins(self):
 
             creator_identifier = creator_class.identifier
             if creator_identifier in creators:
-                self.log.warning((
-                    "Duplicated Creator identifier. "
-                    "Using first and skipping following"
-                ))
+                self.log.warning(
+                    "Duplicated Creator identifier. Using first and "
+                    "skipping following: {}".format(str(creator_class))
```
iLLiCiTiT, line 1795: Not saying it's not important, but how does logging class name of one of the classes help? Shouldn't be logged `creator_identifier` instead?

BigRoy: Well, since the duplication usually arises from different classes by accident using the same identifier - and this shows which of the two classes that were both using the same identifier is being ignored. (The first one will still be loaded)

So at least it's clear which of the two is skipped.

But I do agree that it might be good to also include the identifier itself in the message. Shall I add it?

BigRoy: Changed this with https://github.com/ynput/ayon-core/pull/271/commits/07cc3bba0b61fbb4a7d8260dddd43690de2d2f0b - let me know if you like it this way.

iLLiCiTiT (approved):

PR #272, opened 2024-03-27
Houdini: Validate COP Output Node tweak message formatting

BigRoy:
## Changelog Description

Minor cosmetic tweak. It only ever refers to one note.

Plus improve how Publisher shows the characters - it sometimes trips over the ` character.

## Additional info

n/a

## Testing notes:

1. COP (Publish image from Houdini) node validation when no output node set should be nicely reported.


MustafaJafar:
I think this piece of code 
```python
        try:
            output_node = instance.data["output_node"]
        except KeyError:
            six.reraise(
                PublishValidationError,
                PublishValidationError(
                    "Can't determine COP output node.",
                    title=cls.__name__),
                sys.exc_info()[2]
            )
```

should only be 
```python
output_node = instance.data["output_node"]
```

BigRoy:
> I think this piece of code
> 
> ```python
>         try:
>             output_node = instance.data["output_node"]
>         except KeyError:
>             six.reraise(
>                 PublishValidationError,
>                 PublishValidationError(
>                     "Can't determine COP output node.",
>                     title=cls.__name__),
>                 sys.exc_info()[2]
>             )
> ```
> 
> should be
> 
> ```python
> output_node = instance.data.get("output_node")
> ```

I agree - it's weird code.

MustafaJafar (approved):
Tested. working on my side .

image

BigRoy:
Improve the report just a tad more. 👍

MustafaJafar (commented):
Are there any difference between those two ? 
```python
not isinstance(output_node, hou.CopNode)
```
and 
```python
output_node.type().category().name() != "Cop2"
```

BigRoy:
> Are there any difference between those two ?
> 
> ```python
> not isinstance(output_node, hou.CopNode)
> ```
> 
> and
> 
> ```python
> output_node.type().category().name() != "Cop2"
> ```

I'm not sure. I believe something may be inside Cops and thus be a `CopNode` but be of a different category, e.g. an output node inside Cops. (Or vice versa?) I'm not entirely sure, but I believe there may be a difference.

client/ayon_core/hosts/houdini/plugins/publish/validate_cop_output_node.py
```diff
@@ -71,8 +58,9 @@ def get_invalid(cls, instance):
         # the isinstance check above should be stricter than this category
         if output_node.type().category().name() != "Cop2":
```
MustafaJafar, line 59: I find this confusing. 
sudo code.
```
if the instance not instance of hou.CopNode: 
    print its category type name.  (1)
    mark output node as invalid. 
if the category type name is not Cop2: 
    print that node is not of category Cop2 
    mark output node as invalid.
```
Hypothetically, what if the first condition was valid and at the same time the second one was invalid ? 
then the printed value at (1) will be wrong.

client/ayon_core/hosts/houdini/plugins/publish/validate_cop_output_node.py
```diff
@@ -71,8 +58,9 @@ def get_invalid(cls, instance):
...
-                    "Output node {} is not of category Cop2."
-                    " This is a bug..."
-                ).format(output_node.path()),
-                title=cls.label)
+                f"Output node {output_node.path()} is not of category Cop2.",
+                description=(
+                    "### Invalid COP output node\n\n"
+                    "The output node path for the instance must be set to a "
+                    "valid COP node path. See the log for more details."
+                ))
```
MustafaJafar, line 66: for sake of consistency, 
should this be: 
```suggestion
            cls.log.error(
                f"Output node {output_node.path()} is not of category Cop2.",
            )
            return [output_node.path()]
```

client/ayon_core/hosts/houdini/plugins/publish/validate_cop_output_node.py
```diff
@@ -26,28 +25,16 @@ def process(self, instance):
         invalid = self.get_invalid(instance)
         if invalid:
             raise PublishValidationError(
-                ("Output node(s) `{}` are incorrect. "
+                ("Output node '{}' is incorrect. "
                  "See plug-in log for details.").format(invalid),
                 title=self.label
             )
```
MustafaJafar, line 31: and this be: 
```suggestion
            raise PublishValidationError(
                title=self.label,
                description=(
                    "### Invalid COP output node\n\n"
                    "The output node path for the instance '{}' must "
                    "be set to a valid COP node path. See the log for "
                    "more details.".format(invalid)
                )
            )
```

MustafaJafar (commented):
I wonder if this change should be done for sake of consistency ?

client/ayon_core/hosts/houdini/plugins/publish/validate_cop_output_node.py
BigRoy: The description should not be used to display instance specific data because then by default you will only see if for one of the instances, not all. Description should hence try to be 'static' for the plug-in.

client/ayon_core/hosts/houdini/plugins/publish/validate_cop_output_node.py
BigRoy: Sorry, I don't get what you're saying. As the comment in the code describes - this check _should_ be stricter and the previous one more generic. According to that comment then the first _may_ be valid but be disallowed here. I do agree that it's odd that this directly raises an error instead of logging and returning. It should still behave 'similarly'. So with that, I agree.

MustafaJafar (approved):
Tested. Looks good. 

image
image

antirotor (approved):

PR #274, opened 2024-03-27
Houdini: Alembic Load do not create `unpack`, `normal` and `null`

### LINKED ISSUE #253: Houdini: Load Alembic - disable automatic unpacking

Loading an Alembic into Houdini via the loader [automatically adds an "Unpack" and "Normal" node](https://github.com/ynput/ayon-core/blob/5db37c9590174fd913a51fc006102c32d374076a/client/ayon_core/hosts/houdini/plugins/load/load_alembic.py#L47-L66) which a long time ago was added for ease of access to the data (by artists at our studio).

However, @fabiaserra [stated here](https://github.com/ynput/ayon-core/pull/227#discussion_r1539019570) that it might not be a good idea to keep that around and instead rely on automating less and instead leave it up to Houdini artists to manage the data. In essence, not creating those additional nodes on load.

Since it has been there for so long we might want to consider backwards compatibility. Just removing the logic for creation of those nodes should still work fine with existing loaded entries (they will just keep the nodes around) so it should be trivial to be 'forward' compatible.

The real question becomes, do we ever still want to _keep_ the logic that it automatically creates it for you.

We could add it as "Options" on the Loader (little option box on the loader) or could make it a Project Settings to enable/disable the logic.

_No response_

### Additional context:

_No response_

[cuID:[AY-4796](https://app.clickup.com/t/86by894ge)]

BigRoy:
## Changelog Description

On Alembic loading to SOPs do not automatically create `unpack`, `normal` and `null`

## Additional info

Fixes #253 based on comments by @fabiaserra 

Also removed the assignment to `self[:]` since it's unused.

## Testing notes:

1. Loading and managing loaded Alembics should work
2. Managing "legacy" loaded containers should also work
    - They will still have the unpack, normal and null - only newly loaded containers will not.


MustafaJafar (approved):
It works on my side. 
I wonder if we should add output node or null node at the end.
I think it can save sometime for artists.

BigRoy:
> It works on my side.
> I wonder if we should add output node or null node at the end.
> I think it can save sometime for artists.

I wouldn't do that. We want to provide the minimal. It being the only node inside of it - it's already the "output" and we're also not doing this for e.g. the `bgeo` loader or other SOPs loaders as far as I know.

Wonder what @fabiaserra thinks?

fabiaserra:
> > It works on my side.
> > I wonder if we should add output node or null node at the end.
> > I think it can save sometime for artists.
> 
> I wouldn't do that. We want to provide the minimal. It being the only node inside of it - it's already the "output" and we're also not doing this for e.g. the `bgeo` loader or other SOPs loaders as far as I know.
> 
> Wonder what @fabiaserra thinks?

Yeah I don't think it's necessary, although it would be good to add support to switch to the context where it gets loaded so it's clear to the artist that something was loaded

BigRoy:
> Yeah I don't think it's necessary, although it would be good to add support to switch to the context where it gets loaded with the new node selected so it's clear to the artist that something was loaded

Ah yes, I think that's a different beast to tackle since currently it loads _into the avalon container_ which (at least for us) isn't a workspace one usually works in, but usually gets object merged.

Anyway, definitely different issue.

antirotor (approved):

PR #196, opened 2024-03-19
Chore: Fix typos and linter errors

antirotor:
## Changelog Description
Fixing typos and linter errors in the repository.

## Additional info
Based on #192 this is using codespell and ruff to cleanup the repo. (Un)fortunately any typo-fixed is marking the file as a food for ruff and so it starts triggering linter errors in pre-commit hook preventing the commit. The only way around it is obey the ruff tool - and that is a good thing, isn't it?

> [!NOTE]
> This is using #192 for tools

## Fixing workflow
1) run codespell (you can set it as External tool in PyCharm for example) - `./.poetry/bin/poetry run codespell`
2) take the list of errors, fix them and run ruff on that changed file - `./.poetry/bin/poetry run ruff check /path/to/file.py`
3) fix linter errors
4) commit

PR #280, opened 2024-03-27
Chore: Cleanup imports

iLLiCiTiT:
## Changelog Description
Cleaned up imports in codebase. Removed unused imports, added some missing imports and filled `__all__` where should be. None of the changes should affect the code in any way.

NOTE: PR based on changes in https://github.com/ynput/ayon-core/pull/196 , but not all imports were handled in this PR. Some removed imports in the PR are actually there because of Python 2 type-hinting (e.g. `# type: Anatomy`). I think those require separate PR done by responsible people.


client/ayon_core/hosts/substancepainter/api/pipeline.py
```diff
@@ -76,7 +74,7 @@ def install(self):
         log.info("Installing menu ... ")
         self._install_menu()
 
-        project_settings = get_current_project_settings()
+        project_settings = get_project_settings(get_current_project_name())
```
BigRoy, line 77: Is `get_current_project_settings` deprecated?

BigRoy (approved):
Didn't test but looks good to me at first sight. Just had the one comment about removal of `get_current_project_settings` usage.

client/ayon_core/hosts/substancepainter/api/pipeline.py
iLLiCiTiT: > Is get_current_project_settings deprecated? Is there a reason for this code change?

I had some local changes and didn't realized there was `get_current_project_settings` before. But it is good question. Function `get_current_project_settings` is using envrionment variable for current project, but that may be different from `get_current_project_name`...

PR #283, opened 2024-03-27
Chore: Formatting fixes

iLLiCiTiT:
## Changelog Description
Small enhancements of code formatting.

## Additional info
Removed unused variables. Removed unnecessary f-strings, removed `format` when there is nothing to format, fixed some formattings which used combination of `{}` and `%s` for formatting. Fixed one Python 2 exception capture (`except Exception, e` > `except Exception as e`). Use `isinstance` for type check. Converted lambda to real function.

NOTE: Based on PR https://github.com/ynput/ayon-core/pull/196 .

### DIFF
--- a/client/ayon_core/hosts/hiero/api/otio/hiero_import.py
+++ b/client/ayon_core/hosts/hiero/api/otio/hiero_import.py
@@ -101,7 +101,7 @@ def apply_transition(otio_track, otio_item, track):
     if transition_type == 'dissolve':
         transition_func = getattr(
             hiero.core.Transition,
-            'create{kind}DissolveTransition'.format(kind=kind)
+            "create{kind}DissolveTransition".format(kind=kind)
         )
 
         try:
@@ -109,7 +109,7 @@ def apply_transition(otio_track, otio_item, track):
                 item_in,
                 item_out,
                 otio_item.in_offset.value,
-                otio_item.out_offset.value
+                otio_item.out_offset.value,
             )
 
         # Catch error raised if transition is bigger than TrackItem source
@@ -134,7 +134,7 @@ def apply_transition(otio_track, otio_item, track):
 
         transition = transition_func(
             item_out,
-            otio_item.out_offset.value
+            otio_item.out_offset.value,
         )
 
     elif transition_type == 'fade_out':
@@ -183,9 +183,7 @@ def prep_url(url_in):
 def create_offline_mediasource(otio_clip, path=None):
     global _otio_old
 
-    hiero_rate = hiero.core.TimeBase(
-        otio_clip.source_range.start_time.rate
-    )
+    hiero_rate = hiero.core.TimeBase(otio_clip.source_range.start_time.rate)
 
     try:
         legal_media_refs = (
@@ -212,7 +210,7 @@ def create_offline_mediasource(otio_clip, path=None):
         source_range.start_time.value,
         source_range.duration.value,
         hiero_rate,
-        source_range.start_time.value
+        source_range.start_time.value,
     )
 
     return media
@@ -385,7 +383,8 @@ def create_trackitem(playhead, track, otio_clip, clip):
             # Only reverse effect can be applied here
             if abs(time_scalar) == 1.:
                 trackitem.setPlaybackSpeed(
-                    trackitem.playbackSpeed() * time_scalar)
+                    trackitem.playbackSpeed() * time_scalar
+                )
 
         elif isinstance(effect, otio.schema.FreezeFrame):
             # For freeze frame, playback speed must be set after range
@@ -397,28 +396,21 @@ def create_trackitem(playhead, track, otio_clip, clip):
         source_in = source_range.end_time_inclusive().value
 
         timeline_in = playhead + source_out
-        timeline_out = (
-            timeline_in +
-            source_range.duration.value
-        ) - 1
+        timeline_out = (timeline_in + source_range.duration.value) - 1
     else:
         # Normal playback speed
         source_in = source_range.start_time.value
         source_out = source_range.end_time_inclusive().value
 
         timeline_in = playhead
-        timeline_out = (
-            timeline_in +
-            source_range.duration.value
-        ) - 1
+        timeline_out = (timeline_in + source_range.duration.value) - 1
 
     # Set source and timeline in/out points
     trackitem.setTimes(
         timeline_in,
         timeline_out,
         source_in,
-        source_out
-
+        source_out,
     )
 
     # Apply playback speed for freeze frames
@@ -435,7 +427,8 @@ def create_trackitem(playhead, track, otio_clip, clip):
 
 
 def build_sequence(
-        otio_timeline, project=None, sequence=None, track_kind=None):
+    otio_timeline, project=None, sequence=None, track_kind=None
+):
     if project is None:
         if sequence:
             project = sequence.project()
@@ -509,10 +502,7 @@ def build_sequence(
 
                 # Create TrackItem
                 trackitem = create_trackitem(
-                    playhead,
-                    track,
-                    otio_clip,
-                    clip
+                    playhead, track, otio_clip, clip
                 )
 
                 # Add markers
--- a/client/ayon_core/hosts/hiero/api/startup/Python/StartupUI/otioimporter/OTIOImport.py
+++ b/client/ayon_core/hosts/hiero/api/startup/Python/StartupUI/otioimporter/OTIOImport.py
@@ -254,12 +254,6 @@ def add_markers(otio_item, hiero_item, tagsbin):
         if _tag is None:
             _tag = hiero.core.Tag(marker_color_map[marker.color])
 
-        start = marker.marked_range.start_time.value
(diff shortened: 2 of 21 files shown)


client/ayon_core/hosts/hiero/api/startup/Python/StartupUI/otioimporter/OTIOImport.py
```diff
@@ -101,35 +101,29 @@ def apply_transition(otio_track, otio_item, track):
...
-            transition = transition_func(
-                item_out,
-                otio_item.out_offset.value
-            )
+            transition = transition_func(item_out, otio_item.out_offset.value)
 
         elif transition_type == "fade_out":
             transition_func = getattr(
                 hiero.core.Transition,
                 'create{kind}FadeOutTransition'.format(kind=kind)
```
kalisp, line 124:
```suggestion
                "create{kind}FadeOutTransition".format(kind=kind)
```

client/ayon_core/hosts/houdini/plugins/load/load_alembic.py
```diff
@@ -59,7 +59,7 @@ def load(self, context, name=None, namespace=None, data=None):
 
         normal_node.setInput(0, unpack)
 
-        null = container.createNode("null", node_name="OUT".format(name))
+        null = container.createNode("null", node_name="OUT")
```
kalisp, line 62: As a formatting change it is okay, but I am wondering if it shouldn't be adding name into the node name to make it more unique, but I don't know anything about houdini integration.

BigRoy: This is actually opted for removal in #253 - so I'd say disregard.
Also, it's always "unique" inside the container itself, so no need to add extra names, etc.

This change here is thus OK as is.

client/ayon_core/hosts/hiero/api/otio/hiero_import.py
```diff
@@ -397,28 +396,21 @@ def create_trackitem(playhead, track, otio_clip, clip):
         source_in = source_range.end_time_inclusive().value
 
         timeline_in = playhead + source_out
-        timeline_out = (
-            timeline_in +
-            source_range.duration.value
-        ) - 1
+        timeline_out = (timeline_in + source_range.duration.value) - 1
```
BigRoy, line 399: Do we still care about keeping the brackets?
```suggestion
        timeline_out = timeline_in + source_range.duration.value - 1
```

client/ayon_core/hosts/hiero/api/otio/hiero_import.py
```diff
@@ -397,28 +396,21 @@ def create_trackitem(playhead, track, otio_clip, clip):
...
         # Normal playback speed
         source_in = source_range.start_time.value
         source_out = source_range.end_time_inclusive().value
 
         timeline_in = playhead
-        timeline_out = (
-            timeline_in +
-            source_range.duration.value
-        ) - 1
+        timeline_out = (timeline_in + source_range.duration.value) - 1
```
BigRoy, line 406: Do we still care about keeping the brackets?
```suggestion
        timeline_out = timeline_in + source_range.duration.value - 1
```

kalisp (approved):
Visually it look ok.

client/ayon_core/hosts/hiero/api/startup/Python/StartupUI/otioimporter/OTIOImport.py
```diff
@@ -153,18 +146,13 @@ def prep_url(url_in):
     if url.startswith("file://localhost/"):
         return url.replace("file://localhost/", "")
 
-    url = '{url}'.format(
-        sep=url.startswith(os.sep) and "" or os.sep,
-        url=url.startswith(os.sep) and url[1:] or url
-    )
+    url = "{url}".format(url=url.startswith(os.sep) and url[1:] or url)
```
BigRoy, line 149: Uh - what? There's something wrong with this logic.
This will make `url = "True"` if url startswith with `os.sep`

I think this was intended to be: 

```suggestion
    url = url[1:] if url.startswith(os.sep) else url
```

Or:

```suggestion
    if url.startswith(os.sep):
        url = url[1:]
```

Or:

```suggestion
    url = url.lstrip(os.sep)
```
_this last one may strip more however_

client/ayon_core/hosts/hiero/plugins/publish_old_workflow/precollect_retime.py
```diff
@@ -36,8 +36,8 @@ def process(self, instance):
         speed = track_item.playbackSpeed()
 
         # calculate available material before retime
-        available_in = int(track_item.handleInLength() * speed)
-        available_out = int(track_item.handleOutLength() * speed)
+        # available_in = int(track_item.handleInLength() * speed)
+        # available_out = int(track_item.handleOutLength() * speed)
```
BigRoy, line 40: Were these unused? If so, should we remove the commented code?

client/ayon_core/hosts/maya/plugins/publish/extract_unreal_skeletalmesh_fbx.py
```diff
@@ -74,7 +74,7 @@ def process(self, instance):
             renamed_to_extract.append("|".join(node_path))
 
         with renamed(original_parent, parent_node):
-            self.log.debug("Extracting: {}".format(renamed_to_extract, path))
+            self.log.debug("Extracting: {}".format(renamed_to_extract))
```
BigRoy, line 77: Can rely on `logging` formatting here.
```suggestion
            self.log.debug("Extracting: %s", renamed_to_extract)
```

client/ayon_core/hosts/hiero/api/otio/hiero_import.py
BigRoy: (There are a few more lines of code where this is the case in this PR)

client/ayon_core/hosts/hiero/api/otio/hiero_import.py
iLLiCiTiT: Technically the same, but visually helps to keep it in brackets.

BigRoy: Cool - feel free to mark resolved.

client/ayon_core/hosts/maya/plugins/publish/extract_unreal_skeletalmesh_fbx.py
BigRoy: Sure should.

```python
logging.info("hello %s", ["test", "bar"])
# root : hello ['test', 'bar']
```

client/ayon_core/hosts/hiero/plugins/publish_old_workflow/precollect_retime.py
iLLiCiTiT: I've copied it from the other PR. Looks like it is not needed, and it is in "old workflow" so I guess it is deprecated? Nevermind, removing.

client/ayon_core/hosts/maya/plugins/publish/extract_unreal_skeletalmesh_fbx.py
iLLiCiTiT: But if you would modify the list, the publisher would not show the values that the variable had at the moment it was logged.

```python
my_var = ["test", "bar"]
logging.info("hello1 %s", my_var)
my_var.append("new")
logging.info("hello2 %s", my_var)
```
Both of these logs would show `hello ['test', 'bar', 'new']` in UI.

BigRoy: Well - I haven't looked much into how the `logging` library works but I wouldn't expect that to be the case.

Also, the publisher UI handles the log directly with its handler, so then I definitely wouldn't expect that to be the case.

But hey, happy to be proven wrong.

iLLiCiTiT: We actually did hit these issues when we used the logging formatting. If plugin changed content of logged data it did not show correct output in UI.

client/ayon_core/hosts/hiero/plugins/publish_old_workflow/precollect_retime.py
antirotor: I've commented them out in the original PR because I thought that even though they are not used downstream, it can server as an example how to get those values. Otherwise it is safe to delete

client/ayon_core/hosts/maya/plugins/publish/extract_unreal_skeletalmesh_fbx.py
BigRoy: Interesting - that's very good to know actually; I wasn't aware of that being a potential issue. I understand the mutable arguments, etc. but would have expected a library like `logging` to maybe keep around immutable copies or whatever.

Thanks.

PR #287, opened 2024-03-27
Maya: Yeti - Implement writing and loading user variables with a yeti cache

BigRoy:
## Changelog Description

Implement writing and loading user variables with a yeti cache

## Additional info

Support writing and loading of user variables in a yeti node with the `yeticache` product type.

## Testing notes:

1. Loading legacy yeti products should work
2. Publish a new yeti product without user variables should work + loading it + updating it
3. Publish a new yeti product with user variables should work + loading it + updating it


client/ayon_core/hosts/maya/api/yeti.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            cmds.setAttr(plug, value)
+
+        return True
+    return False
+
+
+def _create_vector_yeti_user_variable(yeti_shape_node: str, attr_name: str):
+    if not attr_name.startswith("yetiVariableV_"):
+        raise ValueError("Must start with yetiVariableV_")
```
moonyuet, line 76: Maybe we can raise the `LoadError` which can be implemented from the loader plugin in ayon pipeline?

client/ayon_core/hosts/maya/api/yeti.py
```diff
@@ -0,0 +1,101 @@
...
+                     longName="{}{}".format(attr_name, axis),
+                     attributeType="double",
+                     parent=attr_name,
+                     cachedInternally=True,
+                     keyable=True)
+
+
+def _create_float_yeti_user_variable(yeti_node: str, attr_name: str):
+    if not attr_name.startswith("yetiVariableF_"):
+        raise ValueError("Must start with yetiVariableF_")
```
moonyuet, line 93: Same above

client/ayon_core/hosts/maya/api/yeti.py
BigRoy: We could - but honestly, this should never actually occur during load because the extraction and load logic already ensures that basically these should be prefixed - specifically because they are called from `create_yeti_variable` which will only ever call one of the two if the names are valid.

The raising of the error is just there so that if someone were to somehow call these private functions and misuse them there would be warned about it.

client/ayon_core/hosts/maya/api/yeti.py
BigRoy: Same as comment [above](https://github.com/ynput/ayon-core/pull/287#discussion_r1569457579)

LiborBatek:
@BigRoy sorry being ignorant, could u elaborate more on those user variables pls? aka testing steps...thx as Im not sure which are u refering to.

LiborBatek (approved):
I have tested yeti with and without any ```user variable``` and all ayon actions works normally...

User being able to ```Load``` ```Manage``` and lastly ```Connect Yeti``` via action in the inventory. No issue were found also with older products published before this PR...

All good!

Here is my asset with user variable loaded and checked if present on the ```yetiRig``` asset. Works fine.
Screenshot 2024-04-19 124307

BigRoy:
> I have tested yeti with and without any `user variable` and all ayon actions works normally...
> 
> User being able to `Load` `Manage` and lastly `Connect Yeti` via action in the inventory. No issue were found also with older products published before this PR...
> 
> All good!
> 
> Here is my asset with user variable loaded and checked if present on the `yetiRig` asset. Works fine. Screenshot 2024-04-19 124307

Actually @LiborBatek this may have been my bad in the testing notes. But they should load to the Yeti Cache product when loading - not just to the loaded yeti rig. So from that yeti rig in your screenshot, publish a cache - that cache should then also on load have the user variables.

LiborBatek (approved):
Thanks to @BigRoy for noting I have to test also ```yeti cache``` product....

So it works fine and with propagating ```user variables```. 

Screenshot 2024-04-19 145000


Can be merged i guess...

moonyuet (approved):

PR #289, opened 2024-03-27
Maya: Validate Node Ids Related report folder path of the other ids that were found

BigRoy:
## Changelog Description

Validate Node Ids Related report folder names of the other ids that were found

## Additional info

Improves the artist facing report.

## Testing notes:

1. Publish a model or rig with `cbId` from different folders - the report should now list which folders those are from.


LiborBatek (approved):
Been tested in ```maya 2024```

Works as expected and without any issues.
image

BigRoy:
> Been tested in `maya 2024`
> 
> Works as expected and without any issues. image

Thanks - still not too happy with how that report turns out. The fact that you have odd UI scaling doesn't help - but those node names really aren't useful in this way.

I've converted the PR to draft and will look into it some more when I have some time.

BigRoy:
Found another issue with this PR. It tries to query the `ayon_api.get_folders` with the ids found on the nodes. However, when the AYON api finds an invalid identifier it results in an error instead of skipping over the invalid ids:

```
// Traceback (most recent call last):
//   File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2403071252_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
//     runner(*args)
//   File "E:\dev\ayon-core\client\ayon_core\hosts\maya\plugins\publish\validate_node_ids_related.py", line 51, in process
//     invalid = self.get_invalid(instance)
//   File "E:\dev\ayon-core\client\ayon_core\hosts\maya\plugins\publish\validate_node_ids_related.py", line 90, in get_invalid
//     folder_paths = {entity["path"] for entity in folder_entities}
//   File "E:\dev\ayon-core\client\ayon_core\hosts\maya\plugins\publish\validate_node_ids_related.py", line 90, in <setcomp>
//     folder_paths = {entity["path"] for entity in folder_entities}
//   File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\server_api.py", line 3721, in get_folders
//     for parsed_data in query.continuous_query(self):
//   File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\graphql.py", line 380, in continuous_query
//     raise GraphQlQueryFailed(
// ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Invalid entity ID 646247c42d3be13057e546d4 on item 'project/folders' (Line 3 Column 5)
```

That behavior is somewhat different to how things worked in OpenPype.
Anyway, as such - we might need to exclude the invalid uuid identifiers so there isn't a hard error:

```python
def is_valid_uuid(uuid_value):
    try:
        uuid.UUID(uuid_value)
    except ValueError:
        return False
    return True
```
And then adding something like:
```python
            # Remove folder ids that are not valid UUID identifiers, these
            # may be legacy OpenPype ids
            other_folder_ids = {folder_id for folder_id in other_folder_ids
                                if is_valid_uuid(folder_id)}
```

@iLLiCiTiT would you say this is also a bug on the AYON API side? Or is this the preferred workaround?

#### OpenPype asset ids

These are asset ids from `OpenPype` by the way that are invalid for the `folder_ids` query it seems.
An example asset id from OpenPype mongo db that fails is: `646247c42d3be13057e546d4`. For example:
```python
project_name = "test"
folder_ids = {"646247c42d3be13057e546d4"}
for f in ayon_api.get_folders(project_name=project_name, folder_ids=folder_ids):
    print(f)
```

Results in:
```
# Error: GraphQl query Failed: Invalid entity ID 646247c42d3be13057e546d4 on item 'project/folders' (Line 3 Column 5)
# #   File "<maya console>", line 4, in <module>
# #   File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\server_api.py", line 3721, in get_folders
# #     for parsed_data in query.continuous_query(self):
# #   File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\graphql.py", line 380, in continuous_query
# #     raise GraphQlQueryFailed(
# # ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Invalid entity ID 646247c42d3be13057e546d4 on item 'project/folders' (Line 3 Column 5)
```

BigRoy:
I check with @martastain regarding the AYON backend entity ids. He stated:

> when ayon creates them, it uses UUID1, but the format is always UUID. all endpoints should be able to normalize the incoming string, so it does not matter whether the id contains hyphens or not, as long there is the correct number of characters (32 i think)

I also tested:
```python
import uuid
import traceback
import ayon_api

for value in [
    "4f3353b4-d69a-4cd1-ac00-b2c7b91d4acc", # OK
    "A" * 32, # OK
    "a" * 32, # OK
    "4f3353b4-d69a-4cd1-ac00", # invalid
    "-" * 32, # Invalid
    "l" * 32, # Invalid
    "z" * 32, # Invalid
]:
    try:
        uuid.UUID(value)
        print(f"OK:               {value}")
    except ValueError:
        print(f"Invalid:          {value}")
        
    try:
        for f in ayon_api.get_folders("McD_Summer24", folder_ids={value}):
            print(f)
        print(f"OK AYON API:      {value}")
    except ayon_api.exceptions.GraphQlQueryFailed:
        print(f"Invalid AYON API: {value}")
```
Which seems to confirm the API at least matches UUIDs.

As such the `is_valid_uuid` check above should suffice:
```python
def is_valid_uuid(uuid_value):
    try:
        uuid.UUID(uuid_value)
    except ValueError:
        return False
    return True
```

tokejepsen (approved):

PR #294, opened 2024-03-27
Houdini: Implement Load Asset LOP HDA

BigRoy:
## Changelog Description

Implements a LOPs (Solaris / USD) asset loader to reference or payload in files into the Solaris stage.

- Has separate parms for project_name, folder_path, product_name, version and representation that describe which product to load. Changing any of those parms will trigger an update to load that particular version, as such - you can dynamically drive these using any way Houdini allows you to, e.g. the project name and folder default to respectively `$AYON_PROJECT_NAME` and `$AYON_FOLDER_PATH` representing the project and folder of your current context.
- Has parms and functionality to automatically attach a 'version' thumbnail if the version entity being loaded has one. Disabled by default but can be enabled, resized, etc. per node.

## Additional info

image

This is a further enhancement from this OpenPype PR https://github.com/ynput/OpenPype/pull/5831 
Some of that can also be seen in the presentation I gave at the Ynput Summit 2024 [here](https://m.youtube.com/watch?v=TpPZFGQOKOc&t=15557).

We have actually used an OpenPype-equivalent in production. Nonetheless I still expect many bugs (and feature requests) so input is greatly welcomed. I'm inviting you to break it and provide feedback. Also provide screen recordings if you get stuck so we can see how "other users" try to use it. @fabiaserra @MustafaJafar @moonyuet @antirotor you are cordially invited.

I did a pretty quick refactor to AYON, so testing is greatly appreciated! :)

---

Other visual reference of a 'similar' implementation by @fabiaserra [here](https://github.com/ynput/OpenPype/pull/5831#issuecomment-1819889172).

## Testing notes:

1. Load and manage USD assets via the Load Asset LOP (and should also work via the manager)


BigRoy:
@iLLiCiTiT this also uses some logic to store a temporary thumbnail alongside the scene so that Houdini can display it - in the screenshot with the PR for example the checkerboard image above the node bottom right. There may be much better ways, but for now I just wanted to highlight this uses some version thumbnails logic from `ayon_api`.

client/ayon_core/hosts/houdini/api/pipeline.py
```diff
@@ -262,7 +262,29 @@ def parse_container(container):
         dict: The container schema data for this container node.
 
     """
-    data = lib.read(container)
+    # Read only relevant parms
+    # TODO: Clean up this hack replacing `lib.read(container)`
```
BigRoy, line 266: @antirotor This might be up for discussion - but data used from a container usually is only these required fields. By parsing only these instead of **all** spare parameters of a node it may greatly reduce the amount of values returned and thus the speed and memory usage of this call. E.g. on nodes with Redshift parameters all the redshift related attributes are "spare parameters" and thus 100s of parms will be read otherwise.

Let me know what you think - of course we'll need to clean up the todo. ;)

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
```diff
@@ -0,0 +1,52 @@
...
+    label = "Load Asset (LOPs)"
+    representations = ["usd", "abc", "usda", "usdc"]
+    order = -10
+    icon = "code-fork"
+    color = "orange"
+
+    def load(self, context, name=None, namespace=None, data=None):
+
+        # Define node name
+        namespace = namespace if namespace else context["asset"]["name"]
```
MustafaJafar, line 19:
```suggestion
        namespace = namespace if namespace else context["folder"]["name"]
```

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
```diff
@@ -0,0 +1,52 @@
...
+        # Create node
+        network = find_active_network(
+            category=hou.lopNodeTypeCategory(),
+            default="/stage"
+        )
+        node = network.createNode("ayon::lop_import", node_name=node_name)
+        node.moveToGoodPosition()
+
+        # Set representation id
+        representation_id = str(context["representation"]["_id"])
```
MustafaJafar, line 31:
```suggestion
        representation_id = str(context["representation"]["id"])
```

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
```diff
@@ -0,0 +1,52 @@
...
+        parm = node.parm("representation")
+        parm.set(representation_id)
+        parm.pressButton()  # trigger callbacks
+
+        nodes = [node]
+        self[:] = nodes
+
+    def update(self, container, representation):
```
MustafaJafar, line 39: After some inspection, I found out that `update` method expects a `context`
```suggestion
    def update(self, container, context):
```

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
```diff
@@ -0,0 +1,52 @@
...
+        node = container["node"]
+
+        representation_id = str(representation["_id"])
```
MustafaJafar, line 42:
```suggestion
        representation_id = str(context["representation"]["id"])
```

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
```diff
@@ -0,0 +1,52 @@
...
+        parm = node.parm("representation")
+        parm.set(representation_id)
+        parm.pressButton()  # trigger callbacks
+
+    def remove(self, container):
+        node = container["node"]
+        node.destroy()
+
+    def switch(self, container, representation):
+        self.update(container, representation)
```
MustafaJafar, line 52:
```suggestion
    def switch(self, container, context):
        self.update(container, context)
```

MustafaJafar (changes requested):
I had to make these changes on my side to make it work.

MustafaJafar (commented):
I have two suggestions to make this HDA more fun. 

1. A button to select folder path (which is the same as the button in the publisher UI)
This logic works on my side.
```python
"""Select a folder path."""
from ayon_core.tools.publisher.widgets.folders_dialog import FoldersDialog
from ayon_core.tools.utils.host_tools import get_tool_by_name
from ayon_core.hosts.houdini.api.lib import get_main_window


main_window = get_main_window()
publisher_window = get_tool_by_name( tool_name="publisher", parent=main_window)

dialog = FoldersDialog(publisher_window.controller, main_window)
dialog.exec_()

selected_folder_path = dialog.get_selected_folder_path()

print ("asset name:", selected_folder_path)
```

2. A button to select a USD product name.
This snippet works. it gets a list of usd products within a particular folderpath 
```python
import ayon_api

project_name = "Robo"
folder_path = "/Assets/Character/bigrobo"
my_product_type = "usd"


id_only = ["id"]
folder_entity = ayon_api.get_folder_by_path(project_name,
                                            folder_path,
                                            fields=id_only)
my_folder_id = folder_entity["id"]

my_usd_products = ayon_api.get_products(
    project_name,
    folder_ids = [my_folder_id],
    product_types = [my_product_type]
    
)

print(list(map(lambda p: p["name"], my_usd_products)))
```

irrelevant to this PR.

3. I think adding a product type selector in the HDA can be used to enhance the previous code snippet. which can be a starting point to have a generic loader. where the digital asset acts as a container and nodes inside it are created based on the data in each loader. 
I'm not sure if I rephrased it correctly. but, I hope you get me.

BigRoy:
> I have two suggestions to make this HDA more fun.

Yes, please. I'd love there to be buttons that show a mini-list dialog to pick from. Feel free to make a PR to my branch with those features. it'd be greatly appreciated 😍 - I looked into it in OpenPype but the dialogs then weren't to trivial to hijack into but if you have a good hack for AYON, go for it!

Would you be able to contribute?

> I had to make these changes on my side to make it work.

Interesting, seems like I might have missed pushing a few commits - because it did work for me.

MustafaJafar:
> Would you be able to contribute?

I was born ready to contribute 🤓..
Hopefully, tomorrow I'll tinker around it.

MustafaJafar (approved):
This is super cool and it works. 

I didn't test some of its features : thumbnails (I don't have products with thumbnails) and load options (look fine as it's just straight forward copied parameters).

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
```diff
@@ -0,0 +1,52 @@
...
+        network = find_active_network(
+            category=hou.lopNodeTypeCategory(),
+            default="/stage"
+        )
+        node = network.createNode("ayon::lop_import", node_name=node_name)
+        node.moveToGoodPosition()
+
+        # Set representation id
+        parm = node.parm("representation")
+        parm.set(context["representation"]["id"])
```
Lyon-Rosenblatt-Ynput, line 33: It would be cool if, in the future, we have the representations as a dropdown, too. 
However, I believe it is best to change this in the HDA. 

Here is an option for doing it as a spare Parm. I didn't do much research for implementation yet (just as a warning)
``` 
dropdown_parm = hou.MenuParmTemplate("representations", "representations", menu_items=self.representations)
if node.parmTemplateGroup() is None:
    node.setParmTemplateGroup(hou.ParmTemplateGroup())
node.addSpareParmTuple(dropdown_parm)
```

Lyon-Rosenblatt-Ynput (changes requested):
Here are some things I found. 

Sorry for the one commit I made; I will leave it as comments from now on.
 
Also, I have one question in case you tested it. In the on_save() function in pipline.py, there is a call to a function that apparently only exists in the Maya lib.py, and it errors on save. I'm not sure if it does that for you, too. 

Lovely stuff btw.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -998,6 +998,232 @@ def add_self_publish_button(node):
     node.setParmTemplateGroup(template)
```
Lyon-Rosenblatt-Ynput, line 1000:
```suggestion
from uuid import UUID
def is_valid_id(entity_id):
    try:
        UUID(entity_id)
        return True
    except ValueError:
        return False
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+            f.write(data.content)
+        return path
+
+
+def set_representation(node, repre_id):
+    file_parm = node.parm("file")
+    if repre_id:
+        project_name = node.evalParm("project_name") or \
+                       get_current_project_name()
+        try:
```
Lyon-Rosenblatt-Ynput, line 140: I believe this might be an option to take out the try-except. The reason why I think that might be important is that the try-except Exception might shadow other errors. This solution, together with what I placed into the lib.py, does survive my testing and removes the try-except.

you probably want to check the indent. my copy past doesn't carry it over
```suggestion
repre_entity = None
if lib.is_valid_id(repre_id):
        repre_entity = get_representation_by_id(project_name, repre_id)

```

Lyon-Rosenblatt-Ynput (changes requested):

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+            repre_entity = get_representation_by_id(project_name, repre_id)
+        except ayon_api.exceptions.GraphQlQueryFailed:
```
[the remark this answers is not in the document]
BigRoy: Thanks - yes, having this be a clearer check is perfect. The `is_valid_id` code you mentioned I think would also be nice since it could avoid even starting to hit the database, which is great. I'm just not entirely sure myself whether the postgres ids for representations are _required_ to be valid `UUID` entries or whether in the backend they just happen to be strings that default to UUIDs. @martastain can likely confirm?

client/ayon_core/hosts/houdini/api/hda_utils.py
BigRoy: Thanks - I'll continue the conversation here: https://github.com/ynput/ayon-core/pull/294#discussion_r1547373265

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+         context (dict): Context of representation
+
+     """
+    # TODO: Avoid 'duplicate' taking over the expression if originally
+    #       it was $OS and by duplicating, e.g. the `folder` does not exist
+    #       anymore since it is now `hero1` instead of `hero`
+    # TODO: Support hero versions
+    version = str(context["version"]["version"])
+
+    # We only set the values if the value does not match the currently
```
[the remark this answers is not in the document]
BigRoy: Not necessarily 'expressions' but if you set e.g. `$AYON_FOLDER_PATH` for the folder and `$AYON_PROJECT` (of whatever the var name is) for project then if the value on change matches the evaluated value of those variables then these will remain as is and not be replaced by their evaluated values (basically keeping them dynamic if we can)

In essence:
_We only change the value if it does not match the *current* value_

Where 'current' value may be a variable that evaluates to the same value. It uses the current "evaluated" value to compare against, not the raw parm value.

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+
+    product_entity = get_product_by_name(
+        project_name,
+        product_name=product_name,
+        folder_id=folder_entity["id"],
+        fields=id_only)
+    if not product_entity:
+        load_message_parm.set(f"Product not found: '{product_name}'")
+        return
+    version_doc = get_version_by_name(
```
iLLiCiTiT, line 323:
```suggestion
    version_entity = get_version_by_name(
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+        project_name,
+        version,
+        product_id=product_entity["id"],
+        fields=id_only)
+    if not version_doc:
+        load_message_parm.set(f"Version not found: '{version}'")
+        return
+    representation_doc = get_representation_by_name(
```
iLLiCiTiT, line 331:
```suggestion
    representation_entity = get_representation_by_name(
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+        return
+
+    try:
+        version = int(version.strip())
+    except ValueError:
+        load_message_parm.set(f"Invalid version format: '{version}'\n"
+                              "Make sure to set a valid version number.")
+        return
+
+    id_only = ["id"]
```
iLLiCiTiT, line 302: Why to use variable?

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -998,6 +998,232 @@ def add_self_publish_button(node):
...
+
+    Similar to: `nodegraphutils.loadBackgroundImages`
+
+    """
+
+    def _serialize(image):
+        """Return hou.NetworkImage as serialized dict"""
+        if isinstance(image, dict):
+            # Assume already serialized, only do some minor validations
+            assert "path" in image
```
iLLiCiTiT, line 1043: Note: `AssertionError` is for tests, should not be used for real error handling.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -998,6 +998,232 @@ def add_self_publish_button(node):
...
+
+    """
+    # Find network editors that are current tab of given category
+    index = 0
+    while True:
+        pane = hou.ui.paneTabOfType(hou.paneTabType.NetworkEditor, index)
+        index += 1
+        if pane is None:
+            break
+
```
iLLiCiTiT, line 1210:
```suggestion
        if pane is None:
            break

        index += 1
```

client/ayon_core/hosts/houdini/api/pipeline.py
```diff
@@ -262,7 +262,29 @@ def parse_container(container):
...
+    import six
+    import json
+    from .lib import JSON_PREFIX
```
iLLiCiTiT, line 269: Why are imports inside the function?

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -998,6 +998,232 @@ def add_self_publish_button(node):
...
+        data = {
+            "path": image.path(),
+            "rect": [rect_min.x(), rect_min.y(), rect_max.x(), rect_max.y()],
+            "brightness": image.brightness(),
+            "relativetopath": image.relativeToPath()
+        }
+        if data["brightness"] == 1.0:
+            del data["brightness"]
+        if data["relativetopath"] == "":
+            del data["relativetopath"]
```
iLLiCiTiT, line 1060:
```suggestion
        }
        if image.brightness() != 1.0:
            data["brightness"] = image.brightness()
        if image.relativeToPath():
            data["relativetopath"] = image.relativeToPath()
```

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -998,6 +998,232 @@ def add_self_publish_button(node):
...
+    if not data:
+        return []
+
+    try:
+        images = json.loads(data)
+        if not raw:
+            images = [_parse(_data) for _data in images]
+    except json.decoder.JSONDecodeError:
+        images = []
+
```
iLLiCiTiT, line 1028:
```suggestion
    try:
        images = json.loads(data)
    except json.decoder.JSONDecodeError:
        images = []

    if not raw:
        images = [_parse(_data) for _data in images]
```

BigRoy: Thanks!

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+    representation_doc = get_representation_by_name(
+        project_name,
+        representation_name,
+        version_id=version_doc["id"],
+        fields=id_only)
+    if not representation_doc:
+        load_message_parm.set(
+            f"Representation not found: '{representation_name}'.")
+        return
+    return representation_doc["id"]
```
BigRoy, line 340:
```suggestion
    if not representation_entity:
        load_message_parm.set(
            f"Representation not found: '{representation_name}'.")
        return
    return representation_entity["id"]
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+        fields=id_only)
+    if not product_entity:
+        load_message_parm.set(f"Product not found: '{product_name}'")
+        return
+    version_doc = get_version_by_name(
+        project_name,
+        version,
+        product_id=product_entity["id"],
+        fields=id_only)
+    if not version_doc:
```
BigRoy, line 328:
```suggestion
    if not version_entity:
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,402 @@
...
+        load_message_parm.set(f"Version not found: '{version}'")
+        return
+    representation_doc = get_representation_by_name(
+        project_name,
+        representation_name,
+        version_id=version_doc["id"],
```
BigRoy, line 334:
```suggestion
        version_id=version_entity["id"],
```

client/ayon_core/hosts/houdini/api/hda_utils.py
BigRoy: Just because it was re-used for each of the multiple queries. Fine with changing it around if you prefer that.

client/ayon_core/hosts/houdini/api/pipeline.py
BigRoy: Just me trying to be quick when I implemented this 'fix' at the time.

client/ayon_core/hosts/houdini/api/hda_utils.py
BigRoy: Implemented with https://github.com/ynput/ayon-core/pull/294/commits/bef67347ecbc990813a3b933c7c5342815bfabb7

client/ayon_core/hosts/houdini/api/lib.py
BigRoy: Implemented with https://github.com/ynput/ayon-core/pull/294/commits/bef67347ecbc990813a3b933c7c5342815bfabb7 - thanks again!

client/ayon_core/hosts/houdini/api/hda_utils.py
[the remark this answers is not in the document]
BigRoy: Implemented with https://github.com/ynput/ayon-core/pull/294/commits/bef67347ecbc990813a3b933c7c5342815bfabb7

client/ayon_core/hosts/houdini/api/hda_utils.py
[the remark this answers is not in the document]
BigRoy: Tweaked it slightly with https://github.com/ynput/ayon-core/pull/294/commits/49a8ef420c80e78bf0c813ce0f0773e3041a6098

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
BigRoy: @MustafaJafar just wondering if this still on your radar to look into potentially?

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+def _unlocked_parm(parm):
+    """Unlock parm during context; will always lock after"""
+    try:
+        parm.lock(False)
+        yield
+    finally:
+        parm.lock(True)
+
+
+def get_available_versions(node):
```
iLLiCiTiT, line 47: Where is this used? If it is used on multiple nodes at the same time I would recommend to do it in bulk as this may be slow when done one by one.

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+
+    if not all([
+        project_name, folder_path, product_name
+    ]):
+        return []
+
+    id_only = ["id"]
+    folder_entity = get_folder_by_path(project_name,
+                                       folder_path,
+                                       fields=id_only)
```
iLLiCiTiT, line 69: Please keep formatting consistency. I don't think you need `id_only` variable, especialy when contains only `"id"`.
```suggestion
    folder_entity = get_folder_by_path(
        project_name,
        folder_path,
        fields={"id"})
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+        folder_id=folder_entity["id"],
+        fields=id_only)
+    if not product_entity:
+        return []
+
+    # TODO: Support hero versions
+    versions = get_versions(project_name=project_name,
+                            product_ids=[product_entity["id"]],
+                            fields=["version"],
+                            hero=False)
```
iLLiCiTiT, line 84:
```suggestion
    versions = get_versions(
        project_name,
        product_ids={product_entity["id"]},
        fields={"version"},
        hero=False)
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+    folder_entity = get_folder_by_path(project_name,
+                                       folder_path,
+                                       fields=id_only)
+    if not folder_entity:
+        return []
+    product_entity = get_product_by_name(
+        project_name,
+        product_name=product_name,
+        folder_id=folder_entity["id"],
+        fields=id_only)
```
iLLiCiTiT, line 76:
```suggestion
        fields={"id"})
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+    project_name = node.evalParm("project_name") or get_current_project_name()
+    folder_path = node.evalParm("folder_path")
+    product_name = node.evalParm("product_name")
+
+    if not all([
+        project_name, folder_path, product_name
+    ]):
+        return []
+
+    id_only = ["id"]
```
iLLiCiTiT, line 66:
```suggestion
```

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+                version_id = repre_entity["versionId"]
+                thumbnail_dir = node.evalParm("thumbnail_cache_dir")
+                thumbnail_path = _get_thumbnail(project_name, version_id,
+                                                thumbnail_dir)
+                set_node_thumbnail(node, thumbnail_path)
+    else:
+        # Clear filepath and thumbnail
+        with _unlocked_parm(file_parm):
+            file_parm.set("")
+        set_node_thumbnail(node, None)
```
iLLiCiTiT, line 181: Small refactor of conditions
```suggestion
    if not repre_id:
        # Clear filepath and thumbnail
        with _unlocked_parm(file_parm):
            file_parm.set("")
        set_node_thumbnail(node, None)
        return

    project_name = (
        node.evalParm("project_name")
        or get_current_project_name()
    )

    # Ignore invalid representation ids silently
    if not is_valid_uuid(repre_id):
        return

    repre_entity = get_representation_by_id(project_name, repre_id)
    if not repre_entity:
        return

    context = get_representation_context(project_name, repre_entity)
    update_info(node, context)
    path = get_representation_path_from_context(context)
    # Load fails on UNC paths with backslashes and also
    # fails to resolve @sourcename var with backslashed
    # paths correctly. So we force forward slashes
    path = path.replace("\\", "/")
    with _unlocked_parm(file_parm):
        file_parm.set(path)

    if node.evalParm("show_thumbnail"):
        # Update thumbnail
        # TODO: Cache thumbnail path as well
        version_id = repre_entity["versionId"]
        thumbnail_dir = node.evalParm("thumbnail_cache_dir")
        thumbnail_path = _get_thumbnail(
            project_name, version_id, thumbnail_dir
        )
        set_node_thumbnail(node, thumbnail_path)
```

client/ayon_core/hosts/houdini/api/hda_utils.py
iLLiCiTiT: Yes please.

client/ayon_core/hosts/houdini/api/pipeline.py
```diff
@@ -262,7 +266,26 @@ def parse_container(container):
...
+    for name in ["name", "namespace", "loader", "representation", "id"]:
+        parm = container.parm(name)
+        if not parm:
+            return {}
+
+        value = parm.eval()
+
+        # test if value is json encoded dict
+        if isinstance(value, six.string_types) and \
+                value.startswith(JSON_PREFIX):
```
iLLiCiTiT, line 282:
```suggestion
        if isinstance(value, str) and value.startswith(JSON_PREFIX):
```

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -1077,6 +1077,233 @@ def sceneview_snapshot(
...
+                       for image in images)
+            data = json.dumps([_serialize(image) for image in images])
+            node.setUserData("backgroundimages", data)
+        else:
+            node.destroyUserData("backgroundimages", must_exist=False)
+
+
+def set_node_thumbnail(node,
+                       image_path,
+                       rect=None):
```
iLLiCiTiT, line 1155:
```suggestion
def set_node_thumbnail(node, image_path, rect=None):
```

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
MustafaJafar: Hey, Some updates. 
I added few dropdown menus in this PR to make it easier for users to select the product to load.
https://github.com/BigRoy/ayon-core/pull/3

> the representations as a dropdown, too.

we can do the same for ~~`representations`~~ `product type` dropdown. (I think it should be renamed as mentioned in https://github.com/BigRoy/ayon-core/pull/3#issuecomment-2077866687 )

This code should update the menu entries. 
I don't know if there's a prettier version of it.

```python
import ayon_api


def get_available_products_types():
    host = registered_host()
    current_context = host.get_current_context()
    project_name = current_context["project_name"]
    folder_path = current_context["folder_path"]        
    
    id_only = ["id"]
    folder_entity = ayon_api.get_folder_by_path(project_name,
                                                folder_path,
                                                fields=id_only)
    my_folder_id = folder_entity["id"]
    
    my_products = ayon_api.get_products(
        project_name,
        folder_ids = [my_folder_id],
        
    )
    
    available_types = [product["productType"] for product in my_products if product["productType"] != "workfile"]
    
    available_types = list(dict.fromkeys(available_types))
    
    result = []

    for i in available_types:
        result.append(i)
        result.append(i)

    return(result)
```

client/ayon_core/hosts/houdini/api/hda_utils.py
MustafaJafar: It's run once when a user clicks this button in the Houdini digital asset.
image

BigRoy: It's as @MustafaJafar states - it's run once, one request by the node itself on a user clicking a button. Unfortunately it's not something I can batch across multiple nodes.

client/ayon_core/hosts/houdini/api/hda_utils.py
BigRoy: Love it, thanks!

client/ayon_core/hosts/houdini/api/hda_utils.py
```diff
@@ -0,0 +1,414 @@
...
+    """
+    Usually used as callback to the project, folder, product, version and
+    representation parms which on change - would result in a different
+    representation id to be resolved.
+
+    Args:
+        node (hou.Node): Node to update.
+    """
+    project_name = node.evalParm("project_name") or get_current_project_name()
+    representation_id = get_representation_id(
```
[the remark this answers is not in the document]
BigRoy: No - and this is on purpose actually. We want a product name change on the node to dynamically reroute the representation id - so that the artists can set up Houdini expression/variables or alike to potentially even dynamically remap say the folder path to a different context in a multi-shot workflow in Houdini.

They could say have Houdini context option set to "folderPath = 010" and switch it to 020 in a single file to have the full node tree use the context variable and get re-evaluated to the new asset or alike. The **biggest** issue yet though, is that this would only really be 'logical' with unversioned (hero versions) of products and I'm not sure yet how to really deal with multi-shot workflows in a single graph in Houdini whilst still easily maintaining version decisions per shot, etc.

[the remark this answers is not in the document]
iLLiCiTiT: Ok new question, can this be done in bulk instead of node by node?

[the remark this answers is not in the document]
BigRoy: It's a callback that triggers if a parm (=attribute) changes on one particular node in Houdini. That happens per node per parm so there isn't much we can do to batch these unfortunately.

Since it also only ever triggers if you tweak a specific node then performance is totally fine. However, if you were to somehow select 100s of nodes and tweak the parms all at once, then yes - you'd likely get hit by the bottleneck that it's actually doing one by one.

I guess technically we could wait in a thread until all parms are set, then go over the scene and collect any 'dirty' nodes and batch 'em in one go. But there's a big complexity overhead there - it'll be much slower for the most used case (changing a single parm on a single node) since we'd then always need to go over the full scene to find all nodes (unless we store it in a houdini session variable on change and check just that).

I understand the question - unfortunately I don't see a trivial way to resolve this particular case. Also, I don't think it'll become a production issue on short notice.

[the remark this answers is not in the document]
fabiaserra: in my prior studio we solved this by having a singleton cascading cache system that would only invalidate the downstream fields from the field that changed so upstream entities wouldn't need to be re-queried. Having a cache system is pretty crucial when dealing with Houdini callbacks as they can get triggered quite a lot and you want to be smart on lowering the amount of queries needed. Even with such system we had a few occasions where we would sometimes bring the server to a haul because of the amount of queries hitting it (of course AYON isn't being used at that scale of users but still worth considering) so we had to do some other things to break expressions on parms when submitting scenes to the farm to avoid extra queries doing rendering/caching.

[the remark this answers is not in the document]
BigRoy: That could be doable - sure. Could be a separate PR however to keep this streamlined? I do like that - even though it does come with its own potential 'side effects' of course.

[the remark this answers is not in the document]
fabiaserra: yeah up to you, just try to be as minimal as possible on repeating the `get_*` calls. I see a few of them in @MustafaJafar code snippet he shared so just try be smart on not repeating those unnecessarily

client/ayon_core/hosts/houdini/api/pipeline.py
antirotor: uh, I've missed that one, but sure - it make sense to read just what we need

BigRoy:
- Refactored for moved client code + bumped addon version.
- Optimizations @fabiaserra recommended would be left for a follow-up PR: https://github.com/ynput/ayon-core/pull/294#discussion_r1609830499
- Implementation of nicer "pick a representation" from the node itself could be up to a follow-up PR - nice prototype by @MustafaJafar here: https://github.com/BigRoy/ayon-core/pull/3
    - If we prefer to include that we could spend some time on that PR first before merging this one.

client/ayon_core/hosts/houdini/plugins/load/load_asset_lop.py
BigRoy: Should I mark this resolved in favor of moving it to a follow-up PR?

Lyon-Rosenblatt-Ynput:
Hi Roy excellent tool. 

I want to mention that there might be a bug regarding thumbnail creation. 
I also want to mention that I'm very confused about where this path gets from, '/home/workh', because my $Home path is /home/lyonh. I have a system with /home/workh, but that's not the system I'm using your code on. 
```
  File "ayon::Lop/lop_import::1.0/show_thumbnail", line 1, in <module>
  File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 224, in on_thumbnail_show_changed
    on_representation_id_changed(node)
  File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 245, in on_representation_id_changed
    set_representation(node, repre_id)
  File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 185, in set_representation
    thumbnail_path = _get_thumbnail(
  File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 141, in _get_thumbnail
    os.makedirs(thumbnail_dir_expanded, exist_ok=True)
  File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 215, in makedirs
    makedirs(head, exist_ok=exist_ok)
  [Previous line repeated 4 more times]
  File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 225, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/home/workh'
```


Also, while I'm here. Some brain food and thoughts while testing. 
1. maybe we want to prepend all AYON tools with a prefix so that a User can just hit Tab+AY to see all the Ayon tools in case they forgot the name or want to see what's available. PS: I know this could be done via a submenu entry, but we might want to add more than one submenu entry at some point. 
2. Implementing all .hda files in the code and discarding the binary files might be interesting. This would allow us to see changes in the .hda directly in the PR. 
3. I'm a bit surprised or confused about your repo layout because your [hosts](https://github.com/BigRoy/ayon-core/tree/enhancement/houdini_load_asset_lop/client/ayon_core/hosts) directory is "empty," and it looks like you implement all the Houdini code as a server plugin. Is there a reason for that? It's blocking "hot" reloading off the loader, so I was just wondering why you went this path.

BigRoy:
> I want to mention that there might be a bug regarding thumbnail creation.
I also want to mention that I'm very confused about where this path gets from, '/home/workh', because my $Home path is /home/lyonh. I have a system with /home/workh, but that's not the system I'm using your code on.

Thanks so much - I was expecting logic in that area to break somewhere but hadn't heard anyone hitting issues with it so I have secretly not looked at that logic in a long time. I'll investigate what's going on here because this definitely sounds wrong. So, thanks for reporting!

> maybe we want to prepend all AYON tools with a prefix so that a User can just hit Tab+AY to see all the Ayon tools in case they forgot the name or want to see what's available. PS: I know this could be done via a submenu entry, but we might want to add more than one submenu entry at some point.

Yes, I think that'd be great - so just adding `AYON` to the start of the label?

> 2\. Implementing all .hda files in the code and discarding the binary files might be interesting. This would allow us to see changes in the .hda directly in the PR.

Yes please - do you know how I can implement the HDA more than this in the code? As far as I know we're still required to in the end generate an HDA? Unless we make it a `subnet` with spare parameters by code but that's much harder to maintain over updates down the line (to propagate the updated version of that tool into the houdini scene since we're basically replicating HDA logic).

I'd actually love for SideFX Houdini to be able to create `.hda` in plain text for better version control.

Do you have any ideas?

> I'm a bit surprised or confused about your repo layout because your [hosts](https://github.com/BigRoy/ayon-core/tree/enhancement/houdini_load_asset_lop/client/ayon_core/hosts) directory is "empty," and it looks like you implement all the Houdini code as a server plugin. Is there a reason for that? It's blocking "hot" reloading off the loader, so I was just wondering why you went this path.

Actually - this is the _new_ way of doing this. It's not actually the server addon but it's the **moved out** client code of houdini into `ayon_houdini` out of `ayon_core`. See PR #516 - the same is happening for other integrations like Maya #546 , Nuke #556 , etc.

Lyon-Rosenblatt-Ynput:
> Yes, I think that'd be great - so just adding `AYON` to the start of the label?

Yeah, I'm not sure if it should be AYON or like AY or something. Maybe @dee-ynput should have an opinion about that. 
My main reason for thinking about AY is that it's shorter. like mtlx vs materialX 
But on the other hand, AYON is like 4 letters, so yeah, I'm not sure what's optimal there. 


> Yes please - do you know how I can implement the HDA more than this in the code? As far as I know we're still required to in the end generate an HDA? Unless we make it a `subnet` with spare parameters by code but that's much harder to maintain over updates down the line (to propagate the updated version of that tool into the houdini scene since we're basically replicating HDA logic).

yeah that's gonna be an interesting question. my idea would have been to have python code that creates a subnet and then adds the interface items to it. maybe have it as a class with some easy to use node to node link options. but your not wrong this will be complicated in the long run. 
It might be worth brainstorming about that in a Hou-WG call with a few more people. Maybe some off them have already dellt with it. 
the reason I'm bringing it up is that I already know when we keep this system for a long time, we will end up with logic in places where it doesn't belong because nobody saw it, and now it's there. Aka: I personally feel like the .hda system is not really that nicely scalable for a pipeline company. 


> For what it's worth, the `hda_utils` module in this PR is intended to be "all" of the code related to the HDAs, and then in the HDA itself, there's just import of that and running the functions in there. Just wasn't sure how to push even more into the code.

i mean that's a really nice approach tbh. and if we don't find a good way to "replace" .hda files this might very well be our best bet. 


> I'd actually love for SideFX Houdini to be able to create `.hda` in plain text for better version control.
> 
> Do you have any ideas?

I have kinda 2 thoughts. 
1. maybe we could have a script setup that we generally use where we say, when you want a .hda, it needs to be defined as code, and this script then builds the .hda from it. A bit like IAC or the way our creat_package.py works. This is probably not a nice, rounded approach, but it may be easy to implement. I am not sure, tho. 
2. I know that one of the people at the Hou-WG meeting mentioned that they implemented shelf buttons as tab menu options by having a shelf button that was not assigned to a shelf. Maybe we can combine that with the usage of a general AYON-Creat-Node.py setup where we define a common structure for node creation that is a bit more opinionated and simpler than the default Houdini way. I think this setup might allow us to describe node setups as code while also not having huge scaling issues. If we allow node creation as classes, we might even be able to describe node graphs as a "tree" "-like structure using dependencies. 

but those are all just thoughts, so let's treat them with a bucket of salt. 



> Actually - this is the _new_ way of doing this. It's not actually the server addon but it's the **moved out** client code of houdini into `ayon_houdini` out of `ayon_core`. See PR #516 - the same is happening for other integrations like Maya #546 , Nuke #556 , etc.
> 
> Also it should **not** block hot reloading of the loader - you just need to make sure you also set up the `ayon_houdini` plugin to that client folder in your dev bundle!

oh man, thanks for the info. I totally missed that your repo already uses/contributes to the plugin Split.

Lyon-Rosenblatt-Ynput:
> I think being able to nicely version control HDAs is a topic we could also raise with SideFX in a call with them @dee-ynput @antirotor - I suppose it's also something they must be dealing with internally in their releases or have heard about from the community, so maybe there are already solutions they can offer for this.
 
That's true. I didn't even think about it, to be honest. But yeah, they are handling many HDAs, so they better have something in place. 

> I love both of these ideas. I think those may actually warrant creating a separate issue for just that topic since it'd be a big thing to work on and probably not reasonable for this particular PR. But we should definitely try to align on some ideas there.

Absolutely, let's create a GH issue and brainstorm a bit about what we would like to see from such a system.  Issue Link https://github.com/ynput/ayon-core/issues/591

fabiaserra:
> I think being able to nicely version control HDAs is a topic we could also raise with SideFX in a call with them @dee-ynput @antirotor - I suppose it's also something they must be dealing with internally in their releases or have heard about from the community, so maybe there are already solutions they can offer for this.

That's already a solved problem and it's called expanding HDAs and you can do so with the `hotl` binary that ships with Houdini: https://www.sidefx.com/docs/houdini/ref/utils/hotl.html

We expand all of our internal HDAs that we store in version control by running `hotl -t directory hdafile`, something like this:
```
hotl -t driver_ax.ax_publisher.1.0 driver_ax.ax_publisher.1.0.hda
rm driver_ax.ax_publisher.1.0.hda
mv driver_ax.ax_publisher.1.0 driver_ax.ax_publisher.1.0.hda
```
Of course this can be easily automated through the build system of your repo (we did automate it in my prior company) but I haven't created that many HDAs yet to need to spend the time for doing that.

Moreover, in order to be able to reuse as much of the backend code of the nodes, similarly to you in the HDA python scripts module I simply import this class wrapper factory of the node so I can then just call the functions defined in my repository.

In my prior studio we had a more complex system that using monkey patching it would override the base hou.Node python class so our code looked just like an extension to the existing python code of Houdini, that way our functions looked like native in Houdini (i.e. `hou.node("/out/geometry").my_custom_function()` or ``hou.node("/out/geometry").render()`) so we could extend functionality of native houdini nodes and wrap the backend of any node by simply calling the class the same name as the node.

fabiaserra:
> Yeah, I'm not sure if it should be AYON or like AY or something. Maybe @dee-ynput should have an opinion about that.

Similar to this request, all of our internal nodes we prepend with our company code `AX` (Alkemy-X) so artists can easily find them. For AYON I would suggest `AY`...

BigRoy:
> That's already a solved problem and it's called expanding HDAs and you can do so with the `hotl` binary that ships with Houdini: https://www.sidefx.com/docs/houdini/ref/utils/hotl.html

Thanks @fabiaserra - perfect. I'd have imagined there was a solution to this.

Does this mean we would ship the extended state in the repository, but somehow turn those into `.hda` just before houdini launch? Or when do we convert back? Do we even convert back? Or is Houdini capable of loading the expanded `hda` directly and did you end up doing so? I would assume there's a performance bottleneck there.

dee-ynput:
> > Yeah, I'm not sure if it should be AYON or like AY or something. Maybe @dee-ynput should have an opinion about that.
> 
> Similar to this request, all of our internal nodes we prepend with our company code `AX` (Alkemy-X) so artists can easily find them. For AYON I would suggest `AY`...

I would usually favor using the company name than the product as namespace, but it indeed wouldn't fit here.

That being said, 2 letters feel a little weak. 3 would be better at reaching our goal (fitltering to only AYON nodes)
We don't have an official trigram and "AYON" is one letter away... Why not go for AYON then? 😃

BigRoy:
> I want to mention that there might be a bug regarding thumbnail creation. I also want to mention that I'm very confused about where this path gets from, '/home/workh', because my $Home path is /home/lyonh. I have a system with /home/workh, but that's not the system I'm using your code on.
> 
> ```
>   File "ayon::Lop/lop_import::1.0/show_thumbnail", line 1, in <module>
>   File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 224, in on_thumbnail_show_changed
>     on_representation_id_changed(node)
>   File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 245, in on_representation_id_changed
>     set_representation(node, repre_id)
>   File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 185, in set_representation
>     thumbnail_path = _get_thumbnail(
>   File "/home/lyonh/.local/share/AYON/addons/houdini_0.3.1/ayon_houdini/api/hda_utils.py", line 141, in _get_thumbnail
>     os.makedirs(thumbnail_dir_expanded, exist_ok=True)
>   File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 215, in makedirs
>     makedirs(head, exist_ok=exist_ok)
>   File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 215, in makedirs
>     makedirs(head, exist_ok=exist_ok)
>   File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 215, in makedirs
>     makedirs(head, exist_ok=exist_ok)
>   [Previous line repeated 4 more times]
>   File "/opt/hfs20.0.710/python/lib/python3.10/os.py", line 225, in makedirs
>     mkdir(name, mode)
> PermissionError: [Errno 13] Permission denied: '/home/workh'
> ```

So the path it uses, [defined here](https://github.com/ynput/ayon-core/pull/294/files#diff-1b3780962f1520531824aa772f84578d3c3bfd5e04453c760f87b3c03d7907a7R184) should be based on a `thumbnail_cache_dir` parm on the node itself. It's a hidden parm by default.
The default value on the HDA should be `$JOB/.houdini_loader_thumbnails`.
Using that information, can you figure out how it to that path maybe @Lyon-Rosenblatt-Ynput ?

server_addon/houdini/client/ayon_houdini/startup/otls/ayon_lop_import.hda/ayon_8_8Lop_1lop__import_8_81.0/Tools.shelf
```diff
@@ -0,0 +1,19 @@
...
+Note, that two definitions of the same element are not allowed in
+a single file. -->
+<tool name="$HDA_DEFAULT_TOOL" label="$HDA_LABEL" icon="$HDA_ICON">
+    <toolMenuContext name="viewer">
+    <contextNetType>LOP</contextNetType>
+    </toolMenuContext>
+    <toolMenuContext name="network">
+    <contextOpType>$HDA_TABLE_AND_NAME</contextOpType>
+    </toolMenuContext>
+    <toolSubmenu>CB Pipeline</toolSubmenu>
```
MustafaJafar, line 14: I expected the HDA to exist under `AYON` in tab menu
```suggestion
    <toolSubmenu>AYON</toolSubmenu>
```

MustafaJafar:
The new dialog looks cool! 
image

MustafaJafar:
The new dialog looks cool! 
image

server_addon/houdini/client/ayon_houdini/api/hda_utils.py
```diff
@@ -0,0 +1,498 @@
...
+            self.folder_widget = folder_widget
+
+        def get_selected_folder_path(self):
+            return self.folder_widget.get_selected_folder_path()
+
+        def on_confirm(self):
+            self.close()
+
+    # Note: The following dialog doesn't support changing `the project_name`
+    #         But, having a semi-functional dialog is better than nothing.
```
MustafaJafar, line 478: Should this TODO removed ?  I can change the project.
```suggestion
```

image

BigRoy: Yes.

MustafaJafar:
is this an expected behavior ? clicking **X** returns the selection not canceling it.

Animation_68

BigRoy:
> is this an expected behavior ? clicking **X** returns the selection not canceling it.
> 
> Animation_68 [ Animation_68 ](https://private-user-images.githubusercontent.com/20871534/338363919-4be73749-6933-4917-b7fb-9c35c973c07b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTgwNTc0NjUsIm5iZiI6MTcxODA1NzE2NSwicGF0aCI6Ii8yMDg3MTUzNC8zMzgzNjM5MTktNGJlNzM3NDktNjkzMy00OTE3LWI3ZmItOWMzNWM5NzNjMDdiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA2MTAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNjEwVDIyMDYwNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTJmOWVhMTQ5MTg3ZWFmZjdkMGEwOTdjMmJjZjRlNjc3NjI0ZTFjODNlYmMwOTE2NmY1ZGE4NTQ2Mjk0ZTJhYzcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.shZpjubHiIAiV3B6fQjGvhyKTsJsLJjLwtDcFaISfJM) [ ](https://private-user-images.githubusercontent.com/20871534/338363919-4be73749-6933-4917-b7fb-9c35c973c07b.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTgwNTc0NjUsIm5iZiI6MTcxODA1NzE2NSwicGF0aCI6Ii8yMDg3MTUzNC8zMzgzNjM5MTktNGJlNzM3NDktNjkzMy00OTE3LWI3ZmItOWMzNWM5NzNjMDdiLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA2MTAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNjEwVDIyMDYwNVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTJmOWVhMTQ5MTg3ZWFmZjdkMGEwOTdjMmJjZjRlNjc3NjI0ZTFjODNlYmMwOTE2NmY1ZGE4NTQ2Mjk0ZTJhYzcmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.shZpjubHiIAiV3B6fQjGvhyKTsJsLJjLwtDcFaISfJM)

Thank you, fixed with https://github.com/ynput/ayon-core/pull/294/commits/d332dd204c07585711a6c1e946589ea153ea60e1

BigRoy:
Getting this PR tested more would be greatly appreciated! :)

MustafaJafar:
Hey, @BigRoy 
I already sent this to you but let me add it here as well.

I was looking at `PickDialog` and it encouraged me to a little about `qtpy`. 
This thing functions. but, its layout needs to be fixed. 
we can use the same picker on `Project` and `Folder Path` then the dialog will update both.

```python
from qtpy import QtWidgets
import ayon_api

from ayon_core.tools.utils import SimpleFoldersWidget
from ayon_core.style import load_stylesheet
from ayon_houdini.api import lib


class PickDialog(QtWidgets.QDialog):

    def __init__(self, parent=None):
        super(PickDialog, self).__init__(parent)
        self.setWindowTitle("Select project and folder path..")
        self.setStyleSheet(load_stylesheet())

        self.project_name = get_current_project_name()

        # First Row, First Column
        project_list = QtWidgets.QListWidget()
        project_list.itemPressed.connect(self.on_list_itemPressed)

        # Build list.
        for p in self.get_projects():
            project_list.addItem(p)
        
        # First Row, Second Column
        filter_widget = QtWidgets.QLineEdit()
        filter_widget.setPlaceholderText("Filter folders..")
        folder_widget = SimpleFoldersWidget(parent=self)

        filter_widget.textChanged.connect(folder_widget.set_name_filter)
        folder_widget.double_clicked.connect(self.on_confirm)
        self.folder_widget = folder_widget

        # Second Row
        accept_button = QtWidgets.QPushButton("Accept")
        accept_button.clicked.connect(self.on_confirm)

        # Set Layout
        # This doesn't work yet
        layout_1 = QtWidgets.QVBoxLayout(self)
        layout_1.addWidget(project_list)

        layout_2 = QtWidgets.QVBoxLayout(self)
        layout_2.addWidget(filter_widget)
        layout_2.addWidget(folder_widget)
    
        main_layout = QtWidgets.QGridLayout()
        main_layout.addLayout(layout_1, 0, 0)
        main_layout.addLayout(layout_2, 0, 1)
        main_layout.addWidget(accept_button, 1, 0,1,2)
        self.setLayout(main_layout)

    def get_selected_folder_path(self):
        return self.folder_widget.get_selected_folder_path()
              
    def get_projects(self):
        projects = ayon_api.get_projects(fields=["name"])
        return [p["name"] for p in projects]
    
    def on_list_itemPressed(self, item):
        self.project_name = item.text()
        self.folder_widget.set_project_name(self.project_name)

    def on_confirm(self):
        self.close()

main_window = lib.get_main_window()
dialog = PickDialog(parent=main_window)
dialog.exec_()
print(dialog.project_name, dialog.get_selected_folder_path())
```

BigRoy:
> Hey, @BigRoy I already sent this to you but let me add it here as well.
> 
> I was looking at `PickDialog` and it encouraged me to a little about `qtpy`. This thing functions. but, its layout needs to be fixed. we can use the same picker on `Project` and `Folder Path` then the dialog will update both.

Thanks, cleaned it up, fixed it, tested it - and pushed it 🚀 
So, test-run please and we can merge it?

client/ayon_core/hosts/houdini/api/pipeline.py
BigRoy: This has worked out fine in our pipeline as we've been using this in production over the past months. I'll resolve this conversation; the TODO is there to see if we should clean up elsewhere over time.

server_addon/houdini/client/ayon_houdini/startup/otls/ayon_lop_import.hda/ayon_8_8Lop_1lop__import_8_81.0/DialogScript
```diff
@@ -0,0 +1,326 @@
...
+            label   "Heading"
+            type    label
+            default { "Choose Product" }
+            parmtag { "sidefx::look" "heading" }
+        }
+        parm {
+            name    "project_name"
+            label   "Project"
+            type    string
+            default { "$AYON_PROJECT_NAME" }
```
MustafaJafar, line 34: This adds the same select button next to proejcts
```suggestion
            default { "$AYON_PROJECT_NAME" }
            parmtag { "script_action" "from ayon_houdini.api.hda_utils import select_folder_path;select_folder_path(kwargs['node'])" }
            parmtag { "script_action_icon" "BUTTONS_reselect" }
```

I was concerned what if user clicked the selection button next project and and the one next to folder path. 
and it was funny that Houdini prevented that from happening because selection button forces the user to make a selection or close the window.


Animation_79

---

Also, I loved how Houdini propagated the changes live, once I modified the file it changed the definition while I was using the HDA

moonyuet (approved):
Tested successfully in Houdini 19.5

server_addon/houdini/client/ayon_houdini/api/pipeline.py
```diff
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 """Pipeline tools for OpenPype Houdini integration."""
 import os
+import six
```
MustafaJafar, line 4: ruff is complaining about unused import
```suggestion
```

MustafaJafar:
It is so cool!
Animation_81

BigRoy:
@moonyuet @MustafaJafar I re-requested a review because I did some massive refactoring (function renaming + moving code around) based on some comments @iLLiCiTiT had here and some in DMs on Discord. I did a simple test run on my end and it's working now - but I think it's good to do some other tests because it was a big move.

BigRoy:
Sorry @moonyuet @MustafaJafar I accidentally re-requested the review on the wrong PR.

MustafaJafar (approved):
It's working as cool as before

moonyuet:
> @moonyuet @BigRoy any last words before merge ?

no, it can be merged.

PR #197, opened 2024-03-19
Chore: Use folder in cli arguments

iLLiCiTiT:
## Changelog Description
Use `--folder` instead of `--asset` argument in command line interface.

## Additional info
Use folder over asset argument in command line interface. ~~Modified royal render and deadline to support new argument. Deadline plugin does support both options, tries to use `--asset` if first attempt fails.~~ This PR affects only command `selectcontext` used in royal render.

PR #297, opened 2024-03-28
Chore: Applications addon

iLLiCiTiT:
## Changelog Description
Move applications logic from lib to addon. And modified related code accordingly.

## Additional info
Applications addon does contain everything what `applications.py` in lib did. Moved `extractenvironments` command to the addon as it is his responsibility to implement it. Moved collect plugin which cares about collection of app name/label, or host name based on application name. Modified deadline and royalrender to support both options of arguments. Marked global command `extractenvironments` as deprecated.

Core changes: Created folder `addons` inside core addon which can be used for converted addons that are not yet in different repository. The first subfolder is name of an addon which is added to `sys.path`. So any content in the folder will be added to sys path and can be imported as with addons client code. This approach was chosen as future prove approach, to be able to skip addons which already are separated into repository, but older version of ayon core is used. This approach does not add the addons to PYTHONPATH so it may have issues with host addons as they might not be able to just import their code directly.

Opened questions:
- Should global prelaunch hooks be in applications addon or in core addon?
- Should we split prelaunch hooks that are specific for limited amount of hosts into the hosts?

## Testing notes:
There is no visible change, all should work as before.
1. For farm publishing update plugins accordingly.
2. Launch application of your choice.
3. Run publishing.
4. Run publishing through farm.
5. All should work.

PR partially collides with changes in https://github.com/ynput/ayon-core/pull/197 . The PR will be modified accordingly after this PR merge.


BigRoy:
Do we need to already set reviewers for this PR, or still WIP?

iLLiCiTiT:
> Do we need to already set reviewers for this PR, or still WIP?

I forgot to set them.

kalisp (approved):
Tested in Maya, Fusion, on farm

client/ayon_core/cli_commands.py
```diff
@@ -149,36 +143,35 @@ def publish(path: str, targets: list=None, gui:bool=False) -> None:
...
+        from ayon_core.addon import AddonsManager
 
-        if all((project, asset, task, app)):
-            env = get_app_environments_for_context(
-                project,
-                asset,
-                task,
-                app,
-                env_group=env_group,
-                launch_type=LaunchTypes.farm_render
```
BigRoy, line 171: This PR changed this launch type from `LaunchTypes.farm_render` to `LaunchTypes.farm_publish` for all calls to `extractenvironments` CLI - including Deadline Render jobs which I suppose is an incorrect change?

Reported as issue here: https://github.com/ynput/ayon-applications/issues/2

PR #299, opened 2024-03-28
Maya: Improve validate current renderlayer renderable artist report

BigRoy:
## Changelog Description

Improve validate current renderlayer renderable artist report

## Additional info

n/a

## Testing notes:

1. Set up a render scene
2. Set up no renderable camera in the current renderlayer
3. Set up a renderable camrea in another renderlayer (the one you are submitting)
4. Publish (submit scene)

Validation should be clear to the artist.


tokejepsen:
> Set up no renderable camera in the current renderlayer

How would you do that?
Could do it maybe with code and iterate through all cameras, but unsure what you mean here.

BigRoy:
> > Set up no renderable camera in the current renderlayer
> 
> How would you do that? Could do it maybe with code and iterate through all cameras, but unsure what you mean here.

Select camera in the attribute editor, under Output Settings disable Renderable.

image

LiborBatek:
Hmm, ok, I also didnt get but meanwhile you have added that comment with ```Renderable``` attrib on ```camera```

however it did not show me human aka artist friendly Validation error msg but this

image

BigRoy:
> Hmm, ok, I also didnt get but meanwhile you have added that comment with `Renderable` attrib on `camera`
> 
> however it did not show me human aka artist friendly Validation error msg but this
> image

Could you share the report? I have a feeling that this may actually be an error from a **different** validator.
In your case you likely also had no active renderable camera in the renderlayer - and thus it _may_ fail on Collect Render (which I think I have a fix for somewhere in our codebase as well which would then be a different PR).

Anyway, let me break this down:

We sometimes have render camera overrides per Render Setup Layer. This means making a collection with an override to the "renderable" attribute (disabling one, enabling another). Or, disable all cameras in the master layer by default, but just enable the camera you want as an override in the particular renderlayers.

Once you do that I suspect you'll not hit the hard error when you're publishing whilst in the master layer. Because, as the validator describes, there's a bug in Redshift where if you are in the masterlayer then in batch renders it doesn't want to render another layer because it seems to check 'is there a renderable camera' in the layer the scene starts in - not in the layer you're about to render.

LiborBatek:
here is that JSON report for that error which I have encountered above,

meanwhile I will adapt my testing via those camera overrides per layer and  update my review with those...

[publish-report-240329-11-15.json.txt](https://github.com/ynput/ayon-core/files/14802772/publish-report-240329-11-15.json.txt)

BigRoy:
> here is that JSON report for that error which I have encountered above,
> 
> meanwhile I will adapt my testing via those camera overrides per layer and update my review with those...
> 
> [publish-report-240329-11-15.json.txt](https://github.com/ynput/ayon-core/files/14802772/publish-report-240329-11-15.json.txt)

This does indeed fail on `CollectRender` so is unrelated to this PR.

```python
    runner(*args)
  File "C:\Work\REPO\ayon-core\client\ayon_core\hosts\maya\plugins\publish\collect_render.py", line 115, in process
    assert render_products, "no render products generated"
AssertionError: no render products generated
```

However, I think I have a fix for that somewhere - so you might see a PR for that soon.

LiborBatek (commented):
So I have used override for that ```camera``` and ```renderable``` attrib and my findings so far are that it reports nicely only if ```defaultRenderLayer``` does have ```disabled``` state of that attribute, as seen in this scrngrab

Screenshot 2024-03-29 112316

The issue here besides this particular ```renderLayer``` not being used for render publishing  is when using overriden ```Renderable``` parameter aka ```disabled``` on actual ```renderLayer``` used for publishing...then Im getting error without any user friendly msg like this:

Screenshot 2024-03-29 112554touch

```Traceback (most recent call last):
    runner(*args)
  File "C:\Work\REPO\ayon-core\client\ayon_core\hosts\maya\plugins\publish\collect_render.py", line 115, in process
    assert render_products, "no render products generated"
AssertionError: no render products generated```

BigRoy:
> The issue here besides this particular `renderLayer` not being used for render publishing is when using overriden `Renderable` parameter aka `disabled` on actual `renderLayer` used for publishing...then Im getting error without any user friendly msg like this:

This is actually the correct behavior. This validator checks the _current renderlayer_ - not necessarily the one that you render with. The validation report should make that clear right hand side here:

> Screenshot 2024-03-29 112316

Let me know if that's not clear enough.

---

Then the other error that if a renderlayer that is to be submitted for rendering has no renderable camera crashing without a nice validation error is a separate issue, definitely one to concern us about and as stated I believe I have code for that in my OpenPype repository - I'll look into setting up a PR asap.

tokejepsen (approved):

tokejepsen:
Since this is specific to Redshift, should we query whether the Redshift plugin is being used or the current renderer is Redshift?

Might be confusing for artists that do not touch Redshift to get this message.

moonyuet (approved):

moonyuet:
@BigRoy, this PR can be merged after adding the related condition of querying the redshift.

tokejepsen (changes requested):
Yeah, lets get this more specific to Redshift.

BigRoy:
> Since this is specific to Redshift, should we query whether the Redshift plugin is being used or the current renderer is Redshift?

Should be done with https://github.com/ynput/ayon-core/pull/299/commits/2bf1ca2a2df2fc5009cdd4d4d704e4a4c791b8d9

I didn't test it myself yet.

tokejepsen (changes requested):
Is this plugin not accounting for no renderable cameras already?

https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/publish/validate_render_single_camera.py#L69-L71

BigRoy:
> Is this plugin not accounting for no renderable cameras already?
> 
> https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/publish/validate_render_single_camera.py#L69-L71

It does - but that validates whether the renderlayer for the layer being submitted (that actually gets rendered) has a render camera set. This particular validator validates the _current active renderlayer in the scene file._ So this validator validates the renderlayer that the scene opens in, because that's where the Redshift issue came up. The redshift error would report about the current renderlayer even if in mayabatch it was going to render another layer that actually did have render cameras set.

So them both validators being present is intended and correct.

tokejepsen (approved):

PR #301, opened 2024-03-28
Maya: Validate Look Shading Group ignore if no material

BigRoy:
## Changelog Description

Do not error with confusing message if shadingEngine has no material for whatever reason.

## Additional info

We hit this in production once and took a few minutes to debug and find the issue. By having this warning it's much easier to find what might be causing the issue.

## Testing notes:

1. Prepare a look
2. For one material, break the connection between material and `shadingEngine.surfaceShader`
3. Publish look
4. Validator should detect it without crashing.

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_look_shading_group.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_look_shading_group.py
@@ -47,10 +47,18 @@ class ValidateShadingEngine(pyblish.api.InstancePlugin,
                 shape, destination=True, type="shadingEngine"
             ) or []
             for shading_engine in shading_engines:
-                name = (
-                    cmds.listConnections(shading_engine + ".surfaceShader")[0]
-                    + "SG"
+                materials = cmds.listConnections(
+                    shading_engine + ".surfaceShader",
+                    source=True, destination=False
                 )
+                if not materials:
+                    cls.log.warning(
+                        "Shading engine '{}' has no material connected to its "
+                        ".surfaceShader attribute.".format(shading_engine))
+                    continue
+
+                material = materials[0]  # there should only ever be one input
+                name = material + "SG"
                 if shading_engine != name:
                     invalid.append(shading_engine)
 


LiborBatek (approved):
Works ok for me....
did try to disconnect shader from SG and just this warning popped up. No any failure.

image

BigRoy:
Thanks, yes - that's great! It's quite technical but it really is quite an edge case that should really rarely happen 🗡️

tokejepsen:
> Thanks, yes - that's great! It's quite technical but it really is quite an edge case that should really rarely happe

Should it not error rather than just a warning?

BigRoy:
> > Thanks, yes - that's great! It's quite technical but it really is quite an edge case that should really rarely happe
> 
> Should it not error rather than just a warning?

It could _technically_ be valid for some shading engines, e.g. those that are only intended to be used as volume shader or whatever - again, edge case.

Also, if this were an error I don't think an artist would really know _what to do_. So having the warning is mostly for that _if_ this happens to generate some issues down the line a TD might see the warning and think "hey, maybe it's due to that." which is what we used the warning for. It had really only been one intern that hit this - but then hit this a few times within a week, which was odd and never found out really why that particular person kept hitting this issue.

Anyway, does that answer your question?

tokejepsen:
> Anyway, does that answer your question?

Yeah, it does answer it. Personally I would throw an error on an optional plugin since warning are ignored by artists often, but if a warning solves your case I dont mind merging.

BigRoy:
> > Anyway, does that answer your question?
> 
> Yeah, it does answer it. Personally I would throw an error on an optional plugin since warning are ignored by artists often, but if a warning solves your case I dont mind merging.

I believe since the shader doesn't end up with a 'material' there's another validator that ensures all look meshes have materials or whatever - and as such, if this were to throw an error you would be hitting an publish validation error more than once for what is essentially the same source issue.

Since this Validator intends to just validate shader names (and nothing else!) I don't see why it should tell you anything in your scene is wrong if a shading engine does not have a material. If we do want to error on that - then it's not up to this validator that is there just to check material and shading engine names.

tokejepsen:

PR #303, opened 2024-03-28
Maya: Support SelectInvalidAction with pyblish ContextPlugin

BigRoy:
## Changelog Description

Support `SelectInvalidAction` with pyblish ContextPlugin

## Additional info

I don't believe there's currently any `ContextPlugin` in the maya integration that is actively using the `SelectInvalidAction`. However we had some in our codebase and it would be nice to port at least this functionality upstream to avoid any surprises in the future.

We might want to match the logic in other host integrations if they also have a `SelectInvalidAction` so they are also prepared.

## Testing notes:

1. Select Invalid actions should still work in publisher UI.

### DIFF
--- a/client/ayon_core/hosts/maya/api/action.py
+++ b/client/ayon_core/hosts/maya/api/action.py
@@ -4,7 +4,10 @@ from __future__ import absolute_import
 import pyblish.api
 import ayon_api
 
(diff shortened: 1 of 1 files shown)


tokejepsen (approved):

Easy testing for `SelectInvalidAction` is to make a `model` without the `_GEO` suffix.

PR #305, opened 2024-03-28
Houdini: Optimize collect inputs by caching scene containers once

BigRoy:
## Changelog Description

Optimize collect inputs by caching scene containers once

## Additional info

Behavior should be the same - it just doesn't list all scene containers and involved nodes per instance, but does it only once in the publish session, making it much faster when running over many instances.

## Testing notes:

1. Publish a few different Houdini products, preferably with loaded content used in the inputs - should work without errors.


client/ayon_core/hosts/houdini/plugins/publish/collect_inputs.py
```diff
@@ -108,13 +107,32 @@ def process(self, instance):
...
-        nodes.append(output)
-
-        # Collect containers for the given set of nodes
-        containers = collect_input_containers(nodes)
+        # For large scenes the querying of "host.ls()" can be relatively slow
+        # e.g. up to a second. Many instances calling it easily slows this
+        # down. As such, we cache it so we trigger it only once.
+        # todo: Instead of hidden cache make "CollectContainers" plug-in
+        cache_key = "__cache_containers"
+        scene_containers = instance.context.data.get(cache_key, None)
```
moonyuet, line 115: just a small question, is the `cache_key` always named as `__cache_containers`? It seems quite hard-coded to me but if it is always named like this, it is still okay.

client/ayon_core/hosts/houdini/plugins/publish/collect_inputs.py
```diff
@@ -108,13 +107,32 @@ def process(self, instance):
...
+        if scene_containers is None:
+            # Query the scenes' containers if there's no cache yet
+            host = registered_host()
+            scene_containers = list(host.ls())
+            for container in scene_containers:
+                # Embed the members into the container dictionary
+                container_members = set(get_container_members(container))
+                container["_members"] = container_members
+            instance.context.data["__cache_containers"] = scene_containers
```
moonyuet, line 124:
```suggestion
            instance.context.data[cache_key] = scene_containers
```

moonyuet:
Tested with houdini, looks good and please review comment above.

client/ayon_core/hosts/houdini/plugins/publish/collect_inputs.py
BigRoy: Good question. In the past I had sometimes prefixed it also with the class name to ensure better uniqueness. But yes, in a way it is hardcoded. Feel too unsafe? Should I prefix it more?

BigRoy:
> I'm not quite familiar with this plugin.
> So, we trace of any loaded nodes that contribute to the output data and save their ids in `"inputRepresentations"` ?
> if my description is correct, then I'm wondering how could we benefit from that piece of information ?

That's correct - it detects what input loaded content was used in your output so they could turn into "Links" in your publish. I believe there's a `IntegrateInputLinks` plug-in that uses this data.

The data gets converted here: https://github.com/ynput/ayon-core/blob/70306a1229c54cffce0a2513fbd784a47d1bce81/client/ayon_core/plugins/publish/collect_input_representations_to_versions.py#L5

Then those turn into "generative links": https://github.com/ynput/ayon-core/blob/c4f74a6aedc12fe7d74adedbb6403eb2dca51eda/client/ayon_core/plugins/publish/integrate_inputlinks.py#L115

client/ayon_core/hosts/houdini/plugins/publish/collect_inputs.py
moonyuet: > @moonyuet let me know what you'd like me to do here to make this ready for merging

we can merge this first as I guess there would be enhancement when building the universal node for publishing.

moonyuet (approved):

PR #306, opened 2024-03-28
Launcher: Use selection object for action

iLLiCiTiT:
## Changelog Description
Prepared selection object that helps to handle selection in Launcher actions.

## Additional info
The object should backwards compatible for unconverted actions.

## Testing notes:
1. Launcher actions are working (e.g. applications)
2. It would be nice to test backwards compatibility of uncorverted actions but I don't have any.

### DIFF
--- a/client/ayon_core/pipeline/actions.py
+++ b/client/ayon_core/pipeline/actions.py
@@ -102,6 +102,12 @@ class LauncherActionSelection:
         return False
 
     def get(self, key, default=None):
+        """
+
+        Deprecated:
+            Added for backwards compatibility with older actions.
+
+        """
         warnings.warn(
             (
                 "Using deprecated access to selection data. Please use"
@@ -116,6 +122,12 @@ class LauncherActionSelection:
             return default
 
     def items(self):
+        """
+
+        Deprecated:
+            Added for backwards compatibility with older actions.
+
+        """
         for key, value in (
             ("AYON_PROJECT_NAME", self.project_name),
             ("AYON_FOLDER_PATH", self.folder_path),
@@ -125,10 +137,22 @@ class LauncherActionSelection:
                 yield (key, value)
 
     def keys(self):
+        """
+
+        Deprecated:
+            Added for backwards compatibility with older actions.
+
+        """
         for key, _ in self.items():
             yield key
 
     def values(self):
+        """
+
+        Deprecated:
+            Added for backwards compatibility with older actions.
+
+        """
         for _, value in self.items():
             yield value
 
--- a/client/ayon_core/pipeline/actions.py
+++ b/client/ayon_core/pipeline/actions.py
@@ -64,9 +64,9 @@ class LauncherActionSelection:
         )
         if key in {"AYON_PROJECT_NAME", "AVALON_PROJECT"}:
             return self.project_name
-        if key == {"AYON_FOLDER_PATH", "AVALON_ASSET"}:
+        if key in {"AYON_FOLDER_PATH", "AVALON_ASSET"}:
             return self.folder_path
-        if key == {"AYON_TASK_NAME", "AVALON_TASK"}:
+        if key in {"AYON_TASK_NAME", "AVALON_TASK"}:
             return self.task_name
         raise KeyError(f"Key: {key} not found")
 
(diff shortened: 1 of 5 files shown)


client/ayon_core/pipeline/actions.py
```diff
@@ -10,6 +13,224 @@
...
+        """
+        return self._project_name is not None
+
+    @property
+    def is_folder_selected(self):
+        return self._folder_id is not None
+
+    @property
+    def is_task_selected(self):
+        return self._task_id is not None
```
BigRoy, line 221: Missing docstring.
```suggestion
    @property
    def is_folder_selected(self):
        """Return whether a folder is selected.

        Returns:
            bool: Whether a folder is selected.

        """
        return self._folder_id is not None

    @property
    def is_task_selected(self):
        """Return whether a task is selected.

        Returns:
            bool: Whether a task is selected.

        """
        return self._task_id is not None
```

client/ayon_core/pipeline/actions.py
```diff
@@ -10,6 +13,224 @@
...
+            "AVALON_TASK",
+        }:
+            return self._task_id is not None
+        return False
+
+    def get(self, key, default=None):
+        try:
+            return self[key]
+        except KeyError:
+            return default
```
BigRoy, line 104: I feel like this is here mostly just for backwards compatibility - and I wonder, if moving forward we want to keep this. How about only implementing backwards compatibility here with deprecation warning so that we push for using the explicit property/getter methods?

client/ayon_core/pipeline/actions.py
```diff
@@ -10,6 +13,224 @@
...
+            return None
+        if self._folder_path is None:
+            self._folder_path = self.folder_entity["path"]
+        return self._folder_path
+
+    def get_task_id(self):
+        """Selected task id.
+
+        Returns:
+            Union[str, None]: Selected task id.
```
BigRoy, line 141: Isn't optional clearer here?
```suggestion
            Optional[str]: Selected task id.
```

client/ayon_core/pipeline/actions.py
```diff
@@ -10,6 +13,224 @@
...
+    def __init__(
+        self,
+        project_name,
+        folder_id,
+        task_id,
+        folder_path=None,
+        task_name=None,
+        project_entity=None,
+        folder_entity=None,
+        task_entity=None
```
BigRoy, line 42: Could you point me to where we're actually setting these from the launcher? I can't seem to find that.

client/ayon_core/pipeline/actions.py
iLLiCiTiT: To be honest, I don't like using `Optional` in return type hint.

client/ayon_core/pipeline/actions.py
iLLiCiTiT: Nowhere.

BigRoy: In that case - we shouldn't expose it as if it'll be available? :) Or will you be adding commits to fix it? :)

BigRoy: Ah wait, I only _now_ see that it queries it whenever you access it. Nevermind. I'm blind.

iLLiCiTiT: Why not? It might be used for testing.

BigRoy:
> Launcher actions are working (e.g. applications)

Confirmed.

> It would be nice to test backwards compatibility of uncorverted actions but I don't have any.

[This plugin](https://github.com/BigRoy/ayon-colorbleed/blob/b8bbfcee92ecea171327f6b463e09e43a50fefc6/client/ayon_colorbleed/plugins/launcher_actions/debug_shell.py#L80) doesn't seem to get listed for me.

iLLiCiTiT:
> [This plugin](https://github.com/BigRoy/ayon-colorbleed/blob/b8bbfcee92ecea171327f6b463e09e43a50fefc6/client/ayon_colorbleed/plugins/launcher_actions/debug_shell.py#L80) doesn't seem to get listed for me.

Should be fixed

BigRoy (approved):
My custom actions now show up and launch ok. 👍

PR #307, opened 2024-03-28
Houdini: Fix generic filepath loader 

BigRoy:
## Changelog Description

Refactor imports + update method arguments

## Additional info

Not sure how these sneaked through in #222 - especially without the import change this should've been broken.

## Testing notes:

1. Load, update and switch a generic filepath container.


moonyuet (approved):
Looks good for me. It doesn't fix the bug after the update of the loaderPlugin

PR #314, opened 2024-03-28
Nuke: simplified deadline submission on write node - AY-974

tokejepsen:
## Changelog Description
This adds a `Render Farm` button to the instance node, next to the `Render Local`.

To use as much existing code as possible, we publish headless to submit to the farm with the instance from the node and the workfile. There are some slight hacking to point the job submission to a timestamped copy of the workfile.

## Testing notes:
1. In the settings `ayon+settings://nuke/create/CreateWriteRender/instance_attributes` add `Render On Farm` attribute.
2. Create write instance in Nuke.
3. Hit the `Render Farm` buttton.
4. Validate job is submitted to Deadline and renders correctly.


moonyuet (approved):
The button shows up in the write instance:) 
image

and submit to render farm successfully
image
If my webservice for deadline has gone, it raise the error as expected too.
image

jakubjezek001:
> The button shows up in the write instance:) image
> 
> and submit to render farm successfully image If my webservice for deadline has gone, it raise the error as expected too. image

This seems like your webservice is not working. @tokejepsen could we add some better UX communication in case the sanity check for the connection is failing?

client/ayon_core/hosts/nuke/api/lib.py
```diff
@@ -1268,6 +1291,9 @@ def create_write_node(
                 link.setFlag(0x1000)
                 GN.addKnob(link)
 
+    # Adding render farm submission button.
+    add_button_render_farm(GN)
```
jakubjezek001, line 1295: lets add the button on demand. How about to add it as feature item into `ayon+settings://nuke/create/CreateWriteRender/instance_attributes`. Something like `Instant Farm submission` or something similar?

jakubjezek001 (approved):
The code looks solid and feature was working for me. I would add some webservice sanity check and button on demand and then it is good to be merged. Thanks @tokejepsen for the speed!

tokejepsen:
> could we add some better UX communication in case the sanity check for the connection is failing?

@jakubjezek001 could you elaborate on the what you want to happen? The error for webservice should not happen if setup correctly, so should not really be artists facing.

client/ayon_core/hosts/nuke/api/lib.py
BigRoy: This feature sounds similar to a feature implemented in Houdini by @MustafaJafar where I believe he named it something like "Self-publish" or alike.

tokejepsen: I think `Headless Farm Submission` would work.

kjpoli:
This matches pretty closely the functionality we have at Hornet, but retains the paradigm of having a publish for every submission if i'm understanding the call to `util.integrate` in context here, I should have time to test tomorrow, but is this process blocking for Nuke ?. The UX is more or less exactly what we have and was hugely popular with the compers.

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,160 @@ def is_headless():
...
+
+    if not success:
+        show_message_dialog(
+            "Extraction Errors", error_report, level="critical"
+        )
+        return
+
+    # Save the workfile.
+    host = registered_host()
+    host.save_file(host.current_file())
```
BigRoy, line 277: Doesn't the publishing logic of 'submitting to the farm' take care of this?

Also, quite a hack to do this here outside of the publishing logic itself?

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,160 @@ def is_headless():
...
+
+    directory = os.path.join(os.path.dirname(base), "farm_submissions")
+    if not os.path.exists(directory):
+        os.makedirs(directory)
+
+    filename = "{}_{}{}".format(
+        os.path.basename(base), formatted_timestamp, ext
+    )
+    path = os.path.join(directory, filename).replace("\\", "/")
+    context.data["currentFile"] = path
```
BigRoy, line 292: Do we really want to hijack the publishing logic this much? This looks like an anti-pattern?

tokejepsen:
> but is this process blocking for Nuke ?

@kjpoli yes, the processing of the submission is blocking Nuke, but should not take long to process since you are just copying the workfile and submitting that.

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,160 @@ def is_headless():
...
+        _submit_headless_farm(node)
+    else:
+        # If not in root context, move to the root context and then execute the
+        # code.
+        with nuke.root():
+            _submit_headless_farm(node)
+
+
+def _submit_headless_farm(node):
+    context = util.collect()
```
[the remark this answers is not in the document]
tokejepsen: > I feel that in a way these two features are rather similar and it would be great to unify where we can and streamline the "publishing API" where it appears to be lacking?

Think this will be outside the scope of this PR unfortunately.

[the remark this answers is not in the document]
tokejepsen: > This not going through the CreateContext means you might lack instance data that it needs from the new publisher Create Context. I really wonder whether this logic is 'safe' and should be more re-written to at least started of with the Create Context.

@iLLiCiTiT ?

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,160 @@ def is_headless():
...
+    for count, Instance in enumerate(context):
+        if Instance.data["family"] == "workfile":
+            instance_workfile = Instance
+            continue
+
+        instance_node = Instance.data["transientData"]["node"]
+        if node.name() == instance_node.name():
+            instance = Instance
+        else:
+            indexes_to_remove.append(count)
```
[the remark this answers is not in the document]
tokejepsen: Yeah, will do.

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: Good point. Extraction stage saves the workfile.

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: Got any suggestions to improve it?

client/ayon_core/hosts/nuke/api/utils.py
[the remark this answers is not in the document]
jakubjezek001: Do we really need it to be using **CreateContext** @iLLiCiTiT, this functionality's aim is not publishing. It only needs to simply submit job to farm, with no validation what so ever. This should be crude as much as possible.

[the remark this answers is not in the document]
tokejepsen: > It only needs to simply submit job to farm, with no validation what so ever.

Not entirely true. We want validations and go through the publishing code as much as possible, so we do not duplicate code or reinvent some logic.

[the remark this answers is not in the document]
BigRoy: I really have my doubts whether it's worth maintaining this much code for whatever this seems to be doing. I think this code is a clear sign of how "bad" the situation is currently with the publishing system and how much of an easy API lacks for that - as others have commented frequently. I believe @dee-ynput has been made aware of that and @fabiaserra has been vocal about that lack of API as well.

Not saying I have any issues with merging this (we don't use Nuke anyway) - but just saying that this shows the issue at hand. This looks like something that in essence should've been a few lines of code away to implement.

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,150 @@ def is_headless():
...
+    if nuke.root() == nuke.thisNode():
+        _submit_headless_farm(node)
+    else:
+        # If not in root context, move to the root context and then execute the
+        # code.
+        with nuke.root():
+            _submit_headless_farm(node)
+
+
+def _submit_headless_farm(node):
```
BigRoy, line 198: Can we add a docstring please - kind of explaining the what and why?

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,150 @@ def is_headless():
         bool: headless
     """
     return QtWidgets.QApplication.instance() is None
+
+
+def create_error_report(context):
```
BigRoy, line 154: It's too bad this is in `nuke.api.utils` - this is really something that should be "easy" to retrieve from the publishing API and should not need a 'nuke specific implementation' I would say.

Anyway, docstring would be welcome.

client/ayon_core/hosts/nuke/plugins/create/create_write_render.py
```diff
@@ -40,11 +40,15 @@ def get_pre_create_attr_defs(self):
...
+        settings = self.project_settings["nuke"]["create"]["CreateWriteRender"]
+        settings = settings["instance_attributes"]
+
         # add fpath_template
         write_data = {
             "creator": self.__class__.__name__,
             "productName": product_name,
-            "fpath_template": self.temp_rendering_path_template
+            "fpath_template": self.temp_rendering_path_template,
+            "headless_farm_submission": "headless_farm_submission" in settings
```
BigRoy, line 51: Is this being present in settings really what makes it a headless farm submission? Shouldn't it check a value or whatever? Or how does this work?

client/ayon_core/hosts/nuke/api/utils.py
BigRoy: I feel like:

- Collect instances
- Find the relevant instance
- Set the relevant instance attribute value (toggle the active state, etc.)
- Continue publish with only that instance active

Which would be along the lines of:

```python
# pseudocode
create_context = CreateContext(host)
for instance in create_context.instances:
    instance.data["active"] = is_the_relevant_intsance(instance)
    instance.data["render_mode"] = "farm"  # or whatever the attribute definition and value is
try:
    create_context.publish()
except PublishFailed:
    report = create_context.get_report()
    show_message_dialog("Publish failed", report, level="critical")
```

This should essentially all be handled by 'default publishing logic' - we just want to trigger it simply for this one instance, right?

client/ayon_core/hosts/nuke/api/utils.py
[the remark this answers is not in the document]
fabiaserra: > Think this will be outside the scope of this PR unfortunately.

Behind this argument and pressure from clients you guys end up merging so much half-done patchy stuff and then the proper fix is never done... arguing that there's no time or resources to do it right and everytime you do this it gets harder and harder to realign and clean the design, even more so because of the policy you guys have of creating simple small PRs that disencourage doing anything big, like I said here https://github.com/ynput/ayon-shotgrid/pull/83#issuecomment-2045353485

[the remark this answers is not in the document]
fabiaserra: Clear example of what I'm saying is this comment I added warning about the changes, https://github.com/ynput/OpenPype/pull/4903#issuecomment-1738343017 it was merged and never done anything else about it

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,150 @@ def is_headless():
...
+    if not success:
+        show_message_dialog(
+            "Collection Errors", error_report, level="critical"
+        )
+        return
+
+    # Find instance for node and workfile.
+    instance = None
+    instance_workfile = None
+    for Instance in context:
```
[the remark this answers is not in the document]
tokejepsen: Done

client/ayon_core/hosts/nuke/api/utils.py
BigRoy: > The client has requested to have a timestamped workfile instead of rendering from the current workfile, hence the hijacked publishing process.

I mean - that sounds _just_ as useful for any farm submission. Why not implement it in the regular publishing logic?

Even more so - wasn't there like tons of logic already that allowed farm rendering to basically render from a **published workfile** which in essence is exactly that - a 'stamped' static version for the farm? (Timestamp is then the publish datetime, etc.)

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: Done.

fabiaserra:
I really don't get what this is doing. Is the `Render farm` also doing a headless publish or just rendering out?

Our integration of OP in Nuke separates very clearly the rendering vs the publishing, on the write node we only expose the options to streamline the render and write either locally or in the farm (completely split from any publish process), like this screenshot shows: https://community.ynput.io/t/nuke-creators-as-nodes/124/4?u=fabiaserra for the Deadline submission to use a copy of the script and not the workfile you simply need to avoid passing the `SceneFile` arg in the plugin info and it will do the copy automatically.

And then to publish a write node directly we have a util command they can run by simply selecting the write node and typing `Publish`, running the following code:
```
def publish():
    """Publish utility to publish selected Write node"""
    
    write_selected_nodes = [
        selected_nodes for selected_nodes in nuke.selectedNodes()
        if selected_nodes.Class() == "Write"
    ]
    if not write_selected_nodes:
        nuke.message("You must select a write node")
        return
    # For now it's been decided to limit the publish to only a single Write node
    # at a time but if in the future we decide we want to be able to publish
    # all selected write nodes it would simply be matter of removing this
    elif not len(write_selected_nodes) == 1:
        nuke.message("You must select only one write node")
        return

    user_input = PublishPanel()
    submission_notes = None

    # Until user sets submission notes it keeps prompting the dialog unless they
    # click cancel
    while not submission_notes:
        if not user_input.showModalDialog():
            return

        submission_notes = user_input.submission_notes_knob.value()
        if not submission_notes:
            nuke.message("You must enter submission notes before publishing")

    host = registered_host()
    create_context = create.CreateContext(host, reset=True)
    instance_exists = False
    for create_instance in create_context.instances:
        node = create_instance.transient_data["node"]
        if create_instance.family == "workfile":
            create_instance["active"] = True
            data_to_update = {
                "publish_attributes": {
                    "ValidateVersion": {"active": False}
                }
            }
            api.update_node_data(node, "publish_instance", data_to_update)
            log.info(api.get_node_data(node, "publish_instance"))
            continue
        
        if node in write_selected_nodes:
            instance_exists = True
            create_instance["active"] = True
            data_to_update = {
                # Remove comment from any instance. Ensure comment is from context
                "comment": "",
                "version": int(node["ver"].value()),
                "creator_attributes": {
                    "render_target": "frames_farm",
                    "review": True
                }
            }
            api.update_node_data(node, "publish_instance", data_to_update)
        else:
            create_instance["active"] = False

    if not instance_exists:
        nuke.message("Couldn't find publish instance on selected write node")
        return

    create_context.save_changes()

    main_window = lib.get_main_window()
    publisher_window = host_tools.get_tool_by_name(
        tool_name="publisher", parent=main_window
    )
    publisher_window.show_and_publish(submission_notes)
```

It would be very straightforward to add this command as a button directly on the Write node. What am I missing? It really is not beneficial to tie rendering and publishing together as I have said multiple times here and in Houdini discussions because you really want to reduce the issues that arise when rendering and adding Ayon to the mix just creates more issues so I hope this is not doing that?

client/ayon_core/hosts/nuke/plugins/create/create_write_render.py
tokejepsen: The `settings` variable here is an enum; `ayon+settings://nuke/create/CreateWriteRender/instance_attributes?project=demo_Commercial`

Screenshot 2024-04-12 155650

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,160 @@ def is_headless():
...
+
+    # Disable version validation.
+    instance.data.pop("latestVersion")
+    instance_workfile.data.pop("latestVersion")
+
+    # Remove all other instances.
+    indexes_to_remove.sort(reverse=True)
+    for i in indexes_to_remove:
+        if 0 <= i < len(context):
+            del context[i]
```
BigRoy: The is in range change would only be obscuring an actual bug - since you're removing from the back of the indices that should never occur. And if it does, then you have bigger issues to worry about because it means your indices would be misaligned.
```suggestion
    # Remove all other instances.
    for i in reversed(indexes_to_remove):
        del context[i]
```

client/ayon_core/hosts/nuke/api/utils.py
BigRoy: > Client does not want a published workfile to clutter the database.

This feels quite related to [this Ynput Community forum discussion on not publishing directly, but first making a 'reviewable' first](https://community.ynput.io/t/publishing-renders-from-houdini/1359/2).

Anyway, since your logic is still not removing the workfile instance, nor removing it - isn't it still being published?

client/ayon_core/hosts/nuke/plugins/create/create_write_render.py
BigRoy: I see now - ok, then maybe `settings` as a variable name isn't very explicit and `instance_attributes` as variable name may be more accurate?

tokejepsen:
> I really don't get what this is doing. Is the Render farm also doing a headless publish or just rendering out?

@fabiaserra the `Render farm` button does a headless publish of a timestamped workfile, to re-use as much as possible of the publishing code for farm submission. The job on the farm is only the Nuke render job. Nothing gets published.

> And then to publish a write node directly we have a util command they can run by simply selecting the write node and typing Publish, running the following code:

@fabiaserra thanks for the code! This is doing almost the same thing, but without showing the publishing window. Just error messages if there is any during "publishing" (in quotes here, cause we are not actually publishing anything) or a message if the submission was successful.

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: Not sure where you got this from, but this code was removed (on your request :) )

fabiaserra:
> @fabiaserra the `Render farm` button does a headless publish of a timestamped workfile, to re-use as much as possible of the publishing code for farm submission. The job on the farm is only the Nuke render job. Nothing gets published.

How is that not contradictory? hahah you say the button does a headless **publish** but it doesn't do any publish? Please explain

> @fabiaserra thanks for the code! This is doing almost the same thing, but without showing the publishing window. Just error messages if there is any during "publishing" (in quotes here, cause we are not actually publishing anything) or a message if the submission was successful.

You are just reusing the publish code to do the render submission but skip the publish dependency task?

client/ayon_core/hosts/nuke/api/utils.py
BigRoy: Interesting - github was still showing it to me somehow. Ignore me then.

BigRoy:
> button does a headless publish of a timestamped workfile, to re-use as much as possible of the publishing code for farm submission. The job on the farm is only the Nuke render job. Nothing gets published.

This really sounds like they just want a button that does:

- Save my file as a timestamped filename
- Submit to deadline for rendering
(And that is all)?

fabiaserra:
I appreciate the philosophy of reusing code but as @BigRoy pointed out, it would be far better to create an agnostic Deadline submit module (like we did here https://github.com/fabiaserra/ayon-core/blob/release/alkemyx/client/ayon_core/modules/deadline/lib/submit.py#L21) and start reusing it across all the DCCs (making it more useful and stable on every iteration of adding support to the other DCCs) instead of patching on top of an already messed up design and adding extra overhead

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: > Anyway, since your logic is still not removing the workfile instance, nor removing it - isn't it still being published?

True, did not notice that. The publishing does not actually publish the workfile when testing, and I dont know why, but I've deactivated the instance just in case.

jakubjezek001:
> I appreciate the philosophy of reusing code but as @BigRoy pointed out, it would be far better to create an agnostic Deadline submit module (like we did here [fabiaserra/ayon-core@`release`/alkemyx/client/ayon_core/modules/deadline/lib/submit.py#L21](https://github.com/fabiaserra/ayon-core/blob/release/alkemyx/client/ayon_core/modules/deadline/lib/submit.py?rgh-link-date=2024-04-12T15%3A12%3A40Z#L21)) and start reusing it across all the DCCs (making it more useful and stable on every iteration of adding support to the other DCCs) instead of patching on top of an already messed up design and adding extra overhead

We do not have this available in our codebase. I am guessing that if we had got any PR with the solution you are proposing we would probably consider it but for now we are sticking to what we are having. This is client request and we are usually trying to evaluate solution based on particular tier we need to work with. At the moment to adapt your solution and use it in this particular situation would inflate the allowed time. Lets continue with this please and create an issue with proposal for changes for now @fabiaserra since our client is already expecting this to be deployed as soon as possible.

client/ayon_core/hosts/nuke/plugins/create/create_write_render.py
iLLiCiTiT: > I see now - ok, then maybe settings as a variable name isn't very explicit and instance_attributes as variable name may be more accurate?

100% agree
```
        instance_attributes = (
            self.project_settings
            ["nuke"]
            ["create"]
            ["CreateWriteRender"]
            ["instance_attributes"]
        )
```

client/ayon_core/hosts/nuke/api/utils.py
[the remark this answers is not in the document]
iLLiCiTiT: Ad not using `CreateContext`. The context is auto-created during collection, so it is not really necessary, but what I'm worried about is the approach how it tells to use farm publishing, and skip some other stuff.

Clearing `"families"` is dangerous, change `"farm"` out of plugins is dangerous, changeing `"transfer"` (to `False` for some reason?) is dangerous. Forcing `"use_published_workfile"` to `False` is strange too.

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,174 @@ def is_headless():
...
+        show_message_dialog(
+            "Collection Error",
+            "Could not find the instance from the node.",
+            level="critical"
+        )
+        return
+
+    # Enable for farm publishing.
+    instance.data["farm"] = True
+    instance.data["transfer"] = False
```
iLLiCiTiT, line 257: What `"transfer"` does? I didn't find any usage of it anywhere in codebase.

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,174 @@ def is_headless():
...
+
+    # Clear the families as we only want the main family, ei. no review etc.
+    instance.data["families"] = []
```
iLLiCiTiT, line 260: This can cause a lot of issues.

tokejepsen: Have you got a solution in mind or maybe what issues it could cause?

BigRoy:
> We don't have this in our codebase yet. If we received a pull request (PR) with your suggested solution, we might consider it. But for now, we're sticking with what we have. This is what the client wants, and we try to pick solutions that fit the tier we're working on. Right now, using your solution would take too much time. Let's stick with the current approach and create an issue to suggest changes later, @fabiaserra. Our client is waiting for this deployment, so we need to move fast.

I don't want to be too harsh - but admittedly this feels like a case where this should be rolled out to that client - and to that client only. Like @fabiaserra states I don't feel like this should be the state of code and workflows we should allow into the main branch?

At the same time, I don't care too much since it's Nuke addon code which we personally don't use. But just wanted to remark that if something is 'just urgent for some studio' doesn't mean that's the best state of taking that upstream into the main codebase.

client/ayon_core/hosts/nuke/api/utils.py
iLLiCiTiT: Nothing from this should happen out of publish plugins. So the question is how to affect publish logic from here.

You can create pyblish context object and set a specific key on it so some collector can do what are you doing here. So you can set `context.data["singleWriteSubmitNode"] = node`, take this as an example, can be done by instance id, node path etc. . Then add collector which in fact disables other instances and changes the attributes as necessary (can be order `pyblish.api.CollectorOrder - 0.4999`).

Also I think you should create `CreateContext` and pass it to the context too, to make sure the instance you're going to publish is enabled, because it won't be collected if is disabled.

client/ayon_core/hosts/nuke/api/lib.py
```diff
@@ -1144,6 +1154,19 @@ def create_write_node(
...
+    special_characters = set("!@#$%^&*()=[]{}|\\;',.<>/?~+-")
+    found_special_characters = []
+
+    # Check each character in the node name
+    for char in name:
+        if char in special_characters:
+            found_special_characters.append(char)
+
+    msg = f"Special characters found in name \"{name}\": "
+    msg += f"{' '.join(found_special_characters)}"
```
iLLiCiTiT, line 1167:
```suggestion
    msg = (
        f"Special characters found in name \"{name}\": "
        f"{' '.join(found_special_characters)}"
    )
```

dee-ynput:
That's a balance we need to keep in mind, indeed.
But our clients found the open-source dev and not only their private whish list. This is a cultural choice based on our values.
I totaly think we could set up finer staging options, and you are right in the concequences of the current situation. We'er hiring a Head of R&D for this among other needs, feel free to pass the info around and help us <3

Until then, we priorities fast delivery of client critical features.
We do know a new API is required, and we do need to schedule its implementation.
But this does not depends as much on "the best way" to do it than on "when will we be founded enough" to do it.

If this is creating more issues than help, you may want to do some lobbying on our founding clients in favor of the implementation of this API :)
We would be founded for it, we would have "best" aligned with "when", and you would be ecstatically happy :D
I think that would be excelent use for your debating talents ;)

fabiaserra:
> If this is creating more issues than help, you may want to do some lobbying on our founding clients in favor of the implementation of this API :) We would be founded for it, we would have "best" aligned with "when", and you would be ecstatically happy :D I think that would be excelent use for your debating talents ;)

Thank you for the extra context and I understand that doing that balance is not easy. However, I'd hope that as part of the funding you contemplate that the code written has to be as robust, stable and future-proof as possible and bad practices end up being way more expensive (for the funding clients too) than not. There's been already quite a few different client funded requests to provide Deadline submission across different DCCs, if you had spent the time on the second (or even third) request to align the submission logic into an abstracted API, the later requests would have been much easier to accomplish and it would make maintaining the codebase much cheaper than patching it in multiple locations, specially when the community (and your team) also benefits from having abstracted modules easier to dev with. I don't think creating a simple API entry point to do Deadline submissions is really that costly or time consuming... it doesn't need to be perfect on the first release, it's something that grows naturally the more things make use of it, but anyway, this is my opinion and I'm not a funding client so I will respect whatever you decide to do, I just hope this resonates some bells and try to do an effort for the upcoming development!

BigRoy:
Just wanted to 👍 with @fabiaserra 's comment here. 

I think another point to make could be that if this _were_ more globally implemented in a more usable way across the board that others would also adopt it and you'd also benefit from the community working with it as well. In a way you'renow not only wasting your own time (which you get funding for from the client so I guess it's not a pure waste) but it also takes 'our' time. In reverse, if this were part of the more global solution to the root of the problem then I'd be happy to spend some of my community time as well - which may gain you something. Or maybe that example is better made using a developer beside myself that actually saves you time ;)

antirotor:
I guess the point is that we are aligned and we want to do it. It is just a matter of when. Sometimes finding even one hour is difficult when there is so much to do as you all know for sure from the production. I do remember all those decisions like "let's do those changes directly in the render scene because there's no time to push it through the pipeline properly" just to find out month later that it was not a good idea at that moment.

dee-ynput:
I'm the first advocate for more formalisation, APIs, decoupling and all the shebang.
I've spend 25 years in production, I do know why we - and you - need it 😃
That's why lobbying for it is good 😉

But it's going to take the time it needs.
A Ondrej say, that's something we learn from production experience ✨

Right now our main priority is to run the last yards of the "OP to AYON" marathon.
This marathon is completely meant to help us adding APIs and moving out of monolith.
We do this for us, and for you.
And implementing short term solution is not just funding it, but also showing us (and you) where the need for it is. It's agile, it matches our clients needs, and most probably match other user needs too.

The ones paying the cost are us and you: the devs - as you pointed out.
That's what we do: we put the users first and work on the grant scheme of things. That's our strategy, and that's our values.
But I reckon you and your users see and enjoy the benefit of this every(work)day.
Seeing some downsides of it doesn't erase all the work that has been done before you can say "Ho! look! There... that sucks".

Please bear in mind that when you see something needed but not worked on, it doesn't mean we don't see it too.
Pointing it out is nice and helpfull because it can consolidates our position or triggers nice debates. But once the debate is done and we all agree on the need, it's rude, unfruitfull, and time consuming for everyone to keep on argueing around it.

When there is as much help than there is blames, we feel on the same team. And we are amazingly thankfull for it.
When blame is all we get, we're just ruining the team motivation.

### TL;DR
We agree, and we do our best. We're hoping you trust us.
Now go and thank the Ynput devs for all the good thing they've achieved yet, and all the good things that they're waiting to achieve next ❤️

fabiaserra:
> ### TL;DR
> We agree, and we do our best. We're hoping you trust us. Now go and thank the Ynput devs for all the good thing they've achieved yet, and all the good things that they're waiting to achieve next ❤️

I'm sorry if my comments felt rude! if I didn't like what the team has built so far I would have abandoned it long ago and I wouldn't participate on the community, big kudos to the team ❤️ My intention is not to demotivate anyone but to raise the bigger issues that I see, specially new ones being introduced, as Roy and you said, the ones being most impacted at the end are all of us, since we are the ones building on this product and supporting it so I want to make sure I'm protecting my bases and I understand the roadmap before investing my time and resources further. I wish I was closer to the dev team and some of the decision making to actually know what your plans are so that perhaps we could collaborate together or align our dev iterations so we don't duplicate efforts and work closer as a dev community

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: @iLLiCiTiT how do you pass the `CreateContext` to the publishing?
Tried the below but it does not work;
```
from pyblish import util

from ayon_core.pipeline.create import CreateContext


host = registered_host()
create_context = CreateContext(host)
context = util.collect(create_context)
```

client/ayon_core/hosts/nuke/plugins/create/create_write_render.py
tokejepsen: Done

client/ayon_core/hosts/nuke/api/utils.py
iLLiCiTiT:
```
import pyblish.api
from pyblish import util

from ayon_core.pipeline.create import CreateContext


host = registered_host()
create_context = CreateContext(host)
publish_context = pyblish.api.Context()
publish_context.data["create_context"] = create_context
util.collect(context=publish_context)
```

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -1,11 +1,17 @@
 import os
 import re
+import traceback
 
 import nuke
 
-from ayon_core import resources
+from pyblish import util, api
```
iLLiCiTiT, line 7: Please use full import of `pyblish.api` and `pyblish.util`.

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class CollectHeadlessFarm(pyblish.api.InstancePlugin):
+    """Setup instances for headless farm submission."""
+
+    order = pyblish.api.CollectorOrder + 0.4999
+    label = "Collect Headless Farm"
+    hosts = ["nuke"]
+
+    def process(self, instance):
```
iLLiCiTiT, line 11: Does this have to be instance plugin?

client/ayon_core/hosts/nuke/plugins/publish/increment_script_version.py
```diff
@@ -13,6 +13,8 @@ class IncrementScriptVersion(pyblish.api.ContextPlugin):
     hosts = ['nuke']
 
     def process(self, context):
+        if context.data.get("headless_farm", False):
```
iLLiCiTiT, line 16: Should this really be skipped?

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: We need to iterate over the instances to find the right and also the workfile instance. Could be context plugin, but we would be needing to iterate anyways.
Why should it not be instance plugin?

client/ayon_core/hosts/nuke/plugins/publish/increment_script_version.py
tokejepsen: Yeah, the workflow is to have a timestamped workfile and continue from the currently versioned workfile.

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: Done.

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
iLLiCiTiT: The plugin does not have any instance filters. Making in context plugin would make it run once. You can actually validate if there is any instance matching the node name, current implementation does not validate it in any way, just turns off all that don't match the conditions, but don't validate there is one which does match.

tokejepsen: Done

client/ayon_core/hosts/nuke/api/lib.py
```diff
@@ -1144,6 +1154,21 @@ def create_write_node(
...
         node (obj): group node with avalon data as Knobs
     '''
+    # Ensure name does not contain any invalid characters.
+    special_characters = set("!@#$%^&*()=[]{}|\\;',.<>/?~+-")
+    found_special_characters = []
+
+    # Check each character in the node name
+    for char in name:
+        if char in special_characters:
+            found_special_characters.append(char)
```
iLLiCiTiT, line 1164: This might require `import re`.
```suggestion
    special_chars = re.escape("!@#$%^&*()=[]{}|\\;',.<>/?~+-")
    special_chars_regex = re.compile(f"[{special_chars}]")
    found_special_characters = list(special_chars_regex.findall(name))
```

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,95 @@ def is_headless():
...
+    Args:
+        node (Node): The node for which the farm submission is being made.
+    """
+
+    host = registered_host()
+    create_context = CreateContext(host)
+    context = pyblish.api.Context()
+    context.data["create_context"] = create_context
+    # Used in pyblish plugin to determine which instance to publish.
+    context.data["node_name"] = node.name()
```
iLLiCiTiT, line 227: We should also find the instance in `create_context` and make sure it is enabled.

client/ayon_core/hosts/nuke/api/utils.py
iLLiCiTiT: Could this be at least renamed to `create_publish_error_report`?

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,42 @@
+import pyblish.api
+
+
+class CollectHeadlessFarm(pyblish.api.ContextPlugin):
+    """Setup instances for headless farm submission."""
+
+    order = pyblish.api.CollectorOrder + 0.4999
```
iLLiCiTiT, line 7: Why this happens so late?

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,42 @@
...
+    def process(self, context):
+        if not context.data.get("headless_farm", False):
+            return
+
+        for instance in context:
+            if instance.data["family"] == "workfile":
+                instance.data["active"] = False
+
+                # Disable version validation.
+                instance.data.pop("latestVersion")
```
iLLiCiTiT, line 20: Does not have to be removed when instance is disabled.
```suggestion
```

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,42 @@
...
+            if node.name() != instance.context.data["node_name"]:
+                instance.data["active"] = False
+                continue
+
+            # Enable for farm publishing.
+            instance.data["farm"] = True
+
+            # Clear the families as we only want the main family, ei. no review
+            # etc.
+            instance.data["families"] = []
```
iLLiCiTiT, line 34: I would add a family here so the instance can be recognized.

Why:
- Change of `"NukeSubmitDeadline"` should not happen in this plugin, it would actually crash if you don't have deadline addon enabled, should be moved to deadline plugin instead (with instance families filtering). Why we're actually changing "settings", there is no way how to tell to not use published workfile?
- The same applies to `ExtractHeadlessFarm` where should be families filtering.

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: To be the last plugin to run.

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: Done

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,39 @@
+import pyblish.api
+
+
+class CollectHeadlessFarm(pyblish.api.ContextPlugin):
```
iLLiCiTiT, line 4: Does the headless mean that it will be published on farm from nuke? Because for that was chosen "Remote publishing" (run publishing on farm from a DCC).

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
iLLiCiTiT: But it can be in fact one of the first plugins instead. It can run after `CollectFromCreateContext` which has order `pyblish.api.CollectorOrder - 0.5`, or?

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: `headless` means publishing without the Publisher UI.

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: Done

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: We need to make sure `instance.data["families"] = ["headless_farm"]` only. Cause we do not want any `review` or other families.

iLLiCiTiT: But all the instances you're disabling here will be collecte and will go through all collectors before you actually disable them. So this can be moved to `pyblish.api.CollectorOrder - 0.49`.

iLLiCiTiT: And to be honest I don't think it is good idea to clear `"families"` this late, the instance won't be integrated to ftrack.

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
iLLiCiTiT: So why is `"farm"` set to `True`?

tokejepsen: `farm` will make the publishing process skip integration.

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: > But all the instances you're disabling here will go through all collectors before you actually disable them, my question is why to not disable them as soon as possible? This can be moved to pyblish.api.CollectorOrder - 0.49.

Have split the plugin to disable instances as early as possible now.

> And if that is intended behavior then I suggest to move it out of this plugin to another one which can be enabled/disabled, so this logic can be in fact used as "standard" publishing.

I dont get what you mean here?

jakubjezek001:
@tokejepsen can we finalize this PR. It seems that it would be good to merge after some cosmetic suggestions are applied.

client/ayon_core/modules/deadline/plugins/publish/submit_nuke_deadline.py
```diff
@@ -115,8 +115,11 @@ def process(self, instance):
         render_path = instance.data['path']
         script_path = context.data["currentFile"]
 
-        use_published_workfile = instance.data["attributeValues"].get(
-            "use_published_workfile", self.use_published_workfile
+        use_published_workfile = instance.data.get(
```
iLLiCiTiT, line 118: Why this changed. Just set `"use_published_workfile"` on `"attributeValues"` where you change it...

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,47 @@
...
+    def process(self, instance):
+        # Enable for farm publishing.
+        instance.data["farm"] = True
+
+        # Clear the families as we only want the main family, ei. no review
+        # etc.
+        instance.data["families"] = ["headless_farm"]
+
+        # Use the workfile instead of published.
+        instance.data["use_published_workfile"] = False
```
iLLiCiTiT, line 47: Related to comment https://github.com/ynput/ayon-core/pull/314/files#r1609616503
```suggestion
        attribute_values = instance.data.setdefault("attributeValues", {})
        attribute_values["use_published_workfile"] = False
```

client/ayon_core/hosts/nuke/api/utils.py
iLLiCiTiT: 👆

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
iLLiCiTiT: > I dont get what you mean here?

If clearing of `"families"` is for some reason intended behavior, then it probably should be in specific plugin that does that (later in the order).

client/ayon_core/hosts/nuke/api/lib.py
```diff
@@ -1024,6 +1024,16 @@ def script_name():
     return nuke.root().knob("name").value()
 
 
+def add_button_headless_farm_submission(node):
+    name = "headlessFarmSubmission"
+    label = "Headless Farm Submission"
+    value = "from ayon_core.hosts.nuke.api.utils import submit_headless_farm;"
+    value += "submit_headless_farm(nuke.thisNode())"
```
iLLiCiTiT, line 1031:
```suggestion
    value = (
        "from ayon_core.hosts.nuke.api.utils import submit_headless_farm;"
        "submit_headless_farm(nuke.thisNode())"
    )
```

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
```diff
@@ -0,0 +1,47 @@
+import pyblish.api
+
+
+class CollectHeadlessFarm(pyblish.api.ContextPlugin):
+    """Setup instances for headless farm submission."""
+
+    # Needs to be after CollectFromCreateContext
+    order = pyblish.api.CollectorOrder - 0.4
```
iLLiCiTiT, line 8: Can it be even earlier?
```suggestion
    order = pyblish.api.CollectorOrder - 0.49
```

client/ayon_core/hosts/nuke/plugins/create/create_write_prerender.py
```diff
@@ -46,11 +46,15 @@ def get_pre_create_attr_defs(self):
...
+        settings = self.project_settings["nuke"]["create"]
+        settings = settings["CreateWritePrerender"]["instance_attributes"]
+
         # add fpath_template
         write_data = {
             "creator": self.__class__.__name__,
             "productName": product_name,
-            "fpath_template": self.temp_rendering_path_template
+            "fpath_template": self.temp_rendering_path_template,
+            "headless_farm_submission": "headless_farm_submission" in settings
```
iLLiCiTiT, line 57: Should this be
```suggestion
            "headless_farm_submission": settings.get("headless_farm_submission", False)
```

client/ayon_core/hosts/nuke/api/utils.py
```diff
@@ -142,3 +149,86 @@ def is_headless():
...
+
+    # Ensure CreateInstance is enabled.
+    for instance in create_context.instances:
+        if node.name() != instance.transient_data["node"].name():
+            continue
+
+        instance.data["active"] = True
+
+    context = pyblish.api.Context()
+    context.data["create_context"] = create_context
```
iLLiCiTiT, line 216: If I get it correctly, the instances are disabled during publishing, so it is not necessary to disable then in create context, and create context is auto-created during publishing, so it does not have to be handled here at all.
```suggestion
    context = pyblish.api.Context()
```

iLLiCiTiT:
I have nasty question, why it is named `"headless_farm"`. It is combination of 2 words that might be misleading as I would expect it will create remote publish job (run publishing of the workfile on farm), but it just renders the output. Could we rename it to e.g. `"render_on_farm"`?

fabiaserra:
> I have nasty question, why it is named `"headless_farm"`. It is combination of 2 words that might be misleading as I would expect it will create remote publish job (run publishing of the workfile on farm), but it just renders the output. Could we rename it to e.g. `"render_on_farm"`?

This is not a nasty question at all, this is something that I raised already in the very beginning of this PR https://github.com/ynput/ayon-core/pull/314#issuecomment-2051911240 on how confusing and convoluted the implementation of this is. Sorry to play devil's advocate (again), you can ignore me if you want but please hear me out, the fact that this PR is still going and the numerous comments, validate my initial point. You are trying to fit a solution into a framework that wasn't designed for this and we will keep suffering from it and wasting more and more time unless you pivot and re-design the framework to be not only for publishing but having separate APIs. Deadline submissions are something that's already been solved a long time ago and it works in most of the DCCs that it's supported for, we don't need to reinvent the wheel and add more technical debt to the codebase. It's the exact same issue you are having in Houdini and that I have raised quite a few times, you are trying to hijack the publish system to do rendering and do just a subset of the things you can do already with vanilla Deadline implementation. 

You could argue that you'd want to run the same validation plugins that you run during publish during the render submission but then just abstract the validation parts of the framework so you can run them before submitting the render to Deadline using an abstracted API that's just for that and don't make the system even more confusing.

The argument you said for the reason to do it this way was to do it for "simplicity" and to have something ready before you have resources to do it properly... but the fact that this PR has been opened for 2 months prove that's not simple and quick!! Are we going to keep creating PRs such as this to support deadline submission on every publish type? Studios can already benefit from all the things supported by Deadline without us having to provide it. You did the same for point caches in Houdini (and I already raised this issue then https://github.com/ynput/OpenPype/pull/4903), I did it myself when I had to spend quite a bit of time to be able to support splitting extraction and rendering for render submissions in Houdini (https://github.com/ynput/OpenPype/pull/5420) - which is when I realized it was pointless to do all of that work if Deadline already had it working for that and a lot of other things it would take me quite a long time to do the same for in OP - and we are doing it here... All of these things work in Deadline, please take advantage of it!!!

client/ayon_core/hosts/nuke/plugins/publish/collect_headless_farm.py
tokejepsen: Looks like we need to use `publish_attributes` since `attributeValues` comes from `self.get_attr_values_from_data`; https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/modules/deadline/plugins/publish/submit_nuke_deadline.py#L95

client/ayon_core/hosts/nuke/plugins/create/create_write_prerender.py
tokejepsen: No, the `instance_attributes` is a list, not dict.

client/ayon_core/hosts/nuke/api/utils.py
tokejepsen: This bit of code, is ensuring the instance is active in create context. We are not disabling any instances in create context.

tokejepsen:
> Could we rename it to e.g. "render_on_farm"?

@iLLiCiTiT done

dee-ynput:
Hello Fabia :)
As we said previously, we're quite on the same page.
Your concerns are being addressed with our work on the Houdini side, and will probably trigger implementation of APIs as you, and I, recommend.
Gradual adoption of those APIs will come later on, per host most probably.

We're doing it at the pace and in the order dictated by client needs. Not only because it is how it's funded but also because it's a better strategy than changing everything at once.

Technical debt isn't solved by disrupting the system but by making it evolve...

Thank you for the passion you put in this. It helps validate our intentions. But I do hope you will come to understand that we have very different constraints as a multi studio technology than what each studio has in isolation :)

dee-ynput:
Hello Fabia :)
As we said previously, we're quite on the same page.
Your concerns are being addressed with our work on the Houdini side, and will probably trigger implementation of APIs as you, and I, recommend.
Gradual adoption of those APIs will come later on, per host most probably.

We're doing it at the pace and in the order dictated by client needs. Not only because it is how it's funded but also because it's a better strategy than changing everything at once.

Technical debt isn't solved by disrupting the system but by making it evolve...

Thank you for the passion you put in this. It helps validate our intentions. But I do hope you will come to understand that we have very different constraints as a multi studio technology than what each studio has in isolation :)

iLLiCiTiT (approved):

jakubjezek001 (approved):

Issue #133, opened 2024-02-28
Enhancement: Callbacks and groups with Publisher attributes


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

In publisher, both creator and publish plugins can define attributes that can be set to control specifics of the publishing process. Right now, all those attributes are basically standalone without having any connection to others.

We need to define relationship between individual attributes, so changing value of one can affect the whole group (or other groups, or other controls).

### How would you imagine the implementation of the feature?

Since attributes are defined in individual plugins there are few things that needs to be implemented to overcome this:

* **attribute identifier:** to be able to reference one attribute from somewhere else
* **callbacks:** ability for plugin to define custom callback to be called when some attribute state is changed (connected to event system). Those callbacks should be able to report back to UI - for example to report invalid state.
* **labeling:** show what controls belongs together, either plugin-wise, or logical group related

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Rreferences
https://github.com/ynput/ayon-workgroups/issues/4

[cuID:[AY-2420](https://app.clickup.com/t/6658547/OP-2420)]

antirotor:
So, based on our discussion with @BigRoy and @iLLiCiTiT , this is example mockup code that might be solution:

```python
import subprocess
from ayon_core.lib import EnumDef, BoolDef, TextDef, ButtonDef
from  ayon_core.pipeline.create import CreateContext
import pyblish.api


class ExamplePlugin(pyblish.api.InstancePlugin):
    """Example plugin for demonstration purposes.
    
    Attributes:
        id (str): Unique identifier for this plugin.
            It is using reverse domain name notation.
    
    """

    @classmethod
    def get_attribute_defs(cls):
        """Get the attribute definitions for this plugin.
        
        Every attribute can be uniquely identified by its id. To get it's value,
        you need also instance itself.
        """
        return [
            BoolDef(
                "groupEnabled",  # io.ayon.publish.examplePlugin.attributes.groupEnabled
                label="Group Enabled",
                default=True,
            ),
            EnumDef("groupEnum",  # io.ayon.publish.examplePlugin.attributes.groupEnum
                    label="Enum Example",
                    items=["Foo", "Bar", "Baz"],
                    default="Foo"),
            TextDef("groupText",  # io.ayon.publish.examplePlugin.attributes.groupText
                    label="Text Example",
[...]

PR #315, opened 2024-03-28
Deadline: Remove legacy `suspend_publish` attribute definition in favor of `publishJobState`

BigRoy:
## Changelog Description

Remove legacy `suspend_publish` attribute definition in favor of `publishJobState`.

The Publish Job State became exposed some time ago through a dedicated `EnumDef` replacing the 'legacy' `suspend_publish`:
image

## Additional info

n/a

## Testing notes:

1. Publishing to Deadline should be able to set publish job state and it should work:
    - [x] Houdini
    - [x] Maya (should not actually be altered)
    - [x] Fusion
    - [x] Nuke


MustafaJafar (approved):
works in Houdini

kalisp:
Isn't working in Fusion, as instance there is just regular `render`. `render.farm` is attached only in `collect_render` plugin.

LiborBatek (commented):
When tested in ```Maya 2024``` and set my ```PublishJobState``` to ```Suspended``` it just suspends the actual ```publish job``` and not the ```render job``` and immedeately start to render those...

Is that correct behaviour?? As I was expecting the job simply wont start and being suspended till user resume it...

As seen here that the render job being queing and not ```suspended``` as the publish job
Screenshot 2024-04-02 160850

BigRoy:
> When tested in `Maya 2024` and set my `PublishJobState` to `Suspended` it just suspends the actual `publish job` and not the `render job` and immedeately start to render those...
> 
> Is that correct behaviour?? As I was expecting the job simply wont start and being suspended till user resume it...

Yes, that is the correct behavior of that particular toggle. It's only about the Publish Job - not the Render Job state.

kalisp:
I think that in Nuke it has same issue as in Fusion, the toggle is not visible as `render.farm` is not there yet.

BigRoy:
Ok, so I took a look at Fusion (and a bit at reading code for Nuke). I wonder how we can actually best 'target' this plug-in so that the attribute definitions show for the instances.

Fusion now has two creators, product type "image" and product type "render". To show the attribute definitions for both of those the `families` for this plug-in must contains those product types as families. However, "image" may conflict with product types in other hosts where we may NOT want to show it, maybe?

Anyway, I've added `render` and `image` and also `prerender` which I think is relevant for Nuke? The actual processing of this plug-in is skipped anyway if `instance.data["farm"]` is not true anyway.

Implemented with https://github.com/ynput/ayon-core/pull/315/commits/03cfed2c7957538a9c72afe4e31c010cb1d606ab

Would love to know if anyone sees any issues - and/or whether it works. it worked for me in Fusion, but definitely give it a test run.
And of course, test others hosts as well that use any family `render`, `image` or `prerender`.

moonyuet:
Tested with the image instance in SP, looks good.
image

kalisp:
This is weird (and will be weird for all local `render` or `image` publishes).
image

I don't think this is correct solution, I would prefer to move assignment of `render.farm` to creation phase for Nuke and Fusion, removal `render` product type in DL submissions altogether. 

The issue there might be if artist would switch between `Local` and `Farm` publishing, I don't think we are "repainting" toggles after that (maybe we do after `Save`? That would be enough, I think.)

BigRoy:
> But it **does** create a publish job for it.

This is good - if the publish job works. The toggle doesn't decide whether it gets created or not - it only decides whether it starts "active" (which will be pending at first) or "suspended"

BigRoy:
> This is weird (and will be weird for all local `render` or `image` publishes). image
> 
> I don't think this is correct solution, I would prefer to move assignment of `render.farm` to creation phase for Nuke and Fusion, removal `render` product type in DL submissions altogether.
> 
> The issue there might be if artist would switch between `Local` and `Farm` publishing, I don't think we are "repainting" toggles after that (maybe we do after `Save`? That would be enough, I think.)

No - the attribute definitions are static, and only target the main product type - not something during collect stage or any "additional families" set by the creator. As such, unfortunately this is the current "best" behavior. The attribute always shows and does not disable when it becomes irrelevant due to another toggle on the instance.

In short, unfortunately that looks correct. That is if when setting render target to FARM that the publish job toggle actually does something. I believe @antirotor has set up some issues tracking the Publisher improvements to be able to make toggles dependent on each other, etc. - I believe it's this one: https://github.com/ynput/ayon-core/issues/133

kalisp (approved):
I tested Nuke and Fusion, seems ok (with regard of my last comment).

MustafaJafar (approved):
works in Houdini

BigRoy:
Ready to merge?

PR #317, opened 2024-03-29
Maya: ExtractGpuCache and ExtractModel (maya scene) optional state in settings

BigRoy:
## Changelog Description

Add support for changing the `enabled`, `active` and `optional` state in settings of:
- Extract GPU Cache
- Extract Model (Maya Scene)

## Additional info

n/a

## Testing notes:

1. Default values for the settings should be 'backwards compatible' (behave like before)
2. Changing settings for Extract GPU Cache should work
3. Changing settings for Extract Model (Maya Scene) should work


tokejepsen (approved):

PR #318, opened 2024-03-29
Settings: Add `usd` to product type enum for Loader Product Type Filter

BigRoy:
## Changelog Description

Add `usd` to product type enum for Loader Product Type Filter

Used in `ayon+settings://core/tools/loader/product_type_filter_profiles`

## Additional info

I didn't bump server addon version because I can imagine those many won't care about this particular missing entry - and they would then see it whenever new addon version is released anyway. 

## Testing notes:

1. `usd` should appear in the product types list for `ayon+settings://core/tools/loader/product_type_filter_profiles`


MustafaJafar (approved):
works on my side. 
I think it makes sense not to bumping ayon-core version. 
image

antirotor (approved):

PR #319, opened 2024-03-29
Substance Painter: Allow loading `usd` mesh to substance painter

BigRoy:
## Changelog Description

Substance Painter: Allow loading `usd` mesh to substance painter

## Additional info

n/a

## Testing notes:

1. Load mesh in loader should now also show (and work) for USD representations.


moonyuet (approved):
Tested in my side. Looks good.
image

PR #321, opened 2024-03-29
Maya: Ignore invalid references in `get_container_members`

BigRoy:
## Changelog Description

Do not error if there are references that maya considers invalid, instead ignore them

## Additional info

Admittedly I'm not sure how to easily reproduce this maya issue where references are 'invalid' but we usually get this in scenes where the artists just quickly wanted to "copy paste" some stuff from one scene to another - likely bringing along the reference node in an incomplete manner.

This ignores those.

## Testing notes:

1. To reproduce an 'invalid' reference see the explanation [here](https://github.com/ynput/ayon-core/pull/321#discussion_r1557419625)
2. But with this PR:
    - Loading references and updating should still work
    - Assigning looks + updating loaded looks should still work
    - Select Containers Inventory action should still work
    - Recreate animation instance Inventory action should still work


client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -1832,6 +1832,29 @@ def apply_attributes(attributes, nodes_by_id):
...
+         reference_node (str): Reference node name
+
+    Returns:
+        bool: Whether reference node is a valid reference
+
+    """
+    sel = OpenMaya.MSelectionList()
+    sel.add(reference_node)
+    depend_node = sel.getDependNode(0)
+    return OpenMaya.MFnReference(depend_node).isValidReference()
```
MustafaJafar, line 1855: I doubt this code works (I tested this function with some node as input)
it turned out that `getDependNode` expects another argument of type `MObject` 
This [link](http://evgeniyzaitsev.com/2015/11/16/get-all-extra-attributes-maya-python-api/) helped me.

```python
# This one works

ref = "boxRN1"

sel = OpenMaya.MSelectionList()
sel.add(ref)

m_node = OpenMaya.MObject()
sel.getDependNode(0, m_node)

m_refrence = OpenMaya.MFnReference(m_node)
print(m_refrence.isLoaded())
print(m_refrence.isValidReference())
```

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -1857,7 +1880,16 @@ def get_container_members(container):
         if ref.rsplit(":", 1)[-1].startswith("_UNKNOWN_REF_NODE_"):
             continue
 
-        reference_members = cmds.referenceQuery(ref, nodes=True, dagPath=True)
+        try:
+            reference_members = cmds.referenceQuery(ref,
+                                                    nodes=True,
+                                                    dagPath=True)
```
MustafaJafar, line 1886: I tried making this fail but it only gave me `RuntimeError` when using wrong flags otherwise it always return empty any thing or `None`.

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -1857,7 +1880,16 @@ def get_container_members(container):
...
+        except RuntimeError:
+            # Ignore reference nodes that are not associated with a
+            # referenced file on which `referenceQuery` command fails
```
MustafaJafar, line 1889: is it possible to have a reference node without an associated filepath ?

MustafaJafar (commented):
I couldn't test this PR so I thought about testing its individual parts. 

---
I also tried creating a reference manually then delete/move the file associated with the reference and it turned out it was considered a valid reference 🙂

BigRoy:
> Also, are there anything that check if files associated with references actually exists ? I made a reference and deleted/moved its file from my disk and file path no longer exists. does it make sense to be able to publish a layout with that reference as child ?
> 
> ```python
> import os
> 
> ref = "boxRN1"
> ref_filepath = cmds.referenceQuery(ref, filename=True)
> 
> print(os.path.exists(ref_filepath))  # False on my side because the file was deleted.
> ```

I don't think there is - but at the same time we don't really care whether the file exists or not. We just want to detect the nodes that 'according to maya' are loaded - even if the file was removed by any means in the meantime - maya would still have them loaded.

client/ayon_core/hosts/maya/api/lib.py
BigRoy: I believe it is technically possible, but this maya bug isn't exactly the same. I'll look into whether I can create a reproducable scene.

client/ayon_core/hosts/maya/api/lib.py
BigRoy: You are looking at the Maya Python API 1.0 -> `import maya.OpenMaya`
My code is for the Maya Python API 2.0 -> `import maya.api.OpenMaya`

MustafaJafar: Yes. (I have maya 2022 py3.7)
```
# Error: TypeError: file <maya console> line 5: MSelectionList_getDependNode expected 3 arguments, got 2 # 
```

BigRoy: Thanks for the error  - I'll look into it. It sounds like you're still using the older Python API of Maya.

This will fail:
```python
from maya import OpenMaya
reference_node = "robo_roboangry_modeling_v001RN"
sel = OpenMaya.MSelectionList()
sel.add(reference_node)
depend_node = sel.getDependNode(0)
print( OpenMaya.MFnReference(depend_node).isValidReference())
```

This should work:
```python
from maya.api import OpenMaya
reference_node = "robo_roboangry_modeling_v001RN"
sel = OpenMaya.MSelectionList()
sel.add(reference_node)
depend_node = sel.getDependNode(0)
print( OpenMaya.MFnReference(depend_node).isValidReference())
```

moonyuet: I tested the code above in both maya 2023.3 and maya 2024. both work with from `maya.api import OpenMaya`.
Guess we can implement the backward compatibility for maya 2022.
Like:
```
#Backward Compatibility for importing OpenMaya Library
maya_version = int(cmds.about(version=True))
if maya_version >= 2023:
    from maya.api import OpenMaya
else:
    from maya import OpenMaya
```

moonyuet (commented):
I did some simple test in maya 2023.3 and 2024 with the code by loading reference, updating reference and loading look and assigning look. Looks okay. But not sure about the more complicated tests.
But I do think we need to implement some functions for importing OpenMaya with backward compatibility so that it is also compatible with the OpenMaya API in maya 2022.

client/ayon_core/hosts/maya/api/lib.py
BigRoy: There's no need for backwards compatibility. The logic works also in those versions. The `maya.api.OpenMaya` library has already been present a long time as far as I know.

client/ayon_core/hosts/maya/api/lib.py
BigRoy: It's not about the arguments being wrong - it's about the reference node being passed along being an 'invalid' reference.

I was able to reproduce it like this.

1. Create a maya reference (e.g. load something)
2. Select the reference node in the outliner (you might need to enable Display > Reference Nodes in outliner)
3. Duplicate the node (press CTRL+D)

The duplicated reference node will not display in the outliner, but it is a valid reference node:
```python
from maya import cmds
print(cmds.ls(type="reference"))
```

This will then also print the duplicated reference.
However, now if we use that reference node in our queries like this:

```python
from maya import cmds
for reference in cmds.ls(type="reference"):
    cmds.referenceQuery(reference, nodes=True)
```

We get an error:
```
# Error: Reference node 'testRN1' is not associated with a reference file.
# #   File "<maya console>", line 4, in <module>
# # RuntimeError: Reference node 'testRN1' is not associated with a reference file.
```

This PR solves that by first checking whether the reference node is 'valid' (associated with a reference file).
I believe there are other ways to get reference nodes to this state than just duplicating it. But this makes at least an easy reproducable.

client/ayon_core/hosts/maya/api/lib.py
BigRoy: See https://github.com/ynput/ayon-core/pull/321#discussion_r1557419625

BigRoy:
> But I do think we need to implement some functions for importing OpenMaya with backward compatibility so that it is also compatible with the OpenMaya API in maya 2022.

As stated [here](https://github.com/ynput/ayon-core/pull/321#discussion_r1557253363) I'm quite confident this should work as is. Please test in Maya 2022 and provide me an error report if it does not.

LiborBatek (approved):
Did multiple tests with different actions from the ```Inventory``` and all working fine and as expected!

image

tokejepsen (changes requested):
To replicate the issue mentioned;

1. Load a `reference`, say `camera`.
2. Duplicate the reference node associated with  the loaded `camera`.
3. Update the container or switch the version.

```
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 279, in <lambda>
#     lambda: self._show_version_dialog(items))
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 695, in _show_version_dialog
#     self._update_containers(items, version)
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, in _update_containers
#     update_container(item, item_version)
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\pipeline\load\utils.py", line 532, in update_container
#     return Loader().update(container, context)
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\plugins\load\load_reference.py", line 237, in update
#     super(ReferenceLoader, self).update(container, context)
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\plugin.py", line 823, in update
#     reference_node = lib.get_reference_node(members, self.log)
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\lib.py", line 4278, in get_reference_node
#     highest = min(references,
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\lib.py", line 4279, in <lambda>
#     key=lambda x: len(get_reference_node_parents(x)))
#   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\lib.py", line 4303, in get_reference_node_parents
#     parent = cmds.referenceQuery(ref,
# RuntimeError: Reference node 'shots_cameraStatic_01_RN1' is not associated with a reference file.
```

BigRoy:
> To replicate the issue mentioned;
> 
>     1. Load a `reference`, say `camera`.
> 
>     2. Duplicate the reference node associated with  the loaded `camera`.
> 
>     3. Update the container or switch the version.
> 
> 
> ```
> # Traceback (most recent call last):
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 279, in <lambda>
> #     lambda: self._show_version_dialog(items))
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 695, in _show_version_dialog
> #     self._update_containers(items, version)
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, in _update_containers
> #     update_container(item, item_version)
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\pipeline\load\utils.py", line 532, in update_container
> #     return Loader().update(container, context)
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\plugins\load\load_reference.py", line 237, in update
> #     super(ReferenceLoader, self).update(container, context)
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\plugin.py", line 823, in update
> #     reference_node = lib.get_reference_node(members, self.log)
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\lib.py", line 4278, in get_reference_node
> #     highest = min(references,
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\lib.py", line 4279, in <lambda>
> #     key=lambda x: len(get_reference_node_parents(x)))
> #   File "C:\Users\tokejepsen\ayon-core\client\ayon_core\hosts\maya\api\lib.py", line 4303, in get_reference_node_parents
> #     parent = cmds.referenceQuery(ref,
> # RuntimeError: Reference node 'shots_cameraStatic_01_RN1' is not associated with a reference file.
> ```

Haha, nice - now we know how to reproduce it we can find the issue elsewhere as well. Included a fix with https://github.com/ynput/ayon-core/pull/321/commits/065906a5ab69274f04aefdc140dc3b9a540d0254

tokejepsen (approved):

PR #323, opened 2024-03-29
Maya: Set scene frame range - apply playback options in one call

BigRoy:
## Changelog Description

Don't set time slider values one by one, but set all at once.

## Additional info

Less maya calls is always good.

This really shouldn't break anything. :)

## Testing notes:

1. Updating scene frame range should work. (e.g. reset frame range) or by the 'fix' in the prompt on scene save if your scene is wrong frame range.

### DIFF
--- a/client/ayon_core/hosts/maya/api/lib.py
+++ b/client/ayon_core/hosts/maya/api/lib.py
@@ -2444,12 +2444,10 @@ def set_scene_fps(fps, update=True):
     cmds.currentUnit(time=unit, updateAnimation=update)
 
     # Set time slider data back to previous state
-    cmds.playbackOptions(edit=True, minTime=start_frame)
-    cmds.playbackOptions(edit=True, maxTime=end_frame)
-
-    # Set animation data
-    cmds.playbackOptions(edit=True, animationStartTime=animation_start)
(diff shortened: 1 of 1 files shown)


m-u-r-p-h-y (changes requested):
Should not we already take the values from the task, not the shot? It was true for OP but in Ayon we already have task level attributes.

Also can I ask for improving the info message we print to the script editor? We only inform about the fps change, it would be great to include all attributes we touch (frameStart, frameEnd)

image

MustafaJafar (approved):
It works on my side.

MustafaJafar:
> Should not we already take the values from the task, not the shot? It was true for OP but in Ayon we already have task level attributes.

Yeah, I traced that and found the problem which is not related to this PR.
Going to make a PR for that.

tokejepsen (changes requested):
Could you update the title and description as this only affects the time range and not the FPS?

BigRoy:
> Could you update the title and description as this only affects the time range and not the FPS?

Well spotted. Thanks!



> > Should not we already take the values from the task, not the shot? It was true for OP but in Ayon we already have task level attributes.
> 
> Yeah, I traced that and found the problem which is not related to this PR. I made a PR #338

Nice work and good thinking!

tokejepsen (approved):

PR #324, opened 2024-03-29
Maya: Optimize `get_related_sets`

BigRoy:
## Changelog Description

Optimize `get_related_sets` by opting-out earlier if no valid sets to perform less maya calls

## Additional info

This is a bit of a micro-optimization. Just happened to have this in our OpenPype codebase.

## Testing notes:

1. Maya look collecting and publishing should still work

### DIFF
--- a/client/ayon_core/hosts/maya/api/lib.py
+++ b/client/ayon_core/hosts/maya/api/lib.py
@@ -2155,20 +2155,20 @@ def get_related_sets(node):
         return sets
 
     # Exclude deformer sets (`type=2` for `maya.cmds.listSets`)
-    deformer_sets = cmds.listSets(object=node,
-                                  extendToShape=False,
-                                  type=2) or []
-    deformer_sets = set(deformer_sets)  # optimize lookup
-    sets = [s for s in sets if s not in deformer_sets]
+    exclude_sets = cmds.listSets(object=node,
+                                 extendToShape=False,
+                                 type=2) or []
+    exclude_sets = set(exclude_sets)  # optimize lookup
+
+    # Default nodes to ignore
+    exclude_sets.update({"defaultLightSet", "defaultObjectSet"})
+
+    # Filter out the sets to exclude
+    sets = [s for s in sets if s not in exclude_sets]
 
     # Ignore when the set has a specific suffix
     ignore_suffices = ("out_SET", "controls_SET", "_INST", "_CON")
     sets = [s for s in sets if not s.endswith(ignore_suffices)]
-
-    # Default nodes to ignore
-    defaults = {"defaultLightSet", "defaultObjectSet"}
-    sets = [s for s in sets if s not in defaults]
-
     if not sets:
         return sets
 
--- a/client/ayon_core/hosts/maya/api/lib.py
+++ b/client/ayon_core/hosts/maya/api/lib.py
@@ -2125,22 +2125,6 @@ def get_related_sets(node):
 
     """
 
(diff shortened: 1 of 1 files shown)


tokejepsen (approved):

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -2168,12 +2162,20 @@ def get_related_sets(node):
     sets = [s for s in sets if s not in deformer_sets]
 
     # Ignore when the set has a specific suffix
-    sets = [s for s in sets if not any(s.endswith(x) for x in ignore_suffices)]
+    ignore_suffices = ("out_SET", "controls_SET", "_INST", "_CON")
```
iLLiCiTiT, line 2165: There are 3 list comprehensions, could that be simplified to:
```python
    # Ignore when the set has a specific suffix
    ignore_suffices = ("out_SET", "controls_SET", "_INST", "_CON")
    # Default nodes to ignore
    defaults = {"defaultLightSet", "defaultObjectSet"}
    sets = [
        s
        for s in sets
        if (
            s not in deformer_sets
            and s not in defaults
            not s.endswith(ignore_suffices)
        )
]
```

iLLiCiTiT: Or you can use tuple in `endswith`?

BigRoy: I could combine them - sure. 

Note that `sets = [s for s in sets if not any(s.endswith(x) for x in ignore_suffices)]` can also be written as ` sets = [s for s in sets if not s.endswith(ignore_suffices)]`. `str.endswith` can actually take a `tuple`. 🤯

iLLiCiTiT: Cool, modified the snippet.

BigRoy: Changed the code with https://github.com/ynput/ayon-core/pull/324/commits/f3324c3863598f1946525aef02cbd8abd5444637

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -2159,21 +2151,31 @@ def get_related_sets(node):
...
+    sets = [s for s in sets if not s.endswith(ignore_suffices)]
+    if not sets:
+        return sets
 
     # Ignore viewport filter view sets (from isolate select and
     # viewports)
+    view_sets = get_isolate_view_sets()
     sets = [s for s in sets if s not in view_sets]
-    sets = [s for s in sets if s not in defaults]
```
iLLiCiTiT, line 2176: Can't tell if this is ok, @antirotor can you confirm?

BigRoy: The `defaults` filtering is still there - just higher up added into the `exclude_sets` `set`.

iLLiCiTiT: But https://github.com/ynput/ayon-core/pull/324/files#diff-d3957520925f7a55982d51ee662c9f5432aaa46f67cdff06858b2d3a9575598eR2167 reset the `sets` variable, or?

BigRoy: Note the `sets` but the `view_sets`

All the logic is just filtering the `sets` to exclude whatever we don't want.
Or maybe I'm just misreading what you might be pointing out to be a bug - I just don't see it :D

PR #325, opened 2024-03-29
Maya: Follow up for prompt on saving to another context

BigRoy:
## Changelog Description

Fix for implementation of #261

- Remove logic that would always update before save/open anyway (and as such the dialog itself didn't make any sense)
- Also update `task` on the instances instead of only `folderPath`

## Additional info

This also fixes that before opening a scene in another context that maya was actually resetting time range, fps, etc. always even though it would be opening a file anyway - optimizing opening workfiles in other contexts because of that.

## Testing notes:

1. Save into a different folder or task
2. When picking no options (or cancelling) on the prompt - nothing should update
3. When clicking ok with certain options, only those options should update.


MustafaJafar (approved):
works on my side.

client/ayon_core/hosts/maya/api/pipeline.py
```diff
@@ -653,10 +653,6 @@ def on_task_changed():
             "Can't set project for new context because path does not exist: {}"
         ).format(workdir))
 
-    with lib.suspended_refresh():
```
iLLiCiTiT, line 656: I think only @antirotor can tell if this is ok.

BigRoy: The logic removed here was basically replaced by an artist facing prompt on file save to another context in the PR here https://github.com/ynput/ayon-core/pull/261

PR #327, opened 2024-03-29
Workfile Templates tweaks: plugin discovery, event system, maya fixes

BigRoy:
## Changelog Description

- Implement registering and discovering of Placeholder Plugins dynamically
- Implement event system in workfile templates
- Maya: Implement fixes for outliner order of loaded placeholders (always place after placeholder)
- Maya: Allow storing more complex data, e.g. `EnumDef` with `multiselection=True`
- Maya: Implement Assign Look placeholder
- Maya: Implement Script placeholder

## Additional info

AYON equivalent to https://github.com/ynput/OpenPype/pull/5830

However, with some things removed which I'll turn into separate PRs
But added refactoring Nuke and After Effects to also use the dynamic plug-in discovery

#### After Effects

Make sure workfile templates in still behave as intended (basically existing templates should continue to work)

1. [ ] Create placeholders
2. [ ] Build workfile templates (from new ones and existing ones)

#### Nuke

Make sure workfile templates in still behave as intended (basically existing templates should continue to work)

1. [ ] Create placeholders
2. [ ] Build workfile templates (from new ones and existing ones)

#### Maya

Make sure workfile templates in still behave as intended (basically existing templates should continue to work)

1. [ ] Create placeholders
2. [ ] Build workfile templates (from new ones and existing ones)
3. [ ] The Run Python Script Placeholder in Maya should work (including 'ordering' it with the `order` attribute)
4. [ ] The Assign Look Placeholder in Maya should work - any contents inside the placeholder should get shaders assigned, even if e,g. the loaded content is itself loaded from a placeholder.
5. [ ] Ordering in Maya outliner should be preserved for placeholders, also when "keep placeholders" is enabled.


BigRoy:
I have a feeling that [this](https://github.com/ynput/ayon-core/blob/05cbb887b6005a358e92ae72fc37425986f75483/client/ayon_core/pipeline/workfile/workfile_template_builder.py#L1612-L1616) might also need fixing/refactoring to AYON core.

However, on just building a template I didn't hit a direct error or crash. Any idea what to look out for @tokejepsen @iLLiCiTiT?
And maybe, what it should be refactored to?

No idea what `_reduce_last_version_repre_entities` is supposed to mean or what it's supposed to do.

client/ayon_core/hosts/aftereffects/api/workfile_template_builder.py
```diff
@@ -51,6 +47,8 @@ def import_template(self, path):
 class AEPlaceholderPlugin(PlaceholderPlugin):
     """Contains generic methods for all PlaceholderPlugins."""
 
+    item_type = PlaceholderItem
```
iLLiCiTiT, line 50: Please use method instead of attribute.

```python
@abstractmethod
def _create_placeholder_item(self, item_data):
    pass
```

And then implement it in both plugins...
```python
def _create_placeholder_item(self, item_data):
    return CreatePlaceholderItem(
        scene_identifier=item["uuid"],
        data=item["data"],
        plugin=self
    )
```

iLLiCiTiT:
Can it be split into smaller PRs? This now requires heavy testing of 4 people. If you would split it, then it may happen. The core should also be backwards compatible.

I can create some base PRs based on this if you would like.

BigRoy:
> I can create some base PRs based on this if you would like.

That would of course be greatly appreciated - but don't spend too much time on it if you already have a big to do list. I can look into it - likely just not today but maybe later this week or next week. (I have a bit of a short week this week)

tokejepsen (changes requested):
Yeah, lets split this into smaller PRs.

BigRoy:
> Yeah, lets split this into smaller PRs.

I've started the work - I'll continue the maya part of things after some of them are merged because those are easier for me to convert over as soon as other things are implemented. In the meantime, I'll close this PR in favor of the new ones and the upcoming ones.

These are the only files I still need to 'touch' to implement a fix in that upcoming PR:
image

All other changes of this PR should already be covered by one of the currently existing PRs.

PR #295, opened 2024-03-27
Houdini: Implement USD workflow with global asset/shot contributions plug-in

BigRoy:
## Changelog Description

- Implement USD publishing with asset and shot contributions through the publisher - for Houdini.
- Implement a USD Render logic which will submit to Deadline with Husk Standalone (for which you'll need a [Husk Standalone Deadline Plug-in](https://github.com/BigRoy/HuskStandaloneSubmitter))
- Also implements a USD look publishing where textures are being copied along as resource files and are anchored relative to the USD file (similar to maya look publishing)

## Additional info

Based on https://github.com/ynput/OpenPype/pull/5925 but excluding the Maya logic (that will be a separate PR) - the description and notes in that PR still apply. That PR description also has some "early WIP demo videos" which may help you along.

This is currently a draft refactor from our USD Project's OpenPype logic to AYON:
- [x] Published usd layer from Houdini with asset contribution as variant
- [x] Published usd layer from Houdini with asset contribution without variant set
- [x] Published usd layer from Houdini as shot contribution
- [x] Submitted to Deadline Husk Standalone renders (with split export jobs)
- [x] Publish a look with textures

Note that you currently need to make sure you set your default primitive path on the USD ROPs for correctly outputting the USD files and of course making sure your prims adhere to that hierarchy, e.g. `char_hero/geo/my_cube`.

# Tutorial

_Click these sections to expand/collapse_

---

<details><summary>Background details + example videos</summary>

- [In the original OpenPype PR description it contained quite some 'test' videos](https://github.com/ynput/OpenPype/pull/5925#issue-1995346181)
- [At FMX 2024 an OpenUSD presentation was given that showed usage of what is basically the functionality of this PR](https://www.youtube.com/watch?v=1KqrIRCi_EQ).

</details>

---

<details><summary>Your first USD publish with AYON</summary>

## Creating an asset

The producer has his 🔫 pointing at you and needs the **asset** _now!_

When we are referring to the USD asset, we mean the asset with both the model's geometry and the look's materials and textures. The USD structure (simplified) then becomes:

```
asset.usd
  - look.usd
  - model.usd
```
Where the look is applied over the model inside the asset.

### Creating the model

_From a Maya, Houdini or Blender perspective_

We will want to create our geometry's contribution to the asset.
For the asset structure it's critical we work within the asset's hierarchy so that USD referencing works. (_consider that USD mumbo jumbo for now_)
As such, we'll always work within a root group with the current folder (asset)'s name.
[...]

Issue #238, opened 2024-03-25
Houdini: Better support for single frame outputs


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

A lot of extractor seems to rely on `instance.data["frames"]` to be a list of files:

- https://github.com/ynput/ayon-core/blob/ee6e42ec6c090eb2ded6302b6cb8e09ef2d55dc0/client/ayon_core/hosts/houdini/plugins/publish/extract_bgeo.py#L37-L42
- https://github.com/ynput/ayon-core/blob/ee6e42ec6c090eb2ded6302b6cb8e09ef2d55dc0/client/ayon_core/hosts/houdini/plugins/publish/extract_composite.py#L32-L34
- https://github.com/ynput/ayon-core/blob/ee6e42ec6c090eb2ded6302b6cb8e09ef2d55dc0/client/ayon_core/hosts/houdini/plugins/publish/extract_ass.py#L37-L49
- and more.

However [it's a single file if the instance has no frame range data](https://github.com/ynput/ayon-core/blob/ee6e42ec6c090eb2ded6302b6cb8e09ef2d55dc0/client/ayon_core/hosts/houdini/plugins/publish/collect_frames.py#L43-L59).

If it has frame range data but a it's a single frame then I believe these extractors end up setting a single file list like: `representation["files"] = ["single_file.1001.exr"]` - which the integrator does not tend to like.

For example, a single frame COP2 image publish errors with the following report:

[publish-report-240325-17-47.json](https://github.com/ynput/ayon-core/files/14747011/publish-report-240325-17-47.json)

The integrator wants single files to be a single filename instead of a list.

### Expected Behavior:

There should be simple consistency across the API, so I suppose `instance.data["frames"]` should at best always be a list. We should then however just restructure the `representation["files"]` to be a single filename instead of the list of files.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Publish a single frame of a sequenced product type, e.g. image sequence (COP2), bgeo, vdb

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
Traceback (most recent call last):
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2403061937_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
[...]

PR #328, opened 2024-03-29
Houdini local render support

MustafaJafar:
## Changelog Description
This PR adds support for local rendering in Houdini and marking render as reviewable.

## Additional Info: How it works 
The logic of this PR is based on re-using the collected data for each render instance to perform local rendering. 
this happens by remapping the collected data `expectedFiles` where I use them to create a dedicated instance for each AOV in the `expectedFiles` while keeping the original instance empty. 
I couldn't remove the original instance while this process because it holds the debug data.

- CollectFarmInstances 
- Collect{..}RenderProducts (mantra, karma, redshift, arnold, vray) 
   Compute `expectedFiles` which is a dictionary of AOVs and their files.
- CollectLocalRenderInstances
   if farm disabled append `render.local.hou` to families
   create instance for each AOV in expected files and add essential data and its representation data
- ExtractLocalRender
   if farm disabled, if `skip_render` disabled, render the node
- Deadline stuff should work only if farm is enabled.

## Testing notes:
1. Open Houdini 
2. Create Render instance  while disabling `Submitting to Farm`
3. Publish (it should work)
4. Create Render instance while enabling `Submitting to Farm`
5. Publish (it should work as usual)


client/ayon_core/hosts/houdini/plugins/create/create_mantra_rop.py
```diff
@@ -101,5 +95,24 @@ def get_pre_create_attr_defs(self):
                     label="Override Camera Resolution",
                     tooltip="Override the current camera "
                             "resolution, recommended for IPR.",
-                    default=False)
+                    default=False),
+            UISeparatorDef(key="1"),
+            UILabelDef(label="Farm Render Options:"),
+            BoolDef("farm",
+                    label="Submitting to Farm",
+                    default=True),
```
BigRoy, line 103: Can we make this an `EnumDef` with options that say:
```
Local Render
Farm Render
```

That way we can expand it later to also include:
```
Local Render (Use Existing Frames)
Farm Render (Use Existing Frames)
```
Which is something I blelieve Nuke and Fusion also do.

client/ayon_core/hosts/houdini/plugins/publish/validate_split_render_is_disabled.py
```diff
@@ -0,0 +1,68 @@
...
+            return
+
+
+        split_enabled = creator_attribute["split_render"]
+        if split_enabled:
+            invalid.append(rop_node)
+            cls.log.error(
+                "Split Export must be disabled in local render instances."
+            )
+
```
BigRoy, line 56: I think we shouldn't use a validator for this. Instead - if local render is set we basically have full control over the "render" logic.We should just use a context manager around the `render_rop` call to ensure the settings are as we need to perform the actual render.

client/ayon_core/modules/deadline/plugins/publish/collect_pools.py
```diff
@@ -41,11 +41,7 @@ class CollectDeadlinePools(pyblish.api.InstancePlugin,
                 "renderlayer",
                 "maxrender",
                 "usdrender",
-                "redshift_rop",
-                "arnold_rop",
-                "mantra_rop",
-                "karma_rop",
-                "vray_rop",
+                "render.farm.hou",
```
BigRoy, line 44: I'm pretty sure that by doing this you are losing the Pool selection attribute defs. As such, this seems wrong?

client/ayon_core/hosts/houdini/plugins/create/create_mantra_rop.py
moonyuet: > Can we make this an `EnumDef` with options that say:
> 
> ```
> Local Render
> Farm Render
> ```
> 
> That way we can expand it later to also include:
> 
> ```
> Use Existing Frames (local)
> Use Existing Frames (on farm)
> ```
> 
> Which is something I blelieve Nuke and Fusion also do. Would love to hear other options though.
> 
> That way the `skip_render` toggle is basically part of this combox

we can take references from what the local/farm rendering setting in Nuke hosts. 
1. Local Render
2. Local Render(with existing frames)
3. Farm Render
It is more user-friendly especially when both local render with or without existing frames as the users can always manually choose whether they use the frames which they manually rendered.or not.

client/ayon_core/hosts/houdini/plugins/publish/validate_split_render_is_disabled.py
moonyuet: Maybe we can do split render in somewhere after choosing the local render settings(as it is just for the farm submission). Like right before farm submission and after user's decision on whether they use local or farm rendering.

client/ayon_core/modules/deadline/plugins/publish/collect_pools.py
MustafaJafar: correct! 
Since you have mentioned it. 
This problem also occurs with cache families that depend on `"publish.hou"` and it doesn't happen with `collect_pools` only but also `submit_houdini_render_deadline` and `submit_houdini_cache_deadline` 

The thing is How to add the families to these plugins without making them run `process`.. 
e.g `submit_houdini_render_deadline` will submit a job to deadline where I only want the `get_attribute_defs`. 

Should we move these `get_attribute_defs` to a dedicated collector ? 
It can be another issue (refactor Deadline submission) which is an issue that _Hopfully_ I'm going/be able to tackle soon 

| family add to bools | removed from bools |
|--|--|
| image | image  |

client/ayon_core/hosts/houdini/plugins/create/create_mantra_rop.py
MustafaJafar: implemented in https://github.com/ynput/ayon-core/pull/328/commits/bcb1c2a0ba3b49c43a08507ad52a9fdce07037ef

client/ayon_core/hosts/houdini/plugins/publish/validate_split_render_is_disabled.py
MustafaJafar: Implemented in https://github.com/ynput/ayon-core/pull/328/commits/bca10c7c7d0d4d8457a2ba25730dd37ebd5bbff9

client/ayon_core/modules/deadline/plugins/publish/collect_pools.py
BigRoy: This targeting of families to show specific attributes - I believe @iLLiCiTiT had some opinions on that. So probably best to check with him how he'd recommend approaching this.

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,96 @@
...
+    # this plugin runs after Collect Render Products
+    order = pyblish.api.CollectorOrder + 0.12
+    families = ["mantra_rop",
+                "karma_rop",
+                "redshift_rop",
+                "arnold_rop",
+                "vray_rop"]
+
+    hosts = ["houdini"]
+    targets = ["local", "remote"]
```
moonyuet, line 20:
```suggestion

```

client/ayon_core/hosts/houdini/plugins/publish/extract_local_render.py
```diff
@@ -0,0 +1,50 @@
...
+        rop_node = hou.node(instance.data.get("instance_node"))
+
+        if product_type == "arnold_rop":
+            rop_node.setParms({"ar_ass_export_enable": 0})
+        elif product_type == "mantra_rop":
+            rop_node.setParms({"soho_outputmode": 0})
+        elif product_type == "redshift_rop":
+            rop_node.setParms({"RS_archive_enable": 0})
+        elif product_type == "vray_rop":
+            rop_node.setParms({"render_export_mode": "1"})
```
moonyuet, line 44:
```suggestion
        if product_type == "arnold_rop":
            rop_node.setParms({"ar_ass_export_enable": 0})
        elif product_type == "mantra_rop":
            rop_node.setParms({"soho_outputmode": 0})
        elif product_type == "redshift_rop":
            rop_node.setParms({"RS_archive_enable": 0})
        elif product_type == "vray_rop":
            rop_node.setParms({"render_export_mode": "1"})
```
Maybe it's not a good idea to use collector(s) and extractors which is changing the parameters. 
Maybe we can set up these parameters in creators instead?
What do you think? @BigRoy

BigRoy: Being able to set this on create itself - does help. But it'd then also require 'updating' of the setting if the artist toggled there preference on the instance (if it's an instance attribute definition).

moonyuet: > Being able to set this on create itself - does help. But it'd then also require 'updating' of the setting if the artist toggled there preference on the instance (if it's an instance attribute definition) in publisher.
Make sense. maybe this can be resolved by the universal publishing node. Just leave it here then first.

MustafaJafar: Should I mark this as resolved ?

client/ayon_core/modules/deadline/plugins/publish/collect_pools.py
MustafaJafar: I came up with a solution. I removed the intermediate family `render.farm.hou` it was unnecessary anyways.   
Now, it should work normally.
image

client/ayon_core/hosts/houdini/plugins/publish/extract_local_render.py
```diff
@@ -0,0 +1,63 @@
...
+
+    order = pyblish.api.ExtractorOrder
+    label = "Extract Local Render"
+    hosts = ["houdini"]
+    families = ["mantra_rop",
+                "karma_rop",
+                "redshift_rop",
+                "arnold_rop",
+                "vray_rop"]
+    targets = ["local", "remote"]
```
moonyuet, line 19:
```suggestion

```

client/ayon_core/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
```diff
@@ -71,11 +71,12 @@ class HoudiniSubmitDeadline(
     order = pyblish.api.IntegratorOrder
     hosts = ["houdini"]
     families = ["usdrender",
```
moonyuet, line 73: Guess we need to pack the rop(s) into one published family before the farm submission? It seems to me that it is still submitting to the render farm even if the user chooses the local render.

client/ayon_core/modules/deadline/plugins/publish/submit_publish_job.py
```diff
@@ -92,9 +92,11 @@ class ProcessSubmittedJobOnFarm(pyblish.api.InstancePlugin,
                 "prerender.farm", "prerender.frames_farm",
                 "renderlayer", "imagesequence",
                 "vrayscene", "maxrender",
-                "arnold_rop", "mantra_rop",
-                "karma_rop", "vray_rop",
-                "redshift_rop"]
+                "mantra_rop",
```
moonyuet, line 95: See: https://github.com/ynput/ayon-core/pull/328/files#r1551669930

client/ayon_core/hosts/houdini/plugins/publish/collect_farm_instances.py
```diff
@@ -0,0 +1,54 @@
...
+        # Collect Render Target
+        if creator_attribute.get("render_target") not in {
+            "farm_split", "farm"
+        }:
+            instance.data["farm"] = False
+            self.log.debug("Render on farm is disabled. "
+                           "Skipping farm collecting.")
+            return
+
+        instance.data["farm"] = True
```
moonyuet, line 54: By doing this
```suggestion
        instance.data["farm"] = True
        instance.data["families"] += "render.hou"
```
You can remove all rop-related families and add `render.hou` into families type

BigRoy: Actually - this won't work. Because then attribute definitions on the `CollectPools` plug-ins etc won't show for this product type since the family would not exist on the instance on creation- but only during publishing.

client/ayon_core/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
BigRoy: If that's the case - that's just because `instance.data["farm"]` apparently does not get set? Since this plug-in should get skipped by the `instance.data["farm"]` check in `process()`

client/ayon_core/hosts/houdini/plugins/publish/collect_farm_instances.py
moonyuet: I haven't thought of this scanerio. Just ignore me then.

client/ayon_core/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
moonyuet: > If that's the case - that's just because `instance.data["farm"]` apparently does not get set? Since this plug-in should get skipped by the `instance.data["farm"]` check in `process()`

This can be the way to solve this. I believe it is added into the submit_publish_job if I remember correctly.

MustafaJafar: > Guess we need to pack the rop(s)-related families into one published family before the farm submission?

This was done in https://github.com/ynput/ayon-core/pull/328/commits/941e80dd952864adcd4ba4386d5883434f5cd338 but I found out that we lost the attribute_defs in the publisher as BigRoy mentioned here https://github.com/ynput/ayon-core/pull/328#discussion_r1544893280 
So, I had to remove it in order to get the attribute_defs back. 

Also, the `instance.data["farm"]` check was added to `process()`. 
Hopefully, This skips the job submission.

moonyuet (changes requested):
1. The local renders error out in integrate.py
2. This doesn't pass to extract burnin and thumbnail to make mp4 and thumbnail respectively
image

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,95 @@
...
+                # 'label': label,
+                "task": instance.data["task"],
+                "folderPath": instance.data["folderPath"],
+                "frameStart": instance.data["frameStartHandle"],
+                "frameEnd": instance.data["frameEndHandle"],
+                "productType": product_type,
+                "productName": product_name,
+                "productGroup": product_group,
+                "tags": [],
+                "families": ["render.local.hou"],
```
moonyuet, line 75: Maybe you can try to add review as being parts of families to see if it works as expected.
```suggestion
                "families": ["render.local.hou", "review"],
```

MustafaJafar:
> 1. The local renders error out in integrate.py

oh, it's the single frame issue mentioned here https://github.com/ynput/ayon-core/issues/238

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
moonyuet: > So, After a quick investigation. Review product logic in Houdini was built assuming we are only publishing reviews from review instances. (opengl node)

I recommend you test on the local publishing in Nuke Host or study the logic behind the submit publish job. They are helpful for you to work on this PR.

MustafaJafar: I'm going to dig more. I was just mentioning that the logic is interlaced. 
All review plugins in Houdini are not generic as they expect opengl node. 
and in order to make it work. I needed to add to them something like this
```python 
ropnode = hou.node(instance.data.get("instance_node"))
        if ropnode.type().name() != "opengl":
            self.log.debug("Skipping OpenGl extraction. Rop node {} "
                           "is not of OpenGl type.".format(ropnode.path()))
            return
```

MustafaJafar: @moonyuet  I'm not sure about my approach to implement it. so, I created this PR.
https://github.com/ynput/ayon-core/pull/381

moonyuet: > @moonyuet I'm not sure about my approach to implement it. so, I created this PR. #381

it works nicely with local rendering and I already merged it but I spotted another issue with the sync of the settings from creator attr to instance attr. I will inform with more detail later.

moonyuet (changes requested):
The farm options in the creator tab doesn't align with the value of those in the publish tab. You can try to consider implementing the data through the creator's attributes.
You may find the pesudo codes below helpful:
```python
creator_attributes = instance_data.setdefault(
    "creator_attributes", dict())
for key in ["farm"]:
    if key in pre_create_data:
        creator_attributes[key] = pre_create_data[key]

```
In your collector which decides whether farm or local:
```python
creator_attrs = instance.data["creator_attributes"]
data = {"farm": creator_attrs["farm"]}
instance.data.update(data)
```

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,102 @@
...
+            aov_instance.data.update({
+                # 'label': label,
+                "task": instance.data["task"],
+                "folderPath": instance.data["folderPath"],
+                "frameStart": instance.data["frameStartHandle"],
+                "frameEnd": instance.data["frameEndHandle"],
+                "productType": product_type,
+                "productName": product_name,
+                "productGroup": product_group,
+                "families": ["render.local.hou", "review"],
```
iLLiCiTiT, line 81: Missing `"family"`.

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,102 @@
...
+                        "files": aov_filenames,
+                        "frameStart": instance.data["frameStartHandle"],
+                        "frameEnd": instance.data["frameEndHandle"]
+                    }
+                ]
+            })
+
+        # Remove Mantra instance
+        # I can't remove it here as I still need it to trigger the render.
+        # context.remove(instance)
```
iLLiCiTiT, line 98: I guess that're lated to https://github.com/ynput/ayon-core/pull/328/files#r1557279839 ?
```suggestion
        # Remove Mantra instance
        context.remove(instance)
```

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,102 @@
...
+            self.log.debug("Render on farm is enabled. "
+                           "Skipping local render collecting.")
+            return
+
+        # Create Instance for each AOV.
+        context = instance.context
+        expectedFiles = next(iter(instance.data["expectedFiles"]), {})
+
+        product_type = "render"  # is always render
+        product_group = "render{Task}{productName}".format(
```
iLLiCiTiT, line 34: I guess we should use `get_product_name` for this and for product name of each aov.

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class CollectLocalRenderInstances(pyblish.api.InstancePlugin):
+    """Collect instances for local render.
+
+    Agnostic Local Render Collector.
+    """
+
+    # this plugin runs after Collect Render Products
+    order = pyblish.api.CollectorOrder + 0.12
```
iLLiCiTiT, line 12: Why so late? From the code, this can run `pyblish.api.CollectorOrder - 0.49999`

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
MustafaJafar: what should be the value of `"family"` ?  same as  `productType` ?

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
MustafaJafar: I should have change that comment maybe to 
```
# Remove Original Render instance.
```

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
MustafaJafar: I'm re-using collected data `expectedFiles` and they are being collected at `pyblish.api.CollectorOrder + 0.11`

some examples: 

https://github.com/ynput/ayon-core/blob/8f143e446db78783314db8239c5315116bb9c4c5/client/ayon_core/hosts/houdini/plugins/publish/collect_mantra_rop.py#L29

https://github.com/ynput/ayon-core/blob/8f143e446db78783314db8239c5315116bb9c4c5/client/ayon_core/hosts/houdini/plugins/publish/collect_karma_rop.py#L29

https://github.com/ynput/ayon-core/blob/8f143e446db78783314db8239c5315116bb9c4c5/client/ayon_core/hosts/houdini/plugins/publish/collect_redshift_rop.py#L29

moonyuet:
@MustafaJafar, this is the issue I want to mention. 
https://github.com/ynput/ayon-core/assets/64118225/74162764-35fa-4086-b61b-3abf4bf0ff43

MustafaJafar:
> > > @MustafaJafar, this is the issue I want to mention.
> > 
> > 
> > oh, I got it now! Thanks
> 
> You can find this PR helpful for the setting sync: #385

Many thanks for your help. I slightly edited it as it's helpful to transfer all `pre_create_data` to `creator_attributes`

BigRoy:
> Many thanks for your help. I slightly edited it as it's helpful to transfer all `pre_create_data` to `creator_attributes`

Not entirely true. For example, `use_selection` really is _just_ pre create data not important after the creation. So you may now be "bloating" your data with redundant data that has no real place among creator attributes.

client/ayon_core/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
```diff
@@ -265,20 +266,23 @@ def get_plugin_info(self, job_type=None):
...
                     SceneFile=instance.data["ifdFile"],
                     Version=hou_major_minor,
                 )
-            elif product_type == "vray_rop":
+            elif node_type == "vray_renderer":
                 plugin_info = VrayRenderPluginInfo(
                     InputFilename=instance.data["ifdFile"],
                 )
-            elif product_type == "redshift_rop":
+            elif node_type == "Redshift_ROP":
```
BigRoy, line 285: What's the reason we prefer the node type over the product type now?

BigRoy (commented):
Looking at the amount of duplicated code between the creator I believe it's time for these RenderCreators in Houdini to inherit from a parent `HoudiniRenderCreator` class to avoid duplicating this much logic.

client/ayon_core/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
MustafaJafar: Nothing. 
I should have reverted it after removing the intermediate family. 
https://github.com/ynput/ayon-core/pull/328/commits/4e6bd3d3361708fa7cf84ef69e0e5715ed451df2

client/ayon_core/hosts/houdini/plugins/publish/validate_review_colorspace.py
```diff
@@ -33,6 +33,13 @@ class ValidateReviewColorspace(pyblish.api.InstancePlugin,
 
     def process(self, instance):
 
+        rop_node = hou.node(instance.data["instance_node"])
+
+        if rop_node.type().name() != "opengl":
+            self.log.debug("Skipping Validation. Rop node {} "
+                           "is not an OpenGl node.".format(rop_node.path()))
+            return
```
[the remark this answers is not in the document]
MustafaJafar: Your comment is so true. 
The thing I'm confused about how to add mark an instance as reviewable. 
as far as I know, I need to 
1. append `"review"` to `instance.data["families"]`
2. append `"review"` to `tags` in the `representation`

And, this validator is triggered whenever appending `"review"` to families. which is as you said: *this validator is now just running over the wrong instances*

> Also, when searching the code base, I found that sometimes we `instance.data.update({"review" : True})` and I'm not sure how this works.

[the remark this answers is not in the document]
BigRoy: Note that:

```python
instance.data.update({"review" : True})
```

Is the same as:
```
instance.data["review"] = True
```

Because `instance.data` is just a `dict` and thus:
```python
new = {"review": True}
instance.data.update(new)
```
Is basically the same.

---



> as far as I know, I need to
> 
>     1. append `"review"` to `instance.data["families"]`
> 
>     2. append `"review"` to `tags` in the `representation`

How "marking something for review" is confusing has been mentioned quite a few times before. I personally don't know all too much about it and potentially @jakubjezek001 or @antirotor are more knowledgeable about the reasoning or exact steps needed to "mark for review" which I think for you just means to also get burnins and previews generated, yes? I know @iLLiCiTiT did some work on the burning/review logic so might be able to chime in as well.

client/ayon_core/hosts/houdini/plugins/publish/extract_local_render.py
```diff
@@ -0,0 +1,62 @@
...
+            rop_node.setParms({"RS_archive_enable": 0})
+        elif product_type == "vray_rop":
+            rop_node.setParms({"render_export_mode": "1"})
+
+        ropnode = hou.node(instance.data.get("instance_node"))
+        render_rop(ropnode)
+
+        # Check missing frames.
+        # Frames won't exist if user cancels the render.
+        expected_files = next(iter(instance.data["expectedFiles"]), {})
```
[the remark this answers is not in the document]
MustafaJafar: It's a list of one dict. 
I'm using this line to return the first item (the only dict in that list) or return and empty dict if the list is empty. 
example 1:
```python
[
    {
        "": [
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.1001.exr",
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.1002.exr"
        ]
    }
]
```
Example 2: 
```python
[
    {
        "beauty": [
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.beauty.1001.exr",
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.beauty.1002.exr"
        ],
        "Z": [
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.Z.1001.exr",
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.Z.1002.exr"
        ],
        "N": [
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.N.1001.exr",
            "//storage/work/ayon_projects/Robo/Assets/Character/warrior_robo/work/cfx/pyblish/renders/redshift_ropMain/redshift_ropMain.N.1002.exr"
        ]
    }
]
```

client/ayon_core/hosts/houdini/plugins/publish/extract_local_render.py
```diff
@@ -0,0 +1,62 @@
...
+        expected_files =  sum(expected_files.values(), [])
```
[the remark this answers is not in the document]
MustafaJafar: I tried to pick the shortest solution to keep minimize the code size. writing so much code can distract me from seeing the big picture of what this plugin should do. 

Maybe we can add a TODO and fix it after making the plugin work correctly ? 
```suggestion
        # TODO: enhance the readability.
        expected_files = sum(expected_files.values(), [])
```

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -0,0 +1,102 @@
...
+                ]
+            })
+
+        # Remove Mantra instance
+        # I can't remove it here as I still need it to trigger the render.
+        # context.remove(instance)
+
+    @staticmethod
+    def _capitalize(word):
+        return word[:1].upper() + word[1:]
```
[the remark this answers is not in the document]
MustafaJafar: yeah, but I think I'm going to remove it anyways. it's no longer needed when using 
https://github.com/ynput/ayon-core/pull/328#discussion_r1557281684

client/ayon_core/hosts/houdini/plugins/publish/collect_review_data.py
```diff
@@ -55,3 +57,17 @@ def process(self, instance):
...
+        }:
+            return ropnode.parm("camera").eval()
+
+        elif ropnode.type().name() == "Redshift_ROP":
+            return ropnode.parm("RS_renderCamera").eval()
+
+        elif ropnode.type().name() == "vray_renderer":
+            return ropnode.parm("render_camera").eval()
+
+        return ""
```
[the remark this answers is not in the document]
MustafaJafar: I'm going to do so as it turns out that `hou.node(None)` is fine.

client/ayon_core/hosts/houdini/plugins/publish/collect_farm_instances.py
```diff
@@ -0,0 +1,54 @@
...
+                rop_node.setParms({"render_export_mode": "2"})
+        else:
+            if product_type == "arnold_rop":
+                rop_node.setParms({"ar_ass_export_enable": 0})
+            elif product_type == "mantra_rop":
+                rop_node.setParms({"soho_outputmode": 0})
+            elif product_type == "redshift_rop":
+                rop_node.setParms({"RS_archive_enable": 0})
+            elif product_type == "vray_rop":
+                rop_node.setParms({"render_export_mode": "1"})
```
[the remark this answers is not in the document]
MustafaJafar: Collector responsibility is to collect data and not to alter the instance node state, correct ? 

But, Let's think of this riddle. 
I need to update the node state/parameters to align with user preference before the validation phase.
and I only have 
- collect
- validate
- extract 

ehmmm, what about `save changes` in the publisher UI? 
How could I use it to update my instance state/parameters?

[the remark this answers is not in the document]
BigRoy: > ehmmm, what about save changes in the publisher UI?
> How could I use it to update my instance state/parameters?

That would be up to the Creator - in e.g. `Creator.update_instances` or alike.

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
MustafaJafar: I've done it in this commit  https://github.com/ynput/ayon-core/pull/328/commits/90e2c1f1b5235e0d6fd408588a74f69f9cb7ac51 
let me know what do you think.

client/ayon_core/modules/deadline/plugins/publish/submit_houdini_render_deadline.py
MustafaJafar: done in https://github.com/ynput/ayon-core/pull/328/commits/75879f54be31e293cad18a1c4a8a60005d62ba70

MustafaJafar:
> > Many thanks for your help. I slightly edited it as it's helpful to transfer all `pre_create_data` to `creator_attributes`
> 
> Not entirely true. For example, `use_selection` really is _just_ pre create data not important after the creation. So you may now be "bloating" your data with redundant data that has no real place among creator attributes.

Thanks for pointing that out. I missed it!

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
iLLiCiTiT: But those are not creating new instances.

MustafaJafar: yes, but I'm reusing their collected data to create my new instances.

client/ayon_core/hosts/houdini/plugins/publish/validate_review_colorspace.py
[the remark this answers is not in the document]
antirotor: `instance.data["review"] = True` is saying: when this instance, or derived instance is processed on farm, it will add "review" **family** so it will be processed by *Extract Review*. Mostly Deadline based logic that we'll be happy to refactor.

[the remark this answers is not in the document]
MustafaJafar: Should I then add the render product types by default to this plugin + add something like ?
```python
if not instance.data["review"]:
    # skipping
    return
```
https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/houdini/plugins/publish/collect_review_data.py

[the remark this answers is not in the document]
antirotor: to make it more complicated, you need to take care wheter instance is actually reviewable -> not all renders are - for example if you have instance, that is producing normal, point position and depth renders, there is nothing you can use for reviews.

client/ayon_core/hosts/houdini/plugins/publish/extract_local_render.py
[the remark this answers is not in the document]
MustafaJafar: I think we can use
```python
        expected_files = instance.data["expectedFiles"][0]
```
which is similar to https://github.com/ynput/ayon-core/blob/0d3fb5e1bd2addc0eb374475255974ef34be1ecb/client/ayon_core/modules/deadline/plugins/publish/submit_publish_job.py#L389

MustafaJafar:
## Current Stats of this PR

This PR adds support for different render targets
- `Local machine rendering`
- `Use existing frames (local)`
- `Farm Rendering`
- `Farm Rendering - Split export & render jobs`

Also, adds support for marking render as reviewable. it doesn't work yet when user have multiple **separated** AOVs.

image

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
MustafaJafar: It's hard because there is some dependency in the collectors. 

- `CollectHoudiniReviewData - CollectorOrder + 0.13` needs to run after `CollectRopFrameRange - CollectorOrder` and `CollectLocalRenderInstances - CollectorOrder + 0.12`
- `CollectLocalRenderInstances - CollectorOrder + 0.12` needs to run after Collect Render Products e.g. `CollectMantraROPRenderProducts - CollectorOrder + 0.11`
- `CollectMantraROPRenderProducts - CollectorOrder + 0.11` needs to run after `CollectFrames - CollectorOrder + 0.1`
- `CollectFrames - CollectorOrder + 0.1` needs to run after `CollectRopFrameRange  - CollectorOrder`

Mostly, I used `0.01` step.

moonyuet:
> > Many thanks for your help. I slightly edited it as it's helpful to transfer all `pre_create_data` to `creator_attributes`
> 
> Not entirely true. For example, `use_selection` really is _just_ pre create data not important after the creation. So you may now be "bloating" your data with redundant data that has no real place among creator attributes.

we can take reference on this commit, where I only tpok the render target. https://github.com/ynput/ayon-core/pull/385/commits/c5df561c970bc1cfc0498f8b1adf52392f5fe969

moonyuet:
> ## Current Stats of this PR
> This PR adds support for different render targets
> 
> * `Local machine rendering`
> * `Use existing frames (local)`
> * `Farm Rendering`
> * `Farm Rendering - Split export & render jobs`
> 
> Also, adds support for marking render as reviewable. it doesn't work yet when user have multiple **separated** AOVs.
> 
> image

not sure if it is anti-pattern but guess we can put the aov into multiple instances if it is in separated AOV option, similar to the image instances in substance painter hosts? Any thought? @BigRoy @antirotor

MustafaJafar:
> we can take reference on this commit, where I only took the render target. [c5df561](https://github.com/ynput/ayon-core/commit/c5df561c970bc1cfc0498f8b1adf52392f5fe969)

I think it's a matter of UX. Making a decision about it is not easy as it's hard to make all people happy. 
The decision I made so far in this PR was as simple as 
```
Don't use pre create data that can't be used after the creation.
```
That's why I only removed `use_selection` 
Because no way it can be used after the creation.

client/ayon_core/hosts/houdini/plugins/publish/collect_arnold_rop.py
```diff
@@ -85,6 +86,14 @@ def process(self, instance):
             files_by_aov[label] = self.generate_expected_files(instance,
                                                                aov_product)
 
+            # Set to False as soon as we have a separated aov.
+            multipartExr = False
```
BigRoy, line 90: I think this logic is invalid. It's totally fine for an arnold render to have AOVs but write them into a single EXR as multilayers - currently this logic looks like; If any AOVs are enabled then `multipart=False` which seems invalid logic?

MustafaJafar: If any AOVs are enabled and _rendered separately_ then then `multipart=False` 
ROPs have AOVs options but render them as one EXR file. 

I don't know if there's an option in arnold to specify to render as one layer or not but the logic in arnold collector always consider AOVs as separated products, so I assumed they are rendered in separated files. 

https://github.com/ynput/ayon-core/blob/f64bb5c1ea6182b3b3ec251f8aaf3785ddeda38c/client/ayon_core/hosts/houdini/plugins/publish/collect_arnold_rop.py#L71-L86

_For information: I didn't test arnold but I followed the same as mantra and Karma_

BigRoy: I see what you mean.

I think the logic here for Arnold is wrong in that it assumes that _when_ there are AOVs that it must be non-merged / non-multilayer since it's totally valid to render multiple AOVs into a single file - which is exactly what multipart exr is for?

Not sure how the multipart/multilayer logic works in Houdini for AOVs but the Maya render logic does that fine where arnold with aovs can render to both multilayer or separate products.

client/ayon_core/hosts/houdini/plugins/publish/validate_review_colorspace.py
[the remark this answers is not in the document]
MustafaJafar: > to make it more complicated, you need to take care wheter instance is actually reviewable -> not all renders are 

Revisiting the AOVs filters.
settings location: 
- `ayon+settings://deadline/publish/ProcessSubmittedJobOnFarm/aov_filter` 

In the current state of this PR, I found out that these filters work with Houdini farm publishing. 
image

But, I faced some problems because of the regex filter doesn't match the default Houdini behavior. 
Because , by default Not all ROPs support `$AOV` in render paths.
So, I actually had to add AOV names manually in the mantra rop
image

Every product type can have its special cases. 
For example: Redshift 
It has `$AOV` which is cool but once I enable `RS_aovMultipart` and add cryptomatte AOV
in that case: redshift will render cryptomatte separately and ignore `beauty` AOV name suffix thus no review will be created as `beauty` will no longer exist in the file name.

client/ayon_core/hosts/houdini/plugins/publish/collect_arnold_rop.py
MustafaJafar: > how the multipart/multilayer logic works in Houdini for AOVs 

As far as I tested render rops in Houdini are not consisted. 
Some has `multipart` check box explicitly and some add the check box per AOV.
So, to align with the current logic in deadline plugin I had to deduce the `multipart` value.
In a nutshell the logic is as follow: 
```
if `review` key in instance data is enabled: 
    if the render is multipart Exr (One merged exr) then add review tag.
    if not then check against AOV regex filter in settings, if matches the regex then add review tag.
```

client/ayon_core/hosts/houdini/plugins/publish/validate_review_colorspace.py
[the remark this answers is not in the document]
MustafaJafar: >For example: Redshift
It has $AOV which is cool but once I enable RS_aovMultipart and add cryptomatte AOV
in that case: redshift will render cryptomatte separately and ignore beauty AOV name suffix thus no review will be created as beauty will no longer exist in the file name.

I've just discovered that I should add `renderer` key in instance data 
which solves redshift issue mentioned above. 

https://github.com/ynput/ayon-core/blob/27dea6c97a5bac4fc5ba7ca4a0932f4d6a993109/client/ayon_core/pipeline/farm/pyblish_functions.py#L633-L637

client/ayon_core/hosts/houdini/plugins/publish/extract_local_render.py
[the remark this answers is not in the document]
MustafaJafar: The readability should be enhanced in https://github.com/ynput/ayon-core/pull/328/commits/60468e4d7410ff5021d771a09b1e2e16494e6380

MustafaJafar:
Self Reminder: 
`CollectRedshiftROPRenderProducts` doesn't work as expected  in this PR (It works in develop).
ExpectedFiles logic is some how broken.
I suspect the problem happens because of using `get_instance_attr_defs` and the houdini nodes parameters are not synced with the creator attributes properly. 

Fortunately, That problem doesn't exist in https://github.com/ynput/ayon-core/pull/418
Also, in my opinion: having a solution like https://github.com/ynput/ayon-core/pull/418 will eliminate some other problems here as well e.g. this problem https://github.com/ynput/ayon-core/pull/328#discussion_r1558137542

client/ayon_core/hosts/houdini/plugins/publish/collect_arnold_rop.py
MustafaJafar: I added a todo about it, therefore I'll mark this as resolved. 
Feel free to open an issue thou,

MustafaJafar:
So far this PR works but I have some considerations on this PR from my point of view:  

1. using the instance node as the source of truth: which can no longer happen because the exposed `creator_attributes` can change without updating the instance node itself.
So, collectors won't the collect the correct data. 
Currently, I use `creator_attributes` as the source of truth in collect render product plugins and use an extractor plugin to modify the Houdini node parameters accordingly.
This also limits what `creator_attributes` that can be exposed in publish tab. 
Therefore, I created https://github.com/ynput/ayon-core/pull/418 as an attempt to use instance node as the source of truth by syncing instance nodes and publisher. 


2. Marking instance as reviewable, It works on farm and local but it also runs review plugins dedicated to opengl node. This problem is shown here https://github.com/ynput/ayon-core/pull/328/files#r1558145584 
Currently, I solved it temporarily by adding `if rop_node.type().name() != "opengl"` in the unrelated plugins.

3. Creating new instances: tbh, I'm not sure if I should use a dedicated function to create instances. 
Currently, I create them and fill the necessary data keys manually. It works so far.

client/ayon_core/hosts/houdini/plugins/publish/validate_review_colorspace.py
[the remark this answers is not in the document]
MustafaJafar: > Hmm - yeah. This doesn't feel entirely right. I feel like this validator is now just running over the wrong instances - which should be solved in a way that it just does not run to begin with when it shouldn't. I understand why it's done now, since you're likely adding "review" family to something that isn't actually an OpenGL review instance - but maybe the regular review instances should instead have `opengl_rop` family or whatever and have these validators target that?

I added a TODO about it. I think it should be fixed in a separate PR.
Therefore, I'm marking this as resolved.

moonyuet (commented):
I tested with local machine rendering, use existing frames and farm rendering withs single AOVrender.
So far so good
image

moonyuet (approved):
I tested with AOV renders in Mantra. Looks good. Maybe there should be some reminders that each renderer has different AOV suffix in terms of filenames, and we won't misuse the `$AOV` into the renderers other than into Redshift.

MustafaJafar:
> I am still not convinced with `farm_split` name as it's not very descriptive and then there is `splitRender` used.

I couldn't pick better names. 
I'd appreciate any suggestions.

moonyuet:
Should we merge this ticket or we still need to work on something else? @MustafaJafar @antirotor

client/ayon_core/hosts/houdini/plugins/create/create_vray_rop.py
```diff
@@ -143,20 +147,41 @@ def remove_instances(self, instances):
...
+        return [
+            BoolDef("review",
+                    label="Review",
+                    tooltip="Mark as reviewable",
+                    default=True),
+            EnumDef("render_target",
+                    items=render_target_items,
+                    label="Render target",
+                    default=self.render_target)
+        ]
```
BigRoy, line 174: Can we please make a follow-up PR that abstract this to a parent class so this logic is basically re-used for all render creators in Houdini?

This PR clearly shows that basically all these creators share 80% of logic between each other? Can't we unify this in a parent "CreateRender" class?

antirotor: We should indeed, tagging @MustafaJafar

MustafaJafar: I've created an issue to track it. Thank you!
https://github.com/ynput/ayon-core/issues/507

krishna8008:
Can we also have a feature to publish through the Solaris tree renders? Currently, the karma rop is working well. but, all the people use only the solaris tree as we have more control there on stage. Please let me know your thoughts on this.

BigRoy:
> Can we also have a feature to publish through the Solaris tree renders? Currently, the karma rop is working well. but, all the people use only the solaris tree as we have more control there on stage. Please let me know your thoughts on this.

Are you saying you would like to publish or submit using the USD Render ROP inside `/stage/..` and not a USD Render ROP in `/out`? 

For what it's worth, I believe #295 implements supporting USD Render ROP for Deadline submissions using Husk standalone for rendering - however, I think it's still aimed at working in `/out`.

I'm working on some more houdini-wide efforts to support more generalized publishing and submissions that is a more global 'rethink' of how things work in Houdini. This may affect also your question here over time.

Could you elaborate on your exact use case?

PR #331, opened 2024-03-29
Maya: Optimize collect file dependencies

BigRoy:
## Changelog Description

Only collect file dependencies if enabled in settings and only collect if renderlayer family is present

## Additional info

Now also excludes "duplicate" entries.

Still has the issue that it doesn't work correctly for all udim tiles, etc. The logic is just too simple for that and I believe has been reported before. Anyway, for now this is a good optimization.

The setting is `deadline > publish > MayaSubmitDeadline > asset_dependencies `

## Testing notes:

1. Collect file dependencies should only run if a renderlayer instance is active.
2. When enabled in settings, file dependencies should be collected and submitted along to Deadline job
3. When disabled in settings, file dependencies should not be collected

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/collect_file_dependencies.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/collect_file_dependencies.py
@@ -1,5 +1,3 @@
-import json
-
 from maya import cmds
 
 import pyblish.api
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Tested in Maya 2023 successfully and the asset dependencies wont be in the job info parameters when the relevant setting has been disabled.

LiborBatek (approved):
LGTM, already tested by @moonyuet

PR #332, opened 2024-03-29
Maya: Render Improve Validation Report if no cameras are set

BigRoy:
## Changelog Description

Provide a validation report when no renderable cameras are set instead of erroring during collectors.

Resolves the issues mentioned [here](https://github.com/ynput/ayon-core/pull/299#issuecomment-2026900524).

## Additional info

image

Note that you may still get an error from another validator which would be fixed by https://github.com/ynput/ayon-core/pull/299

As such, it actually makes easier testing if you merge both - and validate both together.

## Testing notes:

1. Create a renderlayer
2. In the renderlayer create an override that disables all renderable cameras.
3. Remain in the default layer with a renderable camera - otherwise you might hit another error that gets fixed by another PR https://github.com/ynput/ayon-core/pull/299
4. Publish/Submit the renderlayer

It should now show you a validation that the renderlayer has no renderable camera.

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/collect_render.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/collect_render.py
@@ -1,24 +1,19 @@
 # -*- coding: utf-8 -*-
 """Collect render data.
 
-This collector will go through render layers in maya and prepare all data
-needed to create instances and their representations for submission and
-publishing on farm.
+This collector will go through renderlayer instances and prepare all data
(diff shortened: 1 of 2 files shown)


MustafaJafar (approved):
Tested as user. 
> I used `RemoveRenderableCameras` to test.
image

moonyuet (approved):
Looks good for me. This is more concise than before as the users would know they need to assign renderable camera for getting "render products".

PR #334, opened 2024-03-29
Maya: USD workflow with maya usd

BigRoy:
## Changelog Description

Implement more of Maya USD workflows:

- Add more options to the Maya USD Extractor
- Export a layer of a Maya USD proxy, e.g. lookdev done with Maya's lookdevx.
- Prototype loading of USD references into a Maya USD proxy while keeping it managed by the pipeline
- Prototype loading of Maya references into a Maya USD proxy while keeping it managed by the pipeline

## Additional info

Some of the logic here has been used in our production - but there's also quite a bit that we didn't use at all in the actual production. The logic here can thus be considered mostly a prototype and up for discussion.

We mostly used the loading of USD files to a proxy - for previewing purposes. And the Maya USD export from Maya.

I also had [this prototype](https://gist.github.com/BigRoy/7d73ec1bfcc5944a6aa887ee31b1bf44) to export native maya model + lookdev done with arnold to just a "look" layer override for a USD stage. (So basically a look export without the geometry present so that you could overlay it over the asset; however we did not use that.)

Anyway, input is more than welcome!

This is basically the Maya part of this equivalent OpenPype PR https://github.com/ynput/OpenPype/pull/5925 but with some more improvements we did while using some parts in production.

## Testing notes:

1. For now, play around with the features! (Until PR is ready and out of draft!)


client/ayon_core/hosts/maya/plugins/load/load_maya_usd_add_maya_reference.py
```diff
@@ -0,0 +1,155 @@
...
+    identifier_key = "ayon_identifier"
+
+    def load(self, context, name=None, namespace=None, options=None):
+
+        selection = list(iter_ufe_usd_selection())
+        assert len(selection) == 1, "Select only one PRIM please"
+        ufe_path = selection[0]
+        path = self.filepath_from_context(context)
+
+        import mayaUsdAddMayaReference
```
moonyuet, line 65: Guess need to make sure mayaUsdPlugin needed to be loaded before all the usd loading actions?
```suggestion
         # Make sure we can load the plugin
        cmds.loadPlugin("mayaUsdPlugin", quiet=True)
        import mayaUsdAddMayaReference
```

BigRoy: I'm not entirely sure - so better safe than sorry I suppose. :)

antirotor:
So I've been playing with it. Found particular case I don't know how we should handle. I was trying to reference shot USD consisting of just sublayers:

```usd
#usda 1.0
(
    framesPerSecond = 25
    metersPerUnit = 1
    subLayers = [
        @...:SDF_FORMAT_ARGS:layer_id=lighting&order=600@,
        @...:SDF_FORMAT_ARGS:layer_id=animation&order=300@,
        @...:SDF_FORMAT_ARGS:layer_id=layout&order=200@,
        @...:SDF_FORMAT_ARGS:layer_id=model&order=100@
    ]
    timeCodesPerSecond = 25
    upAxis = "Y"
)
```
Loader gave me:
```
Warning: In </root>: Unresolved reference prim path @...@<defaultPrim> introduced by @anon:00000116D1C58160:anonymousLayer1@</root>
```
and nothing was referenced (so it is more error than warning). When trying to do it manually the error is more clear:

```
Warning: The file [...usd] does not contain a default prim, the reference will be invalid
```
Which make sense.It basically means we can reference in only USD with defined `defaultPrim`. In this case, how to handle it? With different family? Just failing during the load?

BigRoy:
> So I've been playing with it. Found particular case I don't know how we should handle. I was trying to reference shot USD consisting of just sublayers...

@antirotor Could you elaborate a bit more on how to reproduce this particular behavior?

I suppose you were just publishing some "shot contributions" using `Maya USD` similar to this:
image

Or could you share your instance publish settings that produced these shot USD files?

iLLiCiTiT:
Please move to https://github.com/ynput/ayon-maya if is still relevant.

PR #153, opened 2024-03-06
Publisher: Reset all attributes to default feature

bradenjennings:
## Changelog Description
Add a button to publisher to reset all creator and publisher attributes.
https://app.clickup.com/t/6658547/AY-3429

## Additional info
Paragraphs of text giving context of additional technical information or code examples.

## Testing notes:
1. start with this step
2. follow this step

PR #336, opened 2024-04-01
Maya: abc options for Pointcache/Animation family - AY-970

tokejepsen:
## Changelog Description
Port of https://github.com/ynput/OpenPype/pull/5173

- [x] Validate alembic defaults settings


tokejepsen:
@BigRoy with this PR do we still need the bake attributes for alembic extraction?

BigRoy:
> @BigRoy with this PR do we still need the bake attributes for alembic extraction?

I'd say so, yes. One might want to allow the attributes to be specified by a user without being able to disable certain attributes, which is what `bake_attributes` is for. It's not "customizable ever" and "always included" no matter what - however one might still allow other attributes to be editable/optionally written to the output file. So yes, I'd say they are still valuable.

BigRoy (changes requested):
> All working fine and also with `custom attribs`. Did try to produce some custom attribs and also `Validate` of attribs and `Repair` action works well too.
> 
> My settings and `alembic` product with matching `custom attribs` written in...
> 
> Screenshot 2024-04-02 140935
> 
> Screenshot 2024-04-02 141504

Can we make the custom attribute fields a list of strings? Similar to the Bake Attributes settings?

tokejepsen:
> Can we make the custom attribute fields a list of strings? Similar to the Bake Attributes settings?

@BigRoy not sure how that would translate to the publisher? Dont think there is support for lists in the publisher, so we might end up with the same string with separators.

BigRoy:
> > Can we make the custom attribute fields a list of strings? Similar to the Bake Attributes settings?
> 
> @BigRoy not sure how that would translate to the publisher? Dont think there is support for lists in the publisher, so we might end up with the same string with separators.

Ah, very valid point. The publisher UI indeed does not expose that currently. In that case - ignore me. Might be worth tracking an 'issue' for that for future work to see if we can improve this.

server_addon/maya/server/settings/publishers.py
```diff
@@ -35,6 +35,76 @@ def angular_unit_enum():
...
+        {"value": "visibleOnly", "label": "Visible Only"},
+        {"value": "wholeFrameGeo", "label": "wholeFrameGeo"},
+        {"value": "worldSpace", "label": "worldSpace"},
+        {"value": "writeColorSets", "label": "writeColorSets"},
+        {"value": "writeCreases", "label": "writeCreases"},
+        {"value": "writeFaceSets", "label": "writeFaceSets"},
+        {"value": "writeNormals", "label": "writeNormals"},
+        {"value": "writeUVSets", "label": "writeUVSets"},
+        {"value": "writeVisibility", "label": "writeVisibility"}
+    ]
```
BigRoy, line 105: Is there a reason that Pre Roll Start Frame is the only one that has a 'nice' label?

server_addon/maya/server/settings/publishers.py
```diff
@@ -38,26 +38,26 @@ def angular_unit_enum():
...
+        {"value": "uvWrite", "label": "UV Write"},
+        {"value": "uvsOnly", "label": "UVs Only"},
+        {"value": "verbose", "label": "Verbose"},
+        {"value": "wholeFrameGeo", "label": "Whole Frame Geo"},
+        {"value": "worldSpace", "label": "World Space"},
+        {"value": "writeColorSets", "label": "Write Color Sets"},
+        {"value": "writeFaceSets", "label": "Write Face Sets"},
+        {"value": "writeNormals", "label": "Write Normals"},
+        {"value": "writeUVSets", "label": "Write UV Sets"},
+        {"value": "writeVisibility", "label": "Write Visibility"}
```
BigRoy, line 60: I may have OCD - but can we align both of the `enum()` methods to either for both start the entries with the `value` or the `label`. They now seem the opposite. It's cosmetics, but makes it easier to parse if things are consistent throughout the file.

server_addon/maya/server/settings/publishers.py
```diff
@@ -387,13 +393,17 @@ class ExtractAlembicModel(BaseSettingsModel):
...
-    userAttr: str = SettingsField(title="userAttr")
-    userAttrPrefix: str = SettingsField(title="userAttrPrefix")
+    pythonPerFrameCallback: str = SettingsField(
+        title="Python Per Frame Callback"
+    )
+    pythonPostJobCallback: str = SettingsField(
+        title="Python Post Job Callback"
+    )
+    userAttr: str = SettingsField(title="User Attr")
+    userAttrPrefix: str = SettingsField(title="User Attr Prefix")
```
BigRoy, line 406: Can we add a `description=""` argument to these explaining the formatting, so that it's clear one needs to use e.g. `,` (comma) separated values. (Or whatever the separator is? maybe it's `;`?)

BigRoy: Ah it's `;` according to [this code](https://github.com/ynput/ayon-core/pull/336/files#diff-9ac4195c9cccdcfca6bae8ddc782d74a45ff93e0764f05edd34fff3a98055475R72). Definitely add a description tooltip! Even nicer if we could somehow put in `placeholder` text like `attr1;attr2` and `prefix1;prefix2`. Is that possible in settings?

BigRoy: Example:
```suggestion
    userAttr: str = SettingsField(
        title="User Attr",
        placeholder="attr1;attr2",
        description=(
            "Attributes matching by name will be included in the Alembic export. "
            "Attributes should be separated by semi-colon `;`"
        )
    )
    userAttrPrefix: str = SettingsField(
        title="User Attr Prefix",
        placeholder="attr1;attr2",
        description=(
            "Attributes starting with these prefixes will be included in the Alembic export. "
            "Attributes should be separated by semi-colon `;`"
        )
    )
```

BigRoy: Same goes for `attr` and `attrPrefix` - and even better if the description could be used to explain the difference between the two. You can include markdown links as well to maybe some Alembic documentation if needed.

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +202,147 @@ def process(self, instance):
...
+        defs.extend([
+            UISeparatorDef("sep_alembic_options"),
+            UILabelDef("Alembic Options"),
+        ])
+
+        # The Arguments that can be modified by the Publisher
+        overrides = set(getattr(cls, "overrides", set()))
+
+        # What we have set in the Settings as defaults.
+        flags = set(getattr(cls, "flags", set()))
```
BigRoy, line 296: Since these are exposed as regular attributes on the class by its definition, why use `getattr`?

```suggestion
        # The Arguments that can be modified by the Publisher
        overrides = set(cls.overrides)

        # What we have set in the Settings as defaults.
        flags = set(cls.flags)
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +202,147 @@ def process(self, instance):
...
+    @classmethod
+    def get_attribute_defs(cls):
+        override_defs = {
+            "attr": {
+                "def": TextDef,
+                "kwargs": {
+                    "label": "Custom Attributes",
+                    "placeholder": "attr1; attr2; ...",
+                }
+            },
```
BigRoy, line 214: Is there a reason why this is structured like this - why are we not just creating the `TextDef` itself but only returning it from `get_attribute_defs` if it should be exposed? Seems odd that we're going for this data intermediate structure in between? Seems more complex than needed?

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,157 @@
...
+            BoolDef("farm", label="Submit to Farm"),
+            NumberDef("priority", label="Farm job Priority", default=50),
+            BoolDef("refresh", label="Refresh viewport during export"),
+            BoolDef(
+                "includeParentHierarchy", label="Include Parent Hierarchy"
+            ),
+            BoolDef(
+                "includeUserDefinedAttributes",
+                label="Include User Defined Attributes"
+            ),
```
BigRoy, line 26: Can we preserve the tooltips for these attribute definitions?

E.g. we're currently losing [this](https://github.com/ynput/ayon-core/pull/336/files#diff-34154b3e70a9a31f619f8fe214e6eccec7234359cc015fce4181124c3190d279L48).

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +202,147 @@ def process(self, instance):
...
         # Collect the out set nodes
         out_sets = [node for node in instance if node.endswith("out_SET")]
         if len(out_sets) != 1:
-            raise RuntimeError("Couldn't find exactly one out_SET: "
-                               "{0}".format(out_sets))
+            raise RuntimeError(
+                "Couldn't find exactly one out_SET: " "{0}".format(out_sets)
+            )
         out_set = out_sets[0]
         roots = cmds.sets(out_set, query=True)
```
BigRoy, line 339: Should this have a failsafe for `cmds.sets` returning None if the set is empty as well? (Supposedly a validator catches that already, but maybe it doesn't?)

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +202,147 @@ def process(self, instance):
...
 
     def get_members_and_roots(self, instance):
-
         # Collect the out set nodes
         out_sets = [node for node in instance if node.endswith("out_SET")]
         if len(out_sets) != 1:
-            raise RuntimeError("Couldn't find exactly one out_SET: "
-                               "{0}".format(out_sets))
+            raise RuntimeError(
+                "Couldn't find exactly one out_SET: " "{0}".format(out_sets)
```
BigRoy, line 336:
```suggestion
                "Couldn't find exactly one out_SET: {0}".format(out_sets)
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -58,49 +90,71 @@ def process(self, instance):
...
         with suspended_refresh(suspend=suspend):
             with maintained_selection():
                 cmds.select(nodes, noExpand=True)
-                extract_alembic(
-                    file=path,
-                    startFrame=start,
-                    endFrame=end,
-                    **options
+                self.log.debug(
+                    "Running `extract_alembic` with the arguments: {}".format(
```
BigRoy, line 153:
```suggestion
                    "Running `extract_alembic` with arguments: {}".format(
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -124,21 +178,17 @@ def process(self, instance):
...
         with suspended_refresh(suspend=suspend):
             with maintained_selection():
                 cmds.select(instance.data["proxy"])
-                extract_alembic(
-                    file=path,
-                    startFrame=start,
-                    endFrame=end,
-                    **options
-                )
+                extract_alembic(**args)
```
BigRoy, line 191: You're passing the `args` as `**kwargs` (keyword arguments). As such, to avoid confusion I'd recommend renaming the variable accordingly to `kwargs` to match python conventions.

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,143 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+        title="Custom Attributes",
+        placeholder="attr1;attr2",
+        description=(
+            "Attributes matching by name will be included in the Alembic "
+            "export. Attributes should be separated by semi-colon `;`"
+        )
+    )
+    attrPrefix: str = SettingsField(
+        title="Custom Attributes Prefix",
+        placeholder="attr1;attr2",
```
BigRoy, line 451:
```suggestion
        placeholder="prefix1;prefix2",
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,143 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+        title="User Attr",
+        placeholder="attr1;attr2",
+        description=(
+            "Attributes matching by name will be included in the Alembic "
+            "export. Attributes should be separated by semi-colon `;`"
+        )
+    )
+    userAttrPrefix: str = SettingsField(
+        title="User Attr Prefix",
+        placeholder="attr1;attr2",
```
BigRoy, line 508:
```suggestion
        placeholder="prefix1;prefix2",
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -35,6 +35,82 @@ def angular_unit_enum():
...
+        },
+        {"label": "Renderable Only", "value": "renderableOnly"},
+        {"label": "Strip Namespaces", "value": "stripNamespaces"},
+        {"label": "User Attr", "value": "userAttr"},
+        {"label": "User Attr Prefix", "value": "userAttrPrefix"},
+        {"label": "UV Write", "value": "uvWrite"},
+        {"label": "UVs Only", "value": "uvsOnly"},
+        {"label": "Verbose", "value": "verbose"},
+        {"label": "Visible Only", "value": "visibleOnly"},
+        {"label": "Whole FrameGeo", "value": "wholeFrameGeo"},
```
BigRoy, line 103:
```suggestion
        {"label": "Whole Frame Geo", "value": "wholeFrameGeo"},
```

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,162 @@
...
+)
+from ayon_core.pipeline import CreatedInstance
+
+
+def _get_animation_attr_defs(cls):
+    """Get Animation generic definitions."""
+    defs = lib.collect_animation_defs()
+    defs.extend(
+        [
+            BoolDef("farm", label="Submit to Farm"),
```
BigRoy, line 17: Does this farm logic still work?

tokejepsen: Not sure, but I think that might be outside of the scope of this PR. Will make an issue to investigate.

BigRoy:
I won't be able to do a test run of this PR before next Tuesday. So maybe others can review, or I'll look at it then.

antirotor (approved):
Fine by me, I can't test it right now.

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,143 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+            "Write Face Sets - Write face sets with the geometry.\n"
+
+            "Write Normals - Write normals with the deforming geometry.\n"
+
+            "Write UV Sets - Write all uv sets on MFnMeshes as vector 2 "
+            "indexed geometry parameters with face varying scope.\n"
+
+            "Write Visibility - Visibility state will be stored in the Alembic"
+            " file. Otherwise everything written out is treated as visible."
+        )
```
BigRoy, line 439: Can we make the labels bold by using `**` around them (markdown bold)
```suggestion
        description=(
            "**Auto Subd** - If this flag is present and the mesh has crease "
            "edges, crease vertices or holes, the mesh (OPolyMesh) would now "
            "be written out as an OSubD and crease info will be stored in the "
            "Alembic  file. Otherwise, creases info won't be preserved in "
            "Alembic file unless a custom Boolean attribute SubDivisionMesh "
            "has been added to mesh node and its value is true.\n"

            "**Dont Skip Unwritten Frames** - When evaluating multiple "
            "translate jobs, this decides whether to evaluate frames between "
            "jobs when there is a gap in their frame ranges.\n"

            "**Euler Filter** - Apply Euler filter while sampling rotations.\n"

            "**No Normals** - Present normal data for Alembic poly meshes will"
            " not be written.\n"

            "**Pre Roll** - This frame range will not be sampled.\n"

            "**Renderable Only** - Only export renderable visible shapes.\n"

            "**Strip Namespaces** - Namespaces will be stripped off of the "
            "node before being written to Alembic.\n"

            "**UV Write** - Uv data for PolyMesh and SubD shapes will be "
            "written to the Alembic file.\n"

            "**UVs Only** - If this flag is present, only uv data for PolyMesh"
            " and SubD shapes will be written to the Alembic file.\n"

            "**Verbose** - Prints the current frame that is being evaluated.\n"

            "**Whole Frame Geo** - Data for geometry will only be written out "
            "on whole frames.\n"

            "**World Space** - Any root nodes will be stored in world space.\n"

            "**Write Color Sets** - Write vertex colors with the geometry.\n"

            "**Write Face Sets** - Write face sets with the geometry.\n"

            "**Write Normals** - Write normals with the deforming geometry.\n"

            "**Write UV Sets** - Write all uv sets on MFnMeshes as vector 2 "
            "indexed geometry parameters with face varying scope.\n"

            "**Write Visibility** - Visibility state will be stored in the "
            "Alembic file. Otherwise everything written out is treated as "
            "visible."
        )
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,143 @@ class ExtractAlembicModel(BaseSettingsModel):
         description="List of attribute prefixes for attributes that will be "
                     "included in the alembic export.",
     )
+    flags: list[str] = SettingsField(
+        enum_resolver=extract_alembic_flags_enum,
+        title="Export Flags",
```
BigRoy, line 390: Even though I like the description of what each flag means, just looking at the settings I'm clueless what the "Export flags" setting is for. It seems to be the 'default state' for that particular boolean flag (enabled or disabled). When not 'exposed' to the user the value is _always_ at this default state, otherwise it's exposed to the artist and this would define the default state.

server_addon/maya/server/settings/publishers.py
BigRoy: Also, I'm pretty sure the description for **Pre Roll** is incorrect? _This frame range will not be sampled_ seems pretty wrong?

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +201,226 @@ def process(self, instance):
...
+                " parameters with face varying scope."
+            ),
+            "writeVisibility": (
+                "Visibility state will be stored in the Alembic file. "
+                "Otherwise everything written out is treated as visible."
+            )
+        }
+        tooltip = ""
+        for flag in flags:
+            tooltip += "{} - {}\n".format(flag, tooltips[flag])
```
BigRoy, line 380: This introduces a trailing new line currently.

image


```suggestion
        tooltip = "\n".join(f"{flag} - {tooltips['flag']}" for flag in flags)
```

Also, because this is based on sets you're losing the sorted nature of the flags in the tooltip:
image

As such, you're better off making sure the flags are sorted for readability.
e.g.
```suggestion
        tooltip = "\n".join(f"{flag} - {tooltips['flag']}" for flag in sorted(flags))
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +201,226 @@ def process(self, instance):
...
+        )
+
+        for key, value in override_defs.items():
+            if key not in overrides:
+                continue
+
+            defs.append(value)
+
+        defs.append(
+            UISeparatorDef("sep_alembic_options")
```
BigRoy, line 399: As far as I know the actual keys here _should_ be unique. Right @iLLiCiTiT ? So technically this matching the key/name used above would be invalid?

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +201,226 @@ def process(self, instance):
...
 
         # Include all descendants
-        nodes = roots + cmds.listRelatives(roots,
-                                           allDescendents=True,
-                                           fullPath=True) or []
+        nodes = (
+            roots
+            + cmds.listRelatives(roots, allDescendents=True, fullPath=True)
+            or []
+        )
```
BigRoy, line 424: To me this new code is way more unreadable. Anyone else has that?

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,143 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+    pythonPostJobCallback: str = SettingsField(
+        title="Python Post Job Callback",
+        description=(
+            "When the translation has finished the string specified is "
+            "evaluated as a python command."
+        )
+    )
+    preRollStartFrame: int = SettingsField(
+        title="Pre Roll Start Frame",
+        description=(
```
BigRoy, line 492: It may be good in this description to add that `Pre Roll` may need to be added to the Export Flags? (It's a bit unclear at first sight what 'activates' or deactivates this pre roll setting?

BigRoy (commented):
So overall, the logic seems to work but I do have some cosmetic notes.

I also wonder whether an artist will understand this:

image
Better than just a list of checkboxes for these. Any opinions anyone?
Benefit of separate checkboxes is also that they can get their individual tooltips, etc.

LiborBatek (commented):
Did quick tests if ```animation``` and ```pointcache``` publishing and loading works and it is working fine.

There is a small glitch with already present ```publish instance```s that some attribs has broken status switches aka knobs missing. This being true for both animation and pointcache instances...

Screenshot 2024-04-11 091026

when recreated from scratch then it start working and looking normal:

Screenshot 2024-04-11 091320

Question being if we shouldnt do some "convert legacy instances" as we do in different occasions...similarly

tokejepsen:
> Better than just a list of checkboxes for these. Any opinions anyone?

@BigRoy lol, this goes back to a version we had ages ago, but will leave this up to @antirotor or @mkolar

server_addon/maya/server/settings/publishers.py
tokejepsen: > Also, I'm pretty sure the description for Pre Roll is incorrect? This frame range will not be sampled seems pretty wrong?

Think this was from the alembic export help descriptions. Think they mean that the pre roll range will not become part of the export, so not written to disk.

server_addon/maya/server/settings/publishers.py
tokejepsen: Sorry, not sure what you are wanting to change here?

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +201,224 @@ def process(self, instance):
...
+            "writeUVSets": (
+                "Write all uv sets on MFnMeshes as vector 2 indexed geometry"
+                " parameters with face varying scope."
+            ),
+            "writeVisibility": (
+                "Visibility state will be stored in the Alembic file. "
+                "Otherwise everything written out is treated as visible."
+            )
+        }
+        tooltip = "\n".join(f"{flag} - {tooltips['flag']}" for flag in flags)
```
tokejepsen, line 378:
```suggestion
        tooltip = "\n".join(f"{flag} - {tooltips['flag']}" for flag in sorted(flags))
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
tokejepsen: Does not seem to error and the UI looks correct, but will leave to @iLLiCiTiT to confirm on code practice.

server_addon/maya/server/settings/publishers.py
BigRoy: It was more of a remark - I went to settings, saw the Export flags. And even with tooltips it didn't click for me what "Export flags" was for and what it was supposed to do.

It was unclear to me that it would define the default state for that flag. As such, my remark was: "Are we ok with it being unclear?" and "Do we see any good way to improve this user experience?"

Usually tooltips explain what the attribute does - but now it's explaining the flags themselves and totally skips explaining what "Export flags" is for. Which is valuable information, but due to it being unclear what "Export flags" was intended to do it did leave me in the dark for a bit.

BigRoy:
> @BigRoy lol, this goes back to a version we had ages ago, but will leave this up to @antirotor or @mkolar

Yes, totally understand.

Just wanted to add - by having them as checkboxes it also simplifies 'multiselection' editing for a specific flag on multiple instances. With the multiselection enum when having multiple instances selected you're essentially in multiselection editing picking ALL flags for ALL selected instances - but with individual checkboxs it is actually possible to select multiple instances and ONLY enable 'worldSpace' export on the selected instances - leaving the others as they are per instance.

Additionally, with features like #153 and #146 it also becomes much clearer which of the flags is non-default instead of the full "Export flags" enum being 'non-default'.

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +201,224 @@ def process(self, instance):
...
 
         # Include all descendants
-        nodes = roots + cmds.listRelatives(roots,
-                                           allDescendents=True,
-                                           fullPath=True) or []
+        nodes = (
+            roots
+            + cmds.listRelatives(roots, allDescendents=True, fullPath=True)
+            or []
+        )
```
tokejepsen, line 422:
```suggestion
        nodes = roots
        nodes += cmds.listRelatives(
            roots, allDescendents=True, fullPath=True
        ) or []
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
tokejepsen: Yup, agree. Suggested something else now.

server_addon/maya/server/settings/publishers.py
BigRoy: Interesting - since the flag I suppose would be the reverse, when enabled then it _does_ get written? Anyway, let's rewrite it to be clearer: When enabled, the pre roll start frame is used to pre roll the export for.

Side question - is the "pre roll" just sampled for the export so that it runs from there. Or is it actually included and written into the alembic as well?

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,144 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+            "The frame to start scene evaluation at.  This is used to set the "
+            "starting frame for time dependent translations and can be used to"
+            " evaluate run-up that isn't actually translated."
```
tokejepsen, line 496:
```suggestion
            " evaluate run-up that isn't actually translated.\n"
            "NOTE: preRoll needs to be enabled in the export flags for this "
            "start frame to be considered."
```

server_addon/maya/server/settings/publishers.py
tokejepsen: > As such, my remark was: "Are we ok with it being unclear?" and "Do we see any good way to improve this user experience?"

Judge... Leading the witness... :)

I think before doing anything about this, we should reach a decision on the multi-enum in the publisher, cause if that changes we might as well change this to check boxes as well.

tokejepsen:
> Just wanted to add - by having them as checkboxes it also simplifies 'multiselection' editing for a specific flag on multiple instances. With the multiselection enum when having multiple instances selected you're essentially in multiselection editing picking ALL flags for ALL selected instances - but with individual checkboxs it is actually possible to select multiple instances and ONLY enable 'worldSpace' export on the selected instances - leaving the others as they are per instance.

Yeah, I agree, but will leave to @antirotor and @mkolar

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -3,15 +3,24 @@
...
+    NumberDef,
+    EnumDef,
+    UISeparatorDef,
+    UILabelDef,
+)
+from ayon_core.pipeline.publish import OpenPypePyblishPluginMixin
 
 
-class ExtractAlembic(publish.Extractor):
+class ExtractAlembic(publish.Extractor, OpenPypePyblishPluginMixin):
```
iLLiCiTiT, line 23:
```suggestion
class ExtractAlembic(publish.Extractor, AYONPyblishPluginMixin):
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +201,223 @@ def process(self, instance):
...
     families = ["animation"]
 
     def get_members_and_roots(self, instance):
-
         # Collect the out set nodes
         out_sets = [node for node in instance if node.endswith("out_SET")]
         if len(out_sets) != 1:
-            raise RuntimeError("Couldn't find exactly one out_SET: "
-                               "{0}".format(out_sets))
+            raise RuntimeError(
```
iLLiCiTiT, line 411:
```suggestion
            raise KnownPublishError(
```

client/ayon_core/hosts/maya/api/alembic.py
```diff
@@ -0,0 +1,306 @@
+import json
+import logging
+import os
+
+from maya import cmds  # noqa
+
+from openpype.hosts.maya.api.lib import evaluation
```
iLLiCiTiT, line 7:
```suggestion
from ayon_core.hosts.maya.api.lib import evaluation
```

iLLiCiTiT:
Note: At this moment we can't make sure that correct version of maya code will get correct version of settings. This might need to wait for maya to be moved.

server_addon/maya/server/settings/publishers.py
tokejepsen: > Side question - is the "pre roll" just sampled for the export so that it runs from there. Or is it actually included and written into the alembic as well?

Think pre roll is just sampled for the export. Think its meant for simulation run up.

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +385,146 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+            "**Dont Skip Unwritten Frames** - When evaluating multiple "
+            "translate jobs, this decides whether to evaluate frames between "
+            "jobs when there is a gap in their frame ranges.\n"
+
+            "**Euler Filter** - Apply Euler filter while sampling rotations.\n"
+
+            "**No Normals** - Present normal data for Alembic poly meshes will"
+            " not be written.\n"
+
+            "**Pre Roll** - This frame range will not be sampled.\n"
```
tokejepsen, line 408:
```suggestion
            "Pre Roll - When enabled, the pre roll start frame is used to pre "
            "roll the export for.\n"
```

server_addon/maya/server/settings/publishers.py
tokejepsen: > Interesting - since the flag I suppose would be the reverse, when enabled then it does get written? Anyway, let's rewrite it to be clearer: When enabled, the pre roll start frame is used to pre roll the export for.

Done

antirotor:
> > Just wanted to add - by having them as checkboxes it also simplifies 'multiselection' editing for a specific flag on multiple instances. With the multiselection enum when having multiple instances selected you're essentially in multiselection editing picking ALL flags for ALL selected instances - but with individual checkboxs it is actually possible to select multiple instances and ONLY enable 'worldSpace' export on the selected instances - leaving the others as they are per instance.
> 
> Yeah, I agree, but will leave to @antirotor and @mkolar

I vote for the checkboxes. We can then revisit this when doing UI changes in the publisher.

client/ayon_core/hosts/maya/plugins/publish/validate_alembic_options_defaults.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @classmethod
+    def repair(cls, instance):
+        # Find create instance twin.
+        create_context = instance.context.data["create_context"]
+        create_instance = None
+        for Instance in create_context.instances:
+            if Instance.data["instance_id"] == instance.data["instance_id"]:
+                create_instance = Instance
+                break
```
BigRoy, line 81: Don't think make it uppercase first letter is a good python practice.
```suggestion
        for context_instance in create_context.instances:
            if context_instance .data["instance_id"] == instance.data["instance_id"]:
                create_instance = context_instance
                break
```

Or:
```suggestion
        create_instance = next((
            inst for inst in create_context.instances
            if inst.data["instance_id"] == instance.data["instance_id"]),
            None
        )
```

We could also remove the assert itself and instead just let `next()` error if we are pretty sure it _should_ exist.

```suggestion
        create_instance = next(
            inst for inst in create_context.instances
            if inst.data["instance_id"] == instance.data["instance_id"]
        )
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +386,166 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+        title="No Normals",
+        description=(
+            "Present normal data for Alembic poly meshes will not be written."
+        )
+    )
+    preRoll: bool = SettingsField(
+        title="Pre Roll",
+        description=(
+            "When enabled, the pre roll start frame is used to pre roll the "
+            "export for."
```
BigRoy, line 430: Can we add in the description whether that 'pre roll' is included in the exported data or not. I'm a bit unsure myself. Is it just starting from there for simulation purposes for the export, or is it included in the exported data?

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +386,166 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+        )
+    )
+    writeVisibility: bool = SettingsField(
+        title="Write Visibility",
+        description=(
+            "Visibility state will be stored in the Alembic file. Otherwise "
+            "everything written out is treated as visible."
+        )
+    )
+    overrides: list[str] = SettingsField(
```
BigRoy, line 542: I wonder if this key `overrides` is too generic? Should it be `user_editable` or `user_configurable` or `editable_in_publisher`?

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -58,49 +107,81 @@ def process(self, instance):
...
+            ),
+            "renderableOnly": attribute_values.get(
+                "renderableOnly", self.renderableOnly
+            ),
+            "root": root,
+            "selection": True,
+            "startFrame": start,
+            "step": instance.data.get(
+                "creator_attributes", {}
+            ).get("step", 1.0), #missing
```
BigRoy, line 139:
```suggestion
            "step": instance.data.get(
                "creator_attributes", {}
            ).get("step", 1.0), # missing
```

Do we still need to do something about this `missing`?

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,156 @@
...
+            ),
+        ]
+    )
+
+    return defs
+
+
+def extract_alembic_attributes(node_data, class_name):
+    """This is a legacy transfer of creator attributes to publish attributes
+    for ExtractAlembic/ExtractAnimation plugin.
```
BigRoy, line 40: To me this function name does not describe what it does. e.g. `convert_legacy_alembic_creator_attributes` would be better maybe?

client/ayon_core/hosts/maya/plugins/publish/validate_alembic_options_defaults.py
```diff
@@ -0,0 +1,94 @@
...
+
+        assert create_instance is not None
+
+        # Set the settings values on the create context then save to workfile.
+        publish_attributes = instance.data["publish_attributes"]
+        plugin_name = cls._get_plugin_name(publish_attributes)
+        attributes = cls._get_publish_attributes(instance)
+        settings = cls._get_settings(instance.context)
+        create_publish_attributes = create_instance.data["publish_attributes"]
+        for key in attributes.keys():
```
BigRoy, line 91:
```suggestion
        for key in attributes:
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +386,166 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+            "Present normal data for Alembic poly meshes will not be written."
+        )
+    )
+    preRoll: bool = SettingsField(
+        title="Pre Roll",
+        description=(
+            "When enabled, the pre roll start frame is used to pre roll the "
+            "export for."
+        )
+    )
```
tokejepsen, line 432:
```suggestion
    preRoll: bool = SettingsField(
        title="Pre Roll",
        description=(
            "When enabled, the pre roll start frame is used to pre roll the "
            "When enabled, the pre roll start frame is used to being the "
            "evaluation of the mesh. From the pre roll start frame to the "
            "alembic start frame, will not be written to disk. This can be "
            "used for simulation run up."
        )
    )
```

server_addon/maya/server/settings/publishers.py
tokejepsen: Suggested an expand description.

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +229,288 @@ def process(self, instance):
...
+        # The Arguments that can be modified by the Publisher
+        overrides = set(cls.overrides)
+        for key, value in override_defs.items():
+            if key not in overrides:
+                continue
+
+            defs.append(value)
+
+        defs.append(
+            UISeparatorDef("sep_alembic_options")
```
BigRoy, line 490: Should we include a check whether any overrides were enabled to begin with - otherwise one might see an "empty alembic export options" section?

server_addon/maya/server/settings/publishers.py
tokejepsen: Dont mind changing it to something else, just need a decision.

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
tokejepsen:
```suggestion
            "step": instance.data.get(
                "creator_attributes", {}
            ).get("step", 1.0),
```

server_addon/maya/server/settings/publishers.py
BigRoy: @iLLiCiTiT @antirotor @mkolar any preference?

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,156 @@
...
+
+        for node in cached_subsets.get(self.identifier, []):
+            node_data = self.read_instance_node(node)
+
+            node_data = extract_alembic_attributes(
+                node_data, "ExtractAnimation"
+            )
+
+            created_instance = CreatedInstance.from_existing(node_data, self)
+            self._add_instance_to_context(created_instance)
```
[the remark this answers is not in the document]
tokejepsen: doen

client/ayon_core/hosts/maya/plugins/publish/collect_animation.py
```diff
@@ -59,3 +59,7 @@ def process(self, instance):
         if instance.data.get("farm"):
             instance.data["families"].append("publish.farm")
 
+        # User defined attributes.
+        instance.data["includeUserDefinedAttributes"] = (
+            instance.data["creator_attributes"]["includeUserDefinedAttributes"]
+        )
```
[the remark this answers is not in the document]
tokejepsen: done

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
tokejepsen: done

client/ayon_core/hosts/maya/plugins/publish/collect_pointcache.py
```diff
@@ -45,3 +45,8 @@ def process(self, instance):
         if proxy_set:
             instance.remove(proxy_set)
             instance.data["setMembers"].remove(proxy_set)
+
+        # User defined attributes.
+        instance.data["includeUserDefinedAttributes"] = (
+            instance.data["creator_attributes"]["includeUserDefinedAttributes"]
+        )
```
BigRoy, line 52:
```suggestion
```

client/ayon_core/hosts/maya/plugins/publish/collect_user_defined_attributes.py
```diff
@@ -14,7 +14,10 @@ class CollectUserDefinedAttributes(pyblish.api.InstancePlugin):
     def process(self, instance):
 
         # Collect user defined attributes.
-        if not instance.data.get("includeUserDefinedAttributes", False):
+        if not instance.data.get(
+            "includeUserDefinedAttributes",
+            instance.data["creator_attributes"]["includeUserDefinedAttributes"]
+        ):
```
BigRoy, line 20: This basically enforces the `["creator_attributes"]["includeUserDefinedAttributes"]` key to exist. I don't see any other usages of `instance.data["includeUserDefinedAttributes"]` in the code base that still require it to exist in `instance.data` so I would just refactor this to using the creator attributes value instead.

```suggestion
        if not instance.data["creator_attributes"].get(
            "includeUserDefinedAttributes"
        ):
```

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,135 @@
...
+                "includeParentHierarchy",
+                label="Include Parent Hierarchy",
+                tooltip=(
+                    "Whether to include parent hierarchy of nodes in the "
+                    "publish instance."
+                )
+            ),
+            BoolDef(
+                "includeUserDefinedAttributes",
+                label="Include User Defined Attributes"
```
BigRoy, line 30: Would a tooltip be nice? Maybe this describes it?
```suggestion
                label="Include User Defined Attributes",
                tooltip=(
                    "Whether to include all custom maya attributes found "
                    "on nodes as attributes in the Alembic data."
                )
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
tokejepsen: done

client/ayon_core/hosts/maya/plugins/publish/validate_alembic_options_defaults.py
```diff
@@ -0,0 +1,94 @@
...
+    def repair(cls, instance):
+        # Find create instance twin.
+        create_context = instance.context.data["create_context"]
+        create_instance = None
+        create_instance = next(
+            inst for inst in create_context.instances
+            if inst.data["instance_id"] == instance.data["instance_id"]
+        )
+
+        assert create_instance is not None
```
BigRoy, line 83:
```suggestion
        create_instance = next(
            inst for inst in create_context.instances
            if inst.data["instance_id"] == instance.data["instance_id"]
        )
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +386,169 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+    preRollStartFrame: int = SettingsField(
+        title="Pre Roll Start Frame",
+        description=(
+            "The frame to start scene evaluation at.  This is used to set the "
+            "starting frame for time dependent translations and can be used to"
+            " evaluate run-up that isn't actually translated.\n"
+            "NOTE: preRoll needs to be enabled in the export flags for this "
+            "start frame to be considered."
```
BigRoy, line 443:
```suggestion
            "NOTE: Pre Roll needs to be enabled for this start frame "
            "to be considered."
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +386,169 @@ class ExtractAlembicModel(BaseSettingsModel):
         description="List of attribute prefixes for attributes that will be "
                     "included in the alembic export.",
     )
+    dataFormat: str = SettingsField(
```
BigRoy, line 389: I understand that these are somewhat Alphabetically sorted - but I do feel it's more understandable to take some liberty and sort them somewhat more by category:

image

E.g. I would:
- move the attribute related settings together: bake attributes, custom attributes and user attributes
- move python and mel script callbacks together

That would likely also make the list of checkboxes sit nicer together without text fields in-between making it a cleaner list and easier to read.

Any thoughts?

BigRoy: And potentially doing the same in the Publisher UI:

image

May be nicer having the attributes settings lower there?

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -152,24 +229,291 @@ def process(self, instance):
...
+    @classmethod
+    def get_attribute_defs(cls):
+        if not cls.overrides:
+            return []
+
+        override_defs = {
+            "attr": TextDef(
+                "attr",
+                label="Custom Attributes",
+                placeholder="attr1;attr2",
```
BigRoy, line 241: Match placeholder text of `attrPrefix`
```suggestion
                placeholder="attr1; attr2; ...",
```

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -27,8 +37,37 @@ class ExtractAlembic(publish.Extractor):
...
+    preRoll = False
+    preRollStartFrame = 0
+    pythonPerFrameCallback = ""
+    pythonPostJobCallback = ""
+    renderableOnly = False
+    stripNamespaces = True
+    uvsOnly = False
+    uvWrite = False
+    userAttr = ""
+    userAttrPrefix = ""
```
BigRoy, line 61: I think we are lacking the usage of these attributes in the extraction logic?

server_addon/maya/server/settings/publishers.py
```diff
@@ -1371,6 +1610,56 @@ class PublishersModel(BaseSettingsModel):
...
+        "eulerFilter": False,
+        "melPerFrameCallback": "",
+        "melPostFrameCallback": "",
+        "noNormals": False,
+        "overrides": [
+            "attr",
+            "attrPrefix",
+            "renderableOnly",
+            "userAttr",
+            "userAttrPrefix",
```
BigRoy, line 1638: I believe `userAttr` is new from this PR and it can be quite confusing what the difference is between `attr` and `userAttr` in Alembic's export logic. I wonder whether we should keep the `user*` attribute equivalent out of the defaults from being exposed to be more backwards compatible and also avoid confusing artists over time seeing both `attr` and `userAttr` fields in the publisher UI.

I for one here wouldn't 'quickly' know where to type the attribute I want to export:
image

Is it a User attribute or a custom attribute? :)

BigRoy:
```suggestion
            "attr",
            "attrPrefix",
            "renderableOnly",
```

server_addon/maya/server/settings/publishers.py
tokejepsen: Sure, done.

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -473,8 +376,106 @@ def get_attribute_defs(cls):
...
+                "preRollStartFrame",
+                label="Pre Roll Start Frame",
+                tooltip=(
+                    "The frame to start scene evaluation at. This is used"
+                    " to set the starting frame for time dependent "
+                    "translations and can be used to evaluate run-up that"
+                    " isn't actually translated."
+                ),
+                default=cls.preRollStartFrame
+            ),
```
BigRoy, line 390: Can we put `Pre Roll` toggle in front of this one? They are closely related? (I know they were together before this last commit - just anticipated that if they were moved that they would be moved together.)

tokejepsen: Done

server_addon/maya/server/settings/publishers.py
```diff
@@ -309,6 +482,73 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+            "specified is evaluated as a Mel command."
+        )
+    )
+    melPostFrameCallback: str = SettingsField(
+        title="Mel Post Frame Callback",
+        description=(
+            "When the translation has finished the string specified is "
+            "evaluated as a Mel command."
+        )
+    )
```
BigRoy, line 530: I suspect this needs to be post job callback?
```suggestion
    melPostJobCallback: str = SettingsField(
        title="Mel Post Job Callback",
        description=(
            "When the translation has finished the string specified is "
            "evaluated as a Mel command."
        )
    )
```

also see: https://github.com/alembic/alembic/blob/master/maya/AbcExport/AbcExport.cpp#L449C48-L449C66

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
```diff
@@ -41,66 +81,122 @@ def process(self, instance):
...
             # Set the root nodes if we don't want to include parents
             # The roots are to be considered the ones that are the actual
             # direct members of the set
-            options["root"] = roots
+            root = roots
 
-        if int(cmds.about(version=True)) >= 2017:
-            # Since Maya 2017 alembic supports multiple uv sets - write them.
-            options["writeUVSets"] = True
+        kwargs = {
```
BigRoy, line 130: Are we lacking the usage of the python and mel per frame and post job callbacks? If so, let's fix and also include in testing notes that's something we want to specifically test?

server_addon/maya/server/settings/publishers.py
```diff
@@ -1371,6 +1610,54 @@ class PublishersModel(BaseSettingsModel):
...
+        "attr": "",
+        "attrPrefix": "",
+        "autoSubd": False,
+        "bake_attributes": [],
+        "bake_attribute_prefixes": [],
+        "dataFormat": "ogawa",
+        "dontSkipUnwrittenFrames": False,
+        "eulerFilter": False,
+        "melPerFrameCallback": "",
+        "melPostFrameCallback": "",
```
BigRoy, line 1631:
```suggestion
        "melPostJobCallback": "",
```

server_addon/maya/server/settings/publishers.py
```diff
@@ -299,6 +349,129 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+        )
+    )
+    dontSkipUnwrittenFrames: bool = SettingsField(
+        title="Dont Skip Unwritten Frames",
+        description=(
+            "When evaluating multiple translate jobs, this decides whether to "
+            "evaluate frames between jobs when there is a gap in their frame "
+            "ranges."
+        )
+    )
```
BigRoy, line 370: When is this setting useful? As far as I know, the way we are exporting we are never combing a single `AbcExport` call with multiple export jobs at once - as such, this would never be used?

server_addon/maya/server/settings/publishers.py
```diff
@@ -299,6 +349,129 @@ class ExtractAlembicModel(BaseSettingsModel):
...
+    eulerFilter: bool = SettingsField(
+        title="Euler Filter",
+        description="Apply Euler filter while sampling rotations."
+    )
+    noNormals: bool = SettingsField(
+        title="No Normals",
+        description=(
+            "Present normal data for Alembic poly meshes will not be written."
+        )
+    )
```
BigRoy, line 380: Out of curiosity - what's the difference between **No Normals** and **Write Normals**? I feel like they are both the same toggle, but the inverse?

A quick investigation shows that the argument is unused [in this function](https://github.com/ynput/ayon-core/blob/8dcda0a6b9e4f179e7a61d8f9075a6e2bb3035c4/client/ayon_core/hosts/maya/api/alembic.py#L73)

There also is no `writeNormals` in [`alembic/maya/AbcExport.cpp`](https://github.com/alembic/alembic/blob/master/maya/AbcExport/AbcExport.cpp).

I have a feeling that originally we had `writeNormals` inside OpenPype as the equivalent (but inverse) of `noNormals`?

As such, one of these two should be removed? (Likely the one that wasn't used before for backwards compatibility?)

BigRoy (commented):
Settings look good - I did have some remaining comments on what I think may be actual bugs/issues.

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
tokejepsen: Good spot! Have added some extra missing ones.

server_addon/maya/server/settings/publishers.py
tokejepsen: gone

server_addon/maya/server/settings/publishers.py
tokejepsen: > As such, one of these two should be removed? (Likely the one that wasn't used before for backwards compatibility?)

So remove `NoNormals`?

tokejepsen: done

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
iLLiCiTiT: Yep, multiselection would change the look of it.

server_addon/maya/server/settings/publishers.py
iLLiCiTiT: I have no idea what is this used for so can't make correct opinion, but agree that `overrides` is too generic.

BigRoy:
The fact that this PR comes with over 220 commits and an interesting list of participants:
image

Is that something we should worry about? No idea if that influences changelogs in any way, etc. Likely the commits aren't an issue - just wondering because Github was showing e.g. some commits to blender code but no actual changes to files there). Likely comes from OpenPype -> AYON port and how maybe some thing were 'cherry picked' or included other code that got reverted in this PR or whatever?

Anyway, if the participants is of no concern - then ignore my question.

client/ayon_core/hosts/maya/api/alembic.py
```diff
@@ -0,0 +1,306 @@
...
+
+log = logging.getLogger(__name__)
+
+# The maya alembic export types
+ALEMBIC_ARGS = {
+    "attr": (list, tuple),
+    "attrPrefix": (list, tuple),
+    "autoSubd": bool,
+    "dataFormat": str,
+    "dontSkipUnwrittenFrames": bool,
```
BigRoy, line 17: Should we remove this? `extract_alembic` takes no `kwargs` and also does not take this argument?
```suggestion
```

client/ayon_core/hosts/maya/api/alembic.py
```diff
@@ -0,0 +1,306 @@
...
+            object with the namespace taco:foo:bar appears as bar in the
+            Alembic file.
+
+        verbose (bool): When on, outputs frame number information to the
+            Script Editor or output window during extraction.
+
+        preRollStartFrame (float): The frame to start scene
+            evaluation at.  This is used to set the starting frame for time
+            dependent translations and can be used to evaluate run-up that
+            isn't actually translated. Defaults to 0.
```
BigRoy, line 175: Should we care about the fact that the arguments of the function are sorted different than the arguments shown in the docstring?

```suggestion
    file,
    attr=None,
    attrPrefix=None,
    dataFormat="ogawa",
    endFrame=None,
    eulerFilter=True,
    frameRange="",
    noNormals=False,
    preRoll=False,
    preRollStartFrame=0,
    renderableOnly=False,
    root=None,
    selection=True,
    startFrame=None,
    step=1.0,
    stripNamespaces=True,
    uvWrite=True,
    verbose=False,
    wholeFrameGeo=False,
    worldSpace=False,
    writeColorSets=False,
    writeCreases=False,
    writeNormals=False,
    writeFaceSets=False,
    writeUVSets=False,
    writeVisibility=False
):
    """Extract a single Alembic Cache.

    This extracts an Alembic cache using the `-selection` flag to minimize
    the extracted content to solely what was Collected into the instance.

    Arguments:
        file (str): The filepath to write the alembic file to.

        attr (list of str, optional): A specific geometric attribute to write
            out. Defaults to [].

        attrPrefix (list of str, optional): Prefix filter for determining which
            geometric attributes to write out. Defaults to ["ABC_"].

        dataFormat (str): The data format to use for the cache,
                          defaults to "ogawa"

        endFrame (float): End frame of output. Ignored if `frameRange`
            provided.

        eulerFilter (bool): When on, X, Y, and Z rotation data is filtered with
            an Euler filter. Euler filtering helps resolve irregularities in
            rotations especially if X, Y, and Z rotations exceed 360 degrees.
            Defaults to True.

        frameRange (tuple or str): Two-tuple with start and end frame or a
            string formatted as: "startFrame endFrame". This argument
            overrides `startFrame` and `endFrame` arguments.

        noNormals (bool): When on, normal data from the original polygon
            objects is not included in the exported Alembic cache file.

        preRoll (bool): This frame range will not be sampled.
            Defaults to False.

        preRollStartFrame (float): The frame to start scene
            evaluation at.  This is used to set the starting frame for time
            dependent translations and can be used to evaluate run-up that
            isn't actually translated. Defaults to 0.

        renderableOnly (bool): When on, any non-renderable nodes or hierarchy,
            such as hidden objects, are not included in the Alembic file.
            Defaults to False.

        root (list of str): Maya dag path which will be parented to
            the root of the Alembic file. Defaults to [], which means the
            entire scene will be written out.

        selection (bool): Write out all all selected nodes from the
            active selection list that are descendents of the roots specified
            with -root. Defaults to False.

        startFrame (float): Start frame of output. Ignored if `frameRange`
            provided.

        step (float): The time interval (expressed in frames) at
            which the frame range is sampled. Additional samples around each
            frame can be specified with -frs. Defaults to 1.0.

        stripNamespaces (bool): When on, any namespaces associated with the
            exported objects are removed from the Alembic file. For example, an
            object with the namespace taco:foo:bar appears as bar in the
            Alembic file.

        uvWrite (bool): When on, UV data from polygon meshes and subdivision
            objects are written to the Alembic file. Only the current UV map is
            included.

        verbose (bool): When on, outputs frame number information to the
            Script Editor or output window during extraction.

        wholeFrameGeo (bool): Data for geometry will only be written
            out on whole frames. Defaults to False.

        worldSpace (bool): When on, the top node in the node hierarchy is
            stored as world space. By default, these nodes are stored as local
            space. Defaults to False.

        writeColorSets (bool): Write all color sets on MFnMeshes as
            color 3 or color 4 indexed geometry parameters with face varying
            scope. Defaults to False.

        writeCreases (bool): If the mesh has crease edges or crease
            vertices, the mesh (OPolyMesh) would now be written out as an OSubD
            and crease info will be stored in the Alembic file. Otherwise,
            creases info won't be preserved in Alembic file unless a custom
            Boolean attribute SubDivisionMesh has been added to mesh node and
            its value is true. Defaults to False.

        writeFaceSets (bool): Write all Face sets on MFnMeshes.
            Defaults to False.

        writeUVSets (bool): Write all uv sets on MFnMeshes as vector
            2 indexed geometry parameters with face varying scope. Defaults to
            False.

        writeVisibility (bool): Visibility state will be stored in
            the Alembic file.  Otherwise everything written out is treated as
            visible. Defaults to False.
```

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,139 @@
...
+        node_data = super(CreatePointCache, self).read_instance_node(node)
+        node_data = convert_legacy_alembic_creator_attributes(
+            node_data, "ExtractAlembic"
+        )
+        return node_data
+
+    def get_instance_attr_defs(self):
+        super(CreatePointCache, self).get_instance_attr_defs()
+        defs = _get_animation_attr_defs(self)
+        return defs
```
iLLiCiTiT, line 129:
```suggestion
        return _get_animation_attr_defs(self)
```

client/ayon_core/hosts/maya/plugins/publish/validate_alembic_options_defaults.py
```diff
@@ -0,0 +1,91 @@
...
+                return key
+
+    @classmethod
+    def _get_settings(cls, context):
+        maya_settings = context.data["project_settings"]["maya"]
+        settings = maya_settings["publish"]["ExtractAlembic"]
+        return settings
+
+    @classmethod
+    def _get_publish_attributes(cls, instance):
```
iLLiCiTiT, line 35: this looks like a duplicate of `get_attr_values_from_data_for_plugin`.

client/ayon_core/hosts/maya/plugins/publish/validate_alembic_options_defaults.py
```diff
@@ -0,0 +1,91 @@
...
+            raise PublishValidationError("\n".join(errors))
+
+    @classmethod
+    def repair(cls, instance):
+        # Find create instance twin.
+        create_context = instance.context.data["create_context"]
+        create_instance = next(
+            inst for inst in create_context.instances
+            if inst.data["instance_id"] == instance.data["instance_id"]
+        )
```
iLLiCiTiT, line 80:
```suggestion
        create_instance =  create_context.get_instance_by_id(
            instance.data["instance_id"])
        )
```

client/ayon_core/hosts/maya/api/alembic.py
BigRoy: By the way, this also still takes both `noNormals` and `writeNormals`? One should be removed?

`writeNormals` still seems unused in the function and as such does nothing?

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def read_instance_node(self, node):
+        node_data = super(CreatePointCache, self).read_instance_node(node)
+        node_data = convert_legacy_alembic_creator_attributes(
+            node_data, "ExtractAlembic"
+        )
+        return node_data
+
+    def get_instance_attr_defs(self):
+        super(CreatePointCache, self).get_instance_attr_defs()
```
BigRoy, line 127: Whatever the `super` call returns here is unused - even if it did return any attribute definitions the returned result is unused.

client/ayon_core/hosts/maya/plugins/publish/collect_animation.py
```diff
@@ -58,4 +58,3 @@ def process(self, instance):
 
         if instance.data.get("farm"):
             instance.data["families"].append("publish.farm")
-
```
BigRoy, line 61: I think this may have removed the new line by accident?

client/ayon_core/hosts/maya/plugins/publish/collect_animation.py
```diff
@@ -17,7 +17,7 @@ class CollectAnimationOutputGeometry(pyblish.api.InstancePlugin):
 
     order = pyblish.api.CollectorOrder + 0.4
     families = ["animation"]
-    label = "Collect Animation Output Geometry"
+    label = "Collect Animation"
```
BigRoy, line 20:
```suggestion
    label = "Collect Animation Output Geometry"
```

With the collect user defined attributes logic removed from here again I think we can just leave this file untouched and remove the changes?

client/ayon_core/hosts/maya/plugins/publish/extract_pointcache.py
BigRoy:
```suggestion
            UISeparatorDef("sep_alembic_options_end")
```

client/ayon_core/hosts/maya/plugins/publish/validate_alembic_options_defaults.py
BigRoy: It does indeed - but it does so by string instead of plugin `class` which can't really be used here since this Validator in particular wouldn't have access to the `ExtractAlembic` and `ExtractAnimation` classes because usually these plug-ins aren't on pythonpath as an importable file, right?

Anyway, plus the fact that this decides on which of the two plug-in it should apply to instead of running for one of them.

Admittedly, it's definitely not a nice pattern what is done here that way. The only other way I see if creating two Validator plug-ins that dedicate to run over `animation` and `pointcache` family individually. Then at least it's clearer which family belong to which extractors and we do not need the "dynamic" decision on which of the two to use to get the settings from?

tokejepsen: This validator has been split now.

client/ayon_core/hosts/maya/api/alembic.py
tokejepsen: `writeNormals` has been removed now.

client/ayon_core/hosts/maya/plugins/create/create_animation_pointcache.py
tokejepsen: Fixed.

client/ayon_core/hosts/maya/plugins/publish/collect_animation.py
tokejepsen: New line is still in file.

BigRoy:
Massive work everyone! :)

BigRoy:
Getting this error after this PR:
```
  File "C:\Users\Maqina-02\AppData\Local\Ynput\AYON\dependency_packages\ayon_2404022047_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "C:\Users\Maqina-02\AppData\Local\Ynput\AYON\addons\core_0.3.1-dev.3\ayon_core\hosts\maya\plugins\publish\validate_alembic_options_defaults.py", line 46, in process
    attributes = self._get_publish_attributes(instance)
  File "C:\Users\Maqina-02\AppData\Local\Ynput\AYON\addons\core_0.3.1-dev.3\ayon_core\hosts\maya\plugins\publish\validate_alembic_options_defaults.py", line 33, in _get_publish_attributes
    cls.plugin_name(
TypeError: 'str' object is not callable
```

BigRoy:
Next error after that is:
```python
DEBUG: Extracting pointcache..
  File "C:\Users\Maqina-02\AppData\Local\Ynput\AYON\dependency_packages\ayon_2404022047_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "C:\Users\Maqina-02\AppData\Local\Ynput\AYON\addons\core_0.3.1-dev.3\ayon_core\hosts\maya\plugins\publish\extract_pointcache.py", line 165, in process
    "writeCreases", self.writeCreases
AttributeError: 'ExtractAnimation' object has no attribute 'writeCreases'
```

BigRoy:
Next error is:
```python
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2403071252_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-core\client\ayon_core\hosts\maya\plugins\publish\extract_pointcache.py", line 218, in process
    extract_alembic(**kwargs)
TypeError: extract_alembic() got an unexpected keyword argument 'userAttr'
```

I get the feeling this PR was merged too quickly? :D

server_addon/maya/server/settings/publishers.py
```diff
@@ -1371,6 +1600,52 @@ class PublishersModel(BaseSettingsModel):
...
+            "writeNormals"
+        ],
+        "preRoll": False,
+        "preRollStartFrame": 0,
+        "pythonPerFrameCallback": "",
+        "pythonPostJobCallback": "",
+        "renderableOnly": False,
+        "stripNamespaces": True,
+        "uvsOnly": False,
+        "uvWrite": False,
```
BigRoy, line 1638: Do we really feel this is a sensible studio default we should offer from AYON by default. With this `False` by default Alembics will be written **without UVs** - pretty sure we'll want to flip that. Right @tokejepsen @mkolar ?

server_addon/maya/server/settings/publishers.py
```diff
@@ -1371,6 +1600,52 @@ class PublishersModel(BaseSettingsModel):
...
+            "model",
+            "vrayproxy.alembic"
+        ],
+        "attr": "",
+        "attrPrefix": "",
+        "autoSubd": False,
+        "bake_attributes": [],
+        "bake_attribute_prefixes": [],
+        "dataFormat": "ogawa",
+        "eulerFilter": False,
```
BigRoy, line 1618: This defaulted to True before, by making this `False` with this PR this now means there is a difference in default behavior.

server_addon/maya/server/settings/publishers.py
```diff
@@ -1371,6 +1600,52 @@ class PublishersModel(BaseSettingsModel):
...
+        "userAttrPrefix": "",
+        "verbose": False,
+        "visibleOnly": False,
+        "wholeFrameGeo": False,
+        "worldSpace": True,
+        "writeColorSets": False,
+        "writeFaceSets": False,
+        "writeNormals": True,
+        "writeUVSets": False,
+        "writeVisibility": False
```
BigRoy, line 1649: This defaulted to True before this PR - with this now False any animated visibilities (or even any visilities whatsoever) will not be published along. Do we feel that's a sensible default value for AYON?

client/ayon_core/hosts/maya/api/alembic.py
```diff
@@ -0,0 +1,305 @@
...
+    file,
+    attr=None,
+    attrPrefix=None,
+    dataFormat="ogawa",
+    endFrame=None,
+    eulerFilter=True,
+    frameRange="",
+    noNormals=False,
+    preRoll=False,
+    preRollStartFrame=0,
```
BigRoy, line 59: Pretty sure this functionality does not work?

I'm seeing this in my extraction logs:
```
# Warning: Ignoring unsupported flag: -preRollStartFrame
# Warning: Ignoring unsupported flag: 0
```

BigRoy: It seems `preRollStartFrame` is not a job argument but a direct argument to `AbcExport`. See logic [here](https://github.com/tokejepsen/maya-alembic-export/blob/master/alembic_export.py#L544). So the current implementation here is broken.

PR #338, opened 2024-04-01
Maya: Use frame range from task attributes

MustafaJafar:
## Changelog Description
A bug fix for the issue reported here https://github.com/ynput/ayon-core/pull/323#pullrequestreview-1970044683
Frame Range values should be taken from task level attributes. 
image

## Testing notes:
1. Launch Maya 
2. Reset Frame Range

### DIFF
--- a/client/ayon_core/hosts/maya/api/lib.py
+++ b/client/ayon_core/hosts/maya/api/lib.py
@@ -2525,7 +2525,7 @@ def get_fps_for_current_context():
 
 
 def get_frame_range(include_animation_range=False):
-    """Get the current folder frame range and handles.
+    """Get the current task frame range and handles.
 
     Args:
         include_animation_range (bool, optional): Whether to include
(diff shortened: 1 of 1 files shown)


moonyuet (approved):

m-u-r-p-h-y (approved):
works as expected

image

BigRoy:
Would be great to merge this relatively close to this PR https://github.com/ynput/ayon-core/pull/39 so the validator aligns with this as well.

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -2565,11 +2571,8 @@ def get_frame_range(include_animation_range=False):
...
         # keys. That is why these are excluded by default.
-        task_name = get_current_task_name()
+
         settings = get_project_settings(project_name)
-        task_entity = ayon_api.get_task_by_name(
-            project_name, folder_entity["id"], task_name
-        )
         task_type = None
         if task_entity:
             task_type = task_entity["taskType"]
```
BigRoy, line 2578: Looking at the the code above - where we now get attributes directly from the task entity I don't think there's ever a moment where `if task_entity` should be empty or `None` as such this if statement makes no sense anymore?
```suggestion
        task_type = task_entity["taskType"]
```

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -2525,33 +2525,39 @@ def get_fps_for_current_context():
...
     # Set frame start/end
     project_name = get_current_project_name()
     folder_path = get_current_folder_path()
+    task_name = get_current_task_name()
+
     folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
-    folder_attributes = folder_entity["attrib"]
+    task_entity = ayon_api.get_task_by_name(
+            project_name, folder_entity["id"], task_name
+        )
```
BigRoy, line 2548: It would be great if we could get to this task with 1 query instead of first requiring to query the current folder by name just for its id. @iLLiCiTiT any possibilities to do that?

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -2525,33 +2525,39 @@ def get_fps_for_current_context():
...
+        dict: Task's expected frame range values.
 
     """
 
     # Set frame start/end
     project_name = get_current_project_name()
     folder_path = get_current_folder_path()
+    task_name = get_current_task_name()
+
     folder_entity = ayon_api.get_folder_by_path(project_name, folder_path)
```
BigRoy, line 2545: We only need the `id`
```suggestion
    folder_entity = ayon_api.get_folder_by_path(
        project_name, 
        folder_path,
        fields={"id"})
```

client/ayon_core/hosts/maya/api/lib.py
iLLiCiTiT: Not right now, but I agree.

MustafaJafar: should we create an issue to track that ?

BigRoy (approved):
Didn't test this - but code looks good to me.

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -2525,33 +2525,42 @@ def get_fps_for_current_context():
...
+    task_entity = ayon_api.get_task_by_name(
+            project_name, folder_entity["id"], task_name
+        )
 
-    frame_start = folder_attributes.get("frameStart")
-    frame_end = folder_attributes.get("frameEnd")
+    task_attributes = task_entity["attrib"]
+
+    frame_start = task_attributes.get("frameStart")
+    frame_end = task_attributes.get("frameEnd")
```
BigRoy, line 2556: Actually - I believe these can be:
```suggestion
    frame_start = task_attributes["frameStart"]
    frame_end = task_attributes["frameEnd"]
```

Since those should always exist on AYON task entities. And thus the if statement below also becomes redundant.
Same goes for `handleStart` and `handleEnd` I believe.

PR #342, opened 2024-04-02
Max: Using task attributes for validate frame range and validate resolution setting

moonyuet:
## Changelog Description
Align to use the data from task entity in regard to other hosts

## Additional info
n/a

## Testing notes:
1. Launch Max
2. Create Render Instance/Review Instance
3. Publish


client/ayon_core/hosts/max/api/lib.py
```diff
@@ -356,6 +357,26 @@ def get_max_version():
...
+    task_name = get_current_task_name()
+    folder_entity = ayon_api.get_folder_by_path(
+        project_name, folder_path, fields={"id"})
+    task_entity = ayon_api.get_task_by_name(
+        project_name,
+        folder_entity["id"],
+        task_name,
+        fields={"attrib"}
+    )
+    return task_entity
```
BigRoy, line 377: I understand the re-use need of this - but I feel it's a bit odd that this is inside the max lib specifically. If we really need to do this often I feel like this is better implemented more globally in ayon core or in the ayon api. @iLLiCiTiT thoughts?

Also, the fact that this library function defaults to filtering fields to `attrib` only feels a bit confusing because the function name seems to hint it'll just return the task entity.

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -221,18 +226,15 @@ def reset_scene_resolution():
...
+    height = int(task_attributes["resolutionHeight"])
 
     set_scene_resolution(width, height)
 
 
-def get_frame_range(folder_entiy=None) -> Union[Dict[str, Any], None]:
+def get_frame_range(task_entity=None) -> Union[Dict[str, Any], None]:
     """Get the current folder frame range and handles.
 
     Args:
```
BigRoy, line 240: Docstring needs updating with the argument change.

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -242,20 +244,19 @@ def get_frame_range(folder_entiy=None) -> Union[Dict[str, Any], None]:
...
-    frame_start = folder_attributes.get("frameStart")
-    frame_end = folder_attributes.get("frameEnd")
+    if task_entity is None:
+        task_entity = get_current_task_entity()
+    task_attributes = task_entity["attrib"]
+    frame_start = task_attributes.get("frameStart")
+    frame_end = task_attributes.get("frameEnd")
 
     if frame_start is None or frame_end is None:
         return {}
```
BigRoy, line 254: These should really exist on AYON entities, if not - then that's a bug and should be crashed on as far as I know.
```suggestion
    frame_start = task_attributes["frameStart"]
    frame_end = task_attributes["frameEnd"]
```

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -242,20 +244,19 @@ def get_frame_range(folder_entiy=None) -> Union[Dict[str, Any], None]:
...
 
     frame_start = int(frame_start)
     frame_end = int(frame_end)
-    handle_start = int(folder_attributes.get("handleStart", 0))
-    handle_end = int(folder_attributes.get("handleEnd", 0))
+    handle_start = int(task_attributes.get("handleStart", 0))
+    handle_end = int(task_attributes.get("handleEnd", 0))
```
BigRoy, line 259:
```suggestion
    handle_start = int(task_attributes["handleStart"])
    handle_end = int(task_attributes["handleEnd"])
```

client/ayon_core/hosts/max/api/lib.py
moonyuet: yes, and I also find out that the task entity you suggested in other maya PRs is already getting attributes in the dict I assumed

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -281,9 +282,9 @@ def reset_frame_range(fps: bool = True):
             scene frame rate in frames-per-second.
     """
     if fps:
-        project_name = get_current_project_name()
-        project_entity = ayon_api.get_project(project_name)
-        fps_number = float(project_entity["attrib"].get("fps"))
+        task_entity = get_current_task_entity()
+        task_attributes = task_entity["attrib"]
+        fps_number = float(task_attributes.get("fps"))
```
BigRoy, line 287: The `.get` wouldn't help much since `float(None)` would fail anyway:

```
float(None)
# Error: float() argument must be a string or a real number, not 'NoneType'
# #   File "<maya console>", line 1, in <module>
# # TypeError: float() argument must be a string or a real number, not 'NoneType'
```
Also, these attributes should really exist in AYON entities and as far as I know are not optional:
```suggestion
        fps_number = float(task_attributes["fps"])
```

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -242,20 +244,19 @@ def get_frame_range(folder_entiy=None) -> Union[Dict[str, Any], None]:
...
-    frame_start = folder_attributes.get("frameStart")
-    frame_end = folder_attributes.get("frameEnd")
+    if task_entity is None:
+        task_entity = get_current_task_entity()
+    task_attributes = task_entity["attrib"]
+    frame_start = task_attributes["frameStart"]
+    frame_end = task_attributes["frameEnd"]
 
     if frame_start is None or frame_end is None:
         return {}
```
BigRoy, line 254: This can now be removed because it should never be true.

LiborBatek (commented):
Has been tested in ```max 2024``` and works ok for ```render``` instance

validates well against frame range set in the database for current folder

Screenshot 2024-04-03 103523

and also for ```Resolution```


Screenshot 2024-04-03 103101

My only concern is when using ```Review``` instance it does not validate against resolution and just ignores it...it seems that for this instance type we do not validate for resolution...I think we should include an optional settings for that too (as on ```render``` instance type)

Screenshot 2024-04-03 103403

moonyuet:
> Has been tested in `max 2024` and works ok for `render` instance
> 
> validates well against frame range set in the database for current folder
> 
> Screenshot 2024-04-03 103523
> and also for `Resolution`
> 
> Screenshot 2024-04-03 103101
> 
> My only concern is when using `Review` instance it does not validate against resolution and just ignores it...it seems that for this instance type we do not validate for resolution...I think we should include an optional settings for that too (as on `render` instance type)
> 
> Screenshot 2024-04-03 103403

It is supposed that it should be at the separate PR but since this PR is quite small, I guess I can also implement here as it is adding back review as product type

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -356,6 +351,26 @@ def get_max_version():
...
+        project_name, folder_path, fields={"id"})
+    task_entity = ayon_api.get_task_by_name(
+        project_name,
+        folder_entity["id"],
+        task_name,
+        fields={"attrib"}
+    )
+    return task_entity
+
+
```
BigRoy, line 373: @moonyuet this can now be removed with this PR merged: https://github.com/ynput/ayon-core/pull/353

```suggestion
```

client/ayon_core/hosts/max/api/lib.py
```diff
@@ -221,41 +226,31 @@ def reset_scene_resolution():
...
-
-    if frame_start is None or frame_end is None:
-        return {}
-
-    frame_start = int(frame_start)
-    frame_end = int(frame_end)
-    handle_start = int(folder_attributes.get("handleStart", 0))
-    handle_end = int(folder_attributes.get("handleEnd", 0))
+    if task_entity is None:
+        task_entity = get_current_task_entity()
```
BigRoy, line 248: This can now use the implementation added by this PR: https://github.com/ynput/ayon-core/pull/353

LiborBatek (changes requested):
The Resolution Validator does work for ```Render``` instance but not for ```Review``` instance.

Also the ```Repair``` action only fixes the Render Settings aka Render Resolution but not changing the ```Review``` instance resolution.

When fixed the ```Render``` instance resolution it Validates OK even if the ```Review``` instance still having wrong set resolution.

not sure if related but here is copy paste from Max listener aka console:

```Traceback (most recent call last):
    runner(*args)
  File "C:\Work\REPO\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_resolution_setting.py", line 31, in process
    raise PublishValidationError("Resolution Setting "
ayon_core.pipeline.publish.publish_plugins.PublishValidationError: Resolution Setting not matching resolution set on asset or shot.
>>> [ *** Discovered 1 abstract plugins```

Now ```Review``` having ```Validate Resolution``` which is fine and desired...


![Screenshot 2024-04-05 114539](https://github.com/ynput/ayon-core/assets/112623825/4fe450cf-b1c1-4074-b9af-38171432fa48)

moonyuet:
> The Resolution Validator does work for `Render` instance but not for `Review` instance.
> 
> Also the `Repair` action only fixes the Render Settings aka Render Resolution but not changing the `Review` instance resolution.
> 
> When fixed the `Render` instance resolution it Validates OK even if the `Review` instance still having wrong set resolution.
> 
> not sure if related but here is copy paste from Max listener aka console:
> 
> ```
>   File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2402141620_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
>     runner(*args)
>   File "C:\Work\REPO\ayon-core\client\ayon_core\hosts\max\plugins\publish\validate_resolution_setting.py", line 31, in process
>     raise PublishValidationError("Resolution Setting "
> ayon_core.pipeline.publish.publish_plugins.PublishValidationError: Resolution Setting not matching resolution set on asset or shot.
> >>> [ *** Discovered 1 abstract plugins
> ```
> 
> Now `Review` having `Validate Resolution` which is fine and desired...
> 
> Screenshot 2024-04-05 114539

I have separated validator for resolution setting in terms of render and review product type, as they dont get resolution data for exporting the sequences in a similar method.

iLLiCiTiT:
@moonyuet please update with latest develop. @LiborBatek please test again.

LiborBatek (changes requested):
I did new test round and found that

```Validate Frame Range``` works and also ```Repair``` action too

Screenshot 2024-04-17 105409

but for ```Validate Resolution``` besides validation itself which works ok, the ```Repair``` action can fix global render settings namely for ```Render``` instance types but when using this validator on ```Review``` instance then the repair fails on instance.

Screenshot 2024-04-17 105103

and after triggered ```Repair``` action resolution still the same / invalid on the publish ```Review``` Instance...

Screenshot 2024-04-17 105125

LiborBatek (approved):
Now all ```Validator```s for resolution and frame ranges working with ```Repair``` actions for both ```Render``` and ```Review``` instances. All good

Only minor thing, there is unneccessarily long name for validator of the review resolution as seen on scrngrab here

Screenshot 2024-04-17 134200

Should be imho just ```Validate Resolution``` or ```Validate Review Resolution```. 
Other than that good and can be merged then...

moonyuet:
I already renamed in the updated commit.
image

BigRoy:
> Now all `Validator`s for resolution and frame ranges working with `Repair` actions for both `Render` and `Review` instances. All good
> 
> Only minor thing, there is unneccessarily long name for validator of the review resolution as seen on scrngrab here
> Screenshot 2024-04-17 134200
> 
> Should be imho just `Validate Resolution` or `Validate Review Resolution`. Other than that good and can be merged then...

Just for context:
- In [maya it is **Validate Resolution**](https://github.com/ynput/ayon-core/blob/8549054098355f20296aca0f994c6a253b820377/client/ayon_core/hosts/maya/plugins/publish/validate_resolution.py#L19).**
- It's [**Validate Folder Resolution** in Fusion](https://github.com/ynput/ayon-core/blob/1b5d6f9001f9e6cd1a1ff30e1f228b2f189c7518/client/ayon_core/hosts/fusion/plugins/publish/validate_saver_resolution.py#L17).
- There is [**Validate Write Resolution** in Nuke](https://github.com/ynput/ayon-core/blob/ee6e42ec6c090eb2ded6302b6cb8e09ef2d55dc0/client/ayon_core/hosts/nuke/plugins/publish/validate_output_resolution.py#L26) but I believe that's slightly different validation. 
- In After Effects it's part of a larger [**Validate Scene Settings** validator](https://github.com/ynput/ayon-core/blob/d1a09aa46a3e726a7a6a1c8dff0005154c66984f/client/ayon_core/hosts/aftereffects/plugins/publish/validate_scene_settings.py#L60).

PR #345, opened 2024-04-02
3dsMax: Track by handles then object name

r42-chun:
## Changelog Description
Tracks object by accessing the handles instead of tracking them by the name in 3ds max.

## Additional info
When you load an object into a max scene, and you change the name of the object, when you manage your loaded asset (i.e changing versions of the object), because it is tracking via the name of the object, it errors. Instead of tracking by the name of the object, we can track it by it's unique 3dsmax handle.

## Testing notes:
1. Load object into scene via AYON load
2. Rename object
3. In AYON manage, change the version of the object. Checks if it updates correctly.


client/ayon_core/hosts/max/api/pipeline.py
```diff
@@ -240,10 +240,9 @@ def get_previous_loaded_object(container: str):
         node_list(list): list of nodes which are previously loaded
     """
     node_list = []
-    sel_list = rt.getProperty(container.modifiers[0].openPypeData, "sel_list")
-    for obj in rt.Objects:
-        if str(obj) in sel_list:
-            node_list.append(obj)
+    node_transform_monitor_list = rt.getProperty(container.modifiers[0].openPypeData, "all_handles")
```
BigRoy, line 243: Pretty sure this exceeds the allowed styling line length.
```suggestion
    node_transform_monitor_list = rt.getProperty(
        container.modifiers[0].openPypeData, "all_handles")
```

moonyuet (approved):
@r42-chun thank you  for the contribution, will be merged after the cosmetic fix as mentioned in the above comment

PR #346, opened 2024-04-02
Bug: add `headless` flag to prevent UI interactivity in burnins extraction

antirotor:
## Changelog Description
Add `--headless` flag to execution of burnin script from **Extract Burnins** to prevent any UI interaction.

## Additional info
This is fixing the issue where in certain cases **Extract Burnins** could pop some update windows or similar. Oh, and also fixing some typos.

## Testing notes:
It could be tested in cases where running launcher would cause update for example.

### DIFF
--- a/client/ayon_core/plugins/publish/extract_burnin.py
+++ b/client/ayon_core/plugins/publish/extract_burnin.py
@@ -27,7 +27,7 @@ class ExtractBurnin(publish.Extractor):
     Extractor to create video with pre-defined burnins from
     existing extracted video representation.
 
-    It will work only on represenations having `burnin = True` or
+    It will work only on representations having `burnin = True` or
     `tags` including `burnin`
     """
 
@@ -125,7 +125,7 @@ class ExtractBurnin(publish.Extractor):
 
             burnin_defs = copy.deepcopy(src_burnin_defs)
 
-            # Filter output definition by `burnin` represetation key
+            # Filter output definition by `burnin` representation key
             repre_linked_burnins = [
                 burnin_def
                 for burnin_def in burnin_defs
@@ -378,6 +378,7 @@ class ExtractBurnin(publish.Extractor):
                 # Prepare subprocess arguments
                 args = list(executable_args)
                 args.append(temporary_json_filepath)
+                args.append("--headless")
                 self.log.debug("Executing: {}".format(" ".join(args)))
 
                 # Run burnin script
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/publish/extract_burnin.py
```diff
@@ -378,6 +378,7 @@ def main_process(self, instance):
                 # Prepare subprocess arguments
                 args = list(executable_args)
                 args.append(temporary_json_filepath)
+                args.append("--headless")
```
BigRoy, line 381: Can we add a comment to this? I feel like it's a bit unclear in the future what this `--headless` is trying to do. Is it a flag for ffmpeg? or is it a flag for how we run openpype processes? (It's unclear to me which).

If it's one for the OpenPype runner logic then I feel this shouldn't be an arg to the `run_ayon_launcher_process` but maybe a clear argument one can pass to `run_ayon_launcher_process` to differentiate what is essentially arguments for the command you are running an which are just 'settings' for the ayon process launcher.

If this `--headless` is an FFMPEG argument, then ignore me.

antirotor: well, docstring for `run_ayon_launcher_process` says:

```
Args:
        *args (str): ayon-launcher cli arguments.
        **kwargs (Any): Keyword arguments for subprocess.Popen.
```
In this regard, `--headless` fits to the `*args` definition. It is indeed **ayon-launcher** flag, not related to ffmpeg.

iLLiCiTiT: I would just define it here https://github.com/ynput/ayon-core/pull/346/files#diff-04d034303d010fab8f908ef1699037a6087f7c85b41b0d279267c48fa958d654R216 instead of appending if all the time.

iLLiCiTiT: > clear argument one can pass to run_ayon_launcher_process to differentiate what is essentially arguments for the command you are running an which are just 'settings' for the ayon process launcher.

I don't see any advantage in that.

BigRoy: > well, docstring for `run_ayon_launcher_process` says:
> 
> ```
> Args:
>         *args (str): ayon-launcher cli arguments.
>         **kwargs (Any): Keyword arguments for subprocess.Popen.
> ```
> 
> In this regard, `--headless` fits to the `*args` definition. It is indeed **ayon-launcher** flag, not related to ffmpeg.

Actually - seeing it like that makes much more sense. It's  just an argument to AYON; not to FFMPEG.

iLLiCiTiT (approved):

PR #350, opened 2024-04-02
Applications: Live attributes update

iLLiCiTiT:
## Changelog Description
Update attribute values on settings save.

## Additional info
This helps to update the attribute live on settings change. It does work from this version up. The update does not require restart of server, but the live attributes update is available from server version 1.0.7 .

## Testing notes:
1. Make sure you have server 1.0.7 or newer.
2. Create packages on server addons from this PR.
3. Upload applications addon.
4. Use the version in production/staging bundle.
5. Add new vartiant of applications and save settings.
6. Go to any project anatomy and check if the variant is available in the attribute.


m-u-r-p-h-y (changes requested):
not working for me

application addon 0.1.9
server 1.0.8+202403281707

image

server_addon/applications/server/__init__.py
```diff
@@ -129,6 +148,85 @@ async def create_applications_attribute(self) -> bool:
...
+            "SELECT name, position FROM attributes ORDER BY position"
+        ):
+            # check if the required attributes are present
+            # (in that case, we don't need to add them)
+            # also keep track of the last attribute position
+            if row["name"] == apps_attrib_name:
+                apps_present = True
+            elif row["name"] == tools_attrib_name:
+                tools_present = True
+            last_index = row["position"]
```
BigRoy, line 184: Wouldn't this query be much more optimal to include a filter where name is either of the two values instead of returning all attributes? For example adding `WHERE name IN ("applications", "tools")`?

We seem to be only doing this to get the highest available index? Shouldn't there be better ways to insert as highest index + 1 using the database calls themselves?

server_addon/applications/server/__init__.py
```diff
@@ -129,6 +148,85 @@ async def create_applications_attribute(self) -> bool:
...
+            }
+
+        # when any of the required attributes are not present, add them
+        # and return 'True' to indicate that server needs to be restarted
+        needs_restart = False
+        for name, payload in attributes_to_create.items():
+            await Postgres.execute(
+                "INSERT INTO attributes (name, position, scope, data) VALUES ($1, $2, $3, $4)",
+                name,
+                last_index + 1,
```
BigRoy, line 214: Is there no better way in Postgres to ensure atomicity on an insert like this, ensuring the index is set after the current highest value on the actual insert itself? Or do we just not care because this 'runs on the server'?

martastain (changes requested):
Adding new variants with live updating works, but default variants (both for tools and applications) are present in enums even if i remove them and replace with others. i purged all previous versions of the application addon as well as all settings, but old variants are still there. 

Also. Could you please add:
```python
from ayon_server.entities.core import attribute_library
await attribute_library.load()
```
to the end of update_enums? it is not required for now, but in a future release (soon) this will invalidate attribute cache in redis, which is going to be needed for horizontal scaling of the server. At this point, it won't hurt doing it, so it is just for forwards compatibility.

BigRoy:
> Adding new variants with live updating works, but default variants (both for tools and applications) are present in enums even if i remove them and replace with others. i purged all previous versions of the application addon as well as all settings, but old variants are still there.

Did you remove them in both the production settings AND the staging settings?

iLLiCiTiT:
> to the end of update_enums? it is not required for now, but in a future release (soon) this will invalidate attribute cache in redis, which is going to be needed for horizontal scaling of the server. At this point, it won't hurt doing it, so it is just for forwards compatibility.

Could you add function `reset_attribute_cache` which does it? This is too specific for cases the logic would change in future.
```python
from ayon_server.entities.core import attribute_library
await attribute_library.load()
```

server_addon/applications/server/__init__.py
iLLiCiTiT: I don't know about any other, @martastain ?

martastain: whatever makes you happy:


```
SELECT MAX(position) AS max_position, BOOL_OR(name = 'applications') AS has_applications,  BOOL_OR(name = 'tools') AS has_tools FROM attributes;

 max_position | has_applications | has_tools 
--------------+------------------+-----------
           37 | t                | t
```

m-u-r-p-h-y (commented):
It works only in case you manually refresh the page to force the enum to reload.

image

after CTRL+R it is there
image

iLLiCiTiT:
> It works only in case you manually refresh the page to force the enum to reload.

That's out of scope of this PR and my competence.

m-u-r-p-h-y (approved):
you are right. it was becasue of two windows opened (for the screenshot)

in single window it works as expected!

martastain:
> It works only in case you manually refresh the page to force the enum to reload.
> 
> image
> 
> after CTRL+R it is there image

Yep. But it's better than restarting the server. Also we might be able to do that without page reload, but that's a frontend/backend issue (planned)

martastain (approved):

PR #352, opened 2024-04-03
Chore: remove unused work root function

moonyuet:
## Changelog Description
There was a functionality to modify workfile directory by calling `work_root` on host but it is no longer possible now. Although the functions are still implemented, the workfile tool does not use it.
This PR is to remove the unused function.
Hosts needed to be tested:
- [x] blender
- [ ] flame
- [x] fusion
- [ ] harmony
- [ ] hiero
- [x] maya
- [ ] nuke
- [x] photoshop
- [ ] resolve
- [ ] tvpaint
## Additional info
There might be an issue for maya as it uses the mel file to determine workfile directory

## Testing notes:
1. Launch the app
2. Try to browse, save and open workfile via workfile tool
3. It should be browsing/saving/opening the correct workfile directory as usual

### DIFF
--- a/client/ayon_core/hosts/maya/api/workio.py
+++ b/client/ayon_core/hosts/maya/api/workio.py
@@ -32,35 +32,3 @@ def current_file():
         return None
 
     return current_filepath
-
-
-def work_root(session):
-    work_dir = session["AYON_WORKDIR"]
-    scene_dir = None
-
-    # Query scene file rule from workspace.mel if it exists in WORKDIR
-    # We are parsing the workspace.mel manually as opposed to temporarily
-    # setting the Workspace in Maya in a context manager since Maya had a
-    # tendency to crash on frequently changing the workspace when this
-    # function was called many times as one scrolled through Work Files assets.
-    workspace_mel = os.path.join(work_dir, "workspace.mel")
-    if os.path.exists(workspace_mel):
-        scene_rule = 'workspace -fr "scene" '
-        # We need to use builtins as `open` is overridden by the workio API
-        open_file = __builtins__["open"]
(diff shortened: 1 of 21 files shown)


client/ayon_core/hosts/maya/api/workio.py
```diff
@@ -32,35 +32,3 @@ def current_file():
...
-                    scene_dir = remainder.split('"')[1]
-    else:
-        # We can't query a workspace that does not exist
-        # so we return similar to what we do in other hosts.
-        scene_dir = session.get("AVALON_SCENEDIR")
-
-    if scene_dir:
-        return os.path.join(work_dir, scene_dir)
-    else:
-        return work_dir
```
BigRoy, line 66: I'm actually quite sad to see this go - I'm fine with this particular logic disappearing - but being unable to have workspace definitions where scene files are in a subfolder for us is quite a loss. We're starting production today and only now realized this was actually 'broken' in our maya workspace definitions since the workfiles didn't save into our scenes subfolders.

🤞 Let's hope artists won't kill me by not having this subfolder.

moonyuet: maybe we can move and adjust the code to initialization of AYON plugins in Maya (i.e. the `_set_project` function).

BigRoy: It's not specific to just the maya workspaces for us - since we also used subfolders for our Fusion stuff. Anyway, I'll see how artists cope hehe.

LiborBatek:
been successfully tested in ```maya``` ```blender``` ```fusion``` ```PS``` so far

PR #353, opened 2024-04-03
Chore: Better context entity getter function

iLLiCiTiT:
## Changelog Description
Use more suitable function name to get current folder entity and added function to get current task entity.

## Additional info
The function `get_current_folder_entity` was intended to be also used to get different folder from current project, but it not used that way so that part was removed, which gave option to use more suitable name of the function.

Current task entity funcion is not trivial and must call twice to get the task entity, that might change with new ayon api.

## Testing notes:
1. Validate my reasoning is correct.
2. All usage of the function should be replaced.

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/validate_maya_units.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/validate_maya_units.py
@@ -3,7 +3,7 @@ import maya.cmds as cmds
 import pyblish.api
 
 import ayon_core.hosts.maya.api.lib as mayalib
-from ayon_core.pipeline.context_tools import get_current_project_folder
+from ayon_core.pipeline.context_tools import get_current_folder_entity
 from ayon_core.pipeline.publish import (
     RepairContextAction,
     ValidateSceneOrder,
@@ -132,5 +132,5 @@ class ValidateMayaUnits(pyblish.api.ContextPlugin,
 
         cls.log.info("Setting time unit to match project")
         # TODO replace query with using 'context.data["folderEntity"]'
-        folder_entity = get_current_project_folder()
+        folder_entity = get_current_folder_entity()
         mayalib.set_scene_fps(folder_entity["attrib"]["fps"])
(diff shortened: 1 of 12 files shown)


client/ayon_core/pipeline/context_tools.py
```diff
@@ -384,42 +384,67 @@ def get_current_project_entity(fields=None):
...
-def get_current_project_folder(folder_path=None, folder_id=None, fields=None):
+def get_current_folder_entity(fields=None):
     """Helper function to get folder entity based on current context.
 
     This function should be called only in process where host is installed.
 
-    Folder is found out based on passed folder path or id (not both). Folder
-    path is not used for filtering if folder id is passed. When both
-    folder path and id are missing then current folder path is used.
+    Folder is found out based on current project name and folder path.
```
BigRoy, line 392: I think I prefer a wording like this:
```suggestion
    Folder is based on current context project name and folder path.
```

client/ayon_core/pipeline/context_tools.py
```diff
@@ -384,42 +384,67 @@ def get_current_project_entity(fields=None):
...
     )
 
 
+def get_current_task_entity(fields=None):
+    """Helper function to get task entity based on current context.
+
+    This function should be called only in process where host is installed.
+
+    Task is found out based on current project name, folder path
+        and task name.
```
BigRoy, line 420: I think I prefer a wording like this:
```suggestion
    Task is based on current context project name, folder path
        and task name.
```

client/ayon_core/hosts/maya/plugins/publish/validate_maya_units.py
```diff
@@ -132,5 +132,5 @@ def repair(cls, context):
 
         cls.log.info("Setting time unit to match project")
         # TODO replace query with using 'context.data["folderEntity"]'
-        folder_entity = get_current_project_folder()
+        folder_entity = get_current_folder_entity()
```
BigRoy, line 135: This would be better replaced with `context.data["folderEntity"]` as the `# TODO` describes above it.

BigRoy (approved):
Didn't test but changes look good to me except for some cosmetic notes I had.

I wonder, with us exposing this whether it's worth having a `get_current_context_entities()` which returns the project, folder and task entities in one go in what could potentially be a more optimized call if multiples are needed?

iLLiCiTiT:
> I wonder, with us exposing this whether it's worth having a get_current_context_entities() which returns the project, folder and task entities in one go in what could potentially be a more optimized call if multiples are needed?

Is it used that way somewhere?

PR #354, opened 2024-04-03
Chore: Use project name from context to get settings

iLLiCiTiT:
## Changelog Description
Use `"projectName"` value to get settings instead of `get_current_project_settings`.

## Additional info
This is technical change, we should make sure the project which is set in context is used as source of truth for publishing.

## Testing notes:
1. Settings are collected as before.

### DIFF
--- a/client/ayon_core/plugins/publish/collect_settings.py
+++ b/client/ayon_core/plugins/publish/collect_settings.py
@@ -10,6 +10,9 @@ class CollectSettings(api.ContextPlugin):
 
     def process(self, context):
         project_name = context.data["projectName"]
+        self.log.debug(
+            "Collecting settings for project: {}".format(project_name)
+        )
         project_settings = get_project_settings(project_name)
         context.data["project_settings"] = project_settings
         context.data["projectSettings"] = project_settings
--- a/client/ayon_core/plugins/publish/collect_settings.py
+++ b/client/ayon_core/plugins/publish/collect_settings.py
@@ -1,5 +1,5 @@
 from pyblish import api
-from ayon_core.settings import get_current_project_settings
+from ayon_core.settings import get_project_settings
 
 
 class CollectSettings(api.ContextPlugin):
@@ -9,4 +9,7 @@ class CollectSettings(api.ContextPlugin):
     label = "Collect Settings"
 
     def process(self, context):
-        context.data["project_settings"] = get_current_project_settings()
+        project_name = context.data["projectName"]
+        project_settings = get_project_settings(project_name)
+        context.data["project_settings"] = project_settings
+        context.data["projectSettings"] = project_settings


client/ayon_core/plugins/publish/collect_settings.py
```diff
@@ -9,4 +9,10 @@ class CollectSettings(api.ContextPlugin):
...
 
     def process(self, context):
-        context.data["project_settings"] = get_current_project_settings()
+        project_name = context.data["projectName"]
+        self.log.debug(
+            "Collecting settings for project: {}".format(project_name)
+        )
+        project_settings = get_project_settings(project_name)
+        context.data["project_settings"] = project_settings
+        context.data["projectSettings"] = project_settings
```
BigRoy, line 18: What's the reason this key got added? Is `project_settings` deprecated?

Should we add that key in a separate PR if our intent is to swap the usage around - and in that other PR also refactor all usages to match the newer key if that's the preferred key name. Then just keep `project_settings` for backwards compatibility and mark it with a comment as such.

I feel like having the duplicate key is worse than having just one key that may not be named the way we want it.

iLLiCiTiT:
```suggestion
```

iLLiCiTiT: 100% different PR.

iLLiCiTiT:
> The setting of [context.data["projectName"] here](https://github.com/ynput/ayon-core/blob/f59801774ba2711da548140e6bd65bcccdbc538f/client/ayon_core/plugins/publish/collect_from_create_context.py#L36-L38) happens in an if statement that doesn't seem all too promising that we can rely on the project name being accessible.

Not sure what to do with this, maybe just use `.get(...)`?

> Should that if statement be removed as well?

Which if statement?

BigRoy:
> Which if statement?

The `if project_name` statements in the code I linked.

https://github.com/ynput/ayon-core/blob/f59801774ba2711da548140e6bd65bcccdbc538f/client/ayon_core/plugins/publish/collect_from_create_context.py#L36-L38

https://github.com/ynput/ayon-core/blob/2072f2fbe2bdaa09c96dd3c849bff65785c3b3bb/client/ayon_core/plugins/publish/collect_current_context.py#L28-L29

Both seem to 'hint' that `context.data["projectName"]` MAY not exist.

iLLiCiTiT:
> Both seem to 'hint' that context.data["projectName"] MAY not exis

Yes, I'm asking how to handle that in this plugin. I won't touch the other plugins.

BigRoy:
> Yes, I'm asking how to handle that in this plugin. I won't touch the other plugins.

Ah good question - it seems like we always expected a project name to be 'retrievable' and thus be valid - so I'd actually say we should just make those always being set in the other plug-ins accordingly. Rather prefer that, then now still adding a `get_current_project_entity` fallback in this plug-in.

iLLiCiTiT:
> it seems like we always expected a project name to be 'retrievable' and thus be valid - so I'd actually say we should just make those always being set in the other plug-ins accordingly.

You can have completelly different logic in custom addon which collects `"projectName"`. So I don't think we should enforce it.

> Rather prefer that, then now still adding a get_current_project_entity fallback in this plug-in.

I don't think that is correct behavior. Using current project settings would be against what is inside publish context...

kalisp (approved):
Tried a couple of DCCs, didnt explode.

PR #356, opened 2024-04-03
Maya: Allow specifying raw JSON custom tools menu for Maya.

BigRoy:
## Changelog Description

Adds support for specifying the raw JSON for custom tools, allowing much easier/better access for complex menus.

## Additional info

The non-JSON menu builder in settings also disallows the creation of submenus by its design (as far I could see) and 'pasting' our existing settings from e.g. OpenPype didn't allow it to automatically create the required structure for supporting submenus, etc. because of it.

By having this raw JSON field I could just copy-paste our OpenPype settings for the custom colorbleed menu for our Maya.

image
image
image

## Testing notes:

1. Define a custom scripts menu via the JSON format and enable the JSON definition toggle.


moonyuet:
It shows up the custom tools as expected
image
But is it normal behavior that having the raw json definition enabled overriding the original script definition setting below?
image
I also think that this json definition logic not only appies to Maya but Nuke(I remember it has similar toolbar to store custom scripts.)

BigRoy:
> But is it normal behavior that having the raw json definition enabled overriding the original script definition setting below?

The toggle **overrides** which one of the two to use. It uses **either** the settings UI builder OR the JSON menu. It does not merge the two in any way.

iLLiCiTiT:
Wouldn't be better to use enum? So user can see only one of the inputs when selected...

BigRoy:
> Wouldn't be better to use enum? So user can see only one of the inputs when selected...

If there's a way to have the settings front-end to switch between one of the two. Yes, 100%. Any examples?

iLLiCiTiT:
> If there's a way to have the settings front-end to switch between one of the two. Yes, 100%. Any examples?

When enum has `conditionalEnum=True,` it can show only attribute defined by it's value. Example in `server/settings/publish_plugins.py` model `ExtractThumbnailOIIODefaultsModel` where `colorspace` or `display_and_view` are based on value of `type`.

BigRoy:
> > If there's a way to have the settings front-end to switch between one of the two. Yes, 100%. Any examples?
> 
> When enum has `conditionalEnum=True` it can show only attribute defined by it's value. Example in `server/settings/publish_plugins.py` model `ExtractThumbnailOIIODefaultsModel` where `colorspace` or `display_and_view` are visible, based on value of `type`.

Thanks you very much - this is much better!

image
image
image

moonyuet (approved):
I tried with the latest commit of the branch and it installs the custom tools as expected(tested with both menu builder and full json definition). I like this one better than the last one as it allows user to be clearer which options to use to define my custom tools.

tokejepsen (approved):

tokejepsen (changes requested):
Could we get the addon version incremented?

BigRoy:
> Could we get the addon version incremented?

Done with https://github.com/ynput/ayon-core/pull/356/commits/590f1db63e1254d6fb2be4eac1ea22fb60762afb

tokejepsen (approved):

PR #340, opened 2024-04-02
Maya: Support to validate/repair the task attributes in validate data in same context

moonyuet:
## Changelog Description
Bugfix the validate data in same context not being able to be validated and fix the task attribute in repair actions.
It also updates the code with ayon_api function for resetting resolution setting as the additional fix.

## Additional info
Supposes that this PR needs to fix the repair actions of validate resolution and validate frame range.
And there is the related PR to fix for validate frame range https://github.com/ynput/ayon-core/pull/338

## Testing notes:
1. Launch Maya
2. Create any instance
3. change the folder path and task
4. Publish
5. It will error out` validate data in same context`
6. Perform repair action
7. It will assign the current context folderpath and task as AYON's attributes.

PR #358, opened 2024-04-03
Maya: Update publish instances via CreateContext + use task attributes instead of folder attributes

BigRoy:
## Changelog Description

Update instances using `CreateContext` to correctly update product names if needed

- Also rely on task attributes for resolution and frame ranges instead of folder attributes

## Additional info

This will fix the updating the the workfile product name when switching between tasks and 'updating publish instances' along.

## Testing notes:

1. Saving to another context should correctly update workfile product name (and update folder path and task for the new context)
2. Reset frame range should work (and adhere to task attributes)
3. Reset resolution should work (and adhere to task attributes)


BigRoy:
This overlaps a tiny bit with https://github.com/ynput/ayon-core/pull/340 since it touches the resolution reset taking it from the task attributes.

client/ayon_core/hosts/maya/api/lib.py
```diff
@@ -3244,33 +3244,46 @@ def update_content_on_context_change():
...
+            if (
+                    key not in creator_attributes
+                    or creator_attributes[key] == value
+            ):
+                continue
+
+            # Update instance creator attribute value
+            print(f"Updating {instance.product_name} {key} to: {value}")
+            instance[key] = value
+            has_changes = True
```
BigRoy, line 3283: Any code style tips for this @iLLiCiTiT ? Not sure if the two for loops feel cluttered or whether this makes sense as it is.

moonyuet (approved):
Tested successfully in Maya 2023. Validate Frame Range, Validate Resolution, Validate instance in same context and their repair actions all work. The switch context data and save also works too.
image

LiborBatek (approved):
Besides regular switching of context by workfiles and resaving, I have also tested the ```Set Frame range``` and ```Set Resolution``` on a ```task``` level if set differently as seen here for my ```FX``` task 
image

And all working predictably and as supposed to! So All good!

PR #359, opened 2024-04-03
Maya: Fix: Do not always reset context on task change 

BigRoy:
## Changelog Description

Fix for implementation of #261

- Remove logic that would always update before save/open anyway (and as such the dialog itself didn't make any sense)

## Additional info

This also fixes that before opening a scene in another context that maya was actually resetting time range, fps, etc. always even though it would be opening a file anyway - optimizing opening workfiles in other contexts because of that.

Note that the **Task** does not update on context change/resets. For that you'll need PR: https://github.com/ynput/ayon-core/pull/358

## Testing notes:

1. Save into a different folder or task
2. When picking no options (or cancelling) on the prompt - nothing should update
3. When clicking ok with certain options, only those options should update.


moonyuet (approved):
tested successfully in maya 2023. When changing different context and hit okay, the instance name changed as followed. And when  changing different context and hit cancel, the instance name and context won't be changed as expected.

MustafaJafar (approved):
This is cool! 
Thanks.

PR #360, opened 2024-04-03
Applications: Disable applications addon if does not have settings available

iLLiCiTiT:
## Changelog Description
Disable applications addon if does not have available settings (addon is disabled on server).

## Additional info
The client addon would crash if server would be disabled.

### DIFF
--- a/client/ayon_core/addons/applications/ayon_applications/addon.py
+++ b/client/ayon_core/addons/applications/ayon_applications/addon.py
@@ -11,6 +11,10 @@ from .manager import ApplicationManager
 class ApplicationsAddon(AYONAddon, IPluginPaths):
     name = "applications"
 
+    def initialize(self, settings):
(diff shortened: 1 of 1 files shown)


client/ayon_core/addons/applications/ayon_applications/addon.py
```diff
@@ -11,6 +11,10 @@
 class ApplicationsAddon(AYONAddon, IPluginPaths):
     name = "applications"
 
+    def initialize(self, settings):
+        # TODO remove when addon is removed from ayon-core
+        self.enabled = self.name in settings
```
BigRoy, line 16: How come this is needed? Is it because addons present in ayon core are currently **always** loaded no matter what?

iLLiCiTiT: Yes.

PR #361, opened 2024-04-03
Maya: Expose loaders enabled state to settings

BigRoy:
## Changelog Description

Allow to enable/disable Maya loaders based on studio and project settings.

## Additional info

image

## Testing notes:

1. Loader enable/disable state should be adhered to by the loaders.


moonyuet (approved):
Checked with maya, looks good for me. 
Just wondering if we need to implement the related settings in other hosts in the future?

LiborBatek (approved):
Works like charm... did test multiple loader types and not spotted any issue.

image

iLLiCiTiT:
I wonder what is the usecase to be able to enable/disable loaders?

BigRoy:
> I wonder what is the usecase to be able to enable/disable loaders?

Many loaders may apply to the same representations - but in certain project you may not want the V-Ray loaders or the Multiverse USD loaders because you're either not using that plug-in or do not want to load it like that in the project.

The reasoning behind disabling them is that otherwise the loader list can grow very large in numbers. In Maya on our end it can easily reach up to 5+ loaders for a single product, maybe even around 10 - of which many also apply to multiple representations (e.g. `usd` + `abc`).

m-u-r-p-h-y:
> I wonder what is the usecase to be able to enable/disable loaders?

we need to simplify the UI for artists as much as we can to eliminate space for human errors. It would be great to have a possibility to enable/disable loaders based on context. (Ex. modellers in Maya may need to see different set of loaders than animators on the same project)

BigRoy:
> we need to simplify the UI for artists as much as we can to eliminate space for human errors. It would be great to have a possibility to enable/disable loaders based on context. (Ex. modellers in Maya may need to see different set of loaders than animators on the same project)

I personally don't like that too much since suddenly when you're in modeling you can't access the same features. If it would re-order based on your current context it could be nice, but could also turn out very cruel to your muscle memory.

Anyway, there's something to "filtering it" to your current context - definitely if there are great numbers. I just haven't found a UX for that which I really liked. I want a power user to still be able to access all the things from anywhere - however you want the average joe to click the correct thing 99% of the time as well.

LiborBatek:
If I can add something to UX and loaders....what bothers me more is multiplication of many actions....they could be just single entry with Option box to choose which particular one user wants???

image


I also do agree that limitting ```Loader``` types being pretty handy so very welcomed feature!!

tokejepsen:
> If I can add something to UX and loaders....what bothers me more is multiplication of many actions....they could be just single entry with Option box to choose which particular one user wants??? Or even split the menu for Loader types and Utilities containing Copy... actions not cluttering the main work area of the menu

@LiborBatek I think that is a good idea to discuss. Have made an issue for it.

PR #362, opened 2024-04-03
Chore: Added more keys for folder template data

iLLiCiTiT:
## Changelog Description
Proposal of new keys available in folder dictionary.

## Additional info
Added new keys under folder template data. First obvious is `"type"` where is folder type. Key `"path"` is just a folder path, `"parent"` is name of parent which is project if direct parent is folder, `"parents"` is what `"hirarchy"` was.

Any comments, ideas, confrontations welcommed.

### DIFF
--- a/client/ayon_core/pipeline/template_data.py
+++ b/client/ayon_core/pipeline/template_data.py
@@ -74,7 +74,7 @@ def get_folder_template_data(folder_entity, project_name):
                       project
 
     Required document fields:
-        Folder: 'path' -> Plan to require: 'folderType'
+        Folder: 'path', 'folderType'
 
     Args:
         folder_entity (Dict[str, Any]): Folder entity.
@@ -101,6 +101,10 @@ def get_folder_template_data(folder_entity, project_name):
     return {
         "folder": {
             "name": folder_name,
+            "path": path,
+            "type": folder_entity["folderType"],
+            "parents": hierarchy,
+            "parent": parent_name,
         },
         "asset": folder_name,
         "hierarchy": hierarchy,
--- a/client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
+++ b/client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
@@ -399,6 +399,10 @@ class CollectAnatomyInstanceData(pyblish.api.ContextPlugin):
                 "parent": parent_name,
                 "folder": {
                     "name": folder_name,
+                    "path": folder_path,
+                    "type": folder_entity["folderType"],
+                    "parent": parent_name,
+                    "parents": hierarchy,
                 },
             })
             return
@@ -411,13 +415,18 @@ class CollectAnatomyInstanceData(pyblish.api.ContextPlugin):
             if hierarchy:
                 parent_name = hierarchy.split("/")[-1]
 
-            folder_name = instance.data["folderPath"].split("/")[-1]
+            folder_path = instance.data["folderPath"]
+            folder_name = folder_path.split("/")[-1]
             anatomy_data.update({
                 "asset": folder_name,
                 "hierarchy": hierarchy,
                 "parent": parent_name,
                 "folder": {
                     "name": folder_name,
+                    "path": folder_path,
+                    "type": "Shot",
+                    "parent": parent_name,
+                    "parents": hierarchy,
                 },
             })
 


client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -399,6 +399,10 @@ def _fill_folder_data(self, instance, project_entity, anatomy_data):
                 "parent": parent_name,
```
BigRoy, line 399: Shouldn't this just be calling the function from `ayon_core.pipeline.template_data.get_folder_template_data`?

```python
data = get_folder_template_data(folder_entity)
anatomy_data.update(data)
```

client/ayon_core/pipeline/template_data.py
```diff
@@ -101,6 +101,10 @@ def get_folder_template_data(folder_entity, project_name):
     return {
         "folder": {
             "name": folder_name,
+            "path": path,
+            "type": folder_entity["folderType"],
+            "parents": hierarchy,
+            "parent": parent_name,
```
BigRoy, line 107: Is there really use in having the `parent` data duplicated outside the folder key and inside?

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -411,13 +415,18 @@ def _fill_folder_data(self, instance, project_entity, anatomy_data):
...
+            folder_path = instance.data["folderPath"]
+            folder_name = folder_path.split("/")[-1]
             anatomy_data.update({
                 "asset": folder_name,
                 "hierarchy": hierarchy,
                 "parent": parent_name,
                 "folder": {
                     "name": folder_name,
+                    "path": folder_path,
+                    "type": "Shot",
```
BigRoy, line 427: Yeah - this doesn't really make sense to have this hardcoded type? Types are dynamic and a Shot type may not even be defined in the current project.

iLLiCiTiT: Oh, yeah, that's actually comming from current editorial logic, it's hardcoded there. Should be changed, out of scope of this PR.

client/ayon_core/pipeline/template_data.py
iLLiCiTiT: I would say tha the data which are now in root of dictionary should be considered as deprecated with these changes.

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -411,13 +415,18 @@ def _fill_folder_data(self, instance, project_entity, anatomy_data):
             if hierarchy:
                 parent_name = hierarchy.split("/")[-1]
 
-            folder_name = instance.data["folderPath"].split("/")[-1]
+            folder_path = instance.data["folderPath"]
+            folder_name = folder_path.split("/")[-1]
             anatomy_data.update({
```
BigRoy, line 420: If we implement `get_folder_template_data` in a way that it calls e.g. `get_folder_template_data_from_values(folder_path, folder_type, project_name)` then also here we could re-use the implementation of `ayon_core.template_data.get_folder_template_data_from_values` to ensure each of these end up using the same function instead of duplicating the code everywhere that seem to define these template data. Preferably there's one entry point to this. that returns the valid data - instead of having this duplicated.

iLLiCiTiT: Can we focus more on the data keys?

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -399,6 +399,10 @@ def _fill_folder_data(self, instance, project_entity, anatomy_data):
                 "parent": parent_name,
                 "folder": {
                     "name": folder_name,
+                    "path": folder_path,
+                    "type": folder_entity["folderType"],
+                    "parent": parent_name,
+                    "parents": hierarchy,
```
BigRoy, line 405: I personally think it's confusing that "parent" includes the project name if there are no parents for the folder but `parents` never includes the project name.

We could keep e.g. `hierarchy` to be what it was (all parents excluding project) and add `parents` to be `{project_name}/{hierarchy}` so that `parent` makes sense again?

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
BigRoy: Might be better than to first fix the duplication of logic in a separate PR which would make the size of this PR change much smaller since there's much less to change.

iLLiCiTiT: I created this PR to raise discussion about the keys....

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
iLLiCiTiT: I do agree that if `parent` can be project name then `parents` are confusing. My thinking was that `hierarchy` should probably contain the folder name too, but maybe you're right.

Ad. `"parent"` I don't remember why the project name is used? I do remember there was a discussion but don't remember why we chose this approach?

BigRoy: > My thinking was that hierarchy should probably contain the folder name too, but maybe you're right.

Kind of had the same - but by having `{hierarchy}` separate from the folder name itself you can actually use them separately, plus if they were combined then you'd just have `folder[path]`? :)

client/ayon_core/pipeline/template_data.py
MustafaJafar: Just to check I get it correctly. 
these data should be considered as deprecated
image

MustafaJafar (dismissed):
This is cool it works on my side. I used it with this Houdini feature that uses template data.

> I think we should update [Available template keys](https://ayon.ynput.io/docs/admin_settings_project_anatomy/#available-template-keys)

image

kalisp:
Is there matching update of documentation (most likely https://ayon.ynput.io/docs/admin_settings_project_anatomy/#available-template-keys)?

jakubjezek001:
> Is there matching update of documentation (most likely [ayon.ynput.io/docs/admin_settings_project_anatomy/#available-template-keys](https://ayon.ynput.io/docs/admin_settings_project_anatomy/#available-template-keys))?

I am guessing that once the proposal will turn into implementation then yes it will need to have that step.

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
antirotor: I agree on having `hierarchy` as "hierarchy within the project" - so without project name and without folder name. `parents` should be basically `dirname(folder[path])` and `parent` direct ancestor of the folder?

iLLiCiTiT:
Created PR adding `"type"` and `"path"` https://github.com/ynput/ayon-core/pull/445 .

PR #363, opened 2024-04-03
Maya: Validate node ids in database, query folder ids once

BigRoy:
## Changelog Description

For the Validate Node Ids In Database validator cache the folder ids of the project on the first query so we don't need to query all folder ids in the current project per instance.

## Additional info

Also only query the database if the instance actually contains nodes that should have an id to begin with.

## Testing notes:

1. Validation of whether folder id in `cbId` (part before the `:`) is a valid folder id in current project should still work.
2. It should be faster for many instances.


moonyuet (approved):
tested sucessfully in maya 2023

kalisp (approved):

PR #365, opened 2024-04-03
Maya: Collect Look code cleanup + minor bugfix

BigRoy:
## Changelog Description

- Fix: remove from list while iterating over a copy - fixes potential wrong removal from the list of the node types in `RENDER_SET_TYPES `
- Remove redundant code
    - Remove unused `maketx` attribute on the class
    - Remove duplicate query of `instance_lookup` variable.
- Restructure code for readability + improve docstrings

## Additional info

Mostly cleanup.

## Testing notes:

1. Maya Collect Look should still work
2. Maya Collect Model Render Sets (materials on model publish) should still work for non-shading engines (e.g. object sets)


LiborBatek (approved):
And when loaded ```look``` on the asset:

Screenshot 2024-04-04 163117

client/ayon_core/hosts/maya/plugins/publish/collect_look.py
```diff
@@ -439,14 +372,87 @@ def collect(self, instance):
...
+
+        self.log.debug("Found materials:\n{}".format(materials))
+
+        # Get the entire node chain of the look sets
+        # history = cmds.listHistory(look_sets, allConnections=True)
+        # if materials list is empty, listHistory() will crash with
+        # RuntimeError
+        history = set()
+        if materials:
+            history = set(cmds.listHistory(materials, allConnections=True))
```
moonyuet, line 416: history is already the empty set, guess we can just add the element into the set instead.
```suggestion
            history.add(cmds.listHistory(materials, allConnections=True))
```

antirotor: And the update is somewhat faster:
```python
>>> print(timeit.timeit(stmt='a = set([i for i in range(1,100)])', globals=globals()))
4.729680200049188
>>> a =set()
>>> print(timeit.timeit(stmt='a.update([i for i in range(1,100)])', globals=globals()))
3.7812449999619275
```

mkolar:
@antirotor mergable?

tokejepsen (approved):

PR #372, opened 2024-04-04
Chore: Run command args handling

iLLiCiTiT:
## Changelog Description
Remove start.py or ayon executable instead of script path from sys.argv.

## Additional info
When running `ayon_console.exe run my_script.py` the `sys.argv` will constain `["my_script.py"]` instead of `["ayon_console.exe"]`.

## Testing notes:
You can compare the difference when running script with this content with this PR and before this PR.
```python
import sys
print(sys.argv)
```
1. Command `run` works.

### DIFF
--- a/client/ayon_core/cli.py
+++ b/client/ayon_core/cli.py
@@ -11,7 +11,7 @@ import acre
 from ayon_core import AYON_CORE_ROOT
 from ayon_core.addon import AddonsManager
 from ayon_core.settings import get_general_environments
-from ayon_core.lib import initialize_ayon_connection
+from ayon_core.lib import initialize_ayon_connection, is_running_from_build
 
 from .cli_commands import Commands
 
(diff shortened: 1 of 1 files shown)


client/ayon_core/cli.py
```diff
@@ -167,16 +167,29 @@ def run(script):
...
-        args.remove(script)
-        sys.argv = args
-
-        args_string = " ".join(args[1:])
-        print(f"... running: {script} {args_string}")
-        runpy.run_path(script, run_name="__main__", )
+        comp_path = os.getenv("AYON_EXECUTABLE")
+    # Normalize paths for comparison
+    first_arg = os.path.normpath(first_arg).lower()
+    comp_path = os.path.normpath(comp_path).lower()
```
BigRoy, line 183: Should this be `os.path.realpath` because the argument path _may_ be relative? Or is that not the case? Or use some sort of `pathlib.Path` comparison?

client/ayon_core/cli.py
```diff
@@ -167,16 +167,29 @@ def run(script):
...
+    if first_arg == comp_path:
+        sys.argv.pop(0)
+
+    # Remove 'run' command from sys.argv
+    sys.argv.remove("run")
+
+    args_string = " ".join(sys.argv[1:])
+    print(f"... running: {script} {args_string}")
+    runpy.run_path(script, run_name="__main__", )
```
BigRoy, line 192: Cosmetics
```suggestion
    runpy.run_path(script, run_name="__main__")
```

client/ayon_core/cli.py
iLLiCiTiT: Rather `os.path.abspath` but `pathlib` might be better option.

iLLiCiTiT: Use `Path(...).resolve()` for comaprison. Didn't find better approach.

BigRoy: Note that the resolve only works for existing paths it seems - it doesn't turn relative paths into absolute paths if the path does not exist - which might be unwanted side effect?

iLLiCiTiT: I think that's ok in this case.

Issue #348, opened 2024-04-02
Enhance `KnownPublishError` visualisation


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Enhance `KnownPublishError` to support changing the error label and message.

<table>
<tr>
<td>Current State</td>
<td>My suggestion</td>
</tr>
<tr>
<td><img src=https://github.com/user-attachments/assets/ac94fef7-aa90-4a20-a3f5-93e866c94add></td>
<td><img src=https://github.com/user-attachments/assets/16bf053d-c37f-4ee4-ba9a-8b6f80cc4082></td>
</tr>
</table>

## How would you imagine the implementation of the feature?

I made a draft PR for it https://github.com/ynput/ayon-core/pull/373 to experiment the idea where I was able to reproduce the screenshots above.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

Adding a Generic `PublishError` that works in a similar fashion to `PublishValidationError` which is a better user friendly way to raise and show errors from pyblish plugins like extractors or even from `lib.py` files.

It would like this (I hacked the code to be able to produce this screenshot)
In screenshot my code failed during extraction.
![image](https://github.com/user-attachments/assets/65610501-ac93-4d08-9135-b11741d40ef3)

MustafaJafar:
I ended up experimenting tweaking the `KnownPublishError` here's a test PR https://github.com/ynput/ayon-core/pull/373 .

MustafaJafar:
Issue summary: 
My two suggestions to have more user friendly error reporting: 
1. Use the same error report as `PublishValidationError`  maybe we can call it `PublishPluginError`.
2. Having a better visualization for `KnownPublishError` , I made a PR for it https://github.com/ynput/ayon-core/pull/373

| using `PublishValidationError` with extractor plugin | better `KnownPublishError` visualization |
|--|--|
| ![image](https://github.com/ynput/ayon-core/assets/20871534/3f879770-268d-46e2-8b97-9f411643217e) | ![image](https://github.com/ynput/ayon-core/assets/20871534/f680edf1-32ac-4db2-a472-5cd5de561523) |

PR #373, opened 2024-04-04
Publisher: Enhance KnownPublishError visualisation

MustafaJafar:
## Changelog Description
Tweaking `KnownPublishError` to have more user friendly error handling.
Also, slightly changing error report.

> Here's some demo from Houdini - mantra ifd product type

Animation_44

## Testing notes:
1. Change a `RuntimeError` to `KnownPublishError` and use better message and add a label similar to the provided example.
2. A good practice to add `self.log.error(....)` with more detailed error message.
3. Raise the error intentionally
4. Telling if the new error reporting is more user friendly.


client/ayon_core/tools/publisher/widgets/report_page.py
```diff
@@ -1459,13 +1459,14 @@ class CrashWidget(QtWidgets.QWidget):
...
-        main_label = QtWidgets.QLabel("This is not your fault", self)
+        main_label = QtWidgets.QLabel("Error Occurred!", self)
         main_label.setAlignment(QtCore.Qt.AlignCenter)
         main_label.setObjectName("PublishCrashMainLabel")
 
         report_label = QtWidgets.QLabel(
             (
-                "Please report the error to your pipeline support"
+                "Expand the orange error message below to find more info.\n"
+                "You can report the error to your pipeline support"
```
BigRoy, line 1469: If the information is in the orange bar - we should auto-expand it. Not ask the user to do so.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1369,6 +1382,7 @@ def event_system(self):
             "publish.is_running.changed" - Attr 'publish_is_running' changed.
             "publish.has_crashed.changed" - Attr 'publish_has_crashed' changed.
             "publish.publish_error.changed" - Attr 'publish_error'
+            publish.publish_error_label.changed  - Attr 'publish_error_label'
```
BigRoy, line 1385: To match the others in the docstring.
```suggestion
            "publish.publish_error_label.changed"  - Attr 'publish_error_label'
```

client/ayon_core/tools/publisher/control.py
```diff
@@ -1194,6 +1194,18 @@ def publish_error_msg(self):
 
         pass
 
+    @property
+    @abstractmethod
+    def publish_error_label(self):
+        """Current error label which cause fail of publishing.
+
+        Returns:
+            Union[str, None]: Error Label which will be showed to artist or
```
BigRoy, line 1203:
```suggestion
            Union[str, None]: Error Label which will be shown to artist or
```

BigRoy:
> image

With the "error" wording above the buttons I'm currently unsure what "Save to disk" would save because the text feels less focused on 'reporting it' I wouldn't immediately think that "Save to disk" would be talking about the report. But maybe that's just me.

MustafaJafar:
> With the "error" wording above the buttons I'm currently unsure what "Save to disk" would save because the text feels less focused on 'reporting it' I wouldn't immediately think that "Save to disk" would be talking about the report. But maybe that's just me.

I'm not sure about how to rephrase it in a better way. 
My goal was to write  something informative. click for more info, report the error, etc.  
what do you think could be a suitable informative message ?

BigRoy:
> I'm not sure about how to rephrase it in a better way.
> My goal was to write something informative. click for more info, report the error, etc.
> what do you think could be a suitable informative message ?

Maybe re-reading it again I have less of an issue with it. Just my brain being dumb sometimes. Removing the words "You can" might also put more focus on "Report the error".

client/ayon_core/tools/publisher/widgets/report_page.py
MustafaJafar: you know what.. I'll just remove that part ` "Expand the orange error message below to find more info.\n"`

MustafaJafar: I'll just remove  `"Expand the orange error message below to find more info.\n"` . I think it's redundant. 

I think. It'd be better to have the error message un-expanded because this PR adds message and title to the `KnownPublishError` . and having the error expanded will show the error message.

iLLiCiTiT:
Moving some of internal conversation here.

This is current state of the labels:
image

I don't think user should look at the frame at the bottom to find out what happened, if the information is for user. The bottom panel is short technical information.

First, my "feeling" is that both titles might be `Error happened`. The frame at the bottom would be unchanged. Instead of `label` would `KnownPublishError` had `user_message` which would be by default. `This is not your fault. Please report the error using one of the options below.` which would be shown as message at the top, but the developer could change it with the error?

This will cause that the message `"Please report the error using one of the options below."` could be overriden, and question is if we should put the message elsewhere (e.g. next to buttons). Or change labels of buttons to `Copy report to clipboard` and `Save report to disk`?

I think @mkolar came up with current design so he should have opinion on this.

MustafaJafar:
I've just figured out that I forgot to push my example. 
Here's what this PR does:

I've added an example in Mantra IFD houdini product type. to run it:
- Create a mantra IFD product type, switch off farm rendering and publish

Animation_42

image

image

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -244,13 +245,14 @@ def render_rop(ropnode):
...
                        # Allow Deadline to capture completion percentage
                        output_progress=verbose)
-    except hou.Error as exc:
+    except hou.Error:
         # The hou.Error is not inherited from a Python Exception class,
         # so we explicitly capture the houdini error, otherwise pyblish
         # will remain hanging.
         import traceback
         traceback.print_exc()
-        raise RuntimeError("Render failed: {0}".format(exc))
```
BigRoy, line 253: I don't believe the printed exception will be added to the Publisher UI's logs - and thus, by not including the `exc` in the error or logs itself it would mean that IF the Exception itself has some information about why it failed that information will now not appear in the publish logs, which it should. So we might want to still to `log.error(f"Exception: {exc}")` or alike.

client/ayon_core/hosts/houdini/api/lib.py
```diff
@@ -244,13 +245,14 @@ def render_rop(ropnode):
...
+        raise KnownPublishError("'{}' failed to render!".format(ropnode.path()),
+                                label="Render was interrupted or canceled.")
```
BigRoy, line 255: This isn't 100% true. It _may_ have been canceled or interrupted - but it may also have been an actual error during the ROP render. As such, the message here may be untrue.

Also, what's the difference between canceled and interrupted? Isn't that essentially the same thing here?

client/ayon_core/hosts/houdini/plugins/publish/extract_mantra_ifd.py
```diff
@@ -33,9 +33,9 @@ def process(self, instance):
                 os.path.normpath(os.path.join(staging_dir, frame)))
         ]
         if missing_frames:
-            raise RuntimeError("Failed to complete Mantra ifd extraction. "
-                               "Missing output files: {}".format(
-                                   missing_frames))
+            raise KnownPublishError("Failed to complete Mantra ifd extraction. ",
+                                    # "Missing output files: {}".format(missing_frames),
```
BigRoy, line 37: I guess we'll still need to report this somewhere? E.g. having a clear `log.error()` prior to this?

client/ayon_core/tools/publisher/widgets/report_page.py
```diff
@@ -1459,7 +1459,7 @@ class CrashWidget(QtWidgets.QWidget):
     def __init__(self, controller, parent):
         super(CrashWidget, self).__init__(parent)
 
-        main_label = QtWidgets.QLabel("This is not your fault", self)
+        main_label = QtWidgets.QLabel("Error Occurred!", self)
```
BigRoy, line 1462: I wonder if `Error Occurred!` is too unfriendly/scary? I recall that being a huge design decision as to why once it was chosen to report with `This is not your fault`. So may be one for @mkolar to comment on?

client/ayon_core/hosts/houdini/plugins/publish/extract_mantra_ifd.py
MustafaJafar: Agreed! 
going to push a commit shortly.

image

MustafaJafar: here it is https://github.com/ynput/ayon-core/pull/373/commits/e6c52f0430bd63cf8f3a469d9c7edae94d3ad8fd

client/ayon_core/hosts/houdini/api/lib.py
MustafaJafar: I modified it a little bit here.
here it is https://github.com/ynput/ayon-core/pull/373/commits/e6c52f0430bd63cf8f3a469d9c7edae94d3ad8fd

client/ayon_core/tools/publisher/widgets/report_page.py
MustafaJafar: Tbh, I don't know what should be the message. 
I only changed it to open a discussion about it.

iLLiCiTiT:
As I'm more and more thinking about it I guess we'll need different exception so we know if it is something which is caused by user or it is really a bug.

MustafaJafar:
> As I'm more and more thinking about it I guess we'll need different exception so we know if it is something which is caused by user or it is really a bug. It also affects UI which has messages related to actions and their buttons. Question is if the report buttons should be visible in case it is not user fault etc. That is UX desicion, and I not good at it, so we need someone who can tell.

Cool! 
I think we can close this PR or leave it as it is and continue the discussion in this issue https://github.com/ynput/ayon-core/issues/348

MustafaJafar:
Closing this PR and moving the discussion to https://github.com/ynput/ayon-core/issues/348

Issue #180, opened 2024-03-13
Houdini: workfile tool doesn't switch context properly


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Sometimes context is not changed when changing context with workfile tool

![image](https://github.com/ynput/ayon-core/assets/20871534/1069acdb-d17c-417a-91ed-dd4b8c6df78e)

### Expected Behavior:

It should change the asset context.

### Version

Launcher `1.0.3-dev.1`
Ayon Core `0.3.0-dev.1`
Ayon Server `1.0.4+202402231406`

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Open Houdini 
2. Use workfile tool to open a workfile of another asset

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Additional context:

I'm not sure if this problem happens because I'm constantly testing new PRs and updating my `ayon-core` clone.

[cuID:[AY-3851](https://app.clickup.com/t/86bxwjrxy)]

BigRoy:
I wonder if this may have to do with it relying on the "context" saved in the scene - and not necessarily the context of that asset. E.g. when saving to different assets in the first place:

https://github.com/ynput/ayon-core/blob/d724a666331b042d9908a1c6e4a4271348174c76/client/ayon_core/hosts/houdini/api/pipeline.py#L171-L175

It seems to in the menu show the `os.environ["AYON_FOLDER_PATH"]` so when switched it should've updated that env var.
Might be good to just test and print that environment variable to see whether it has updated since - and if not, then we know at least the bug isn't related to the Houdini menu itself not refreshing. ;)

Do you get any logged/printed errors when switching / opening the other that might be related to updating of the context failing?

MustafaJafar:
> You mention "sometimes" - but have you found a way to reliably reproduce it? Maybe it always occurs on your end if you switch from one particular asset to another particular asset.

You are correct.. It's happening when switching to a particular asset.
any ideas how to examine the workfile, maybe it's corrupted.

BigRoy:
You could try running that `get_context_data` call on Houdini host in that scene file to see if it does anything 'wrong' in what it returns.

Also, print `os.environ["AYON_FOLDER_PATH"]` to confirm whether that's indeed invalid.

And as stated, of course keep an eye out on the console to see if there are any errors whatsoever being reported.

MustafaJafar:
TL;DR 
[...]

PR #374, opened 2024-04-04
Workfiles tool: Fix update of context change

iLLiCiTiT:
## Changelog Description
Context change is propagated correctly, because controller of workfiles tool does capture it correctly.

## Additional info
Change the context also in controller, and unify the context change into single method. Removed `reset` of controller on save as ( I'm not sure why the reset was there, probably only to update current context?).

### How to replicate (maybe)
1. Launch DCC -> Task 1.
2. Open workfiles tool.
3. Keep folder, and change task and open workfile. Must be `Open` not `Save as`. -> Task 2
4. Check current context in AYON menu
5. Open workfiles tool again.
6. Again keep folder, change task back to the first one. -> Task 1
7. Context in AYON menu should show context of Task 2

## Testing notes:
1. Workfile opened in same folder, but different task will change the context.

Resolves https://github.com/ynput/ayon-core/issues/180

### DIFF
--- a/client/ayon_core/tools/workfiles/control.py
+++ b/client/ayon_core/tools/workfiles/control.py
@@ -659,16 +659,7 @@ class BaseWorkfileController(
             folder_id != self.get_current_folder_id()
             or task_name != self.get_current_task_name()
         ):
-            folder_entity = ayon_api.get_folder_by_id(
-                event_data["project_name"],
-                event_data["folder_id"],
-            )
-            task_entity = ayon_api.get_task_by_name(
-                event_data["project_name"],
-                event_data["folder_id"],
-                event_data["task_name"]
-            )
-            change_current_context(folder_entity, task_entity)
+            self._change_current_context(project_name, folder_id, task_id)
 
         self._host_open_workfile(filepath)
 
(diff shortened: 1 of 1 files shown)


BigRoy:
Followed the testing notes and seemed to work fine in Maya, Houdini and Fusion.
However, I'm not entirely sure how to reproduce the actual bug this is trying to fix that would exist without this PR.

BigRoy (approved):
Seems to work fine in Maya, Houdini and Fusion. However see my comment - not sure what this fixes.

MustafaJafar (approved):
Tested in Houdini.. works fine.

I wonder if it fixes this issue https://github.com/ynput/ayon-core/issues/180
I tried replicating the issue but it seems it's fixed.

iLLiCiTiT:
Modified the steps a little. With them I was able to replicate the issue consistently.

Animation

BigRoy:
I could not reproduce this in `develop`. I have a feeling that somehow the issue is actually related to the scene file you're opening and that it's maybe reading the "context" from the `OpenPypeContext` metadata that might be saved in the scene.

So what I did was save the modeling scene into lighting WITHOUT updating the context data (e.g. do not update create context, just save the file there without publisher reset, etc.). Then in `develop` indeed I get the issue that changing task would somehow not be at the right context.

**However** note that I could also reproduce that when that file was saved to another folder - not just the task change.

BigRoy:
As such I think the ACTUAL issue isn't the resetting to match with the current context - but the fact that we're somehow allowing to rely on the 'workfile' metadata to be the ground truth for that particular context and allowing it to override the actual session's context.

So even though this workaround PR works - there apparently still is some logic somewhere where "resetting" means resetting it to the invalid context.

MustafaJafar:
> Anyway, with this PR I was unable to intentionally try and break it - so I guess it's safe for now :D

same here.

PR #375, opened 2024-04-04
Maya: Create multiple "render" instances better error message

BigRoy:
## Changelog Description

Provide better descriptive message on the creation failure when trying to create another "render" instance.

## Additional info

The render instance acts as a Singleton instance. When present all renderlayers in the scene are detected as renderable instances - this might not be clear for newcomers at first and they may expect that clicking Create again will in fact create more Render Setup layers. This improved message provides better guidance to new artists

Before:
image

After:
image

## Testing notes:

1. Create a render instance.
2. Create it again. (You'll get the error message)


LiborBatek (approved):
All working fine (nevermind my odd UI scaling as usual due to Win upscale factor to 150%)

image

PR #383, opened 2024-04-09
Unreal: various fixes

antirotor:
## Changelog Description
This PR is fixing some things in ayon-core and unreal workflows. Main issue is that loading of Layout and Camera didn't work it tried to create the whole hierarchy based on the Folder in AYON, but after each item in the hierarchy it crashed. So when you ran the loader multiple times, it would at the end load. This is an attempt to fix it, but I think we will need to refactor the logic somehow anyway.

> [!WARNING]
> This is still live PR with changes coming and going

## Testing notes:
- Loading of pulished Cameras and Layouts should work


antirotor:
closing this because of https://github.com/ynput/ayon-core/pull/614 and also because it is stale

PR #385, opened 2024-04-09
Houdini: Transfer Render Targets settings from pre create to instance when creating ROP instances

moonyuet:
## Changelog Description
Transfer the render target setting from pre-create data to instance data, making sure the setting in the creator tab aligning to that in the publish tab

## Additional info
n/a

## Testing notes:
1. Create ROP instance
2. Choose your preferred render option(local, farm)
3. Go to Publish Tab and check the related setting
4. It should be aligned with what you set in the creator.


MustafaJafar (approved):
After testing this PR. I think we should transfer all `precreate` data to `creator_attributes`

PR #386, opened 2024-04-09
Chore: Use `set` for representation name attribute in LoadPlugin

iLLiCiTiT:
## Changelog Description
Use `set` instead of `list` to define compatible representation names.

## Additional info
This does not break backwards compatibility, using `list` is still possible. The change just helps to define ideal typehint, which is set.

## Testing notes:
Loaders are shown in UI.


moonyuet (approved):
Tested successfully in Blender, Maya, Max

PR #389, opened 2024-04-10
Chore: Mark known bare except handling with noqa

iLLiCiTiT:
## Changelog Description
Bare except error capture is ignored for linters.

## Additional info
PR related to https://github.com/ynput/ayon-core/pull/196. The bare except is usually used when the raised exception does not inherit from python `BaseException` (e.g. in Houdini).

### DIFF
--- a/client/ayon_core/hosts/photoshop/plugins/create/create_image.py
+++ b/client/ayon_core/hosts/photoshop/plugins/create/create_image.py
@@ -53,7 +53,7 @@ class ImageCreator(Creator):
             stub.select_layers(stub.get_layers())
             try:
                 group = stub.group_selected_layers(product_name_from_ui)
-            except:
+            except:  # noqa E722
                 raise CreatorError("Cannot group locked Background layer!")
             groups_to_create.append(group)
 
--- a/client/ayon_core/tools/adobe_webserver/app.py
+++ b/client/ayon_core/tools/adobe_webserver/app.py
@@ -109,7 +109,7 @@ class WebServerTool:
         try:
             sock.bind((host_name, port))
             result = False
-        except:
+        except:  # noqa E722
             print("Port is in use")
 
         return result
--- a/client/ayon_core/hosts/blender/api/lib.py
+++ b/client/ayon_core/hosts/blender/api/lib.py
@@ -33,7 +33,7 @@ def load_scripts(paths):
         if register:
             try:
                 register()
-            except:
+            except:  # noqa E722
                 traceback.print_exc()
         else:
             print("\nWarning! '%s' has no register function, "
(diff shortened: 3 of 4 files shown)


BigRoy:
I feel like some of these are better of being more explicit Exceptions instead of ignoring them. The linter error in that sense is actually 'correct' and should be fixed - not ignored?

And in the cases that the bare except really is on purpose - we're likely best off commenting _why_ that is the case near that code.

iLLiCiTiT:
Replaced one bare except. I know that create plugin is using bare except because host can raise exception not inheriting from `BaseException`. And I have no idea about the other places, if you do, please make suggestion.

client/ayon_core/hosts/photoshop/plugins/create/create_image.py
```diff
@@ -53,7 +53,7 @@ def create(self, product_name_from_ui, data, pre_create_data):
             stub.select_layers(stub.get_layers())
             try:
                 group = stub.group_selected_layers(product_name_from_ui)
-            except:
+            except:  # noqa E722
```
BigRoy, line 56: This may need to be a `ValueError` check if it's [this error](https://github.com/ynput/ayon-core/blob/78a895b720fda253a37725ce30814ca85abafdb9/client/ayon_core/hosts/photoshop/api/ws_stub.py#L292) but I'm not sure. Maybe @kalisp knows or can confirm?

client/ayon_core/tools/adobe_webserver/app.py
```diff
@@ -104,14 +104,11 @@ def port_occupied(self, host_name, port):
...
-        try:
-            sock.bind((host_name, port))
-            result = False
-        except:
-            print("Port is in use")
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as con:
+            result = con.connect_ex((host_name, port)) == 0
 
+        if result:
+            print("Port is in use")
```
BigRoy, line 111: I wonder if it's worth printing which port is already in use?
```suggestion
        if result:
            print(f"Port {port} is already in use")
```

client/ayon_core/hosts/photoshop/plugins/create/create_image.py
antirotor, line 56: Maybe we can chain it like this. I think it is only `ValueError` here we are interested it.
```suggestion
            except ValueError as e:
                raise CreatorError(e) from e
```

antirotor (changes requested):
Apart that photoshop creator, I'd say that all of them seems reasonable.

client/ayon_core/hosts/photoshop/plugins/create/create_image.py
kalisp: Yep, modified.

kalisp: Moved ValueError exception higher.

kalisp (approved):

kalisp:
Update PS creator to catch expected ValueError

antirotor (approved):

PR #394, opened 2024-04-10
Chore: Small code cleanup

iLLiCiTiT:
## Changelog Description
Small code cleanup. Mark unused variables and imports that have purpose to be ignored by linters, or removed the variables. Added imported variables to `__all__`. Converted lambda to function. Use better variable naming.

## Additional info
PR related to https://github.com/ynput/ayon-core/pull/196.

## Testing notes:
1. Logic did not change. Validate if it's true.

### DIFF
--- a/client/ayon_core/hosts/unreal/api/__init__.py
+++ b/client/ayon_core/hosts/unreal/api/__init__.py
@@ -28,11 +28,11 @@ from .pipeline import (
 )
 
 __all__ = [
-    "install",
-    "uninstall",
     "UnrealActorCreator",
     "UnrealAssetCreator",
     "Loader",
+    "install",
+    "uninstall",
     "ls",
     "publish",
     "containerise",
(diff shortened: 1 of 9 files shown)


client/ayon_core/hosts/unreal/api/__init__.py
```diff
@@ -30,6 +30,8 @@
 __all__ = [
     "install",
     "uninstall",
+    "UnrealActorCreator",
+    "UnrealAssetCreator",
     "Loader",
     "ls",
```
BigRoy, line 36: Should we move `install` and `uninstall` to after the `Loader` so that it matches the imports up top?
```suggestion
    "UnrealActorCreator",
    "UnrealAssetCreator",
    "Loader",
    "install",
    "uninstall",
    "ls",
```

BigRoy (approved):
Didn't test - but changes look good to me.

PR #396, opened 2024-04-10
Ruff: Add list of files to exclude from linting

iLLiCiTiT:
## Changelog Description
Added list of files that should be ignored for ruff when.

## Additional information
PR related to https://github.com/ynput/ayon-core/pull/196 .

## Testing notes:
- Validate if excluded paths make sense to be ignored.
- Validate that ruff is actually ignoring them.


BigRoy:
What is the reason that we're ignoring some instead of fixing them?

iLLiCiTiT:
> What is the reason that we're ignoring some instead of fixing them?

Please do if you have spare time.

antirotor (approved):
Most of the things will be moved to their respective repositories anyway, what's left here should be then definitely fixed

PR #400, opened 2024-04-10
Substance Painter: Allow users to customize the template settings for project creation

moonyuet:
## Changelog Description
This PR is to add the support of template settings in the mesh loaders for Substance project creation. User can customize and add template settings in AYON settings and apply it through the option mode(the button with memo icon).

## Additional info
This PR also refactors the original loader's setting without directly import qargparse.
Prefers user to use options (memo icon) to set up their preferences as the prompt dialog might not set up the project as expected(It is still experimental).

Make sure to update your Substance Painter Addon to `0.1.1`

## Testing notes:
1. Customize the template with a template name in AYON setting
image
2. Launch Substance Painter
3. Click Load..
4. It would pop up the dialog to allow you to choose your template settings
image
5. Click Ok
6. It should apply all the settings you set in AYON settings


LiborBatek (approved):
Does work and as expected. Did a few test with different settings...all fine.

image

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -25,26 +25,35 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
+            BoolDef("import_cameras",
+                    default=True,
+                    label="Import Cameras",
+                    tooltip="Import cameras from the mesh file."
+            ),
+            EnumDef("texture_resolution",
+                    items=[128, 256, 512, 1024, 2048, 4096],
+                    default=1024,
+                    label="Texture Resolution",
+                    tooltip="Set texture resolution for the project")
```
BigRoy, line 46: Should the tooltip elaborate that this only works for "NEW" projects - not for loading into an existing project? Or maybe we put in a "divider" in the definitions with a label like "New Project Settings:" or alike?

client/ayon_core/hosts/substancepainter/api/lib.py
```diff
@@ -624,7 +623,8 @@ def _setup_prompt():
                 f"{mesh_filepath}\n\n"
                 "Creating new project directly with the mesh path instead.")
         else:
-            dialog.done(dialog.Accepted)
+            if not allow_user_setting:
+                dialog.done(dialog.Accepted)
```
BigRoy, line 627: I feel like if we do not allow user settings is there still a reason to go through the convoluted steps of menu entries above, etc. instead of just a "create project" call without dialogs?

moonyuet:
> Does work and as expected. Did a few test with different settings...all fine.
> 
> image

@LiborBatek, as you mentioned in the DM in discord, I have added other settings to allow users to set the project texture resolution and udim settings or any other settings. And the users can choose to turn off the user preferences for directly loading mesh with preset settings and create project.

client/ayon_core/hosts/substancepainter/api/lib.py
BigRoy: I understand - but isn't this particular toggle intended to disable that prompt?

moonyuet: > I understand - but isn't this particular toggle intended to disable that prompt?

maybe this is not a good naming of variable for this toggle. If "allow_user_setting" is disabled, the system would take preset and click ok to the dialog after loading the mesh. If "allow_user_setting" is enabled, the dialog would be display and user can set the setting after loading the mesh.

BigRoy: Exactly. So if this option is False - there is not reason for the dialog to come up to begin with. We could just use substance's API to create the project _without_ even trying with a workaround to bring up the dialog.

moonyuet: > Exactly. So if this option is False - there is not reason for the dialog to come up to begin with. We could just use substance's API to create the project _without_ even trying with a workaround to bring up the dialog.

It is by default True, this PR is to make the dialog not disappearing.

BigRoy: > The option is by default True, this PR is to make the dialog not disappearing, we still need the dialog so that the user can configure the setting and click ok to create project

Let me rephrase:

If there's any options that "disappears" the dialog as if the user should not interact with it. Then we should just _NOT_ show the dialog to begin with instead of popping up the dialog and accepting it. So if the default `True` means "do not allow the  user to interact with a dialog" then we should just not open the dialog but instead use the api to create the project so we don't need to go through the UI. Yes?

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -25,34 +25,50 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
+                    tooltip=("Preserve strokes positions on mesh.\n"
+                             "(only relevant when loading into "
+                             "existing project)")),
+            BoolDef("import_cameras",
+                    default=True,
+                    label="Import Cameras",
+                    tooltip="Import cameras from the mesh file."
+            ),
+            EnumDef("texture_resolution",
+                    items=[128, 256, 512, 1024, 2048, 4096],
```
BigRoy, line 48: With this PR coming up: https://github.com/ynput/ayon-core/pull/407 - I'd expect the project to also be able to go up to 8k. Are we missing an option here?

moonyuet: Yes, but I just make a separate PR for that.

client/ayon_core/hosts/substancepainter/api/lib.py
moonyuet: the default `True` means "it allows the user to interact with a dialog and they can click ok to create project"

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
BigRoy: But that's for the export size, not the project size - right?

moonyuet: > But that's for the export size, not the project size - right?

yes it's different maybe I elaborate the name a little bit

BigRoy: Anyway. So should this PR include a higher project resolution in this `items` list? I might be misunderstanding from your reply why it shouldn't.
```suggestion
                    items=[128, 256, 512, 1024, 2048, 4096, 8192],
```

moonyuet: > Anyway. So should this PR include a higher project resolution in this `items` list? I might be misunderstanding from your reply why it shouldn't.

But in the texture configuration setting, it doesn't have the 8k supports, it is only limited to the texture export.

BigRoy: Interesting - why would you do an 8k export for something that's based on 4k maps? What am I missing?

moonyuet: > Interesting - why would you do an 8k export for something that's based on 4k maps? What am I missing?

there is an option for the texture export. It's in the export setting.
image

LiborBatek: @moonyuet  the ```Project Config``` pop up window shows just for a few milisecs and dissappears :(

@BigRoy SPainter can do recalculation when exporting the 8K even tho the project being just 4K (they mention it in the docs - so it upres the output really)

BigRoy: > @BigRoy SPainter can do recalculation when exporting the 8K even tho the project being just 4K (they mention it in the docs - so it upres the output really)

You just got to love software - well I guess that 'explains' it then.

LiborBatek:
@moonyuet When keep ```Allow User Settings``` ON

image

This Window which is crucial to set up UDIMs if need be just pops and dissappears immedeately

Screenshot 2024-04-11 162657

moonyuet:
@LiborBatek I believe we need to test on the dialog to see if it is possible for user configuring the setting(It is not quite user family for my current commit in my opinion ). My finding is that dialog can disappear depends on what file extension and how big file size of the loaded meshes are. I mark it draft first, if it is not quite doable in the current stage. I guess I will also make the udim settings as part of the settings for the mesh loader in SP.

LiborBatek:
Im for when loading the ```model``` for very first time the ```Project Settings```  should show up always. If it is not doable technically, I would be for popping up our toolbox (the one you are working on in this PR) also automatically and always... not just as an optional tool aka options as it is now...

I wouldnt mind to just dismiss this dialog each time...its better that forget about to set it to your liking and also its a bit hidden feature not much obvious imho

moonyuet:
> Im for when loading the `model` for very first time the `Project Settings` should show up always. If it is not doable technically, I would be for popping up our toolbox (the one you are working on in this PR) also automatically and always... not just as an optional tool aka options as it is now...
> 
> I wouldnt mind to just dismiss this dialog each time...its better that forget about to set it to your liking and also its a bit hidden feature not much obvious imho
> 
> ...only questionable, what should happen if user select "Cancel" on our pop up...probably some SP defaults, something like Resolution 1024, and all other options turned ON

I tweaked the code a little bit and remove the allow user setting. If the prompt dialog cannot create project, it is still using substance api to create project. can you please test it again?

moonyuet:
> Im for when loading the `model` for very first time the `Project Settings` should show up always. If it is not doable technically, I would be for popping up our toolbox (the one you are working on in this PR) also automatically and always... not just as an optional tool aka options as it is now...
> 
> I wouldnt mind to just dismiss this dialog each time...its better that forget about to set it to your liking and also its a bit hidden feature not much obvious imho
> 
> ...only questionable, what should happen if user select "Cancel" on our pop up...probably some SP defaults, something like Resolution 1024, and all other options turned ON

@LiborBatek I have updated the PR which allow the users to set up with the prompt dialog, if the users can't really create the project by this case, they can still use the option settings in the mesh loader

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -11,7 +12,15 @@
...
-import qargparse
+
+
+def get_uv_workflow(uv_option="default"):
+    if uv_option == "default":
+        return substance_painter.project.ProjectWorkflow.Default
+    elif uv_option == "uvTile":
+        return substance_painter.project.ProjectWorkflow.UVTile
+    else:
+        return substance_painter.project.ProjectWorkflow.TextureSetPerUVTile
```
BigRoy, line 23: Let's be more explicit.
```suggestion
def get_uv_workflow(uv_option="default"):
    if uv_option == "default":
        return substance_painter.project.ProjectWorkflow.Default
    elif uv_option == "uvTile":
        return substance_painter.project.ProjectWorkflow.UVTile
    elif uv_option == "textureSetPerUVTile":
        return substance_painter.project.ProjectWorkflow.TextureSetPerUVTile
    else:
        raise ValueError(f"Unsupported UV option: {uv_option}")
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -54,12 +85,16 @@ def load(self, context, name, namespace, data):
...
+                if not substance_painter.project.is_open():
+                    self.log.info("User cancelled new project prompt."
+                                  "Creating new project directly from"
+                                  " Substance Painter API Instead.")
+                    settings = substance_painter.project.create(
+                        mesh_file_path=path, settings=sp_settings
+                    )
+                else:
+                    self.log.info("The project is already created after "
+                                  "the new project prompt action")
```
BigRoy, line 97: How or when does this occur? This seems like we have an error in the logic elsewhere?

moonyuet:
@BigRoy, Guess the codes for prompt dialog can be removed with this PR after the testing(If it is okay)

client/ayon_core/hosts/substancepainter/api/lib.py
```diff
@@ -643,3 +640,86 @@ def _setup_prompt():
...
+            Project was detected
+
+    Returns:
+        python instance: converted python instance of the C++ objects.
+    """
+    if subst_proj_option == "default":
+        return substance_painter.project.ProjectWorkflow.Default
+    elif subst_proj_option == "uvTile":
+        return substance_painter.project.ProjectWorkflow.UVTile
+    elif subst_proj_option == "textureSetPerUVTile":
```
BigRoy, line 665: These are the only three that resolve a lowercase first letter whereas the others match exactly with the python attribute name.

Also, I think it's not a good idea to have on function that remaps a string to attributes of different properties, e.g. `ProjectWorkflow.TextureSetPerUVTile` and `TangentSpace.PerFragment`. That's just waiting for a future update of substance painter suddenly adds `TangentSpace.Default` as well breaking the logic, but worse - calling `convert("default")` is just unclear to what of the attributes it converts.

We could also implement like this:
```python
def _convert(attr):
    value = substance_painter.project
    for key in attr.split("."):
        value = getattr(root, key, default=None)
        if value is None:
            return
    return value
```

And then make the values the string including the dots:
```python
_convert("ProjectWorkflow.TextureSetPerUVTile")
```

client/ayon_core/hosts/substancepainter/api/lib.py
```diff
@@ -643,3 +640,86 @@ def _setup_prompt():
...
+            attributes_data.update(template)
+    attributes_data["normal_map_format"] = convert_substance_object_to_python(
+        subst_proj_option=attributes_data["normal_map_format"])
+    attributes_data["project_workflow"] = convert_substance_object_to_python(
+        subst_proj_option=attributes_data["project_workflow"])
+    attributes_data["tangent_space_mode"] = convert_substance_object_to_python(
+        subst_proj_option=attributes_data["tangent_space_mode"])
+    attributes_data.pop("name")
+    attributes_data.pop("preserve_strokes")
+    return attributes_data
```
BigRoy, line 703: This feels very specific to some part of the settings or logic, don't see why these need to be in `lib`. They don't seem generic enough.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -24,49 +28,38 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
-    ]
+    @classmethod
+    def get_options(cls, contexts):
+        template_enum = [template["name"] for template in cls.project_templates]
+        return [
+            EnumDef("project_template",
+                    items=template_enum,
+                    default="default",
+                    label="Project Template")
+        ]
```
BigRoy, line 41: I personally really wouldn't want "template" defaults to be driven like this from the settings. Instead I'd want to expose the artist to the actual settings for what resolution they need, etc. This just feels like suddenly we'll be getting templates that are just 'unclear' as to what they are actually doing.

```
2048x2048 (UDIM | Texture Set Per UV Tile | Preserve strokes)
```

Also, preserve strokes is not a studio settings template decision - it's an artist decision on how they want to propagate the a mesh **update** into their scene.

BigRoy (commented):
I think the decision to turn this into generalized studio settings templates instead of an artist being able to specify the substance painter project settings themselves is a bad decision. Preferably we're the _thinnest layer possible_ over how Substance Painter works. If Substance Painter exposes certain options and the pipeline functions completely fine if those settings can be configured by an artist - then we should allow them to be configured by an artist - instead of hiding them away in a way where they can't control it.

Note that artists would usually not touch studio or project settings and thus usually don't access the setting up of these templates - basically meaning they can't take control on how they want to configure their substance painter project which I think is a bad thing. No?

moonyuet:
> I think the decision to turn this into generalized studio settings templates instead of an artist being able to specify the substance painter project settings themselves is a bad decision. Preferably we're the _thinnest layer possible_ over how Substance Painter works. If Substance Painter exposes certain options and the pipeline functions completely fine if those settings can be configured by an artist - then we should allow them to be configured by an artist - instead of hiding them away in a way where they can't control it.
> 
> Note that artists would usually not touch studio or project settings and thus usually don't access the setting up of these templates - basically meaning they can't take control on how they want to configure their substance painter project which I think is a bad thing. No?

@LiborBatek any thought on the template settings? 
We should just put into all current settings into all options.

LiborBatek:
@BigRoy  I get your points but let me clarify few things...

Because we arent able to control SP Project Config window and its poping up when needed (as I was told by @moonyuet  that its very problematic via API of SP and done before via some hacky way by you) I have proposed to use just ```Templates``` and select such when first loading model into SP aka starting new work.

Its of course questionable which of the settings to include inside the ```Template profile``` and we can chat about that, but one thing to note, there are few which cant be changed later - mostly uv tile mode (default or uv tiles aka UDIMs) these being crucial to set at the very begining of user work. The rest as for example ```Texture Set``` and its ```Resolution``` can be switched any time back and forward as there arent any restrictions in that manner in the SP. The same goes for output resolution.

So in the end its mostly about the user to choose fitting ```Template``` at particular occasion and rest can be adjusted anytime by user later on.

User also can benefit from templating at the end as not all artist are  very technical and these  could help em to set project correctly just selecting the profile name.

To me its a nice workaround if again ```Project Config Window``` being somewhat problematic in terms of our integration.

BigRoy:
> Because we arent able to control SP Project Config window and its poping up when needed (as I was told by @moonyuet that its very problematic via API of SP and done before via some hacky way by you) I have proposed to use just `Templates` and select such when first loading model into SP aka starting new work. And simply set project and its configuration via command arguments not needing to trigger that UI for that.

In that case - we should be instead popping our own dialog that we do have full control over instead that mimics the Substance Painter dialog settings. That custom Qt UI can then definitely have a "choose preset" functionality like being implemented now from settings - but it should really allow custom user control as well.

I'm not saying we shouldn't set these settings at all - I'm saying: Do not define them by settings, but allow the artist to pick them via OUR tools instead of them picking a 'preset' they have no knowledge of what it contains. We should mimic Substance Painter as much as possible, so that if they are unsure about something they can refer to the substance painter documentation as much as possible. As such, I'd structure our custom dialog exactly like Substance painter's "new project" dialog or alike if we're unable to use their dialog.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -1,23 +1,130 @@
...
+    """
+    if subst_attr in {"Default", "UVTile", "TextureSetPerUVTile"}:
+        return getattr(substance_painter.project.ProjectWorkflow, subst_attr)
+    elif subst_attr in {"PerFragment", "PerVertex"}:
+        return getattr(substance_painter.project.TangentSpace, subst_attr)
+    elif subst_attr in {"DirectX", "OpenGL"}:
+        return getattr(substance_painter.project.NormalMapFormat, subst_attr)
+    else:
+        raise ValueError(
+            f"Unsupported Substance Objects: {subst_attr}")
```
BigRoy, line 39: I think you've now still kept the old issue that it's unclear that `Default` refers to ProjectWorkflow and that `OpenGL` refers to `NormalMapFormat` when passing it to this function. I think we're better off passing it explicit strings like `ProjectWorkflow.Default` and `NormalMapFormat.OpenGL`.

moonyuet: But if I have done something like you mentioned, doesn't it mean the same setting in AYON turned out to be ProjectWorkflow.Default? If yes, this is becoming really abstract for user to recognize what they want to set up.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,120 @@
...
+        dict: data to be used as attributes for reloading mesh with the
+            existing project
+    """
+    attributes_data = {}
+    for template in project_templates:
+        if template["name"] == template_name:
+            for key, value in template.items():
+                if isinstance(value, bool):
+                    attributes_data.update({key: value})
+    return attributes_data
```
BigRoy, line 87: I really have a hard time understanding the differences between these functions.
These are much easier if they just take a template - and are explicit about what keys they take instead of checking on boolean types, etc.
And then having a dedicated 'get_template' function or whatever if really needed.

For example, it's also unclear which of these are now setting the "Default Texture Resolution" value from settings?

```suggestion
def get_project_settings(template):
    """Return project settings based on template from settings.

    Args:
        template (dict): template from settings

    Returns:
        dict: data to be used as attributes for Substance Project Creation
    """
    keys = [
        "normal_map_format",
        "project_workflow",
        "tangent_space_mode",
        "default_texture_resolution"
    ]
    return {
        key: _convert(template[key]) for key in keys
    }

def get_mesh_reloading_settings(template):
    """Function to parse the substances attributes ('import_cameras'
        and 'preserve_strokes') for reloading mesh
        with the existing projects.

    Args:
        template (dict): template from settings

    Returns:
        dict: data to be used as attributes for reloading mesh with the
            existing project
    """
    keys = ["import_cameras", "preserve_strokes"]
    return {
       key: template.get(key, False) for key in keys
    }
```

Note that this first one will only work if `_convert` ignores the integer value from `default_texture_resolution` key.
But - this may be even further simplified if we just explicitly create the `substance_painter.project.Settings` so we don't need these functions at all:

```python
settings = substance_painter.project.Settings(
        normal_map_format=_convert(template["normal_map_format"]),
        project_workflow=_convert(template["project_workflow"]),
        tangent_space_mode=_convert(template["tangent_space_mode"]),
        default_texture_resolution=template["default_texture_resolution"]
)`
This seems quite explicit, quite readable and very clear what is expected from the template - instead of having separate functions that also appear rather complex?

File: client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py [hunk 11]
```diff
@@ -8,10 +9,120 @@
...
-from ayon_core.hosts.substancepainter.api.lib import prompt_new_file_with_mesh
 
 import substance_painter.project
-import qargparse
+
+
+def _convert(subst_attr):
+    """Function to convert substance C++ objects to python instance.
+    It is made to avoid any possible ValueError when C++ objects casting
+    as python instance.
```
BigRoy, line 19: It's not casting C++ objects. It's converting strings to the Substance Painter Python API equivalent objects.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
moonyuet: But does it work when there are multiple templates existing?

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,120 @@
...
+        layout = QtWidgets.QHBoxLayout(self.widgets["buttons"])
+        layout.addWidget(self.widgets["template_options"])
+        layout.addWidget(self.widgets["okButton"])
+        # Build layout.
+        layout = QtWidgets.QVBoxLayout(self)
+        layout.addWidget(self.widgets["label"])
+        layout.addWidget(self.widgets["buttons"])
+
+        self.widgets["okButton"].pressed.connect(self.on_ok_pressed)
+
```
BigRoy, line 120: I would build it so that the "presets" are just that - presets, but in the end each preset still sets the individual attributes in the dialog so that the artist can use a preset as a basis to quickly select a preset - but still have full control and see what the individual settings are.

Currently this just lists the presets and offers no visual difference to the "options" dialog you had before from loader options?

My point was to provide the user ALL the options they want - including picking resolution, UDIM or not, etc. so that to the artist it still feels like how they would configure substance painter.

However, since you mentioned someone wanted presets - then we can expose 'presets' from settings to the dialog to quickly toggle those attributes to "preset values" from the settings so you have both, but still allow customization by the artist. No?

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
BigRoy: > But if I have done something like you mentioned, doesn't it (e.g. Default) mean the same setting in AYON turned out to be ProjectWorkflow.Default? If yes, this is becoming really abstract for user to recognize what they want to set up.

The backend value stored in the settings, etc. - yes. But the labels on the enums or settings could be `Default` just fine. But in the code we would always be working with the explicit strings - that's what I'm saying.

server_addon/substancepainter/server/settings/load_plugins.py
```diff
@@ -0,0 +1,120 @@
...
+            "project_workflow": "UVTile",
+            "tangent_space_mode": "PerFragment",
+            "preserve_strokes": True
+        },
+        {
+            "name": "4K(Custom)",
+            "default_texture_resolution": 4096,
+            "import_cameras": True,
+            "normal_map_format": "OpenGL",
+            "project_workflow": "UVTile",
```
BigRoy, line 115: Shouldn't the project templates settings key be a dict instead of list so that the `name` is ensured to be unique?

moonyuet: It supports multiple template settings. so that's why it is a list of the dicts

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
moonyuet: I have discussed with @LiborBatek about this.  The reason why there is template setting to make the workflow more streamlined. It's necessary for the manager/TD to align the project settings so that it is more unified when the texture exported from substance and used it in other DCCs

BigRoy: > It's necessary for the manager/TD to align the project settings so that it is more unified when the texture exported from substance and used it in other DCCs

I guess fine with me - then maybe later I might expose a setting that allows to enable customization by the artist instead of picking from presets.

moonyuet:
@LiborBatek Can you build and update the SP addon and test it again?
Make sure you tweak the settings in each project template in the screengrab below before the test, otherwises it will take the old data and errors out during loading mesh. 

image

server_addon/substancepainter/server/settings/load_plugins.py
BigRoy: Sorry, how does it support multiple template settings? It support multiple, by name, right?

So in typing it would be:
```python
project_templates: Dict[str, ProjectTemplatesModel]
```

So that you have e.g.

```
{
    "2K(Default)": {... default template ...},
    "2K(UV tile)": {.. other template..}
    "4K(Custom)": { .. custom .. }
}
```

Also, as default I'd have NO idea what 4K(Custom) means - as such, it's just an odd default?

Additionally I'd argue that `preserve_strokes` should really be taken out of the templates, that's really an artist decision on updates. `import_cameras` I'd say is as well. It doesn't usually influence the exports, and it's up to the artists whether cameras could help their workflow. Right?

BigRoy: Hmm, doesn't seem like settings support `Dict` like that @martastain @iLLiCiTiT , correct?
It usually seems hacked [like this](https://github.com/ynput/ayon-core/blob/3d046fc2161929c46e514a71351e7cc528ff1861/server_addon/maya/server/settings/publishers.py#L249-L257).

Is there maybe a way for the backend to 'automatically' convert `Dict[str, value]` to `List[value]` with a unique name validator if that's what is apparently needed to support `dict` like structures? That would at least make defining dict like settings way easier?

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,78 @@
...
+    Args:
+        subst_attr (str): Substance attributes
+
+    Raises:
+        ValueError: Raise Error when unsupported Substance
+            Project was detected
+
+    Returns:
+        python instance: converted python instance of the C++ objects.
```
BigRoy, line 29: It's not related to C++ objects, only string to python object conversions. (The Substance Painter Python API attribute could be a C++ shim but I'd say we don't care about that, but only really about that it matches the python api)

Other changes:
- According to [this example](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) `Raises` goes below `Returns` in the docstring.
- Docstring argument `subst_attr` -> `substance_attr`

```suggestion
def _convert(substance_attr):
    """Return Substance Painter Python API Project attribute from string.
    
    This converts a string like "ProjectWorkflow.Default" to for example
    the Substance Painter Python API equivalent object, like:
        `substance_painter.project.ProjectWorkflow.Default`

    Args:
        substance_attr (str): The `substance_painter.project` attribute,
            for example "ProjectWorkflow.Default"

    Returns:
        Any: Substance Python API object of the project attribute.

    Raises:
        ValueError: If attribute does not exist on the
            `substance_painter.project` python api.
```

server_addon/substancepainter/server/settings/load_plugins.py
```diff
@@ -0,0 +1,120 @@
...
+            "tangent_space_mode": "PerFragment",
+            "preserve_strokes": True
+        }]
```
BigRoy, line 118: The defaults need updating. That likely also fixes the [issue described here](https://github.com/ynput/ayon-core/pull/400#issuecomment-2061667568).
> Make sure you tweak the settings in each project template in the screengrab below before the test, otherwises it will take the old data and errors out during loading mesh.

```suggestion
        "project_templates": [{
            "name": "2K(Default)",
            "default_texture_resolution": 2048,
            "import_cameras": True,
            "normal_map_format": "NormalMapFormat.DirectX",
            "project_workflow": "ProjectWorkflow.Default",
            "tangent_space_mode": "TangentSpace.PerFragment",
            "preserve_strokes": True
        },
        {
            "name": "2K(UV tile)",
            "default_texture_resolution": 2048,
            "import_cameras": True,
            "normal_map_format": "NormalMapFormat.DirectX",
            "project_workflow": "ProjectWorkflow.UVTile",
            "tangent_space_mode": "TangentSpace.PerFragment",
            "preserve_strokes": True
        },
        {
            "name": "4K(Custom)",
            "default_texture_resolution": 4096,
            "import_cameras": True,
            "normal_map_format": "NormalMapFormat.OpenGL",
            "project_workflow": "ProjectWorkflow.UVTile",
            "tangent_space_mode": "TangentSpace.PerFragment",
            "preserve_strokes": True
        }]
```

LiborBatek (approved):
All working fine and predictably. Did test different template configs and all good!
Screenshot 2024-04-18 095753

LiborBatek:
@BigRoy I think it follows also your opinion...not hiding anything from the user and let him choose. And its much streamlined and helping user be faster. If needed, the user is free to dive into SP native ```Project Config``` menu and tweak it to his liking too...

Im really fine with this approach and working nicely imho!

BigRoy:
> @BigRoy I think it follows also your opinion...not hiding anything from the user and let him choose. And its much streamlined and helping user be faster. If needed, the user is free to dive into SP native `Project Config` menu and tweak it to his liking too...
> 
> Im really fine with this approach and working nicely imho!

Cool - if it works fine, let's roll it out. I'll check it out over time. Maybe others are already a big fan and my ideas on how it should work maybe have diminished. Thanks for pushing through. Nice work.

server_addon/substancepainter/server/settings/load_plugins.py
```diff
@@ -0,0 +1,122 @@
...
+                "preserve_strokes": True
+            },
+            {
+                "name": "4K(Custom)",
+                "default_texture_resolution": 4096,
+                "import_cameras": True,
+                "normal_map_format": "NormalMapFormat.OpenGL",
+                "project_workflow": "ProjectWorkflow.UVTile",
+                "tangent_space_mode": "TangentSpace.PerFragment",
+                "preserve_strokes": True
```
BigRoy, line 118: Should we still take the `import_cameras` and `preserve_strokes` out of the template @moonyuet ?

moonyuet: Do you mean adding these two options into the Project Configuration Dialog and let the user choose?

BigRoy: Yes - exactly. Since that came up before. especially because they seem artist choices - and `preserve_strokes` is really only crucial on updates based on what the artist wants to do?

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -24,49 +112,37 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
+        import_cameras = window.import_cameras
+        preserve_strokes = window.preserve_strokes
+        template = get_template_by_name(template_name, self.project_templates)
         sp_settings = substance_painter.project.Settings(
-            import_cameras=import_cameras
+            normal_map_format=_convert(template["normal_map_format"]),
+            project_workflow=_convert(template["project_workflow"]),
+            tangent_space_mode=_convert(template["tangent_space_mode"]),
+            default_texture_resolution=template["default_texture_resolution"]
         )
```
BigRoy, line 133: Can we move the variables only relevant to the `if` statement into the `if` and those only used by the `else` into the `else`?

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,97 @@
...
+
+    def on_ok_pressed(self):
+        if self.widgets["import_cameras"].isChecked():
+            self.import_cameras = True
+        if self.widgets["preserve_strokes"].isChecked():
+            self.preserve_strokes = True
+        self.template_name = (
+            self.widgets["template_options"].currentText()
+        )
+        self.close()
```
BigRoy, line 102: I wonder if it's better to make a `staticmethod` or `classmethod` that just gets the templates and visualizes them and at the end returns a dict of the resulting values. I think the behavior would become way clearer.

Along the lines of:
```python
    @classmethod
    def prompt(cls, templates):
        dialog = cls(templates)
        dialog.exec_()
        return dialog.get_result()
```

So that e.g. when wanting to ask the user for values you can just do:
```python
result = SubstanceProjectConfigurationWindow.prompt(templates)
```

---

By the way, what happens if the user 'cancels' the dialog? I believe currently it will just continue the logic with the default values, not the picked values at that point in time.

Even better maybe if it would stop the full operation completely?

LiborBatek:
> > @LiborBatek @BigRoy I have updated the dialog with the options of preserve strokes and import cameras. image
> 
> In this design I'd expect the OK button at the bottom because now it feels like the OK button is for the combobox somehow. Or does this look fine to you @LiborBatek ?

Yeah, you are probably right, even tho its just a cosmetics.

Would be lovely to follow this kind of dialog (not size wise but more like a style with header ```Project Configuration``` and ```OK``` button at the very bottom as sort of confirmation of all the listed entries.

Not sure if possible or what it takes tho @moonyuet ?

If problematic...I would go with what we have for now....

@BigRoy Does that make sense to you? thx

moonyuet:
> > In this design I'd expect the OK button at the bottom because now it feels like the OK button is for the combobox somehow. > Or does this look fine to you @LiborBatek ?
> 
> Yeah, you are probably right, even tho its just a cosmetics.
> 
> Would be lovely to follow this kind of dialog (not size wise but more like a style with header `Project Configuration` and `OK` button at the very bottom as sort of confirmation of all the listed entries.
> 
> Not sure if possible or what it takes tho @moonyuet ?
> 
> dialog
> 
> If problematic...I would go with what we have for now....
> 
> @BigRoy Does that make sense to you? thx

I mean with the current dialog implementation I can get only this below.
image
I also links to `import_cameras` and `preserve_strokes` to the AYON settings while the artists still have the freedoms to decide if they want to import cameras and preserve strokes.

BigRoy:
> I mean with the current dialog implementation I can get only this below.
> image
> I also links to `import_cameras` and `preserve_strokes` to the AYON settings while the artists still have the freedoms to decide if they want to import cameras and preserve strokes.

If I understand you correctly - then changing the template resets the import cameras and preserve strokes to the configuration in settings?

If so, then the template combobox should be at the top - hinting more that _that_ is what you toggle first before touching the others. By moving the combobox to the top it's also closer to the label at the top making it better to distinguish that the combobox is what describes the 'template' to pick.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,136 @@
...
+        layout.addWidget(self.widgets["import_cameras"])
+        layout.addWidget(self.widgets["preserve_strokes"])
+        # Build combobox
+        layout = QtWidgets.QHBoxLayout(self.widgets["combobox"])
+        layout.addWidget(self.widgets["template_options"])
+
+        # Build buttons
+        layout = QtWidgets.QHBoxLayout(self.widgets["buttons"])
+        layout.addWidget(self.widgets["okButton"])
+        layout.addWidget(self.widgets["cancelButton"])
```
BigRoy, line 96: You could also use a [`QDialogButtonBox`](https://doc.qt.io/qt-5/qdialogbuttonbox.html)? 

Something along the lines of
```python
buttons = QtWidgets.QDialogButtonBox()
buttons.add(QtWidgets.QDialogButtonBox.Ok)
buttons.add(QtWidgets.QDialogButtonBox.Cancel)
```

I believe in a `QDialog` these also directly register as the default accept and cancel buttons so that you can list to the actual dialog's `QDialog.accepted` and `QDialog.rejected` signals. That way you also respond to e.g. a user clicking escape to close the dialog?

moonyuet: Let's me try to see if this is possible for implementing `dialog.accepted` and `dialog.rejected` in prompt along with `QtWidgets.QDialogButtonBox()`

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -24,49 +151,37 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
         # Get user inputs
-        import_cameras = data.get("import_cameras", True)
-        preserve_strokes = data.get("preserve_strokes", True)
+        result = SubstanceProjectConfigurationWindow.prompt(
+            self.project_templates)
+        template_name = result.template_name
+        if template_name is None:
+            return
+        import_cameras = result.import_cameras
+        template = get_template_by_name(template_name, self.project_templates)
```
BigRoy, line 165: I think getting the template resulting values are best to get from the dialog directly - so that the dialog can return the resulting values including any overrides to the import cameras, etc.

```python
template = dialog.get_template()
if not template:
    # user cancelled
    return

...
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -24,49 +163,36 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
-    ]
-
-    def load(self, context, name, namespace, data):
+    def load(self, context, name, namespace, options=None):
 
         # Get user inputs
-        import_cameras = data.get("import_cameras", True)
-        preserve_strokes = data.get("preserve_strokes", True)
+        result = SubstanceProjectConfigurationWindow.prompt(
+            self.project_templates).get_result()
```
BigRoy, line 172: This _may_ fail. The `get_result` command gets the values from the Qt widgets after the dialog is closed - the closing of the dialog MAY delete the qt widgets on the C++ side (if not, then we have a memory leak?)

So technically, we must ensure we query the values from the QtWidgets before they get deleted - which as far as I know means getting it consistently before closing to the dialog. Or set the dialog to not delete on close automatically (which may be the default behavior, but then we currently have the memory leak where we are just generating more dialogs over time in memory all the time and never deleting the previous dialogs) and then get the values and then explicitly mark the dialog for deletion, e.g. `dialog.deleteLater()` or whatever.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,148 @@
...
+
+    def get_boolean_setting(self, template_name):
+        self.import_cameras = next(template["import_cameras"] for
+                                   template in self.project_templates
+                                   if template["name"] == template_name)
+        self.preserve_strokes = next(template["preserve_strokes"] for
+                                     template in self.project_templates
+                                     if template["name"] == template_name)
+        self.widgets["import_cameras"].setChecked(self.import_cameras)
+        self.widgets["preserve_strokes"].setChecked(self.preserve_strokes)
```
BigRoy, line 132: I don't see a reason to store these as attributes on the dialog instance - the Qt widgets themselves are the ground truth for the values we want to end up using.
```suggestion
        template = get_template_by_name(template_name, self.project_templates)
        self.widgets["import_cameras"].setChecked(template["import_cameras"])
        self.widgets["preserve_strokes"].setChecked(template["preserve_strokes"])
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,148 @@
...
+
+    def get_result(self):
+        import copy
```
BigRoy, line 135: Should be imported at top, not here.
```suggestion
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,148 @@
...
+        self.widgets["buttons"].rejected.connect(self.on_cancel_pressed)
+
+    def on_options_changed(self, value):
+        self.get_boolean_setting(value)
+
+    def on_ok_pressed(self):
+        if self.widgets["import_cameras"].isChecked():
+            self.import_cameras = True
+        if self.widgets["preserve_strokes"].isChecked():
+            self.preserve_strokes = True
```
BigRoy, line 114:
```suggestion
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,148 @@
...
+        self.template_name = (
+            self.widgets["template_options"].currentText()
+        )
+        self.close()
+
+    def on_cancel_pressed(self):
+        self.template_name = None
+        self.close()
+
+    def get_boolean_setting(self, template_name):
```
BigRoy, line 124: Function name does not describe that it's actually UPDATING values on the qt widgets. Something like `_update_to_match_template` would be clearer.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,148 @@
...
+    """The pop-up dialog allows users to choose material
+    duplicate options for importing Max objects when updating
+    or switching assets.
+    """
+    def __init__(self, project_templates):
+        super(SubstanceProjectConfigurationWindow, self).__init__()
+        self.setWindowFlags(self.windowFlags() | QtCore.Qt.FramelessWindowHint)
+
+        self.import_cameras = False
+        self.preserve_strokes = False
```
BigRoy, line 62:
```suggestion
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
moonyuet: I am actually adjusting the code now. just give me some time to adjust that

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +9,148 @@
...
+            self.import_cameras = True
+        if self.widgets["preserve_strokes"].isChecked():
+            self.preserve_strokes = True
+        self.template_name = (
+            self.widgets["template_options"].currentText()
+        )
+        self.close()
+
+    def on_cancel_pressed(self):
+        self.template_name = None
```
BigRoy, line 121: We don't need to set it to `None` because it should be `None` by default - we are only setting a value to it if the user accepts the dialog.
```suggestion
```

moonyuet: i would remove this in the upcoming update.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +10,144 @@
...
+            return None
+        template = get_template_by_name(self.template_name, templates)
+        template = copy.deepcopy(template) # do not edit the original
+        template["import_cameras"] = self.widgets["import_cameras"].isChecked()
+        template["preserve_strokes"] = self.widgets["preserve_strokes"].isChecked()
+        for key in template.keys():
+            if key in ["normal_map_format",
+                       "project_workflow",
+                       "tangent_space_mode"]:
+                template[key] = _convert(template[key])
```
BigRoy, line 142: These keys _should_ always exist.

```suggestion
        for key in ["normal_map_format",
                   "project_workflow",
                   "tangent_space_mode"]:
            template[key] = _convert(template[key])
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -24,49 +160,34 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
     order = -10
     icon = "code-fork"
     color = "orange"
+    project_templates = []
```
BigRoy, line 163: I wonder if this will help future developers reading this code:
```suggestion

    # Defined via settings
    project_templates = []
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
moonyuet: I have added` dialog.deleteLater()` after `dialog.exec_()` but not sure if it is the best way to implement it right after the dialog execution.

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +10,133 @@
...
+            template[key] = _convert(template[key])
+        return template
+
+    @classmethod
+    def prompt(cls, templates):
+        dialog = cls(templates)
+        dialog.exec_()
+        dialog.deleteLater()
+
+        return dialog.configuration
```
BigRoy, line 139: > Just to be safe it may be better to get the attribute from the object before marking it for deletion? It should be relatively safe from the python side as it currently is - but I'm not sure what the C++ side of things of Qt might be doing there.
> What do you think?

See [comment](https://github.com/ynput/ayon-core/pull/400#discussion_r1572277502).

```suggestion
        configuration = dialog.configuration
        dialog.deleteLater()
        return configuration
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -8,10 +10,133 @@
...
+        template = get_template_by_name(template_name, self.project_templates)
+        self.widgets["import_cameras"].setChecked(template["import_cameras"])
+        self.widgets["preserve_strokes"].setChecked(
+            template["preserve_strokes"])
+
+    def get_project_configuration(self):
+        templates = self.project_templates
+        template_name = self.widgets["template_options"].currentText()
+        template = get_template_by_name(template_name, templates)
+        template = copy.deepcopy(template) # do not edit the original
```
BigRoy, line 122: Cosmetics / PEP
```suggestion
        template = copy.deepcopy(template)  # do not edit the original
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -91,8 +204,9 @@ def on_mesh_reload(status: substance_painter.project.ReloadMeshStatus):  # noqa
         # We want store some options for updating to keep consistent behavior
         # from the user's original choice. We don't store 'preserve_strokes'
         # as we always preserve strokes on updates.
+        # TODO: update the code
```
BigRoy, line 207: Is this todo still relevant? If so, it's unclear what needs to be updated.
```suggestion
```

client/ayon_core/hosts/substancepainter/plugins/load/load_mesh.py
```diff
@@ -25,48 +150,36 @@ class SubstanceLoadProjectMesh(load.LoaderPlugin):
...
+            # cancelling loader action
+            return
         sp_settings = substance_painter.project.Settings(
-            import_cameras=import_cameras
+            import_cameras=result["import_cameras"],
+            normal_map_format=result["normal_map_format"],
+            project_workflow=result["project_workflow"],
+            tangent_space_mode=result["tangent_space_mode"],
+            default_texture_resolution=result["default_texture_resolution"]
         )
```
BigRoy, line 170: Can we still move this into the `if` branch since it's  only used there?

BigRoy (approved):
Just had the one minor comment - other than that I think it looks fine.

Didn't do a test-run - don't have painter active currently.

LiborBatek (approved):
The new UI works well for me, no issues found. Good to go I guess...

Screenshot 2024-04-22 110224

antirotor (approved):
Can't test it but as far as I can tell it looks good

PR #402, opened 2024-04-10
Chore: Preparation for split of addon from ayon core

iLLiCiTiT:
## Changelog Description
Prepare codebase for first step in separation of addon from ayon core codebase. Moved client code of applications addon next to server code.

## Additional info
This change is preparation for separation of addons into separated repositories. First step will be to move code from modules and hosts next to their coresponding server code under `server_addon` directory. First candidate was applications addon.

Combination of older addons and newer ayon-core,may cause that the client code is not available. I've added a mapping of milestone version of an addon to be able to detect these cases and log them out. For now it is only applications addon, but more will come.

Modified script creating packages in `server_addon` to be able to handle client code and custom `package.py`.

Not sure if logging out the error is enough? Not sure if we should show a message about it?

## Testing notes:
There is so much possibilities how to test this,
1. Run `server_addon/create_ayon_addons.py` and upload the packages to server.
2. Create bundle with the packages.
3. Run AYON launcher.
4. Everything should be working ok.
5. When applications addon in the bundle is lowered to any `0.1.x` version, AYON launcher should log out warning message.


m-u-r-p-h-y (approved):
- created addons
- uploaded to the server
- modified dev bundle to use the latest addons
- launched Ayon in dev mode

all is working as expected, apps are launched correctly (tested, Maya, Nuke, 3dsMax)

PR #217, opened 2024-03-20
Fusion: Add instance in context validator

BigRoy:
## Changelog Description

For Fusion add instance in context validator like in Maya

## Additional info

Also bug fixes that the task on the instance's was always forced to be the task of the context

## Testing notes:

1. Publish to another folder or task
2. Validation should show error
3. Also has a "Select tool" action that should select the saver

PR #405, opened 2024-04-11
Abstract Render Collector: Preserve instance id when replacing with new instance

BigRoy:
## Changelog Description

The abstract render collector converts a Render Instance and adds it to the publishing context, but for After Effects and Fusion these "replaced" existing instances (`pyblish.api.Instance`). However, by nature of removing the original instance and adding a new one this basically meant that now the new instance would not inherit the original instance's id and thus any logs generated on the original instance being replaced would not show in the publisher UI for the new instance.

This PR changes it so that it doesn't replace the original instance at all - but just updates the source instance with the data from the render instance's data.

## Additional info

This implements a different fix as was implemented in this PR https://github.com/ynput/ayon-core/pull/217 

## Testing notes:

1. [x] Submitting renders from Fusion should still work as expected
2. [x] Submitting renders from After Effects should still work as expected.


client/ayon_core/pipeline/publish/lib.py
```diff
@@ -934,3 +934,41 @@ def get_publish_instance_families(instance):
...
+
+    # Replace the instance at the same index in the context
+    for i, instance in enumerate(context):
+        if source_instance is instance:
+            context[i] = destination_instance
+            return True
+
+    # Source was not found
+    context.append(destination_instance)
+    return False
```
iLLiCiTiT, line 974:
```suggestion
    return False

```

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -934,3 +934,41 @@ def get_publish_instance_families(instance):
...
+    the publisher to correctly detect and transfer the logs relevant for the
+    instance.
+
+    Returns:
+        bool: Whether the source instance was replaced or the destination
+            instance was only added because source instance did not exist
+            in the context.
+
+    """
+
```
iLLiCiTiT, line 957:
```suggestion
```

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -934,3 +934,41 @@ def get_publish_instance_families(instance):
...
+    # Transfer the `instance_id` of the new publisher's instances
+    key = "instance_id"
+    if key in source_instance.data:
+        destination_instance.data[key] = source_instance.data[key]
+
+    # Replace the instance at the same index in the context
+    for i, instance in enumerate(context):
+        if source_instance is instance:
+            context[i] = destination_instance
```
iLLiCiTiT, line 969:
```suggestion
    for idx, instance in enumerate(context):
        if source_instance is instance:
            context[idx] = destination_instance
```

client/ayon_core/pipeline/publish/lib.py
iLLiCiTiT: Not sure if there is new-line at the end or not (github may not show it correctly).

BigRoy: There should be according to my IDE.

client/ayon_core/pipeline/publish/abstract_collect_render.py
```diff
@@ -215,7 +219,16 @@ def process(self, context):
             render_instance_dict = attr.asdict(render_instance)
 
             instance = context.create_instance(render_instance.name)
-            instance.data["label"] = render_instance.label
+            if render_instance.source_instance is not None:
+                # remove the new instance, because we want to insert it
```
iLLiCiTiT, line 223: Maybe `replace_instance_in_context` could care about this too?

```python
if src_instance in context:
    context.remove(src_instance)

if dst_instance not in context:
    context.append(dst_instance)
```

BigRoy: > Maybe `replace_instance_in_context` should care about this too?

Doesn't _really_ sound like **replacing** which is an issue the function already suffers from, because it also inserts it if source didn't exist. Maybe we're better of actually _ONLY_ replacing it and erroring otherwise so that it's clear when it fails to replace to begin with?

> NOTE: I don't think the function has to return bool.

I just did so that if you wanted to know whether it actually got replaced (the original removed and the new inserted at that same `instance` index) or was appended to the context. But since it's currently not used anywhere I can remove that. 👍

BigRoy: Removed the `bool` return and made the function only replace and error otherwise.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -933,4 +933,4 @@ def get_publish_instance_families(instance):
         output.append(family)
         families.discard(family)
     output.extend(families)
-    return output
+    return output
```
iLLiCiTiT, line 936:
```suggestion
    return output

```

client/ayon_core/pipeline/publish/abstract_collect_render.py
BigRoy: Used a different solution - completely removed the "replacing" of instances.

iLLiCiTiT (approved):
I like it. @kalisp please re-validate if we won't break stuff.

client/ayon_core/pipeline/publish/abstract_collect_render.py
```diff
@@ -214,8 +217,11 @@ def process(self, context):
             data = self.add_additional_data(data)
             render_instance_dict = attr.asdict(render_instance)
 
-            instance = context.create_instance(render_instance.name)
-            instance.data["label"] = render_instance.label
+            if render_instance.source_instance is not None:
+                instance = render_instance.source_instance
```
iLLiCiTiT, line 221: Hmm, maybe we should remove the source instance from the data.
```suggestion
                instance = render_instance.source_instance
                render_instance_dict.pop("source_instance")
```

iLLiCiTiT: Or maybe just this
```python
instance = remder_instance_dict.pop("source_instance")
if instance is None:
    instance = context.create_instance(render_instance.name)
```

BigRoy: Agreed. Implemented with https://github.com/ynput/ayon-core/pull/405/commits/c538ed58efd1c35849455822c089dc53b2ca8731

client/ayon_core/pipeline/publish/abstract_collect_render.py
```diff
@@ -214,8 +217,11 @@ def process(self, context):
             data = self.add_additional_data(data)
             render_instance_dict = attr.asdict(render_instance)
 
-            instance = context.create_instance(render_instance.name)
-            instance.data["label"] = render_instance.label
```
BigRoy, line 218: Just so it's clear - I removed this particular line because `label` would also be in `render_instance_dict` and already gets merged into the instance later anyway.

kalisp:
I am missing `app_version` in Fusion Deadline submission, investigating when it disappeared
image

kalisp (approved):
Issue in Fusion Deadline submission fixed by https://github.com/ynput/ayon-core/pull/405/commits/cc42c4d5a89978a3ca3be622dc28bbab0e85902f

PR #407, opened 2024-04-11
Substance Painter: add missing 8k option into texture export size setting in the texture creator

moonyuet:
## Changelog Description
This PR is to add missing 8k option into texture size setting in the texture creator

## Additional info
n/a

## Testing notes:
1. Launch Substance Painter
2. The `Size` Setting of Create TextureSet now supports the 8k texture exports.


LiborBatek (approved):
I did test to output in 8K size of textureSet and all good!

Works like charm!
Screenshot 2024-04-11 153226

image

PR #409, opened 2024-04-11
Maya: Improve validators for a mesh without geometry (no faces)

BigRoy:
## Changelog Description

Improve validations for meshes without any faces/geometry.

Easiest reproducable is using this python script to create a mesh: 
```python
from maya import cmds
cmds.createNode("mesh")
```

Or to simplify it for publishing:
```python
from maya import cmds
transform = cmds.createNode("transform", name="empty_GEO")
cmds.createNode("mesh", parent=transform, name="empty_GEOShape")
```

Now publish that mesh as a model with the Validate Mesh NGONs an Validate Mesh UV has map1 UV set validators enabled. Without this PR those validators will error - now they will ignore, or invalidate with PublishValidationError improving the report.

I've also taken the liberty to improve the validation message for the Validate Mesh has map1 UV Set validator

## Additional info

We hit an issue where somehow in production where a `mesh` was broken and had no remaining geometry (likely due to an issue with maya's construction history somewhere). How to reproduce that actual real production scenario, I have no idea. But the code provided above creates a mesh with a 'similar' state.

Note that a mesh without faces/geometry is still invalid - but it's validated by a dedicated validator generating a better report.

## Testing notes:

1. Enable validators:
```
ayon+settings://maya/publish/ValidateMeshUVSetMap1
ayon+settings://maya/publish/ValidateMeshNgons
```

2. Validate the 'broken mesh' from the reproducable
3. A nice report should be visible instead of the "this is not your fault" error.


tokejepsen (changes requested):
In latest develop with those validator enabled I get this when publishing the empty mesh:

Capture

So I dont think the testing notes work.

BigRoy:
> In latest develop with those validator enabled I get this when publishing the empty mesh:
> 
> Capture
> 
> So I dont think the testing notes work.

Are you sure the validators are enabled? Can you see them having run in `Details` tab?

This is what I get (latest develop + Maya 2024):
image
image

Then if I disable NGONs validation, but only enable the other Validate Mesh Has map1 uv set:
image

LiborBatek (approved):
LGTM
Screenshot 2024-04-12 094019

After hitting ```Repair``` the geometry is converted to empty transform which gives me different ```Validate``` error, which seems also correct to me

Screenshot 2024-04-12 094340

moonyuet (approved):
code looks good to me.

tokejepsen (approved):

PR #418, opened 2024-04-15
Houdini: Sync publisher attributes with node parameters

MustafaJafar:
## Changelog Description
Use Houdini node parameters as the source of truth while using metadata as temp values.
This PR adds two methods to creator class which sync the attributes in the publisher to Houdini nodes.
1. `update_node_parameters` update node parameters according the metadata (user preference in the publish tab.) _It's triggered when user click save button in publisher_
2. `read_node_data` update metadata according to the node parameters. _It's triggered when user click refresh button in publisher_

Things left: check these product types:
|Product Type|`update_node_parameters` | `read_node_data` | 
|----------------|---------------|---------------|
|Camera (Abc)| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Arnold ASS| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Arnold ROP| <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> |
|PointCache (Bgeo)| <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> |
|Composite (Image Sequence)| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Houdini Digital Asset (Hda)| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Karma ROP| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Mantra IFD| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Mantra ROP| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|PointCache (Abc)| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Redshift Proxy| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Redshift ROP| <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> |
|Review| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|Static Mesh (FBX) | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|VDB Cache| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|VRay ROP| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |
|USD (experimental)| <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> |

## Demo
https://github.com/ynput/ayon-core/assets/20871534/fbeaff1f-6c1c-494a-8bc2-b0b8af57db01

## Testing notes:
1. Open Houdini
2. Create instances 
3. In publish tab, change some options and hit save button (you should find out that node parameters are synced with the selected options in the publisher)
4. Change the dedicated nodes parameters and hit refresh (you should find out the publisher is synced to the node)


client/ayon_core/hosts/houdini/api/plugin.py
```diff
@@ -253,7 +255,15 @@ def update_instances(self, update_list):
...
                 for key in changes.changed_keys
             }
-            # Update parm templates and values
+
+            # Update Houdini node's parameters based on creator attributes.
+            # This is done by re-using the logic inside the creators.
+            host = registered_host()
+            create_context = CreateContext(host, reset=True)
+            create_plugin = create_context.creators[created_inst.creator_identifier]
+            create_plugin.update_node_parameters(instance_node, new_values["creator_attributes"])
```
BigRoy, line 264: `update_instances` should **NOT** create a new `CreateContext`

You are trying to get the Create plug-in. But that is the **current** plug-in instance!
This should just be:

```suggestion
            self.update_node_parameters(instance_node, new_values["creator_attributes"])
```

Also, I wouldn't limit this to `creator_attributes` but just pass it the full `changes`. Allow any subclass to do with that what they want.

BigRoy: You seem to be lacking logic that does the reverse of updating the node parameters. The creator attributes (or whatever other data you are persisting on the node) should likely also be read from the node again? So that if the user had toggled that on the node itself that the publisher UI also reads that as the ground truth for the creator attribute, etc.

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
```diff
@@ -61,45 +60,56 @@ def create(self, product_name, instance_data, pre_create_data):
...
+            node.evalParm("sopoutput"),
+            allowed_multidot_extensions=[
+                ".ass.gz", ".bgeo.sc", ".bgeo.gz",
+                ".bgeo.lzma", ".bgeo.bz2"
+            ]
+        )
+
+        output = "{file_path}.{ext}".format(
+            file_path=file_path,
+            ext=creator_attributes["bgeo_type"]
```
BigRoy, line 84: You are calling this method with the `creator_attributs` from the `new_values` as such I suspect `bgeo_type` might NOT be available if its not changed and this will error?

BigRoy, line 84: You are calling this method with the `creator_attributs` from the `new_values` as such I suspect `bgeo_type` might NOT be available if its not changed and this will error?

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
```diff
@@ -61,45 +60,56 @@ def create(self, product_name, instance_data, pre_create_data):
...
-            {
-                "value": "bgeo.bz2",
-                "label": "BZip2 compressed bgeo (.bgeo.bz2)"
-            }
-        ]
 
-        return attrs + [
-            EnumDef("bgeo_type", bgeo_enum, label="BGEO Options"),
-        ] + self.get_instance_attr_defs()
+        return attrs + self.get_instance_attr_defs()
```
BigRoy, line 112: Should the `farm` toggle now also be exposed on pre create?

client/ayon_core/hosts/houdini/plugins/create/create_redshift_rop.py
```diff
@@ -118,18 +104,45 @@ def remove_instances(self, instances):
...
         ]
-        multi_layered_mode = [
-            "No Multi-Layered EXR File",
-            "Full Multi-Layered EXR File"
-        ]
-
+        multi_layered_mode = {
+            "1": "No Multi-Layered EXR File",
+            "2": "Full Multi-Layered EXR File"
+        }
```
BigRoy, line 143: Why is this suddenly a string of `"1"`  and `"2"`?

client/ayon_core/hosts/houdini/plugins/create/create_redshift_rop.py
```diff
@@ -141,7 +154,11 @@ def get_pre_create_attr_defs(self):
...
             EnumDef("multi_layered_mode",
-                    multi_layered_mode,
+                    items=multi_layered_mode,
                     default=self.multi_layered_mode,
                     label="Multi-Layered EXR")
         ]
+
+    def get_pre_create_attr_defs(self):
+
+        return self.get_instance_attr_defs()
```
BigRoy, line 164: This now has no `Use selection` on pre create anymore. Is that intentional?

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: `update_instances` is only called when there's a change. and `new_values` include both the un-changed values as well as changed values.

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: tbh, I'm not sure where the `farm` value should exist: 
- should it be a checkbox parameter
- or should it be a `creator_attribute`

client/ayon_core/hosts/houdini/plugins/create/create_redshift_rop.py
MustafaJafar: This is the same as in this PR https://github.com/ynput/ayon-core/pull/328
to me it felt easier to use it in this way instead of adding a dictionary to map these strings to these values. 

from debugging side: I'm not sure if it's a good idea.

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
BigRoy: Are you sure? You are [here](https://github.com/ynput/ayon-core/pull/418/files/ae7097b6bdc43bb31a000272b6f964d910b1ceaf#diff-cd393404eac3ba072c67b8d5f1872e399e07a988c6e60d51227991ace4158090R254-R257) explicitly filtering to only those that have `changed_keys` which should really only be the 'new data' - not all the data. If it does contain all, that sounds like a bug, right @iLLiCiTiT ?

I'd expect `created_inst.data_to_store()` to return all data instead. E.g. like used [here](https://github.com/pypeclub/OpenPype/blob/dc27a9e608688657b726157b4ec633b33407506a/openpype/hosts/maya/api/plugin.py#L267-L272)

client/ayon_core/hosts/houdini/plugins/create/create_redshift_rop.py
BigRoy: So the actual Houdini parm takes a string number?

client/ayon_core/hosts/houdini/plugins/create/create_redshift_rop.py
```diff
@@ -118,18 +104,45 @@ def remove_instances(self, instances):
...
+        multi_layered_mode = creator_attributes.get("multi_layered_mode")
+        multipart = False
+        if multi_layered_mode == "2":
+            multipart = True
+
+        if ext == "exr":
+            parms["RS_outputMultilayerMode"] = multi_layered_mode
+            parms["RS_aovMultipart"] = multipart
+
+        parms["RS_archive_enable"] = creator_attributes.get("split_render")
```
BigRoy, line 132:
```suggestion
        parms = {
            "RS_outputFileFormat": ext_format_index[ext],
            "RS_archive_enable": creator_attributes.get("split_render")
        }

        if ext == "exr":
            multi_layered_mode = creator_attributes.get("multi_layered_mode")
            multipart = multi_layered_mode != "2"
            parms["RS_outputMultilayerMode"] = multi_layered_mode
            parms["RS_aovMultipart"] = multipart
```

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: here's a test from this PR. 
I printed the `creator_attributes` inside `update_node_parameters`
when creating a new redshift instance I got: 
```
{'farm': True, 'split_render': True, 'image_format': 'exr', 'multi_layered_mode': '1'}
```
and when I updated it from the publisher I got
```
{'multi_layered_mode': '2', 'farm': True, 'split_render': True, 'image_format': 'exr'}
```

client/ayon_core/hosts/houdini/api/plugin.py
MustafaJafar: using the rop node as the source of truth is a crucial part in Houdini addon. 
Maybe I was confused when I used `get_instance_attr_defs` for the first time in the other PR.

So, my point here https://github.com/ynput/ayon-core/pull/328/files/a894a9a0897ecc82460af92064527088844409cd#r1558203127 was wrong and I really don't need to update the node state/parameters to align with user preference before the validation phase as collectors should collect/deduce the data?

MustafaJafar: Maybe I'm overthinking. 
But, when applying changes in the publish tab I expected _as a user_ that clicking save button will apply these changes.
Also, knowing the fact that eventually the ROP node will follow the settings in the publish tab makes me willing to use the publisher UI more than doing anything manually in the rop node as the publisher **knows** how to change the parameters properly. 

But, it could be only me.

moonyuet: I agree with @BigRoy. We shouldn't create context when we are updating the instances but instead, we need to find AYON-related parameters which already existed in the node and update them. If I am not wrong, `CreateContext` at this point can probably add unnecessary data within the existing data. This would cause issues when the publisher trying to collect instance data from the instance node(it would probably recognize multiple instances where the un-updated and updated instances live together in the publisher UI).

BigRoy: > Save button in publisher should only update AYON-related parameters.

This isn't entirely a hard rule. There may be reasons to want a native Houdini parameter to be used as the value for a setting exposed in the publisher UI - but I'd also then try to "read" the state of that publisher value from that parameter instead of storing other 'metadata' on the node that gets out of sync.

> Publish instances (Houdini ROPs) should be the source of truth.

As much as possible, yes.

> Collector plugins should collect/deduce/predict the data according to the ROP node state and user settings in the publisher

If we can provide a workflow to the user that is non-destructive and works for our pipeline needs then I think that can be very nice. With Houdini being quite technical software where users can do very advanced stuff there is a lot of merit in staying "thin" on top of Houdini and just trying to allow as many stuff through by the artists (they are in control).

It takes away a bit from being able to automate _everything_ via solely AYON logic -but it does mean things remain flexible for Houdini artists.

There will still be reasons to hijack Houdini and take control ourselves from AYON if we really need to. But if we can avoid it, then I think it's good.

> Extract plugins can be used to alter the state of the node (e.g. Align the node parameters to the instance options in the publisher)

That's totally fine, yes - preferred actually. **BUT** it MUST return it to its original state afterwards. The extractor is allowed to do anything - but at the end of it leave the scene unchanged.

The same is true basically for any pyblish plug-in. If a validator requires putting down an extra node in the node tree, run the node, get the result - just to do the validation, then by all means implement it. As long as at the end of the `process()` method the scene is back exactly how it was before it. Not saying it's good practice to do this heavy stuff, but it's totally fine to do whatever you want within a Pyblish plug-in like that.

> Therefore this PR should be closed.
> What do you think ?

Going from my points above, I can't say a hard yes. I think it's actually **no** - there's merit to this PR. But it totally depends on what it is trying to solve.

If the extractor works fine by changing things 'temporarily' and extracting it - then that's much better than updating the user's scene from the creator I think. However, that logic may not work if e.g. the scene is to be submitted to Deadline and the Deadline job will do whatever the ROP node is configured to. In that case we can't really 'temporarily hijack' the scene state just for the render. As such, for those cases the scene _must_ be in the correct state for the Deadline Job to do what we want it to do. In that case, having it update from the Creator and (also read the state from the node's parameters) could make a much better user experience instead of relying on Validators with Repair functionality.

moonyuet: >If the extractor works fine by changing things 'temporarily' and extracting it - then that's much better than updating the user's scene from the creator I think.

maybe the `imprint` function in `lib` would help for this.
https://github.com/ynput/ayon-core/blob/9299e7e7033da85b26e464494efa15ea8e53a888/client/ayon_core/hosts/houdini/api/lib.py#L259

MustafaJafar:
> Is there any reason why to touch the `update_instances` function to update Houdini node parameters on creator attributes change? Are you trying to align the setting in create tab to the one in publish tab in the publisher UI as the goal of this PR?

No, not really.
I made a quick demo to show what this PR does. 

https://github.com/ynput/ayon-core/assets/20871534/7197b5d0-06c7-4429-9e82-90b8b598b784


I find it confusing.. 
Maybe, we should be more careful about using `get_instance_attr_defs`

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: I believe in the near future, farm checkbox will be removed in favor of putting it inside creator attribute.
As, for consistency we may want to add render targets for caches too like https://github.com/ynput/ayon-core/pull/328 
So, let's ignore farm toggle atm.
Feel free to mark this back as unresolved.

client/ayon_core/hosts/houdini/api/plugin.py
MustafaJafar: I've applied your suggested code.. also, I added a note about it (tbh, I didn't the type of self and it's indeed not needed to create another object) Thanks! 

https://github.com/ynput/ayon-core/pull/418/commits/eda3b88e7c6eed9321196c93ce729a732fa7e3a9

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
BigRoy: > or should it be a creator_attribute

I'd expect it's a toggle on instance attribute defs (and potentially _already_ available to preset on the creator attributes) so that at any point in time one can switch between farm submission and local render - similar to how Fusion an Nuke are doing it with "render local" versus "farm" enum options.

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: > Are you sure? You are [here](https://github.com/ynput/ayon-core/pull/418/files/ae7097b6bdc43bb31a000272b6f964d910b1ceaf#diff-cd393404eac3ba072c67b8d5f1872e399e07a988c6e60d51227991ace4158090R254-R257) explicitly filtering to only those that have changed_keys which should really only be the 'new data' - not all the data.

I checked it again and  this true! 
it only updates the `changed_keys` 
but, it works here well because the changed key is `creator_attributes`

```
changes.changed_keys:  {'creator_attributes'}
new_values:  {'creator_attributes': {'farm': True, 'split_render': True, 'image_format': 'exr', 'multi_layered_mode': '2'}}
```
So, the code will error if I changed any of these attributes for example.

image

BigRoy: > but, it works here well because the changed key is `creator_attributes`

Yup, that's what I expected. I wanted to comment that before, but apparently I forgot. Glad you caught it.

MustafaJafar: > You are calling this method with the creator_attributs from the new_values as such I suspect bgeo_type might NOT be available if its not changed and this will error?

since one key inside `creator_attributes` is changed then `creator_attributes` parameter is considered changed. and it's value is passed even the change is in one of its keys only. 

but, your concern was right as when changing `publish_attributes`, then  `creator_attributes`  won't be passed which leads to the error in my previous comment. 

That case should be solved in https://github.com/ynput/ayon-core/pull/418/commits/ad16bdf43168b8f8727f1bb15ab0d3ae90562602

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: I think there are two problems: 
1. Which parameter should be the source of truth ?
2. How to update related to parameters to follow the source of truth ?

Let's take redshift creator plugin as an example: 
a. In `pre_create_defs`, we add a `BoolDef("farm",...)` to add a toggle to the creator plugin. 
b. In `process`, we add the value of that def in a farm "toggle"
c. At this point the value of the `BoolDef("farm",...)` is saved to two parameters (a `toggle` parameter and `creator_attribtes` parameter)
d. In publisher, I can switch off `Submitting to Farm` which will affect the `"farm"` key in `creator_attributes`. which basically does nothing, and it doesn't affect the farm `toggle`. 
e. I can only switch off the `farm` publishing by clicking the farm `toggle` manually.

So, all plugins expects the farm toggle to be the source of truth (which automatically collected by imprint) and the same with `splitRender` feature which is based on the parameter `RS_archive_enable`  (which is not exposed in the publisher)
at this point the two issues above are not exposed yet.

But, these issues started to show up in https://github.com/ynput/ayon-core/pull/328 when I started to use `get_instance_attr_defs` which allows me to change `creator_attributes` in publish tab.
And, The `render_target` and render mode were exposed in the publisher. 

So, what parameters should be the source of truth. 
- `creator_attributes` parameter which acts as a variable that holds the last selection
- or the actual Houdini parameters that get updated eventually based on the selection in the publisher. 

Which is as you said earlier 
```
but I'd also then try to "read" the state of that publisher value from that parameter instead of storing
 other 'metadata' on the node that gets out of sync.
```

Then let's think of the second question:  How to update related to parameters to follow the source of truth ?
if the Houdini parameters are the source of truth then we should have some inverted logic to deduce the correct `creator_attributes` so that they show up in the publisher correctly which _I assume_ should work on `refresh` action. 

and when opening the publisher we use _maybe this PR_ to apply the changes in the publisher to the instance node. 

I started to feel that I'm overthinking but tbh I consider this a bad issue.
as it touches many problems in other PRs. 
e.g. this one https://github.com/ynput/ayon-core/pull/328#discussion_r1558137542 
because I don't feel it should exist in a collector or extractor. 
The collector should only collect the data. 
The extractor can do any changes but it should revert the changes
However, we want just to update the instance according to our settings and me as a user expected the changes to be applied once I hit save similar to the changes happen in the creator.

MustafaJafar: maybe this commit adding the missing piece of the puzzle https://github.com/ynput/ayon-core/pull/418/commits/101a86d386c7b3812edcd2477aca1fc3dbe646b0

MustafaJafar:
After some overthinking here's another demo where Houdini parameters are used as the single source of truth. 
metadata are used to connect the Houdini parameters to the publisher. 
-  `update_node_parameters` is used to update the node's parameters according to the value of the `creator_attributes` (should it be more generic IDK but in the creator plugin we only use `creator_attributes`)
- `read_node_data` which has the inverted logic of the previous method. the data return from this method will override ayon parameters. 

https://github.com/ynput/ayon-core/assets/20871534/fbeaff1f-6c1c-494a-8bc2-b0b8af57db01

moonyuet:
> After some overthinking here's another demo where Houdini parameters are used as the single source of truth. metadata are used to connect the Houdini parameters to the publisher.
> 
> * `update_node_parameters` is used to update the node's parameters according to the value of the `creator_attributes` (should it be more generic IDK but in the creator plugin we only use `creator_attributes`)
> * `read_node_data` which has the inverted logic of the previous method. the data return from this method will override ayon parameters.
> 
>  2024-04-18.22-46-38.mp4

I can see why now. If you talk about updating the `creator_attributes` parameters on the instance nodes generated by the creator, the latest commit on the adjustment on `update_instances` function does make sense as the goal of Houdini development (which allows more customizations by the user)

client/ayon_core/hosts/houdini/api/plugin.py
```diff
@@ -347,3 +362,45 @@ def apply_settings(self, project_settings):
 
         for key, value in settings.items():
             setattr(self, key, value)
+
+    @staticmethod
```
iLLiCiTiT, line 366: Why are they static methods?

MustafaJafar: because we don't need to pass `self`.. these methods process the given node with given data or return data.

client/ayon_core/hosts/houdini/plugins/create/create_bgeo.py
MustafaJafar: > Should the farm toggle now also be exposed on pre create?

yes, implemented in https://github.com/ynput/ayon-core/pull/418/commits/b865d49f23d44dc21abb8d9202b966955e5813e6

client/ayon_core/hosts/houdini/api/plugin.py
iLLiCiTiT: > because we don't need to pass self

You don't need it "now".

MustafaJafar: > You don't need it "now".

Ouch!

MustafaJafar:
I found some case that doesn't work with this PR. 
e.g. 
when I change an attribute in the publisher and click save (it updates the Houdini node correctly)
Then I revert the changes in the Houdini while keeping the publisher as it's
Then I click again the publisher to apply the changes inside it to the node (but, there are no changes so, it won't update the node)


https://github.com/ynput/ayon-core/assets/20871534/ec247d94-aca4-4250-8274-1701fff1816a

BigRoy:
> I found some case that doesn't work with this PR. e.g. when I change an attribute in the publisher and click save (it updates the Houdini node correctly) Then I revert the changes in the Houdini while keeping the publisher as it's Then I click again the publisher to apply the changes inside it to the node (but, there are no changes so, it won't update the node)
> 
> Because there are no changes, the save button won't even work so in this particular case, the method that's used to update the Houdini nodes won't be even called. so the node and publisher attributes will be out of sync.
> 2024-05-10.21-53-37.mp4

At this stage, I'd say it's expected. If you after "reset" or refresh or save start tweaking stuff inside the Houdini scene without resetting/refreshing again then that'd be user error and should not be an "allowed workflow".

However, changing something in Houdini and then clicking "reset" should actually detect the new values preferably.

MustafaJafar:
> If you after "reset" or refresh or save start tweaking stuff inside the Houdini scene without resetting/refreshing again then that'd be user error and should not be an "allowed workflow".

But, in this case, Should the validators tell if publish attributes and Houdini node parameters are not in sync ? 
The validators won't need to provide/offer any repair actions because the publisher attributes and node parameters will be in sync again once the publisher tool resets.

For reference, in the above screen record, the redshift node will be published with `split export and farm` disabled even though it's enabled in the publisher attributes. (as the node is the source of truth)
or we can just ignore that as the node is the only source of truth in this PR! 


> However, changing something in Houdini and then clicking "reset" should actually detect the new values preferably.

This is the current state of this PR.

antirotor:
Well, you're right that in ideal world, you would have a set of validators checking that those options are in sync - but imho it is lot of work to catch all reliably. Maybe it is enough to check state of the scene, compare checksums of the state before and after - but even that cannot be done 100% reliable. To make it work, you would need to track every individual option in houdini nodes affecting the current publishing process. If this can be done easily, I am completely for it - maybe there are some events we can bind to - something like "on scene change display visual warning that publisher UI can be outdated - but I think this is adding more complexity then there are gains from it. For me it is enough to clearly document how it behaves.

MustafaJafar:
Closing this PR as it adds more complexity then there are gains from it.

PR #423, opened 2024-04-17
Houdini: Deadline job missing Frames attribute

MustafaJafar:
## Changelog Description
Exporting alembics (and maybe other types) from Houdini on Deadline won't take into account frame range as we are not passing Frames attribute.

This PR actually does two things: 
1. Fix the plugin order. 
2. Adds support to publish multi-frames `pointcache (Abc)`.  _Artists can achieve that by adding `$F4` manually in the filepath._

## Testing notes:
1. Create a cache related product type (alembic - Bgeo - vbd - ass)
2. submit to farm (all exported frames should be included in the publish)
    i. in the publish json file, you'll find `"files"` key is a list of the correct files.
    ii. check the publish if it contains all the expected frames


client/ayon_core/hosts/houdini/plugins/publish/collect_frames.py
```diff
@@ -61,6 +61,7 @@ def process(self, instance):
         # todo: `frames` currently conflicts with "explicit frames" for a
         #       for a custom frame list. So this should be refactored.
         instance.data.update({"frames": result})
+        self.log.debug(instance.data["frames"])
```
BigRoy, line 64:
```suggestion
```

client/ayon_core/hosts/houdini/plugins/publish/collect_cache_farm.py
```diff
@@ -7,7 +7,7 @@
 class CollectDataforCache(pyblish.api.InstancePlugin):
     """Collect data for caching to Deadline."""
 
-    order = pyblish.api.CollectorOrder + 0.04
+    order = pyblish.api.CollectorOrder + 0.11
```
BigRoy, line 10: Would it help if we add a comment?
```suggestion
    # Run after Collect Frames
    order = pyblish.api.CollectorOrder + 0.11
```

antirotor (approved):
Just a note about PR description - `$F` isn't related to these changes.

BigRoy:
> Just a note about PR description - `$F` isn't related to these changes.

Aside of that I feel that the PR description doesn't entirely explain what the issue is and how it solves it. I had to read the code (and the plug-ins logic themselves) to get an idea of what's actually changing and why.

MustafaJafar:
> > Just a note about PR description - `$F` isn't related to these changes.
> 
> Aside of that I feel that the PR description doesn't entirely explain what the issue is and how it solves it. I had to read the code (and the plug-ins logic themselves) to get an idea of what's actually changing and why.

Let add what I know about this PR and I'll appreciate any suggestions to update the PR description and may be the PR name.

So, this PR was created to fix an issue with `pointcache (abc)`
Issue: 
By default alembic rop exports one file therefore alembic collectors and extractors all expected one file.
alembic rop also supports using `$F$` token which result in exporting alembic as many frames which is not considered in the collectors. 
Note: This fails on local and farm publish.

Solution: 
fixing the collector and extractor to support the case when exporting alembic as many frames. 
This now works on local and farm publishing.

---

I looked at other product types to check which have the same issue but I couldn't find any thing.

MustafaJafar:
@antirotor For reference, I found a bug while testing deadline.. Also, I found out that this PR solves it.

moonyuet (approved):

PR #431, opened 2024-04-18
Maya: Bake Complex Animation export in FBX format

moonyuet:
## Changelog Description
This PR is to fix the defined type of `Bake Complex Animation` being wrong in `fbx.py` which confuses the system not exporting the animation without baking. Besides, the exported rig/animation is broken if we add constraint as export options and this PR is to fix the bug of animation being exported correctly.
Also, this PR would add joint as validating options to ensure the joints are also being validated and not errored out the validator `ValidateAnimatedReferences`. 

## Additional info
See https://github.com/ynput/OpenPype/pull/5589 which explains how the sets are purposed to do

## Testing notes:
Rig Family:
1. Launch Maya via launcher
2. Create Rig Instance
3. Attach your rigs and geometries to out_set & controls_set as usual
4. Attach your control rigs and/or meshes into `skeletonMesh_SET`
5. Attach your control rigs into  `skeletonAnim_SET` 
6. Publish

Animation Family:
1. Doing all the steps with "Rig Family" testing notes above
2. Load your rigs 
3. Animate your rigs
4. Click "Publish...." via Menu
5. Publish along with the animation family


BigRoy (approved):
Didn't test - but change makes sense.

moonyuet:
@antirotor can you perform some test from your file to see if it fixes the issue?

client/ayon_core/hosts/maya/api/fbx.py
```diff
@@ -102,6 +102,7 @@ def default_options(self):
             "constraints": False,
             "lights": True,
             "embeddedTextures": False,
+            "includeChildren": True,
```
BigRoy, line 105: If adding this - shouldn't this also be in the return of the `options` method?

moonyuet: indeed, it should be also in the options

client/ayon_core/hosts/maya/plugins/publish/extract_fbx_animation.py
```diff
@@ -35,7 +35,7 @@ def process(self, instance):
         fbx_exporter = fbx.FBXExtractor(log=self.log)
         out_members = instance.data.get("animated_skeleton", [])
         # Export
-        instance.data["constraints"] = True
+        instance.data["constraints"] = False
```
BigRoy, line 38: Why does this have to be False now? Is it because we're force baking the keys to the animated skeleton?

_Also, the PR description does not describe that this PR changes this - so if there's reasons to change this within this PR could you add it to the PR description?_

client/ayon_core/hosts/maya/plugins/publish/validate_animated_reference.py
```diff
@@ -16,7 +16,7 @@ class ValidateAnimatedReferenceRig(pyblish.api.InstancePlugin,
     hosts = ["maya"]
     families = ["animation.fbx"]
     label = "Animated Reference Rig"
-    accepted_controllers = ["transform", "locator"]
+    accepted_controllers = ["transform", "locator", "joint"]
```
BigRoy, line 19: Is a `joint` a valid rig 'controller'? Is this correct? Does rig publishing itself even allow that?

This seems to hint that rig controls may only be transforms:
- https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/publish/collect_rig_sets.py#L20
- https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/publish/validate_rig_contents.py#L29

Why did this have to be added?

_Also, the PR description does not describe that this PR changes this - so if there's reasons to change this within this PR could you add it to the PR description?_

client/ayon_core/hosts/maya/plugins/publish/extract_fbx_animation.py
moonyuet: > Why does this have to be False now? Is it because we're force baking the keys to the animated skeleton?
> 
> _Also, the PR description does not describe that this PR changes this - so if there's reasons to change this within this PR could you add it to the PR description?_

it's mainly because when the complex animation is baked, it broke the rig if the constraints option is enabled. But for this case we need to test on multiple scenarios(with both simple and complex rigs)

client/ayon_core/hosts/maya/plugins/publish/validate_animated_reference.py
moonyuet: > Is a `joint` a valid rig 'controller'? Is this correct? Does rig publishing itself even allow that?
> 
> This seems to hint that rig controls may only be transforms:
> 
> * https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/publish/collect_rig_sets.py#L20
> * https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/publish/validate_rig_contents.py#L29
> 
> Why did this have to be added?
> 
> _Also, the PR description does not describe that this PR changes this - so if there's reasons to change this within this PR could you add it to the PR description?_

I have added the description into the changelog. I know it sounds weird, but for some reason the validator errors out the joint when I added the joints into the set which is collected by the collector you mentioned.

BigRoy: "For some reason" sounds like you're fixing something that you're unsure of whether it's intended behavior. To me the bigger question here is - why are joints part of the `controls` to begin with?

This seems like the wrong fix because the attribute naming just seems off - it's not `controls` it's validating? It seems to validate the `animated_skeleton` looking at the set members it's actually validating? If so, then the attribute called `validate_controls` should be `validate_skeleton` and the attribute `accepted_controllers` should maybe be `accepted_skeleton` (and potentially disallow transforms/locators if we do not want those as part of the skeleton?)

client/ayon_core/hosts/maya/plugins/publish/extract_fbx_animation.py
BigRoy: I think we're better of getting to the bottom of why it breaks - so that if needed we can also guide users about that.

In this case, it's unclear _what_ breaks on export and why. 

I think it should now be a matter of finding out why it breaks and THEN decide whether the code with constraints or not? By not including constraints we're also saying we're completely disallowing constraints in FBX animation outputs. Whether that's 'valid' for how users are using the FBX export I'm not sure - I'll leave that up to the community (those users) to answer that. But toggling it 'just because it broke' where it "breaking" is somewhat unclear as to why I would say would mean we're just avoiding to find the right solution to a problem but just 'doing whatever happened to work for that case.'

BigRoy: Also want to add - since this issue + solution seems counterintuitive if this is indeed the resulting _correct_ solution to the problem I'd add a comment to the code as to WHY we're force disabling it, why it's needed and hopefully in a way that also makes it believable that it's the correct fix to the issue.

moonyuet: I guess for the best approach is to remove this line, as the default option is set to `False`. 
There is the scenario where you export the rigs with some constraints set within the maya(e.g. aim constraint) once it gets exported it will break the fbx export.

client/ayon_core/hosts/maya/plugins/publish/validate_animated_reference.py
moonyuet: > "For some reason" sounds like you're fixing something that you're unsure of whether it's intended behavior. To me the bigger question here is - why are joints part of the `controls` to begin with?
> 
> This seems like the wrong fix because the attribute naming just seems off - it's not `controls` it's validating? It seems to validate the `animated_skeleton` looking at the set members it's actually validating? If so, then the attribute called `validate_controls` should be `validate_skeleton` and the attribute `accepted_controllers` should maybe be `accepted_skeleton` (and potentially disallow transforms/locators if we do not want those as part of the skeleton?)

The code you showed is for rig family but not for animation family.(My bad I didn't notice you are showing rig set collector) If we double check on the collector `CollectFbxAnimation`, we are actually collecting everything. Maybe it's also good idea to fix the bug by implementing the ignore type as if we did in collect_animation if the constraint is being the reason to break the animation?

BigRoy: What I'm saying that the attribute names are just invalid on this validator - it refers to "controllers" quite a bit but apparently validates something named `animated_skeleton` which sounds more like it is validating something that is expected to the be the skeleton only?

Then the next step is - if this is validating the skeleton. Then WHAT node types do we allow? I suppose we'd allow bones, of course - maybe parent groups? But do we actually allow locators? Etc.

moonyuet: > What I'm saying that the attribute names are just invalid on this validator - it refers to "controllers" quite a bit but apparently validates something named `animated_skeleton` which sounds more like it is validating something that is expected to the be the skeleton only?
> 
> Then the next step is - if this is validating the skeleton. Then WHAT node types do we allow? I suppose we'd allow bones, of course - maybe parent groups? But do we actually allow locators? Etc.

we just keep it as it is without changing the original code of the validator, all we just need to change the collector to make sure it doesn't collect any maya-related constraint. About the naming of the `animated_skeleton`, the reason of naming `animated_skeleton` contributes to what the related sets have included, we can include set of control rigs inside the animated_skeleton set so that's why it is named like this. If it is not concise name, we can change it with different PRs or continue the discussion later

moonyuet: Update: I guess I would just leave both validator and collector unchanged as the changes can affect the pipeline between fbx animation from maya to unreal.

client/ayon_core/hosts/maya/plugins/publish/extract_fbx_animation.py
moonyuet: Created issue here: https://github.com/ynput/ayon-core/issues/509 we can continue the discussion.

client/ayon_core/hosts/maya/plugins/publish/extract_fbx_animation.py
```diff
@@ -35,7 +35,8 @@ def process(self, instance):
         fbx_exporter = fbx.FBXExtractor(log=self.log)
         out_members = instance.data.get("animated_skeleton", [])
         # Export
-        instance.data["constraints"] = True
```
BigRoy, line 38: @moonyuet was the removal of constraints being `True` still intentional? I thought you mentioned you were to revert that?

moonyuet: yes, it is.

PR #433, opened 2024-04-19
Collect AOV names to anatomy data

antirotor:
## Changelog Description
Simple change to collect AOV names if present on Deadline submission to be used in templates as `aov_name`.

## Additional info
This works only with Deadline submissions, for local renders it needs to be collected elsewhere.

## Testing notes:
1. Submit job with separated AOVs to farm
2. Define anatomy template to use `{aov_name}`

### DIFF
--- a/client/ayon_core/plugins/publish/collect_aovs_to_anatomy_instance_data.py
+++ b/client/ayon_core/plugins/publish/collect_aovs_to_anatomy_instance_data.py
@@ -1,0 +1,22 @@
+import pyblish.api
+
+
+class CollectAOVNameToAnatomyInstanceData(pyblish.api.InstancePlugin):
+    """
+    Collect AOV name to template data.
+
+    Note:
+        This functionality depends on AOV name added by
+        ``submit_publish_job`` plugin and so it only works
+        with Deadline submission.
+
+    """
+    order = pyblish.api.CollectorOrder + 0.499
+    targets = ["farm"]
+    label = "Collect AOV name to template data"
+
+    def process(self, instance):
+        if "aovName" not in instance.data:
+            return
+
+        instance.data["anatomyData"]["aov_name"] = instance.data["aovName"]
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -621,9 +621,11 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
         new_instance = deepcopy(skeleton)
         new_instance["productName"] = product_name
         new_instance["productGroup"] = group_name
+        if aov:
+            new_instance["aovName"] = aov
 
         # toggle preview on if multipart is on
-        # Because we cant query the multipartExr data member of each AOV we'll
+        # Because we can't query the multipartExr data member of each AOV we'll
         # need to have hardcoded rule of excluding any renders with
         # "cryptomatte" in the file name from being a multipart EXR. This issue
         # happens with Redshift that forces Cryptomatte renders to be separate


client/ayon_core/plugins/publish/collect_aovs_to_anatomy_instance_data.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class CollectAOVNameToAnatomyInstanceData(pyblish.api.InstancePlugin):
+    """
+    Collect AOV name to template data.
+
+    Note:
+        This functionality depends on AOV name added by
+        ``submit_publish_job`` plugin and so it only works
+        with Deadline submission.
```
BigRoy, line 11: If so, shouldn't this be inside the `Deadline` addon to begin with?

client/ayon_core/plugins/publish/collect_aovs_to_anatomy_instance_data.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    order = pyblish.api.CollectorOrder + 0.499
+    targets = ["farm"]
+    label = "Collect AOV name to template data"
+
+    def process(self, instance):
+        if "aovName" not in instance.data:
+            return
+
+        instance.data["anatomyData"]["aov_name"] = instance.data["aovName"]
```
BigRoy, line 22: I'm a bit worried about this directly injecting into `anatomyData` - shouldn't that be way more 'static' or 'constant'`? - Why do we need it in anatomy data?

Looking at the overal logic in `pyblish_functions.py` each AOV is already gets separated into its own instance, and thus included in the 'variant' for the instance and thus the subset name. Now also adding a separate `{aov_name}` in there just seems like duplicating that?

The only reason I can think of requiring this `{aov_name}` then is to publish multiple AOVs within a single instance? But we don't seem to ever be doing that.

BigRoy:
> 2\. Define anatomy template to use `{aov_name}`

Having that in your template will fail for any publish that does not have this in `anatomyData` which may very well be almost all publishes with only the exception of non-multilayer renderlayer submissions that split into separate instances? Because even for multilayer instances the `aov_name` should be not there? right?

client/ayon_core/plugins/publish/collect_aovs_to_anatomy_instance_data.py
antirotor: It is stated there, but in essence this can also be utilized inside RoyalRender if needed.

antirotor:
> > 2. Define anatomy template to use `{aov_name}`
> 
> Having that in your template will fail for any publish that does not have this in `anatomyData` which may very well be almost all publishes with only the exception of non-multilayer renderlayer submissions that split into separate instances? Because even for multilayer instances the `aov_name` should be not there? right? Since it's not an AOV?

Well, the intended use case - and I fully admit it is rather quick fix then more solid solution - is to have anatomy template for renders like this:

**Dir**: `{root[work]}/{project[name]}/{hierarchy}/{asset}/publish/{family}/{subset}/{@version}/<{aov_name}/>`
**File**: `{project[code]}_{asset}_{subset}_{@version}<_{output}><.{@frame}>.{ext}`

if your workflow requires aov somewhere. Making it optional won't break cases where aov name is not present. Since this needs to work on farm, where you cannot (without some major rewrites) define host based template because publishing runs outside of host.

So as you've stated, if you set it like this in anatomy template, it will be used only for non-multilayer render submission which is exactly client request and since it is not affecting anything else, I think it is pretty solid.

BigRoy:
> Well, the intended use case - and I fully admit it is rather quick fix then more solid solution - is to have anatomy template for renders like this:
> 
> **Dir**: `{root[work]}/{project[name]}/{hierarchy}/{asset}/publish/{family}/{subset}/{@version}/<{aov_name}/>` **File**: `{project[code]}_{asset}_{subset}_{@version}<_{output}><.{@frame}>.{ext}`
> 
> if your workflow requires aov somewhere. Making it optional won't break cases where aov name is not present. Since this needs to work on farm, where you cannot (without some major rewrites) define host based template because publishing runs outside of host.
> 
> So as you've stated, if you set it like this in anatomy template, it will be used only for non-multilayer render submission which is exactly what client request and since it is not affecting anything else, I think it is pretty solid.

But this doesn't make any sense because `subset` would already include the `aov` since it's a different `variant`? Unless for whatever reason your product name (=subset) does not include the variant at all?

antirotor:
> So yes, you're adding a subfolder named aov_name but they would still not be in the same parent folder, yes?

> Also, nothing is actually ensuring that the aovs go into the same version (unless maybe follow workfile version is enabled) then there is a high chance that it will. Right?

Yeah, you're right, the template would have to be different anyway. But also, silly me, I've got that request totally wrong. Goal is to use ROP node name in template and that is similar, but different interesting problem.

antirotor:
Better to close this, this is not the way.

antirotor:
yeah, but the problem is that the final product name is then mangled here https://github.com/ynput/ayon-core/blob/eb69c5ca7f26d12b83d64fc9ccb3eb330a818db9/client/ayon_core/pipeline/farm/pyblish_functions.py#L572-L597 so you can't get by default to the original one.

BigRoy:
> yeah, but the problem is that the final product name is then mangled here
> 
> https://github.com/ynput/ayon-core/blob/eb69c5ca7f26d12b83d64fc9ccb3eb330a818db9/client/ayon_core/pipeline/farm/pyblish_functions.py#L572-L597
> so you can't get by default to the original one.

Again, sounds like a different problem.

Usually the ROP node itself does not represent the AOV - but it represents the renderlayer's name. So it represents the name the AOVs split from.

It sounds like you may be better of writing an issue that explains the client request in detail so it's clear what the needs are.

Is it just:
- Name the output of an instance using the ROP node name. -> Use `$OS`

Or is it:
- From a single ROP multiple AOVs are rendered and publishes. The studio wants to have **full control** on how those products are named from publishing. If so, what type of control? what are there needs?

This is better discussed/directed based on what the actual needs are then just saying "there will be an issue here" because it's unclear what it would be an issue for.

PR #436, opened 2024-04-19
Maya: Refactor and implement fixes for Load Placeholder

BigRoy:
## Changelog Description

Refactor Maya load placeholder, fixes:

- Implement fixes for outliner order of loaded placeholders (always place after placeholder)
- Allow storing more complex data, e.g. EnumDef with multiselection=True
- Change in behavior: This may fix a bug where previously placeholders may sometimes load only one subset when it should have loaded more. It could thus influence the loading behavior on existing templates
- Implements a PlaceholderPlugin base class to be used by others potential plug-ins

## Additional info

This PR depends on https://github.com/ynput/ayon-core/pull/425

I'll leave it in _draft_ until that PR is merged because the changelog will now appear very big.

## Testing notes:

1. Maya Workfile templates should still work
2. Loading maya workfile templates should put the loaded products correctly next to the placeholder

PR #426, opened 2024-04-17
Workfile Templates: add event system to Workfile Template Builder

BigRoy:
## Changelog Description

Add an `EventSystem` to the workfile template builder for placeholder plug-ins to register to.

From a placeholder plug-in you can now register to events:
```python
def callback(event):
    print(event.data["depth"])
    print(event.data["placeholders_by_scene_id"])

class MyPlaceholderPlugin(...):
    def populate_placeholder(self, placeholder):
        self.register_on_finished_callback(placeholder, callback)
        self.register_on_depth_processed_callback(placeholder, callback)
```

## Additional info

This is separated from #327 to make smaller PRs. Due to the event system being implemented here in isolation it may be hard to do a test run.

Once this is merged and #425 is merged I'll make follow-up PRs for Maya for plug-ins that actually use the event system.

## Testing notes:

1. Workfile templates should still work.
2. Code should be reviewed.

PR #425, opened 2024-04-17
Workfile Templates: Implement registering and discovering of `PlaceholderPlugin`

BigRoy:
## Changelog Description

Implement registering and discovering of Workfile Template Build Placeholder Plugins dynamically.

```python
from ayon_core.pipeline import register_template_placeholder_plugin_path
register_template_placeholder_plugin_path("/path/to/workfile_template_plugins")
```

## Additional info

Separated from https://github.com/ynput/ayon-core/pull/327

## Testing notes:

Make sure workfile templates still behave as intended (basically existing templates should continue to work)

##### After Effects

1. [x] Create placeholders
2. [ ] Build workfile templates (from new ones and existing ones)

##### Nuke

1. [x] Create placeholders
2. [x] Build workfile templates (from new ones and existing ones)

##### Maya

1. [x] Create placeholders
2. [x] Build workfile templates (from new ones and existing ones)

PR #437, opened 2024-04-19
Maya: Workfile Templates implement Assign Look placeholder

BigRoy:
## Changelog Description

Implement a workfile template placeholder plug-in for maya that will create an object set that on populating workfile template will assign the look to all nodes inside the set, including any generated loads from other load placeholders.

This way you can dynamically automatically assign looks on loaded content in workfile templates.

## Additional info

This PR relies on:
- https://github.com/ynput/ayon-core/pull/436
- https://github.com/ynput/ayon-core/pull/425
- https://github.com/ynput/ayon-core/pull/426

I'll keep this PR as a draft until all of those are merged.

## Testing notes:

1. Create a workfile template with the assign look placeholder.
2. The placeholder should be clear to use and do what is described above.
3. The created template should work.


tokejepsen (dismissed):
Looks good, but the `recursive` toggle does not seem to work:

Capture

Even with the toggle off, the look is still assigned to placeholder loaded content. Not whether we should even have this toggle? What is the use case for it?

LiborBatek (approved):
Works for me! Nice little feature...


Screenshot 2024-05-06 164824

Small note on ```Recursive```...didnt find a way to take it into account, as to me it always behaves recursively (its just enough to include parent groups for loaded assets to get look assign to them, which is nice in the end :)

BigRoy:
> Even with the toggle off, the look is still assigned to placeholder loaded content. Not whether we should even have this toggle? What is the use case for it?

> Small note on `Recursive`...didnt find a way to take it into account, as to me it always behaves recursively (its just enough to include parent groups for loaded assets to get look assign to them, which is nice in the end :)

@LiborBatek @tokejepsen I think you both thought that "recursive" means "apply to all children hierarchy" and thus expected children meshes to also get assigned or not? It does not mean that. It instead means that say your workfile template, loads a file, e.g. a maya scene that in itself also has placeholders that those placeholders inside that file will also get the look assigned, recursively.

When disabled, it will only assign the look on placeholders that should've existed after the initial first load of placeholders, and not those that are recursively included from templates that were loaded in the first iteration. So that the child template being loaded it in itself can also use the assign look placeholder to 'encompass' assignments within itself that are not overridden by the parent template (or it would get overridden if the parent template is recursive, etc.

Does that clarify things?

tokejepsen:
> Does that clarify things?

It does clarify it and its a neat feature I hadn't thought of. But think we'll need some documentation, tooltip or label description for that toggle. Would be a good way for people to discover the recursive workflow as well.

LiborBatek:
@BigRoy really cool...also never thought this might be possible! ...I think it would be cool to discuss the templates and workflows in general...and update the user docs, as I can see there is still a lot to cover aka lots of things not covered ! (as I am finding more and more features of the templates workflow every day!)

BigRoy:
I've made the PR compatible with the moved client code of `ayon_maya`
I've also expanded the tooltip slightly for the recursive toggle. It's not much, but may just help enough.

> But think we'll need some documentation, tooltip or label description for that toggle. Would be a good way for people to discover the recursive workflow as well.

Having documentation would be nice - but as far as I know that's a separate repo currently? Is that something you may be able to contribute knowledge for @LiborBatek - especially because recently on discord I recall you sharing some workfile template videos that might be nice there too (or even just some written demo of the general workfile templates workflow maybe?)

LiborBatek (approved):
All works for me!

image

I re did the tests and all seem functional. Also hints present whats the feature for...LGTM

PR #438, opened 2024-04-19
Maya: Workfile Templates implement Run Script Placeholder plug-in

BigRoy:
## Changelog Description

Implement a workfile template placeholder plug-in for maya that can run arbitrary Python scripts at different stages of the workfile template build process.

This allows creating custom logic for a workfile template for which there is no ready to go template plug-in available, but you do want to just force some extra logic to trigger on the workfile build.

## Additional info

This PR relies on:
- https://github.com/ynput/ayon-core/pull/436
- https://github.com/ynput/ayon-core/pull/425
- https://github.com/ynput/ayon-core/pull/426

I'll keep this PR as a draft until all of those are merged.

## Testing notes:

1. Create a workfile template with the Run Script placeholder.
2. The placeholder should be clear to use and do what is described above.
3. The created template should work.


client/ayon_core/hosts/maya/plugins/workfile_build/script_placeholder.py
```diff
@@ -0,0 +1,201 @@
...
+                items={
+                    "spaceLocator": "Locator",
+                    "objectSet": "ObjectSet"
+                },
+                tooltip=(
+                    "The placeholder's node type to be created.<br>"
+                    "<b>Note</b> this only works on create, not on update"
+                ),
+                default=options.get("create_nodetype", "objectSet")
+            ),
```
tokejepsen, line 108: What is the use case for having 2 types of nodes for this?

moonyuet: I have the same question of which toke has asked. 
And also shouldn't we add the `multiselection=True` into Enum Definition when there are multiple options being used or the related function has been updated to allow this happened without the argument?

BigRoy: > What is the use case for having 2 types of nodes for this?

It's so that the placeholder can be either a locator or an objectSet. It being an object set, may mean that it can e.g. contain multiple placeholders for the 'script' to act upon. It being a locator it can be used to e.g. specify a particular point in a hierarchy or whatever. Both may have their own reasons to be used.

A locator may also be more valuable in a 'nested/recursive' load of placeholders since the loaded hierarchy containing a placeholder of its own might already be parented into another hierarchy not known upon creation of the placeholder, etc.

> And also shouldn't we add the `multiselection=True` into Enum Definition when there are multiple options being used or the related function has been updated to allow this happened without the argument?

Not sure what you mean. It can be **either** an object set or a locator - not both at the same time.

moonyuet: > Not sure what you mean. It can be either an object set or a locator - not both at the same time.

I think I misunderstood the code a little bit. 
I get what you mean when you said the default value can be either an object set or a locator.

tokejepsen: So if a look locator is a child of another placeholder, the look will be assigned to its loaded parent container?

BigRoy: > So if a look locator is a child of another placeholder, the look will be assigned to its loaded parent container?

This particular PR is unrelated to looks. It's a placeholder that runs custom Python scripts defined by the user on creation.

tokejepsen: Sorry, I need to be awake before I go on github.

BigRoy: No worries - I'll mark this conversation as resolved. Let me know if there are other issues! :)

tokejepsen (approved):

antirotor (approved):

PR #440, opened 2024-04-22
Addons: Use package py

iLLiCiTiT:
## Changelog Description
Added `package.py` into settings addon so they don't have to be created dynamically.

## Additional info
File `package.py` should be used to define version, name, title and services, so wecan do it now.

## Testing notes:
1. Create server packages.
2. Upload to server.
3. Validate they work as before.


MustafaJafar (approved):
It works.
Ayon server was cool about the new addons. 
I uploaded Houdini, Maya and Nuke. opened them and they seem fine.

PR #443, opened 2024-04-22
Workfiles: Select latest workfile automatically

### LINKED ISSUE #434: Workfiles tool: Workfile selected in different task/folder is unreliable (not picking latest file)

- [X] I have searched the existing issues

### Current Behavior:

When switching in the workfiles tool from one folder or task to another and the user previously had a workfile selected it will also selected a workfile in the new context you've switched to. However, it seems to pick one somewhat randomly?

If I select e.g. v001 file in modeling and go to lookdev - then it selects a different file than if I had say v004 selected in modeling and switch to lookdev. It's a bit unclear why it picks which file, but it being inconsistent is bad.

The issue we're facing is that quite frequently artists click into another task/folder assuming the file picked there automatically is the latest, click open - but only a little later after working for a bit they realize they are actually in a much older file.

### Expected Behavior:

Preferably the behavior is consistent. 

But even better if it automatically selects what you expect it to select, which is usually the latest modified file. I believe that's also the behavior the way it was in OpenPype workfiles tool.

### Version

1.0.0

Windows

### Steps To Reproduce:

1. Switch between different folders and tasks with a workfile selected.

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

_No response_

### Additional context:

_No response_

BigRoy:
## Changelog Description

By default select the file with the latest date modified..

## Additional info

Fixes #434 

Note: in some odd cases where an artists saves over an older version than that file will have the latest date modified and hence will be selected. It's not the latest version, but latest date modified.

## Testing notes:

1. Open workfiles tool
2. Browsing different tasks and folders should always default to selecting the latest workfile.
3. It should feel logical for the majority of production cases.
4. Clicking refresh in the UI should not change the selection.

### DIFF
--- a/client/ayon_core/tools/workfiles/widgets/files_widget_workarea.py
+++ b/client/ayon_core/tools/workfiles/widgets/files_widget_workarea.py
@@ -416,9 +416,11 @@ class WorkAreaFilesWidget(QtWidgets.QWidget):
 
         # Find the row with latest date modified
         latest_index = max(
-            (self._proxy_model.index(i, 0) for
-             i in range(self._proxy_model.rowCount())),
-            key=lambda model_index: model_index.date(DATE_MODIFIED_ROLE)
+            (
+                self._proxy_model.index(idx, 0)
+                for idx in range(self._proxy_model.rowCount())
+            ),
+            key=lambda model_index: model_index.data(DATE_MODIFIED_ROLE)
         )
 
         # Select row of latest modified
--- a/client/ayon_core/tools/workfiles/widgets/files_widget_workarea.py
+++ b/client/ayon_core/tools/workfiles/widgets/files_widget_workarea.py
@@ -414,16 +414,20 @@ class WorkAreaFilesWidget(QtWidgets.QWidget):
         ):
             return
 
-        first_index = self._proxy_model.index(0, 0)
-        last_index = self._proxy_model.index(
-            0, self._proxy_model.columnCount() - 1
+        # Find the row with latest date modified
+        latest_index = max(
+            (self._proxy_model.index(i, 0) for
+             i in range(self._proxy_model.rowCount())),
+            key=lambda model_index: model_index.date(DATE_MODIFIED_ROLE)
         )
-        selection = QtCore.QItemSelection(first_index, last_index)
-        seleciton_model = self._view.selectionModel()
-        seleciton_model.select(
-            selection,
+
+        # Select row of latest modified
+        selection_model = self._view.selectionModel()
+        selection_model.select(
+            latest_index,
             (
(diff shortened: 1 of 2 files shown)


iLLiCiTiT (approved):

client/ayon_core/tools/workfiles/widgets/files_widget_workarea.py
```diff
@@ -394,3 +406,28 @@ def _on_expected_selection_change(self, event):
...
+            not self._change_selection_on_refresh
+            or self._proxy_model.rowCount() < 1
+        ):
+            return
+
+        # Find the row with latest date modified
+        latest_index = max(
+            (self._proxy_model.index(i, 0) for
+             i in range(self._proxy_model.rowCount())),
+            key=lambda model_index: model_index.date(DATE_MODIFIED_ROLE)
```
iLLiCiTiT, line 421:
```suggestion
            (
                self._proxy_model.index(idx, 0) for
                idx in range(self._proxy_model.rowCount())
            ),
            key=lambda model_index: model_index.date(DATE_MODIFIED_ROLE)
```

iLLiCiTiT:
> Does not work for me ATM, is there settings for that in Maya addon? as I have been testing in maya and nothing seems changed in that respect.

This is not settings related, it should select latest workfile when you change context, before this PR it did select random workfile.

LiborBatek:
> > Does not work for me ATM, is there settings for that in Maya addon? as I have been testing in maya and nothing seems changed in that respect.
> 
> This is not settings related, it should select latest workfile when you change context, before this PR it did select random workfile.

Thats why I asked...not working for me at all...still randomly selects workfiles without any pattern (when jumping accros context)

The only thing working atm is staying selected the same workfile when hitting ```Refresh``` function in ```Workfiles``` tool

client/ayon_core/tools/workfiles/widgets/files_widget_workarea.py
iLLiCiTiT: Question: Is last modified file the best option? Shouldn't be use the last of sorted by name?

BigRoy: I'd say so yes - in some cases we copy paste a workfile from another project - as a starting point to work from. If we'd always go by name then it may occur that my version number of my current workfile will never be 'higher alphabetically' than the other filename in the folder. Which is hard to workaround.

Whereas if latest modified, then just saving would intentionally make something the latest.

Would be happy to hear if others disagree though.

LiborBatek (approved):
This time all works like charm!
What a great little feature! Love it...

image

PR #444, opened 2024-04-23
Maya: optional shader stripping for model

antirotor:
## Changelog Description
Shaders are stripped from the model during extraction in Maya. This is adding option to disable that feature to retain shaders if needed.

## Additional info
This is affecting OBJ extractor and Maya Scene model extractor. Default behavior is backwards compatible (stripping on), can be configured in creator settings.

## Testing notes:
1. create model in Maya
2. disable shader stripping option
3. publish
4. load model - shaders should still be there

### DIFF
--- a/client/ayon_core/hosts/maya/plugins/publish/extract_model.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/extract_model.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
 """Extract model as Maya Scene."""
 import os
+from contextlib import nullcontext
 
-from maya import cmds
-
-from ayon_core.pipeline import publish
 from ayon_core.hosts.maya.api import lib
+from ayon_core.lib import BoolDef
+from ayon_core.pipeline import publish
+from maya import cmds
 
 
 class ExtractModel(publish.Extractor,
@@ -70,6 +71,8 @@ class ExtractModel(publish.Extractor,
                           noIntermediate=True,
                           long=True)
 
+        strip_shader = instance.data.get("strip_shaders", True)
+
         with lib.no_display_layers(instance):
             with lib.displaySmoothness(members,
                                        divisionsU=0,
@@ -78,7 +81,7 @@ class ExtractModel(publish.Extractor,
                                        pointsShaded=1,
                                        polygonObject=1):
                 with lib.shader(members,
-                                shadingEngine="initialShadingGroup"):
+                                shadingEngine="initialShadingGroup") if strip_shader else nullcontext():  # noqa: E501
                     with lib.maintained_selection():
                         cmds.select(members, noExpand=True)
                         cmds.file(path,
@@ -106,3 +109,14 @@ class ExtractModel(publish.Extractor,
 
         self.log.debug("Extracted instance '%s' to: %s" % (instance.name,
                                                            path))
+
+    @classmethod
+    def get_attribute_defs(cls):
+        return [
+            BoolDef("strip_shaders",
+                    label="Strip shaders from the model",
+                    default=True,
+                    tooltip=("Remove shaders from the model "
+                             "before extraction and re-apply "
+                             "them afterwards."))
+        ]
--- a/client/ayon_core/hosts/maya/plugins/publish/extract_obj.py
+++ b/client/ayon_core/hosts/maya/plugins/publish/extract_obj.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 import os
+from contextlib import nullcontext
 
-from maya import cmds
 import pyblish.api
-from ayon_core.pipeline import publish
 from ayon_core.hosts.maya.api import lib
+from ayon_core.pipeline import publish
+from maya import cmds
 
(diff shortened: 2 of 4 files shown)


LiborBatek:
@antirotor I did upload new maya addon package (0.1.16) and also removed older ones from the ayon-docker repo but I cant see any settings on the server

Seems there is some issue with the maya addon in this PR.

image

antirotor:
> Seems there is some issue with the maya addon in this PR.

should be fixed

LiborBatek (changes requested):
Hmm no matter if turning ```Strip Shader``` ```On``` or ```Off``` it ends up the pretty much the same
Screenshot 2024-04-23 184038


...both versions (v002) with stripping On and (v003) without any Striping of shaders ends up with same files sizes and also when inspecting those assets....all having InitialShading group assigned.

Im not sure if only should be propagated to ```OBJ``` but even when inspecting those assets I cant see any difference in kilobytes and inside the file...

Screenshot 2024-04-23 184515

And when inspected inside (OBJ file)

image

client/ayon_core/hosts/maya/plugins/publish/extract_obj.py
```diff
@@ -54,12 +68,13 @@ def process(self, instance):
...
                                        polygonObject=1):
                 with lib.shader(members,
-                                shadingEngine="initialShadingGroup"):
+                                shadingEngine="initialShadingGroup") if strip_shader else nullcontext():  # noqa: E501
                     with lib.maintained_selection():
                         cmds.select(members, noExpand=True)
                         cmds.file(path,
                                   exportSelected=True,
                                   type='OBJexport',
+                                  op=';'.join(f"{key}={val}" for key, val in self.obj_options.items()),  # noqa: E501
```
BigRoy, line 77: Would it be nicer to move this `op` formatting earlier instead of inline for readability?

client/ayon_core/hosts/maya/plugins/publish/extract_obj.py
```diff
@@ -54,12 +68,13 @@ def process(self, instance):
                                        pointsShaded=1,
                                        polygonObject=1):
                 with lib.shader(members,
-                                shadingEngine="initialShadingGroup"):
+                                shadingEngine="initialShadingGroup") if strip_shader else nullcontext():  # noqa: E501
```
BigRoy, line 71: If strip shader now forces `materials=0` on the export. Do we still need to force the initial shading group? Or does that help avoiding other stuff to be exported?

client/ayon_core/hosts/maya/plugins/publish/extract_model.py
```diff
@@ -70,6 +70,9 @@ def process(self, instance):
                           noIntermediate=True,
                           long=True)
 
+        strip_shader = instance.data.get("strip_shaders", True)
+        self.log.debug("Stripping shaders: %s" % strip_shader)
```
BigRoy, line 74: Is this debug log too verbose? Does it help debug certain edge cases?

antirotor: I think if you want to debug certain publish output - why your model comes without shaders, having this in the debug log can actually save you some time. But maybe it should be printed out only when strip shaders is enabled.

client/ayon_core/hosts/maya/plugins/publish/extract_obj.py
antirotor: to be honest, options on OBJExport are so undocumented that I left that here just to be safe. But I guess it is slowing extraction down unnecessarily.

client/ayon_core/hosts/maya/plugins/publish/extract_model.py
BigRoy: I also wonder whether "strip shaders" is better than reversing the terminology and naming the option "include shaders" or "export shaders" which is how usually export options are shown (whether something is included or not, instead of whether it's stripped?)

antirotor: That term got there from the clients so I guess that is how they think about it. "Exporting models is stripping any shaders on them". What do you think @LiborBatek

BigRoy: Totally understand it from the perspective of what `model` family has always been, etc. - But just looking at exporters in general and their terminology usage it's usually "what would you like to include"? That's why I wondered whether a `export materials` flag would be better?

Would love to know what Libor thinks as well. And maybe @dee-ynput from a product perspective.

LiborBatek:
Well to me its just really cosmetics...however the term ```strip``` could potentially cause some noise speaking of artists 

```Export Materials```
```Keep Materials```
```Include Materials```
```Remove Materials```

any of these is clear enough for pretty much anyone imho.

Maybe the question is whether we should incorporate the new option in the Publisher as ```Enabled``` or ```Disabled``` by defaults and name it accordingly to it (remove OR keep materials etc) ...but maybe Im just thinking too much about it (backward compatibility and new option)

PR #445, opened 2024-04-23
Chore: Add folder type and path to template data

iLLiCiTiT:
## Changelog Description
It is possible to use `{folder[type]}` and `{folder[path}]}` in anatomy templates.

## Additional info
Added requirement for `"folderType"` field on folder entity. Hardcoded `"Shot"` folder type as default for `"newAssetPublishing"` as that's current hardcoded default (It is out of scope of this PR to change that logic).

## Testing notes:
1. Add `{folder[type]}` to your anatomy templates.
2. Validate they are used. Workfile path, publish path, delivery path etc. should be with the type.


client/ayon_core/pipeline/template_data.py
```diff
@@ -101,6 +101,7 @@ def get_folder_template_data(folder_entity, project_name):
     return {
         "folder": {
             "name": folder_name,
```
fabiaserra, line 103: could you add:
```
"path": path,
```
as well?

antirotor (approved):

PR #447, opened 2024-04-24
Use product name templates for render products on farm publish

antirotor:
## Changelog Description
Product names for render instance were hardcoded in the Deadline logic, not following product name template use in other places. This PR adds ability to use product name templates with all necessary keys to influence how render products are named. It also collects information about renderlayer (ROP node name) to anatomy data along with AOV name so it can be used for template formatting.

## Additional info
Product names for instances to be published were hardcoded directly in the script here:
https://github.com/ynput/ayon-core/blob/8f42c50100bc384a043034e1d723f18cc6c0ceb7/client/ayon_core/pipeline/farm/pyblish_functions.py#L573-L597

This was bypassing any template set in `ayon+settings://core/tools/creator/product_name_profiles`. With this change, one can set template like this:

`{product[type]}<{Renderlayer}><_{Aov}>`

resulting in:

`renderMainlayer_normal`

It is up to TD to set correctly such template to make sure, product name will be stable enough. As a side-effect, optional keys can be used now in product name templates adding flexibility (and perhaps some danger of misuse).

`renderlayer` is now set on instance along with `aov` and is collected to anatomy data, so it can be used in template formatting so you can use template for renders like:

```
{root[work]}/{project[name]}/{hierarchy}/{asset}/publish/{family}/{subset}/{@version}/<{renderlayer}/>
```
> [!NOTE]
> Already collected anatomy data (when available) could be passed to `get_product_name()` to allow even more keys to be used, but is there any real use case for it?

Also, as a point for discussion - there is `variant` that is now ignored in this case. It could be reasonable to use it exactly for this and move templating to "variant templates" but maybe it is breaking settings to even smaller unmanagable bits.

## Testing notes:
1. Set product name template to use `renderlayer` or `aov` keys (can be optional using`<{key}>` syntax)
2. Publish renders from DCC supporting setting of these keys (like Maya and Houdini)
3. Use `{renderlayer}` or  `{aov}` in anatomy templates


moonyuet:
Should we also test to see if it works with 3dsmax multicam renders?

LiborBatek (commented):
I have setup ```product template``` using the ```renderlayer``` token but constantly getting errors when trying ```publish```

Screenshot 2024-05-02 164502

```
    runner(*args)
  File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\collect_anatomy_instance_data.py", line 56, in process
    self.fill_latest_versions(context, project_name)
  File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\collect_anatomy_instance_data.py", line 263, in fill_latest_versions
    product_entities = list(ayon_api.get_products(
  File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\server_api.py", line 4248, in get_products
    parsed_data = query.query(self)
  File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\graphql.py", line 343, in query
    raise GraphQlQueryFailed(response.errors, query_str, variables)
ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Name 'LightingMain_{renderlayer}' does not match regex '^[a-zA-Z0-9_]([a-zA-Z0-9_\.\-]*[a-zA-Z0-9_])?$' on item 'project/products' (Line 3 Column 5)
```

client/ayon_core/hosts/houdini/plugins/publish/collect_karma_rop.py
```diff
@@ -33,6 +33,8 @@ class CollectKarmaROPRenderProducts(pyblish.api.InstancePlugin):
     def process(self, instance):
 
         rop = hou.node(instance.data.get("instance_node"))
+        # to align with maya render layers
+        instance.data["renderlayer"] = rop.name()
```
BigRoy, line 37: Sorry, what? Why is the ROP node's name used as the `renderlayer`? That seems odd?

Isn't it much easier if a studio wants this for them to us `$OS` as part of the `productName` or `subset` attributes on the node itself so that it's editable by the studio, requires no additional changes?

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -352,6 +352,17 @@ def fill_anatomy_data(self, context):
...
+            # make render layer name (or ROP name) available in the
+            # templates
+            render_layer = instance.data.get("renderlayer")
+            if render_layer:
+                anatomy_data["renderlayer"] = render_layer
+
+            # make AOV name if present available for templates
+            aov = instance.data.get("aov")
+            if aov:
+                anatomy_data["aov"] = aov
```
BigRoy, line 364: What's "renderlayer" if not part of the product name already? This seems like an odd decision to allow `renderlayer` (which as far as I know doesn't really have a valid connotation in Houdini?) to be part of the anatomy template? Or?

client/ayon_core/hosts/houdini/plugins/publish/collect_karma_rop.py
antirotor: Yes, but then you can't use it in anatomy data. One thing is product name, the other is to use it somewhere else in path. I agree that render layer is somewhat arbitrary name in Houdini context, but most of the DCCs are using render layer is one way or the other and  I wouldn't spam the anatomy data with every DCC specific term.

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
antirotor: you might want to have render layer as a part of the path (not the whole product name).

BigRoy: > you might want to have render layer as a part of the path (not the whole product name).

When doing so - what would you anticipate the product name be instead? 

Coming back to this from your PR description:

> This was bypassing any template set in `ayon+settings://core/tools/creator/product_name_profiles`. With this change, one can set template like this:
> 
> `{product[type]}<{Renderlayer}><_{Aov}>`
> 
> resulting in:
> 
> `renderMainlayer_normal`
> 
> It is up to TD to set correctly such template to make sure, product name will be stable enough. As a side-effect, optional keys can be used now in product name templates adding flexibility (and perhaps some danger of misuse).
> 
> `renderlayer` is now set on instance along with `aov` and is collected to anatomy data, so it can be used in template formatting so you can use template for renders like:

In practice a user should usually never see this. They see the product name or representation in the loader or the manager. Which raises the question - how come its _this_ crucial in the published filepath which usually isn't something that is displayed to the artists anyway?

antirotor: I think it is just an option to categorize things when you need to have access to them outside loaders/pipeline. We get those requests from time to time when you have part that is not integrated and you need to access the files directly. Then dropping the whole render layer folder to DCC with all of it's AOV sequences can speed things up.

antirotor: You mean to "force" it to context on representation even without it being used in the template so it can be picked up later on by delivery action? Like we are storing `udim` here https://github.com/ynput/ayon-core/blob/4d84ed8696657d976b6939ec1afab044e943818f/client/ayon_core/plugins/publish/integrate.py#L856-L859 (and also `frame` somewhere)

BigRoy: Yes, similar to that. I'd expect this to way more valuable for delivery templates than production templates since in production templates it may be likely you will have duplicity anyway since usually the AOV is part of the product name as well.

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -539,68 +543,60 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
+        # TODO: this is not true, we can have multiple cameras in the scene
+        #       and we should be able to detect them all. Currently, we are
+        #       keeping the old behavior, taking the first one found.
+        if camera:
+            dynamic_data["camera"] = camera[0]
+
+        product_name = get_product_name(
+            project_name=instance.context.data["projectName"],
+            task_name=task_name,
+            task_type=instance.data['taskEntity']['taskType'],
```
iLLiCiTiT, line 580:
```suggestion
            task_type=instance.data["taskEntity"]["taskType"],
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -539,68 +543,60 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
+            host_name=instance.context.data["hostName"],
+            product_type=skeleton['productType'],
```
iLLiCiTiT, line 582:
```suggestion
            product_type=skeleton["productType"],
```

MustafaJafar:
Hey, I was trying to test this PR in Houdini and I'm confused.

which one should I change ? and what is the expected behavior ? 
- `ayon+settings://core/tools/creator/product_name_profiles` 
- `ayon+anatomy://{my-project}/templates/publish/1` _render template_.

As far as I know updating `product_name_profiles` affect the product name in the creator tab. and how does that support renderLayer (rop name) before actually creating the rop ? 

Also, I wonder how does that work with run-time instances that are created in https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py

LiborBatek:
@MustafaJafar  I guess you need to change both...one alters the product name (aka rendered file name) and the other alter or sets the folder path for the rendered sequence (again using tokens like <aov>) which wasnt possible before I guess.

MustafaJafar:
I tested this PR furtherly and Here my results in Houdini: 
On farm it works as expected. 
| Render job | Render after Integration |
|--|--|
| image | image |

On local rendering, I think some work needs to be added to https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
I'll add my local edits from my testing as a PR to this PR.

LiborBatek (approved):
Finally I could properly test it and thanks to @antirotor setup correctly my ```templates``` so it successfully finnished my DL job.

All seems working well.

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -539,68 +546,60 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
+        product_name = get_product_name(
+            project_name=instance.context.data["projectName"],
+            task_name=task_name,
+            task_type=instance.data["taskEntity"]["taskType"],
+            host_name=instance.context.data["hostName"],
+            product_type=skeleton["productType"],
+            dynamic_data=dynamic_data,
+            variant=instance.data.get('variant', ''),
+            project_settings=instance.context.data.get("project_settings"),
+        )
```
BigRoy, line 589: I'm a bit worried about the fact that this now seems to require templates to use `aov` to ensure multiple AOVs become a unique product? Does this mean that with 'older templates' this would now break? Or what is ensuring that they become separate products?

client/ayon_core/hosts/houdini/plugins/publish/collect_local_render_instances.py
```diff
@@ -64,6 +64,8 @@ def process(self, instance):
         expectedFiles = next(iter(instance.data["expectedFiles"]), {})
 
         product_type = "render"  # is always render
+        # TODO: Match the value of product_group with the value of
+        #        group_name in farm pyblish functions.
         product_group = get_product_name(
             context.data["projectName"],
             context.data["taskEntity"]["name"],
```
MustafaJafar, line 71: for reference, 
I gave this PR a test. and was going to mention an issue with the product group name in local Houdini render.
Then, I figured out there was a comment about it in the code.
image

m-u-r-p-h-y: how come it was working for @LiborBatek ?

LiborBatek: > how come it was working for @LiborBatek ?

because tested for ```maya``` I guess...

antirotor:
This need some transitional option in the settings so default template can be configured.

antirotor:
This will be reopened in Houdini addon new repository.

PR #448, opened 2024-04-25
Maya: Solely collect the beauty render when Maya render view is disabled in Arnold 

moonyuet:
## Changelog Description
This PR is to fix the bug of collecting aovs when Maya render view is disabled in the render setting with Arnold as the renderer.

## Additional info
n/a

## Testing notes:
1. Launch Maya
2. Create Render Instance with Arnold as the renderer
3. Set Mode to `disabled` in Maya Render View
image
4. Publish


LiborBatek (approved):
Did test with  mode ```Disabled``` and publish produces just the ```Main.exr``` and not expecting any ```AOVs``` when inspecting the JSON file.

Screenshot 2024-04-26 111023

Screenshot 2024-04-26 111242

Small note: for some reason I didnt get any ```Review``` mp4 but probably not related to this PR anyway...

Seems ok.

PR #456, opened 2024-04-26
Houdini: support hda publishing from non object level and other enhancements

MustafaJafar:
## Changelog Description
This PR:
- Adds support HDA publishing from non object level
- Adds a bug fix related to `setName`
- Don't create a new subnetwork if user selected a subnetwork.
- Adds creator settings to: (users can still override it in the HDA type properties)
    - use project name as submenu path (in tab manu)
    - use their name as the author of the HDA
- Adds default template name profile for Houdini HDA product type.

Things to do: 
- [x] allow publishing from non object level
- [x] check HDA extractor if it still works
- [x] Set tool tab menu location to `AYON/{project_name}`
- [x] Work on HDA Loader to make it possible to load HDAs from non object level.
- [x] Pick a default path for loading VOP HDAs, more info https://github.com/ynput/ayon-core/pull/456#issuecomment-2101326271
  - Note: The default path for loading VOPs is Material Network

https://github.com/ynput/ayon-core/assets/20871534/ec771935-eb70-47e4-813f-e5c6cac328f9

Forgot to mention that 
HDAs TAB location is inside `AYON/project_name` in tab menu
image

## Testing notes:
1. Create and publish HDAs from different Houdini contexts
2. Test selection: 
  i. if selection is a subnetwork -> creator plugin will use it without creating a new subnetwork
  ii.if selection is any types other than subnetwork -> creator plugin will put them a new subnetwork
  iii. if no selection -> creator plugin will deduce your Houdini path context and create a new subnetwork there.


MustafaJafar:
I wanted to add `max_num_outputs` but such a flag doesn't exist in `hou.Node.createDigitalAsset` 
I was able to find its docs in [virtual HOM_Node* HOM_OpNode::createDigitalAsset](https://www.sidefx.com/docs/hdk/class_h_o_m___op_node.html#ae6a159e02c85d0839349435fdf4cec7d). 
Maybe they didn't add it because we can't change the output number in the subnetworks in obj node. 

image

---

For reference: 
I have no idea how to edit maximum number of outputs from houdini python api 
Also, I have no idea how to change the menu entry from houdini python api 
image

moonyuet:
Not quite sure of the purpose of the minimum inputs would be nice in this implementation. If you have two minimum inputs, which means user need to have two inputs connect to the node(s) inside the hda subnet, while if you have four maximum inputs, the hda node would provide four inputs for the nodes to connect into the node(s) inside the subset. I used this kind of subnet for some rigid body classes before, and I don't think it needs to have minimum inputs (which is compulsory for the node) and maximum inputs are already enough. Consider a scenario when the user wants to remove some unnecessary connections between the hda input and the node. There would be an error if you set minimum inputs for mandatory input. 
Here is the example where I set 1 minimum input and 3 maximum inputs.
image

I don't see the reason why we need two outputs for the HDAs(or some other houdini nodes), Usually the node I used have one output (for both dop and sop level), so it's normal Houdini doesn't allow to set the maximum output.

MustafaJafar:
> I don't see the reason why we need two outputs for the HDAs(or some other houdini nodes), Usually the node I used have one output (for both dop and sop level), so it's normal Houdini doesn't allow to set the maximum output.

This happens on Obj level in Houdini but in e.g. `SOP`,  Houdini allows having different outputs.
And since this PR allows creating HDAs in different Houdini levels then _I assume_ we are going to hit this problem more often. 

image

---
Also, some HDAs are built in that way that requires some input to work. 
I was able to modify the type properties of the created HDA manually to adjust the number of inputs and ouputs and then publishing and neither Houdini or Ayon plugins errored.

image

BigRoy:
> Not quite sure of the purpose of the minimum inputs would be nice in this implementation. If you have two minimum inputs, which means user need to have two inputs connect to the node(s) inside the hda subnet, while if you have four maximum inputs, the hda node would provide four inputs for the nodes to connect into the node(s) inside the subset. I used this kind of subnet for some rigid body classes before, and I don't think it needs to have minimum inputs (which is compulsory for the node) and maximum inputs are already enough. Consider a scenario when the user wants to remove some unnecessary connections between the hda input and the node. There would be an error if you set minimum inputs for mandatory input.
> Here is the example where I set 1 minimum input and 3 maximum inputs.

Without having read much of the conversation here I'd say we should support as much as possible what Houdini offers.
If it allows the HDA to set:
- minimum inputs
- maximum inputs
- amount of outputs

We should be supporting each of these. Right?

MustafaJafar:
> We should be supporting each of these. Right?

I agree. 
Also, let's consider that python api or documentation are  limiting. I couldn't find the proper function/method to edit the maximum ouptuts or the HDA location in the tab menu. 

I could only find some docs in the HDK docs! or using `help(hou.Node.createDigitalAsset)`
https://www.sidefx.com/docs/hdk/class_h_o_m___op_node.html#ae6a159e02c85d0839349435fdf4cec7d

moonyuet:
> > Not quite sure of the purpose of the minimum inputs would be nice in this implementation. If you have two minimum inputs, which means user need to have two inputs connect to the node(s) inside the hda subnet, while if you have four maximum inputs, the hda node would provide four inputs for the nodes to connect into the node(s) inside the subset. I used this kind of subnet for some rigid body classes before, and I don't think it needs to have minimum inputs (which is compulsory for the node) and maximum inputs are already enough. Consider a scenario when the user wants to remove some unnecessary connections between the hda input and the node. There would be an error if you set minimum inputs for mandatory input.
> > Here is the example where I set 1 minimum input and 3 maximum inputs.
> 
> Without having read much of the conversation here I'd say we should support as much as possible what Houdini offers. If it allows the HDA to set:
> 
> * minimum inputs
> * maximum inputs
> * amount of outputs
> 
> We should be supporting each of these. Right?

Maybe it is good idea to support each of these . But as far as I notice from the nodes in Houdini, they are mostly not with more than one output. Maybe I am wrong for the latest version of Houdini where some nodes have more than one output( I haven't been using houdini to make scene project for a long time so I am not quite updated). 
@fabiaserra can you give some advice on that?

BigRoy:
> If I dont get it wrong, multiple outputs can only supports sop level but not the obj level.

I believe so, yes. KineFX has a lot of nodes that have multi-input and outputs, for good reason.

MustafaJafar:
> If I dont get it wrong, multiple outputs can only supports sop level but not the obj level.

Also, for reference `hou.Node.createDigitalAsset` doesn't support specifiying the maximum number of outputs. 
let me quote `help(hou.Node.createDigitalAsset)` it's not mentioned in the docs! 

```python
>>> help(hou.Node.createDigitalAsset)
Help on function createDigitalAsset in module hou:

createDigitalAsset(self, name: 'char const *' = None, hda_file_name: 'char const *' = None, description: 'char const *' = None, min_num_inputs: 'int' = 0, max_num_inputs: 'int' = 0, compress_contents: 'bool' = False, comment: 'char const *' = None, version: 'char const *' = None, save_as_embedded: 'bool' = False, ignore_ext
ernal_references: 'bool' = False, compile_asset: 'bool' = False, change_node_type: 'bool' = True, create_backup: 'bool' = True, install_path: 'char const *' = None) -> 'HOM_Node *'
    createDigitalAsset(self, name=None, hda_file_name=None,
    description=None, min_num_inputs=0, max_num_inputs=0,
    compress_contents=False, comment=None, version=None,
    save_as_embedded=False, ignore_external_references=False,
    change_node_type=True, create_backup=True, install_path=None) -> Node
    
        Create a digital asset from this node. You would typically call this
        method on subnet nodes.
    
    
        name
            The name of the node type that the new digital asset will
            define.
    
        hda_file_name
            The name of the hda file where Houdini will save the digital
            asset. If None Houdini will use
            $HOME/houdiniX.Y/hda/OPcustom.hda.
    
        description
            The name that will appear in the tab menu. If None, Houdini will
            use the name for the description.
    
        min_num_inputs
            The minimum number of inputs that need to be wired into
            instances of the digital asset. See
            hou.HDADefinition.minNumInputs for more information.
    
        max_num_inputs
            The number of input connectors available on instances of the
            digital asset for input connections. See
            hou.HDADefinition.minNumInputs for more information.
    
        compress_contents
            Whether or not the contents of this digital asset are compressed
            inside the hda file. See hou.HDAOptions.compressContents for
            more information.
    
        comment
            A user-defined comment string. See hou.HDADefinition.comment for
            more information.
    
        version
            A user-defined version string. See hou.HDADefinition.version for
            more information.
    
        save_as_embedded
            Whether or not the digital asset's definition will be saved with
            the hip file instead of an hda file. When this parameter is
            True, Houdini ignores the hda_file_name parameter. Setting this
            parameter to True is equivalent to setting this parameter to
            False and setting the hda_file_name parameter to Embedded.
    
        ignore_external_references
            If True, Houdini will not generate warnings if the contents of
            this digital asset reference nodes outside the asset.
    
        change_node_type
            Normally, Houdini will change the node creating the digital
            asset into the new digital asset type. Setting this flag to
            false will cause the node to remain unchanged.
    
        create_backup
            Create a backup before modifying an existing hda file.
    
        install_path
            Where to install the new hda. When not specified, it will
            install to either Current HIP File or Scanned Asset Library
            Directories.
```

moonyuet:
> KineFX
Maybe we can just set the condition to inform the message user only one output is being used at the object level. 
(You prove me to be not very updated user, I used houdini for project when it is at around 11-12, the time when Solaris not even released yet)

moonyuet:
> > If I dont get it wrong, multiple outputs can only supports sop level but not the obj level.
> 
> Also, for reference `hou.Node.createDigitalAsset` doesn't support specifiying the maximum number of outputs. let me quote `help(hou.Node.createDigitalAsset)` as it's not mentioned in the docs!
> 
> ```python
> >>> help(hou.Node.createDigitalAsset)
> Help on function createDigitalAsset in module hou:
> 
> createDigitalAsset(self, name: 'char const *' = None, hda_file_name: 'char const *' = None, description: 'char const *' = None, min_num_inputs: 'int' = 0, max_num_inputs: 'int' = 0, compress_contents: 'bool' = False, comment: 'char const *' = None, version: 'char const *' = None, save_as_embedded: 'bool' = False, ignore_ext
> ernal_references: 'bool' = False, compile_asset: 'bool' = False, change_node_type: 'bool' = True, create_backup: 'bool' = True, install_path: 'char const *' = None) -> 'HOM_Node *'
>     createDigitalAsset(self, name=None, hda_file_name=None,
>     description=None, min_num_inputs=0, max_num_inputs=0,
>     compress_contents=False, comment=None, version=None,
>     save_as_embedded=False, ignore_external_references=False,
>     change_node_type=True, create_backup=True, install_path=None) -> Node
>     
>         Create a digital asset from this node. You would typically call this
>         method on subnet nodes.
>     
>     
>         name
>             The name of the node type that the new digital asset will
>             define.
>     
>         hda_file_name
>             The name of the hda file where Houdini will save the digital
>             asset. If None Houdini will use
>             $HOME/houdiniX.Y/hda/OPcustom.hda.
>     
>         description
>             The name that will appear in the tab menu. If None, Houdini will
>             use the name for the description.
>     
>         min_num_inputs
>             The minimum number of inputs that need to be wired into
>             instances of the digital asset. See
>             hou.HDADefinition.minNumInputs for more information.
>     
>         max_num_inputs
>             The number of input connectors available on instances of the
>             digital asset for input connections. See
>             hou.HDADefinition.minNumInputs for more information.
>     
>         compress_contents
>             Whether or not the contents of this digital asset are compressed
>             inside the hda file. See hou.HDAOptions.compressContents for
>             more information.
>     
>         comment
>             A user-defined comment string. See hou.HDADefinition.comment for
>             more information.
>     
>         version
>             A user-defined version string. See hou.HDADefinition.version for
>             more information.
>     
>         save_as_embedded
>             Whether or not the digital asset's definition will be saved with
>             the hip file instead of an hda file. When this parameter is
>             True, Houdini ignores the hda_file_name parameter. Setting this
>             parameter to True is equivalent to setting this parameter to
>             False and setting the hda_file_name parameter to Embedded.
>     
>         ignore_external_references
>             If True, Houdini will not generate warnings if the contents of
>             this digital asset reference nodes outside the asset.
>     
>         change_node_type
>             Normally, Houdini will change the node creating the digital
>             asset into the new digital asset type. Setting this flag to
>             false will cause the node to remain unchanged.
>     
>         create_backup
>             Create a backup before modifying an existing hda file.
>     
>         install_path
>             Where to install the new hda. When not specified, it will
>             install to either Current HIP File or Scanned Asset Library
>             Directories.
> ```

Not sure if the link below helps
https://www.sidefx.com/docs/houdini/hom/hou/NodeType.html

MustafaJafar:
> Not sure if the link below helps
> https://www.sidefx.com/docs/houdini/hom/hou/NodeType.html

it doesn't unfortunately `hou.Node.createDigitalAsset` is only documented in Houdini HDK docs 
and can be found in python via `help` function. 
The code above is the result of `helphou.Node.createDigitalAsset)` which is almost the same as [virtual HOM_Node* HOM_OpNode::createDigitalAsset](https://www.sidefx.com/docs/hdk/class_h_o_m___op_node.html#ae6a159e02c85d0839349435fdf4cec7d)

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
```diff
@@ -84,5 +110,22 @@ def create(self, product_name, instance_data, pre_create_data):
 
     def get_network_categories(self):
         return [
-            hou.objNodeTypeCategory()
+            category for name, category in hou.nodeTypeCategories().items()
+            if name in {
+                "Chop", "Cop2", "Dop", "Driver", "Lop",
+                "Object", "Shop", "Sop", "Top", "Vop"
+            }
```
BigRoy, line 117: Is there a reason we prefer this over explicitly providing it the type categories directly as in ` hou.chopNodeTypeCategory(), hou.cop2NodeTypeCategory(), ...`?

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
```diff
@@ -68,13 +87,20 @@ def create_instance_node(
...
+        # be incremented.
+        hda_node.setName(node_name, unique_name=True)
         self.customize_node_look(hda_node)
         return hda_node
 
     def create(self, product_name, instance_data, pre_create_data):
         instance_data.pop("active", None)
 
+        self.min_num_inputs = pre_create_data["min_num_inputs"]
+        self.max_num_inputs = pre_create_data["max_num_inputs"]
```
BigRoy, line 102: It seems the HDA creator is the only creator in Houdini currently that overrides `create_instance_node` method. As such, I'd actually instead change its signature and have it just pass along `pre_create_data` directly to the `create_instance_node` method since it's just much better design than storing it as attributes on the Creator plug-in class (which is also shared between instances, etc.)

So basically, just add `pre_create_data` as argument to `create_instance_node` method.

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
```diff
@@ -16,6 +17,9 @@ class CreateHDA(plugin.HoudiniCreator):
     icon = "gears"
     maintain_selection = False
 
+    min_num_inputs = 0
+    max_num_inputs = 0
```
BigRoy, line 21: I feel like these class attributes are not needed. See [here](https://github.com/ynput/ayon-core/pull/456/files#r1584444651).

I'd recommend removing these and just setting the defaults directly in the `EnumDef` since this doesn't seems like something that needs to be exposed in settings?

Also, is `0` the good "default"? Wouldn't it usually default to `1`?

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
moonyuet: > Is there a reason we prefer this over explicitly providing it the type categories directly as in ` hou.chopNodeTypeCategory(), hou.cop2NodeTypeCategory(), ...`?

Maybe because we have talked about the hda node can offer more than one outputs at SOP level but only one output at Obj level? But `hou.objNodeTypeCategory()` is already returning to all the node type categories so we dont really need to change this.

MustafaJafar: > Is there a reason we prefer this over explicitly providing it the type categories directly

I found it easier to read and maintain because category names are easier to remember.

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: `0` is always the default for `min_num_inputs` and `max_num_inputs` in `hou.Node.createDigitalAsset(...)` 

```python
createDigitalAsset(self, name=None, hda_file_name=None,
    description=None, min_num_inputs=0, max_num_inputs=0,
    compress_contents=False, comment=None, version=None,
    save_as_embedded=False, ignore_external_references=False,
    change_node_type=True, create_backup=True, install_path=None) -> Node
```

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: Does that mean I need to add `pre_create_data` as in the following ? 

1. on calling `self.create_instance_node` inside super(..).create(..)
https://github.com/ynput/ayon-core/blob/df976a4ab8900b14c70331519afd563e871c8d09/client/ayon_core/hosts/houdini/api/plugin.py#L186-L187

2. in `super(..).create_instance_node(..)`
https://github.com/ynput/ayon-core/blob/df976a4ab8900b14c70331519afd563e871c8d09/client/ayon_core/hosts/houdini/api/plugin.py#L143-L145

3. last but not least in `CreateHDA.create_instance_node`

---
I'm a little bit frustrated because `create_instance_node` is called from `super(..).create(..)` not from `CreateHDA.craete(..)` 

Also, `super(..).create(..)` doesn't support `instance_node` as argument

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
moonyuet: I agree that this should be removed unless we have these attributes in the houdini addon setting.
The one from the enum definition should be enought for now.

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: Are explicit category type more favored ?

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: I also agree, but how can I pass the `pre_create_data` to `create_instance_node` ? 
`create_instance_node` is called inside a super method that doesn't pass that argument.

This mentioned here https://github.com/ynput/ayon-core/pull/456#discussion_r1586038828


what about saving `pre_create_data` to `self.pre_create_data` 
and  use it inside `create_instance_node` ?

BigRoy: Yes, it would be along the lines of https://github.com/BigRoy/ayon-core/commit/6e68802c7617e1aba635be89d60051583e1226ff

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
BigRoy: Exactly, see for example: https://github.com/BigRoy/ayon-core/commit/6e68802c7617e1aba635be89d60051583e1226ff

You could merge that branch into yours to get the commit (or cherry pick the commit via git)

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: oh, thanks!

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: I cherry-picked it, Thanks!

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
MustafaJafar: Fair! 
Thank you for the tip. 
implemented in https://github.com/ynput/ayon-core/pull/456/commits/79139ce0dd4b31c32c98d26403a2f78c6e345f14

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
moonyuet: Are we still keeping these two variables or we can just set the values in the number definitions?

MustafaJafar: They are used inside `create_instance_node` as a fall back value.

https://github.com/ynput/ayon-core/blob/79139ce0dd4b31c32c98d26403a2f78c6e345f14/client/ayon_core/hosts/houdini/plugins/create/create_hda.py#L51-L54

fabiaserra:
the way I've seen HDA management in Houdini has always been integrated as an action menu on the nodes itself:
image

Adding a new menu like `Create Digital Asset (AYON)...` where it then wraps the creation of the HDA by hiding some of the attributes that you don't want exposed (i.e. destination path, main submenu so they only choose between some predefined options...)
image

Those HDA management pipelines usually provide ways to make sure if anyone on the studio starts editing a certain HDA no one else can be editing it at the same time and things like that, but I'd try stick as much as possible to the Houdini vanilla HDA editing workflows than trying to abstract that away. For example, the min/max outputs would be controlled through the operator type properties interface and doesn't need to be something they define ahead of time when the HDA gets created
image

fabiaserra:
> Once they are happy with the local edits they have done to the HDA, they then have a button to "publish" the HDA so other users pick those changes up

As you know since I have said multiple times like a broken record, I think the UX of the create/publish mechanism in Houdini is very cumbersome and counter-intuitive to the vanilla Houdini workflow. I think the `create` step should be implicit on all the vanilla Houdini nodes that we support publishing for (i.e. automatically wrap the vanilla nodes and add a `Publish` tab where they can further control the publish parameters and submit to publish). Although as I said on my prior comment HDA is a bit of a different monster than most other publish types and the UX creation should be treated slightly differently

fabiaserra:
> Once they are happy with the local edits they have done to the HDA, they then have a button to "publish" the HDA so other users pick those changes up

For published HDAs that are loaded in the scene, the action menu that shows wouldn't be `Create Digital Asset (AYON)...` but instead something like `Edit Digital Asset (AYON)...`, which internally would trigger the `Allow editing of contents`, change the destination path to a sandbox local location, set a "mutex" flag so anyone else trying to edit the HDA gets a dialog that the node is already being edited by X person... and from this edit state there would be a new action like `Publish Digital Asset (AYON)...` where they could choose if it's a major or minor version update (for minor updates everyone that has the HDA loaded gets the change automatically and major updates they need to re-import the HDA)

MustafaJafar:
@fabiaserra I love your ideas.

BigRoy:
> Also, for reference `hou.Node.createDigitalAsset` doesn't support specifiying the maximum number of outputs.
> let me quote `help(hou.Node.createDigitalAsset)` as it's not mentioned in the docs!

It's in the docs [here](https://www.sidefx.com/docs/houdini/hom/hou/OpNode.html#assets) on `hou.OpNode`, a child of `hou.Node` since recent versions of Houdini after the introduction of `hou.ApexNode`.

MustafaJafar:
> It's in the docs [here](https://www.sidefx.com/docs/houdini/hom/hou/OpNode.html#assets) on `hou.OpNode`, a child of `hou.Node` since recent versions of Houdini after the introduction of `hou.ApexNode`.

Thank you!

MustafaJafar:
Here's one more demo. 
I think it's considered a step towards what what fabia mentioned here https://github.com/ynput/ayon-core/pull/456#issuecomment-2096575818 
Because the creator plugin includes the necessary code to create/edit the HDA.

https://github.com/ynput/ayon-core/assets/20871534/62713e1a-c961-486e-a811-943a27c7e6da

moonyuet:
I tested with the latest commit. I can create multiple outputs in object level but I can't create multiple output in sop level.

https://github.com/ynput/ayon-core/assets/64118225/add6ee87-75d0-404c-afaf-887a8f9c91cf

I agree with @fabiaserra that we should stick with the HDA editing workflow and the max/min outputs should be updated through the operator properties. Users can customize their own outputs/inputs by accessing the properties editor.
Just from my own perspective, the option of minimum/maximum outputs would give user's perceptions that they are able to create multiple outputs at some level (of which they can't really when they create it in SOP network).

MustafaJafar:
> I tested with the latest commit. I can create multiple outputs in object level but I can't create multiple output in sop level.

it's there and outputs will show up once you add outputs inside the sop hda.  
maybe the plugin should create output nodes to make it easier ? 

> we should stick with the HDA editing workflow and the max/min outputs should be updated through the operator properties. Users can customize their own outputs/inputs by accessing the properties editor.

I don't think exposing these settings in creator conflict with the properties editor.
users still have the full control over their digital assets. 
the creator just aims to save them few clicks.

There is some limitation though. Not all Houdini levels support having min/max. 
e.g. Object level sub-networks have one output. because all object level nodes have just 1 output which is transform! 

So, removing min/max would help avoid such situation. 
Or, I can consider these limitations and write about them in docs / tooltip in creator / Houdini scene view text. 
it can be a matter of UX. 

What do you recommend guys ?

moonyuet:
> > I tested with the latest commit. I can create multiple outputs in object level but I can't create multiple output in sop level.
> 


> it's there and outputs will show up once you add outputs inside the sop hda.
> maybe the plugin should create output nodes to make it easier ?

Maybe I dont explain quite well. I mean that we only have one output node inside the digital asset even though we edit the maximum outputs to more than one. 
This is what inside the digital assets
image
The maximum outputs are 2, it is still always one output.
image
The manual way to fix that is to add the output node under the Boolean node I wanted to output for hda
If we are doing it with the setting, it would be difficult to count which node to be targeted to hook at these output nodes for our hda nodes.

fabiaserra:
I think knowing the input/outputs that you want to expose it's something you realize later on while developing the HDA and not on first creation so I think the best is to just let them edit it like all the other HDA settings that you can control through Houdini type properties

MustafaJafar:
Hey I was checking the hda extractor and I met some problem. I'd appreciate any help. 

problem: I found out that the published HDAs have the menu entry location `Digital Assets` although the location has been set  using 

https://github.com/ynput/ayon-core/blob/4f6a75720963638f45ef4f3a62511f8009343ae1/client/ayon_core/hosts/houdini/plugins/create/create_hda.py#L109-L112

I furtherly explored it. I was able to re-produce the problem _which is not AYON related._ 
I assume the problem occur because the above code is setting the menu entry location in the opened Houdini session but it doesn't save the new location in the HDA itself.
It doesn't work until I open the HDA type properties and hit `Apply`. 

---

You can use the following script to reproduce the problem.

```python
import hou

node_name = "my_example_a"

parent_node = hou.node("/obj")
to_hda = parent_node.createNode(
                "subnet", node_name="{}_subnet".format(node_name))
hda_node = to_hda.createDigitalAsset(
                name=node_name,
                hda_file_name="$HOUDINI_USER_PREF_DIR/otls/{}.hda".format(node_name),
                ignore_external_references=True
            )
            
hda_node.setName(node_name, unique_name=True)
hda_def = hda_node.type().definition()

tool_name = hou.shelves.defaultToolName(
                hda_def.nodeTypeCategory().name(), hda_def.nodeTypeName())
hda_tool = hou.shelves.tool(tool_name)

hda_tool.setToolLocations(("<<<>>>",))  # "<<<>>>"
print("Setting Location", hda_tool.toolMenuLocations())

locations = hda_tool.toolMenuLocations()

hda_options = hda_def.options()

print("Before Setting Options", hda_tool.toolMenuLocations())
hda_options.setSaveInitialParmsAndContents(True)
next_version = "2"
hda_def.setVersion(str(next_version))
hda_def.setOptions(hda_options)
print("After Setting Options", hda_tool.toolMenuLocations())  # "Digital Assets"

hda_tool.setToolLocations(("<<<>>>",))  # "<<<>>>"
print("Setting Location", hda_tool.toolMenuLocations())
print("Before Save", hda_tool.toolMenuLocations())  # "<<<>>>"
hda_def.save(hda_def.libraryFilePath(), hda_node, hda_options)
print("After Save", hda_tool.toolMenuLocations())  # "Digital Assets"

# Set the locations again as saving the HDA resets the Menu entry locations.
hda_tool.setToolLocations(locations)
print("Setting Location", hda_tool.toolMenuLocations())  # "<<<>>>"


"""
Note that the above code works but the menu location is not saved.
So, when closing and opening Houdini again the menu location 
will be the default "Digital Assets"
"""
```

MustafaJafar:
This commit https://github.com/ynput/ayon-core/pull/456/commits/8f1dd6acf6f16f2f1f608fced6be7e416de747ee adds support for loading non object level HDAs. 

One known problem that I couldn't solve yet is loading HDAs that were published from VOPs. 
Because VOPs (VEX Builder) level in Houdini exists in different places ([Vex Contexts](https://www.sidefx.com/docs/houdini/hom/hou/VexContext.html))
e.g.
- Material Network
- Sops e.g. pointsop
- Cop2
- Chop

So, the problem is how to find the vex context 
After some search and testing I found out the following snippet.
But, soon after my commit I found out that it's used with the parent node of the HDA not the HDA itself.
```python
vex_context = hou.vexContextForNodeTypeCategory(hda_def.nodeTypeCategory())
```
Could I save arbitrary data in my published product ? I think I can add the vex context on publishing.

client/ayon_core/hosts/houdini/plugins/load/load_hda.py
```diff
@@ -3,10 +3,28 @@
...
+def get_avalon_container():
+    path = pipeline.AVALON_CONTAINERS
+    avalon_container = hou.node(path)
+    if not avalon_container:
+        # Let's create avalon container secretly
+        # but make sure the pipeline still is built the
+        # way we anticipate it was built, asserting it.
+        assert path == "/obj/AVALON_CONTAINERS"
+
+        parent = hou.node("/obj")
```
moonyuet, line 21: if we can support publishing from non object level, maybe it is also good idea to support digital asset loading in non-object network>

MustafaJafar: > maybe it is also good idea to support digital asset loading in non-object network?

I had this thought too and also loading in the current opened network.
But, currently, artists expects every single loaded product to exist inside avalon container subnetwork.

moonyuet: Maybe we can discuss later if there are some user needs, since HDA is mostly a package of the nodes to be used in external package.

client/ayon_core/hosts/houdini/plugins/publish/extract_hda.py
```diff
@@ -26,6 +26,14 @@ def process(self, instance):
         hda_def.setOptions(hda_options)
         hda_def.save(hda_def.libraryFilePath(), hda_node, hda_options)
 
+        if instance.data["creator_attributes"].get("use_project"):
+            # Set TAB Menu location interactively
+            # This shouldn't be needed if the Tool Location is saved in the HDA.
+            tool_name = hou.shelves.defaultToolName(
+                hda_def.nodeTypeCategory().name(), hda_def.nodeTypeName())
+            hou.shelves.tool(tool_name).setToolLocations(
+                ("AYON/{}".format(self.project_name),))
```
moonyuet, line 35: Encounter this error with this line of code.
```
  File "C:\Users\Kayla\AppData\Local\Ynput\AYON\dependency_packages\ayon_2402141620_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "D:\ayon-core\client\ayon_core\hosts\houdini\plugins\publish\extract_hda.py", line 35, in process
    ("AYON/{}".format(self.project_name),))
AttributeError: 'ExtractHDA' object has no attribute 'project_name'
```
Consider this to fix the bug.
```suggestion
            project_name = instance.context.data["projectName"]
            hou.shelves.tool(tool_name).setToolLocations(
                ("AYON/{}".format(project_name)))
```

moonyuet:
It can publish hda successfully in SOP Level
image

But when I load the hda, it doesn't save the node of which I intended to publish as expected 

When I publish some hda from dop network, it also publishes successfully but it gives me an error.
image

MustafaJafar:
> But when I load the hda, it doesn't save the node of which I intended to publish as expected

Could you provide more info. because I couldn't replicate it.

> When I publish some hda from dop network, it also publishes successfully but it gives me an error.

I couldn't replicate your error.
In your screenshot. the log mentioned that the following line errored because of  wrong type. 
The type here is `name` variable that's passed as argument to load method.
I wonder what is the problem.

https://github.com/ynput/ayon-core/blob/1b70c0a6a4bd08815ab680549b7dd99f199c85ec/client/ayon_core/hosts/houdini/plugins/load/load_hda.py#L59

image

client/ayon_core/version.py
```diff
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 """Package declaring AYON core addon version."""
-__version__ = "0.3.1-dev.2"
+__version__ = "0.3.2-dev.2"
```
moonyuet, line 3: it seems that this is not the commit for this PR.

MustafaJafar: This PR updates default values of `ayon+settings://core/tools/creator/product_name_profiles`

iLLiCiTiT:
```suggestion
__version__ = "0.3.1-dev.2"
```

client/ayon_core/hosts/houdini/plugins/create/create_hda.py
```diff
@@ -68,21 +90,103 @@ def create_instance_node(
...
         self.customize_node_look(hda_node)
+
+        # Set Custom settings.
+        hda_def = hda_node.type().definition()
+
+        if pre_create_data.get("use_ayon_icon"):
+            hda_def.setIcon(get_ayon_icon_filepath())
+
+        if pre_create_data.get("set_user"):
+            hda_def.setUserInfo(getpass.getuser())
```
iLLiCiTiT, line 107: Replace `import getpass` with `from ayon_core.lib import get_ayon_username`
```suggestion
            hda_def.setUserInfo(get_ayon_username())
```

package.py
```diff
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "0.3.2-dev.1"
+version = "0.3.2-dev.2"
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-core/pull/456/commits/888c4ff35434588fc28cc468753a1ac9ce05001f

client/ayon_core/version.py
MustafaJafar: https://github.com/ynput/ayon-core/pull/456/commits/888c4ff35434588fc28cc468753a1ac9ce05001f

MustafaJafar:
In https://github.com/ynput/ayon-core/pull/456/commits/17ec19f5069860be184733254db79f4a86308de5
Shelf locations are now persistent on the HDA and no need to set it interactively on publishing or on loading. 
we only need to set it on creation. 
many thanks to @fabiaserra for his tip. 
image

server/settings/tools.py
```diff
@@ -448,6 +448,17 @@ class GlobalToolsModel(BaseSettingsModel):
...
+            {
+                "product_types": [
+                    "hda"
+                ],
+                "hosts": [
+                    "houdini"
+                ],
+                "task_types": [],
+                "tasks": [],
+                "template": "{folder[label]}_{variant}"
```
BigRoy, line 461: Is it really a good idea to use the folder label in a product name template since it's allowed for a folder label to have characters that are not allowed on a product name, like brackets or whatever.

server_addon/houdini/client/ayon_houdini/plugins/create/create_hda.py
```diff
@@ -68,21 +94,100 @@ def create_instance_node(
...
+        # point to the same hda_file_name. But, their node names will
+        # be incremented.
+        hda_node.setName(node_name, unique_name=True)
         self.customize_node_look(hda_node)
+
+        # Set Custom settings.
+        hda_def = hda_node.type().definition()
+
+        if pre_create_data.get("use_ayon_icon"):
+            hda_def.setIcon(get_ayon_icon_filepath())
```
BigRoy, line 108: Does this mean the icon path is hardcoded into the HDA? Or does this embed the icon as PNG data into the HDA definition?

MustafaJafar: It means that if some user wanted to use the ayon icon, they can toggle that option instead of searching for the ayon icon location and add it manually in the hda type properties.

server/settings/tools.py
MustafaJafar: Oh, I thought the folder label is equivalent to `folder_path.split("/")[-1]`.

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
```diff
@@ -18,38 +39,44 @@ class HdaLoader(plugin.HoudiniLoader):
...
-        self[:] = [hda_node]
+        # Imprint it manually
+        data = {
+            "schema": "openpype:container-2.0",
+            "id": AVALON_CONTAINER_ID,
+            "name": node_name,
+            "namespace": namespace,
+            "loader": str(self.__class__.__name__),
+            "representation": context["representation"]["id"],
+        }
```
moonyuet, line 73: Just wondering why do we imprint the data here instead of using `containerise` function? 
if it is just adding attributes with the data into the nodes, guess we can improve the `containerise` function instead of implementing the repetitive function?

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
```diff
@@ -18,38 +39,44 @@ class HdaLoader(plugin.HoudiniLoader):
...
+        hda_node.moveToGoodPosition()
 
-        self[:] = [hda_node]
+        # Imprint it manually
+        data = {
+            "schema": "openpype:container-2.0",
+            "id": AVALON_CONTAINER_ID,
+            "name": node_name,
+            "namespace": namespace,
+            "loader": str(self.__class__.__name__),
```
moonyuet, line 71:
```suggestion
            "loader": self.__class__.__name__,
```

server/settings/tools.py
antirotor: image

server_addon/houdini/client/ayon_houdini/plugins/create/create_hda.py
BigRoy: I understood that - I was just asking whether the HDA would store it as `/path/to/ayon_icon.png` or whether it would be an embedded PNG data? Because I believe an HDA can do both.

server/settings/tools.py
BigRoy: This would have to be `folder["name"]` to just get `sh010`
If you want the full unique folder path then: `folder["path"]` but note that that includes the `/` like `/shots/sh010`.

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: yeah, I had to do it because each HDA type needs to be loaded in a dedicated place. 

> Which makes me wonder a bit whether it's still worth it to put it inside that container group instead of trying to put it in an active matching panel or alike.

It's fair to support that. but I think it's something that's needs to be supported for other products too.
Maybe, I can add it here. but I'm afraid it becomes too much for this PR.

server_addon/houdini/client/ayon_houdini/plugins/create/create_hda.py
MustafaJafar: As far as I know, it takes a copy and embed it. 
I tested it and I found it's not embedded 😅 
what do you think ? 
image

server/settings/tools.py
MustafaJafar: > image

Omg, what is that! 
let me change it to `folder[name]`

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: good catch! 
It's wired that `str` casting is used in other places in AYON addons although it's already a string.
Thanks!

server_addon/houdini/client/ayon_houdini/plugins/create/create_hda.py
MustafaJafar: 😵 I restarted my Houdini session and then it changed on its own.. but only the one that was published from production bundle? 
image

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
BigRoy: May have been some odd legacy in Py2 where there may somehow have been a difference between `str` and `unicode`, etc?

server_addon/houdini/client/ayon_houdini/plugins/create/create_hda.py
BigRoy: No idea how it works in practice - at least the checkbox to the right is marked.

@fabiaserra do you know when it exactly "embeds" and whether it needs any special sauce to force do that?

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
BigRoy: Agree - likely best for another PR.

server_addon/houdini/client/ayon_houdini/plugins/create/create_hda.py
fabiaserra: not sure when it does... I would expect that as long as you set it and then save it should be embedded if that toggle is on

MustafaJafar: then, I'll just remove it. 
it was not requested anyways. I made because I was excited I knew new things about HDA creation and thought it was cool thing to have.

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: I created an issue for it. and let's move the discussion there.
https://github.com/ynput/ayon-core/issues/650

MustafaJafar:
For reference: 
@BigRoy , your concern here was correct https://github.com/ynput/ayon-core/pull/644#discussion_r1636651849
I found a bug where the loader was loading the wrong HDA def. it happened because I was using the same HDA product name before and after my other PR. 
The solution is to use the last def. https://github.com/ynput/ayon-core/pull/456/commits/71be4e99a7e04356ea3c3912ab9706dbf99b3333

moonyuet (approved):
Tested with houdini successfully and hda loaded correctly with the template name. 
image

server/settings/tools.py
```diff
@@ -448,6 +448,17 @@ class GlobalToolsModel(BaseSettingsModel):
...
+            {
+                "product_types": [
+                    "hda"
+                ],
+                "hosts": [
+                    "houdini"
+                ],
+                "task_types": [],
+                "tasks": [],
+                "template": "{folder[name]}_{variant}"
```
BigRoy, line 461: Wouldn't this technically still make it conflict on folder name? What did this profile solve again? Sorry, I seem to be suffering from short-term memory loss this week.

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
```diff
@@ -1,12 +1,33 @@
...
+    AVALON_CONTAINER_ID
+)
 from ayon_core.pipeline.load import LoadError
 from ayon_houdini.api import (
+    lib,
     pipeline,
     plugin
 )
 
+def get_avalon_container():
```
BigRoy, line 15: I feel like this logic makes more sense in `lib` when taken out of `containerise` into a `get_or_create_avalon_container()` function that `containerise` then also ends up using. That way we ensure the logic remains synced and avoids custom logic here.

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
```diff
@@ -71,4 +93,45 @@ def update(self, container, context):
 
     def remove(self, container):
         node = container["node"]
+        parent = node.parent()
         node.destroy()
+
+        if parent.type().category() == hou.objNodeTypeCategory():
+            return
```
BigRoy, line 100: What does this solve? :) When would this happen?

server/settings/tools.py
MustafaJafar: It doesn't solve anything it's just a regular request. 
I used folder name instead of folder label following onderj's screenshot https://github.com/ynput/ayon-core/pull/456#discussion_r1638156081

> image

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
```diff
@@ -1,12 +1,33 @@
...
 )
 
+def get_avalon_container():
+    path = pipeline.AVALON_CONTAINERS
+    avalon_container = hou.node(path)
+    if not avalon_container:
+        # Let's create avalon container secretly
+        # but make sure the pipeline still is built the
+        # way we anticipate it was built, asserting it.
+        assert path == "/obj/AVALON_CONTAINERS"
```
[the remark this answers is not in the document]
MustafaJafar: I'm exposed 😅 
I copied pasted that from 
https://github.com/ynput/ayon-core/blob/506c9975af1237635827fdc45459ab57c4e2fed1/server_addon/houdini/client/ayon_houdini/plugins/load/load_image.py#L27-L30

OMG. 
image

server/settings/tools.py
BigRoy: Why is it requested? What does the request solve? Seems like an odd default setting to me if it's not there to avoid a certain issue but is a specific studio request? :)

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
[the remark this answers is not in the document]
BigRoy: Nice! :) Even more reason to move that to `lib`. And what do you know it - that logic likely originated from me! 👿 😮

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: This line skip removing the avalon container itself.
this will happen when removing a loaded HDA of type Object while there are no loaded products. 
Therefore, The parent of the Obj HDA will be the Avalon container itself.

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: agreed. 
Also, for reference this logic is found in load image
https://github.com/ynput/ayon-core/blob/506c9975af1237635827fdc45459ab57c4e2fed1/server_addon/houdini/client/ayon_houdini/plugins/load/load_image.py#L24-L35

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
BigRoy: Can't we be more explicit and just skip deletion if it's explicitly that particular case? Like if `parent.path() == lib.AVALON_CONTAINER`?

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: ehm, should it exist in lib or pipeline

server/settings/tools.py
BigRoy: > tbh, supporting the folder name was not particularly requested, however, it's odd that the folder name exists in many templates but not HDA.

It works in the publish filename templates by default, just not the product name itself which by default does not get this data indeed. If it's truly not needed for the HDA logic to work I'd remove it from this PR.

If there's a 'generic' need for `{folder[name]}` to be accessible for product name profiles by default then likely we're better off solving this differently so that it works across the board instead of fixing a particular HDA only case?

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
BigRoy: I'd say `pipeline` :)

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: Thank you https://github.com/ynput/ayon-core/pull/456/commits/320f97c9c68ff5acd220c5da8f632258f935c2dd

server_addon/houdini/client/ayon_houdini/plugins/publish/validate_subset_name.py
```diff
@@ -26,7 +25,8 @@ class ValidateSubsetName(plugin.HoudiniInstancePlugin,
 
     """
 
-    families = ["staticMesh"]
+    families = ["staticMesh", "hda"]
+    hosts = ["houdini"]
```
BigRoy, line 29: Redundant because `plugin.HoudiniInstancePlugin` already sets it.
```suggestion
```

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
MustafaJafar: thank you https://github.com/ynput/ayon-core/pull/456/commits/336d38b21835f02bd04b5a2a4e317d07e5bb94b1

server_addon/houdini/client/ayon_houdini/api/pipeline.py
```diff
@@ -401,6 +397,23 @@ def _enforce_start_frame():
...
+        # but make sure the pipeline still is built the
+        # way we anticipate it was built, asserting it.
+        assert path == "/obj/AVALON_CONTAINERS"
+
+        parent = hou.node("/obj")
+        avalon_container = parent.createNode(
+            "subnet", node_name="AVALON_CONTAINERS"
+        )
+
+    return avalon_container
```
BigRoy, line 414: How about avoiding the assert but just relying on `AVALON_CONTAINERS` completely?
And how about also refactoring to `get_or_create_avalon_container` to be explicit about what the function does?
```suggestion
def get_or_create_avalon_container() -> "hou.OpNode":
    avalon_container = hou.node(AVALON_CONTAINERS)
    if avalon_container:
        return avalon_container

    parent_path, name = AVALON_CONTAINERS.rsplit("/", 1)
    parent = hou.node(parent_path)
    return parent.createNode(
        "subnet", node_name=name
    )
```

server_addon/houdini/client/ayon_houdini/plugins/load/load_hda.py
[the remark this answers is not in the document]
MustafaJafar: Does this commit resolve this conversation? 
https://github.com/ynput/ayon-core/pull/456/commits/0ddfaa5b91a2a6d35dec5f6df70bb6094db582a7

server/settings/tools.py
MustafaJafar: I don't know about the `generic` need. maybe it's a question for @antirotor or @iLLiCiTiT 
---
> If it's truly not needed for the HDA logic to work I'd remove it from this PR.

We still need a default template profile for HDAs. maybe let's stick with the default `{product[type]}{variant}`
and we can wait till some complains about using `{asset}` or `{folder[name]}`

---
Also, I've just discovered that `{folder[name]}` key is being replaced with `{asset}` therefore the creator prints an error message. 
```
    raise TemplateFillError(
ayon_core.pipeline.create.product_name.TemplateFillError: Value for 'asset' key is missing in template '{asset}{variant}'. Available values are {'variant': 'Main', 'family': 'hda', 'task': {'name': 'cfx', 'type': 'CFX'}, 'product': {'type': 'hda'}, 'folder': {'label': 'Moco', 'name': 'moco'}}
```
image

BigRoy: > We still need a default template profile for HDAs

Could you explain why we need it? Why wouldn't the normal default publish template work for HDAs? Again, what does the custom template solve?

MustafaJafar: > Could you explain why we need it? Why wouldn't the normal default publish template work for HDAs? Again, what does the custom template solve?

From a chat with @antirotor, 
```
We will probably need to create new product name template for those HDA so they will have more meaningful names in this context.
```

BigRoy: Thanks. I still feel like I'm missing the crucial bit of the puzzle. Is it the case that the filename defines the HDA node name or type name or whatever and that's why we care so much about the product naming specifically for the HDA compared to other products?

I think I'm lacking the info on why there's this need to differ for published HDAs.

MustafaJafar: Personally, I think it's about the HDA name itself. e.g. in this tab menu, I'd prefer to prefix my HDA names with folder name to make it easy for me to identify them and I don't like the default `hda` prefix. 
image

But, It's better to leave it question for @antirotor.

moonyuet (approved):
Tested successfully in Houdini. Are we gonna merge this ticket or we need some more adjustment upon?

MustafaJafar:
> Tested successfully in Houdini. Are we gonna merge this ticket or we need some more adjustment upon?

As far as I know, the PR is fine. Although there's a conversation I didn't resolve because I don't know what is recommended.

Other than that, the PR works fine on my side.

server/settings/tools.py
MustafaJafar: Mentioning @moonyuet for reference.

iLLiCiTiT:
Where is version bump?

Issue #152, opened 2024-03-06
Bug: 3dsmax - Manage / Set Version malfunctioning


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When loaded asset (max scene loader) in particular version in the active workfile and user wanting to change the ```version``` to any other existing it ```Set Version``` action perform the change but also produce duplicate of the asset leading to unwanted objects appearing in the scene.

<img width="811" alt="Screenshot 2024-03-06 160038" src="https://github.com/ynput/ayon-core/assets/112623825/6a949930-98e2-4a7d-b1a8-2b35261445c4">

### Expected Behavior:

Manage functions should work as expected and not producing any unwanted entities in the active workfile.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1.Load any Asset like model (maxScene Loader)
2. Manage> Set Version (change version to other)
3. If all fine, load another Asset and perform Manage version again
4. Your scene will have unwanted duplicates of assets (as on the screenshot above)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

[cuID:[AY-3686](https://app.clickup.com/t/86bxv7z29)]

PR #461, opened 2024-04-29
Max: Fix the malfunctioning issue in Max Scene loader

moonyuet:
## Changelog Description
Fix https://github.com/ynput/ayon-core/issues/152

## Additional info
Please test with different loaders.

## Testing notes:
1. Launch Max
2. Load the asset with the same name
3. Load the same asset again
4. And set versions for the assets
5. Should update and load accordingly.


moonyuet:
> I have run several tests and all seems working fine.
> 
> Loading multiple assets and also the same ones multiple times and updating / setting its versions.
> 
> All fine.

have you tried with different loaders?

PR #466, opened 2024-04-30
Maya: Preserve time units on load and update with `MayaUSDReferenceLoader`

BigRoy:
## Changelog Description

Preserve time units on load and update with `MayaUSDReferenceLoader`.

Without the current frame, playback range and FPS may be altered by the loading/updating. This ensures to preserve the state of the local scene.

## Additional info

I also tried using the `optionVar` of the import options to disable 'importing' the FPS or frame ranges from the loaded file using option vars `importOptionsImportFramerate` and `animRangeOptDefault` (also see e.g. `C:\Program Files\Autodesk\Maya2024\scripts\others\fileOptions.mel`) but those didn't seem to affect referencing so I ended up more explicit context managers instead.

## Testing notes:

1. Loading or updating a USD file with `Reference Maya USD` loader should not change timeline / current frame. 

Without this PR it usually sets e.g. the start frame to 0 if there's no frame range in the loaded USD file.


moonyuet (approved):
Tested in Maya and it preserve time units in the maya scene.

mkolar (approved):

BigRoy:
With this now part of `ayon_maya` - should a PR like this bump maya addon version?

antirotor (approved):

iLLiCiTiT:
> With this now part of ayon_maya - should a PR like this bump maya addon version @iLLiCiTiT ?

It should.

PR #469, opened 2024-04-30
Max: Fix the error encountered when updating the loaded assets

moonyuet:
## Changelog Description
Fix the error encountered when updating the loaded assets.

## Additional info
n/a

## Testing notes:
1. Launch Max
2. Load any assets with fbx & usd & obj & poincache ornatrix loaders
3. Setting version with these assets
4. Should be updated successfully


LiborBatek (approved):
I can confirm that it works for ```FBX``` and ```OBJ``` and that I was able to set and switch versions when multiple asset versions loaded and no any issues were found.

I m not able to publish ```USD``` for some reason and its giving me some errors (not related to this PR)

So I couldnt test with ```USD``` atm.

PR #480, opened 2024-05-03
Creators: fix undefined task name

antirotor:
## Changelog Description
When there was no `taskEntity` passed, task name was filled correctly but if not task name was `None`, passed down the line and causing crashes. This change is trying to find `taskEntity` and fill task name properly.

## Testing notes:
Difficult to test as it depends on specific creator login at the end. But running something like this in Maya script editor should test it:

```python
import ayon_api
from ayon_core.pipeline.create import CreateContext


host = registered_host() 
context = CreateContext(host)
project_name = host.get_current_project_name()
folder_path = "/path/to/your/test/folder"
task_name = "layout"

creator_identifier = f"io.openpype.creators.maya.layout"
task = ayon_api.get_task_by_name(project_name, folder_entity["id"], task_name)

context_set = context.create(
            creator_identifier,
            variant="Main",
            folder_entity=ayon_api.get_folder_by_path(project_name, folder_path),
            task_entity=ayon_api.get_task_by_name(project_name, folder_entity["id"], task_name),
            pre_create_data={"use_selection": False}
)
```


client/ayon_core/pipeline/create/context.py
```diff
@@ -2022,6 +2026,7 @@ def create(
             "productType": creator.product_type,
             "variant": variant
         }
+        print("Create instance data", instance_data)
```
BigRoy, line 2029:
```suggestion
```

client/ayon_core/pipeline/create/context.py
```diff
@@ -1987,12 +1987,16 @@ def create(
...
-        task_name = None
         if task_entity is None:
             task_name = self.get_current_task_name()
-            task_entity = ayon_api.get_task_by_name(
-                project_name, folder_entity["id"], task_name
-            )
+            if task_name:
+                task_entity = ayon_api.get_task_by_name(
+                    project_name, folder_entity["id"], task_name
+                )
```
BigRoy, line 1995: Why do we need the `if task_name` logic suddenly? Isn't there **always** a current task name?

Also, I wonder if it makes sense to rename the `task_name` variable here to `current_task_name` to avoid confusion with the `task_name` variable further down below.

client/ayon_core/pipeline/create/context.py
```diff
@@ -1987,12 +1987,16 @@ def create(
...
+
+        task_name = None
+        if task_entity:
+            task_name = task_entity["name"]
```
BigRoy, line 1999: Would this be clearer?
```suggestion
        task_name = task_entity["name"] if task_entity else None
```

BigRoy: Or even better yet - thus set that directly [here](https://github.com/ynput/ayon-core/pull/480/files#diff-aef2d4c691a5a8a08e92e98a60998b7636c34882b238c8805b45f34c8ffd7cf5R2025) so that `"task": task_entity["name"] if task_entity else None` to make it clearer it's only used there?

client/ayon_core/pipeline/create/context.py
iLLiCiTiT: > Why do we need the if task_name logic suddenly? Isn't there always a current task name?

No there is not. E.g. in traypublisher is set only project.

> Also, I wonder if it makes sense to rename the task_name variable here to current_task_name to avoid confusion with the task_name variable further down below.

Yes, could be.

iLLiCiTiT (approved):

Issue #485, opened 2024-05-03
Houdini: support `get_publish_families` method


MustafaJafar:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Support `get_publish_families` similar to Maya.

### How would you imagine the implementation of the feature?

It's already implemented as part of this PR https://github.com/ynput/ayon-core/pull/295 
But, not made in a separate PR.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[AY-5225](https://app.clickup.com/t/86bymf6uk)]

MustafaJafar:
For reference, I came across this enhancement when drafting a first iteration of model family in Houdini.
without this enhancement the code logic in my case will be interlaced.

moonyuet:
Just my two cents.
For the long term, it is better to have `self.published_families` since we would implement the custom node(i.e. the universal node) for publishing. If we have published families as method, it would help us to build that node as it allows users to decide which families to publish their instances. Although I am not quite sure about the roadmap of the universal node, I personally think that the `get_published_families  implementation helps to build the structure of the custom node a little bit.

BigRoy:
This has now been merged with #295 - I'll close it but feel free to re-open and transfer to `ayon-houdini` if some is still relevant.

PR #486, opened 2024-05-03
Houdini: add model family (initial support)

MustafaJafar:
## Changelog Description
Adding model family to Houdini. 
The main purpose to support Maya (load/publish model from maya to/from houdini).

## Additional info
Currently, The new creator plugin is almost the same as 'create_pointcache.py'
But renaming the product type to 'model'.
It's considered to support multiple representations in the future.

## Additional info 2
This PR uses `collect_instances_type.py` to `abc` family to publish families of `Model` instances.

## Testing notes:
1. Create Model Family
2. Publish Local and on farm.


client/ayon_core/hosts/houdini/plugins/create/create_model.py
```diff
@@ -0,0 +1,144 @@
...
+
+class CreateModel(plugin.HoudiniCreator):
+    """Create Model"""
+    identifier = "io.openpype.creators.houdini.model"
+    label = "Model"
+    product_type = "model"
+    icon = "cube"
+
+    def get_publish_families(self):
+        return ["model", "abc"]
```
BigRoy, line 29: I wonder whether this is the best solution here. @iLLiCiTiT thoughts?
Or should we do the reverse by e.g. adding the family `model` where we want it to apply instead, similar to e.g. [this](https://github.com/ynput/ayon-core/pull/212/files#diff-6543b87b81db3323d73f825c3cd13d657922b73f812c71b777dcd91ee869630eR82-R90)?

moonyuet: > Or should we do the reverse by e.g. adding the family model where we want it to apply instead, similar to e.g. [this](https://github.com/ynput/ayon-core/pull/212/files#diff-6543b87b81db3323d73f825c3cd13d657922b73f812c71b777dcd91ee869630eR82-R90)?

I personally prefer this, and I applied similar ways in Max host to avoid unnecessary duplicates of the codes. For this case, we can consider doing that if it is just for publishing pointcache model in abc format, as I can see that most of the codes are similar to the one in the creator. Guess we can just refactor the pointcache creator code a little bit, (e.g. put those parameters which sets abc to be static or dynamic into separate function). and the model creator inherited from it. 
But anyways, I am not sure about the scope for Houdini and there might be a reason why applying `get_published families` instead of having child class to inherit from the similar extractor

MustafaJafar: Let me add my 2 cents, please correct me if I'm wrong. 

I think by adding the following snippet to `extract_alembic.py` , it should replace `extract_model.py` (it doesn't exist.)

The extractor will work but it'd be redundant because having `abc` in families is sufficient.
Also, collectors and validators associated with alembic won't be triggered. 

```python
class ExtractModelABC(ExtractAlembic):
    """Extract model as ABC."""

    label = "Extract Alembic (Model)"
    families = ["model"]

    # Driven by settings
    optional = True
```

But, the current solution to add `abc` to families is by using such a collector. 
tbh, I find it redundant if we are capable of specifying publish families inside the creators. 
e.g. in alembic creators, it would be `["pointcache", "abc"]` and in bgeo `["pointcache", "bgeo"]`

https://github.com/ynput/ayon-core/blob/e1f1556ce2c6ab5a15fda488f6af7605ca875867/client/ayon_core/hosts/houdini/plugins/publish/collect_pointcache_type.py#L9-L21


and for model creator. `["model", "abc"]` should be sufficient.

BigRoy: Correct - but there is a current major disadvantage to that approach.

The **new publisher** will only show instance attribute definitions (the editable attributes) for the `product_type` set by the Creator, not by the publish families. So it will show ONLY those for the `model` family, not for the `Extractor` that is targeting only the `abc`.

In this sense the instance attribute definitions (currently) also aren't 'dynamic' per instance, but solely based on the Creator's `product_type` attribute. So if say, the creator would have an option to enable/disable the `abc` family (to enable/disable the alembic export) there would be no way to on toggle or alike update whether the Alembic export options would show or now show on the instance either (during the Create Context stage before publishing).

---

By separating it out into a separate plug-in for the extractor you also get the potential benefit that in settings you can easily specify different defaults in settings for the instance attribute definitions for both the Alembic Extractor (for pointcache) and the Alembic Extractor (for `model` family) separately.

Whether there are better "designs" that are currently possible in AYON or whether there are better ways to design the system so that it's both easy to maintain and easily customizable I'm not sure. I'd have to leave that to @dee-ynput and @iLLiCiTiT most likely.

moonyuet (commented):
it errors out `integrate.py` after exporting the object.
image

MustafaJafar:
> it errors out `integrate.py` after exporting the object.

Yes, the PR needs small addition to work. I added that info to additional info in the PR description.

I can't decide what solution is favored: 
- solution one: implement `get_publish_families ` see https://github.com/ynput/ayon-core/issues/485
- solution two: remove  `get_publish_families `  and add/modify `collect_pointcache_type.py` to add `abc` to families associated with model instances.

Also, this comment https://github.com/ynput/ayon-core/pull/486#discussion_r1591566845 should be considered because it affects the execution flow of publishing models.

moonyuet:
> > it errors out `integrate.py` after exporting the object.
> 
> Yes, the PR needs small addition to work. I added that info to additional info in the PR description.
> 
> I can't decide what solution is favored:
> 
> * solution one: implement `get_publish_families ` see [Houdini: support `get_publish_families` method #485](https://github.com/ynput/ayon-core/issues/485)
> * solution two: remove  `get_publish_families `  and add/modify `collect_pointcache_type.py` to add `abc` to families associated with model instances.
> 
> Also, this comment [#486 (comment)](https://github.com/ynput/ayon-core/pull/486#discussion_r1591566845) should be considered because it affects the execution flow of publishing models.

If you talk about the solution B, you can consider the code below for the model creator. You just need to make sure the identifier is neither `io.openpype.creators.houdini.pointcache` nor `io.openpype.creators.houdini.bgeo`. And you can add one more condition to add abc as part of instance.data["families"] if identifier is "io.openpype.creators.houdini.model". Of course, the collector needs better naming after this (maybe rename it as `collect_instance_types.py` instead of `collect_pointcache_type.py`)

```python
class ExtractModelABC(ExtractAlembic):
    """Extract model as ABC."""
    identifier = "io.openpype.creators.houdini.model"
    label = "Extract Alembic (Model)"
    families = ["model"]

    # Driven by settings
    optional = True
```

MustafaJafar:
> For solution one, I slightly look at how the creator built in the maya host, guess the published families somehow need to be decided during creating the instance. not sure if we can do some tricks on settings(some multiselection one) to let user decide which families type they want to publish?(not sure if it is a good idea.)

**suppose we are allowed to publish multiple representation.** 
Let's take this example I think it can help us: 
In publisher UI I'm able to select publishing my model as (we can select them all)
- alembic
- fbx
- usd

For development,
we'd need to add families based on my selection. 
So, does `get_publish_families` support such a scenario ? maybe @BigRoy can tell us ? 

Personally, I'd prefer to use `get_publish_families` because the logic would be saved in one place which is easier to maintain.
That logic only specifies which publish families to use based on the creator attributes.

MustafaJafar:
In this https://github.com/ynput/ayon-core/pull/486/commits/475d1db69bc518522b57f270f174799842f6d031 I've added @moonyuet's suggestion about using a collector to add `abc` to families. 

I didn't update extractor because it already runs on `Model` instances as they already have `abc` in `families` list.

https://github.com/ynput/ayon-core/blob/d51e7a9423d17d24022b38dfc7648ab3cb7d74f4/client/ayon_core/hosts/houdini/plugins/publish/extract_alembic.py#L11-L16

moonyuet (approved):
General model family publishing looks good
image

client/ayon_core/hosts/houdini/plugins/create/create_model.py
```diff
@@ -0,0 +1,141 @@
+# -*- coding: utf-8 -*-
+"""Creator plugin for creating Model product type.
+
+Note:
+    Currently, This creator plugin is the same as 'create_pointcache.py'
```
fabiaserra, line 5: I think something's wrong on the design if exposing a new type that's basically the same as another requires duplicating all this code...

client/ayon_core/hosts/houdini/plugins/create/create_model.py
```diff
@@ -0,0 +1,141 @@
...
+            "path_attrib": "path",
+            "prim_to_detail_pattern": "cbId",
+            "format": 2,
+            "facesets": 0,
+            "filename": hou.text.expandString(
+                "$HIP/pyblish/{}.abc".format(product_name))
+        }
+
+        if self.selected_nodes:
+            selected_node = self.selected_nodes[0]
```
fabiaserra, line 53: is this intuitive?

client/ayon_core/hosts/houdini/plugins/publish/validate_mesh_is_static.py
```diff
@@ -16,9 +16,13 @@ class ValidateMeshIsStatic(pyblish.api.InstancePlugin,
     """Validate mesh is static.
 
     It checks if output node is time dependent.
+    this avoids getting different output from ROP node when extracted
+    from a different frame than the first frame.
+    (Might be overly restrictive though)
     """
 
-    families = ["staticMesh"]
+    families = ["staticMesh",
```
fabiaserra, line 24: what's the difference between `staticMesh` and `model`?

BigRoy: I believe staticMesh is unreal-specific- but I might be mistaken :/

client/ayon_core/hosts/houdini/plugins/create/create_model.py
MustafaJafar: This is widely adopted in other creator plugins. I'm not justifying following the same idea in this creator. 
But, I just pointing out that it should be fixed in other creator plugins too! 
Therefore it's better discussed in a dedicated issue. 

Would you please create an issue about it so that it can be added to our backlog ?
I'd appreciate any suggested solutions or code snippets to furtherly improve it.

client/ayon_core/hosts/houdini/plugins/publish/validate_mesh_is_static.py
MustafaJafar: [staticMesh](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/houdini/plugins/create/create_staticmesh.py) in Houdini is similar to [staticMesh](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/hosts/maya/plugins/create/create_unreal_staticmesh.py) in maya. 

More info about it here: 
https://ayon.ynput.io/docs/addon_houdini_artist#static-mesh-fbx

client/ayon_core/hosts/houdini/plugins/create/create_model.py
MustafaJafar: I had this thought of product aliases. _I've no idea how to use product aliases in code._
because there's no difference between 'create_pointcache.py' and `create_model.py` other than they have different `product_type`. 

Also, I want to point out that `create_model.py` in this PR is just a first draft. once publishing `multiple representations` is supported in Houdini addon. this creator plugin will be updated!

client/ayon_core/hosts/houdini/plugins/create/create_model.py
fabiaserra: I think most of these ROP nodes allow you passing multiple SOP nodes to export and not just a single one? And you'd expect at least a warning or something telling the user only one of the selected nodes is going to be used. However as I've said before I find the whole AYON/OP create workflow in Houdini very unintuitive so it's not just this small improvement IMO that needs fixing and there needs to be a much bigger redesign. I need to spend some good time trying to see how I can fit our workflow ideas in Houdini in Ayon but I don't really have much time atm as we already have a decently working solution for us

MustafaJafar: > I find the whole AYON/OP create workflow in Houdini very unintuitive so it's not just this small improvement IMO that needs fixing and there needs to be a much bigger redesign. 

Yes, and I agree.. I'd appreciate any suggestions and I'll add them to my notes.

client/ayon_core/hosts/houdini/plugins/publish/validate_export_is_a_single_frame.py
```diff
@@ -0,0 +1,60 @@
...
+    @classmethod
+    def get_invalid(cls, instance):
+
+        invalid = []
+
+        frame_start = instance.data.get("frameStartHandle")
+        frame_end = instance.data.get("frameEndHandle")
+
+        # This happens if instance node has no 'trange' parameter.
+        if (frame_start or frame_end) is None:
```
iLLiCiTiT, line 45:
```suggestion
        if frame_start is None or frame_end is None:
```

antirotor (approved):
Did a model roundtrip between Houdini and Maya without any issue

PR #487, opened 2024-05-06
Chore: Api for thumbnail cache

iLLiCiTiT:
## Changelog Description
Moved cache of thumbnails api to pipeline implementation.

## Additional info
The simple implementation was moved from tools utils to pipeline so it can be used elsewhere without need to import it from tools. The implementation is kept untouched.

## Testing notes:
1. Tools are showing thumbnails as they did.


moonyuet (approved):
show the thumbnails as expected
image
image

PR #488, opened 2024-05-06
Chore: Support adding tools group environment per task level

moonyuet:
## Changelog Description
The PR is to add support of the tools group environment per task level, as the current tools group environment only support for per folder level. If there is empty list of tools group setting up per task level, it will use the tools group per folder level instead if the tools group for the folder level is not empty.

## Additional info
It doesn't work well with Redshift in 3dsmax when both folder and task level has added Redshift as tools group. It will crash 3dsmax during launching it.

## Testing notes:
1. Go to` AYON server -> Studio Settings -> Application addon` to set up your environment for external plugins (e.g. Arnold, Redshift, Yeti)
2. Go to Project -> Select your project
3. Editor -> folder -> task
4. Set up your tools group for your task
image

### DIFF
--- a/server_addon/applications/client/ayon_applications/utils.py
+++ b/server_addon/applications/client/ayon_applications/utils.py
@@ -281,13 +281,15 @@ def prepare_app_environments(
         app.environment
     ]
 
-    folder_entity = data.get("folder_entity")
+    entity = data.get("task_entity")
+    if not entity:
+        entity = data.get("folder_entity")
     # Add tools environments
     groups_by_name = {}
     tool_by_group_name = collections.defaultdict(dict)
-    if folder_entity:
+    if entity:
         # Make sure each tool group can be added only once
-        for key in folder_entity["attrib"].get("tools") or []:
+        for key in entity["attrib"].get("tools") or []:
             tool = app.manager.tools.get(key)
             if not tool or not tool.is_valid_for_app(app):
                 continue
--- a/server_addon/applications/client/ayon_applications/utils.py
+++ b/server_addon/applications/client/ayon_applications/utils.py
@@ -281,28 +281,23 @@ def prepare_app_environments(
         app.environment
     ]
 
-    entity = data.get("task_entity")
-    if not entity:
-        entity = data.get("folder_entity")
+    task_entity = data.get("task_entity")
+    folder_entity = data.get("folder_entity")
     # Add tools environments
     groups_by_name = {}
     tool_by_group_name = collections.defaultdict(dict)
-    if entity:
-        # Make sure each tool group can be added only once
-        for key in entity["attrib"].get("tools") or []:
-            tool = app.manager.tools.get(key)
-            if not tool or not tool.is_valid_for_app(app):
-                continue
-            groups_by_name[tool.group.name] = tool.group
-            tool_by_group_name[tool.group.name][tool.name] = tool
-
-        for group_name in sorted(groups_by_name.keys()):
-            group = groups_by_name[group_name]
-            environments.append(group.environment)
-            for tool_name in sorted(tool_by_group_name[group_name].keys()):
-                tool = tool_by_group_name[group_name][tool_name]
-                environments.append(tool.environment)
-                app_and_tool_labels.append(tool.full_name)
+    if task_entity:
+        groups_by_name, environments, app_and_tool_labels = (
+            get_tool_group_enviornment_by_entities(app, task_entity, groups_by_name,
+                                                   tool_by_group_name, environments,
+                                                    app_and_tool_labels)
+        )
+    elif folder_entity:
+        groups_by_name, environments, app_and_tool_labels = (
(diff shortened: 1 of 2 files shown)


server_addon/applications/client/ayon_applications/utils.py
```diff
@@ -281,13 +281,17 @@ def prepare_app_environments(
         app.environment
     ]
 
+    task_entity = data.get("task_entity")
     folder_entity = data.get("folder_entity")
     # Add tools environments
     groups_by_name = {}
     tool_by_group_name = collections.defaultdict(dict)
-    if folder_entity:
+    if task_entity:
```
iLLiCiTiT, line 289:
```suggestion
    tools = None
    if task_entity:
        tools = task_entity["attrib"].get("tools")

    if tools is None and folder_entity:
        tools = folder_entity["attrib"].get("tools")

    if tools:
        for key in tools:
```

iLLiCiTiT:
You did not add tools attribute to task scope.

iLLiCiTiT:
> You did not add tools attribute to task scope.

Or it is there already?

moonyuet:
> > You did not add tools attribute to task scope.
> 
> Or it is there already?

Guess it is there already as far as I checked from the ayon server which both folder and task includes tools as attributes.,,,,,

antirotor (approved):
In one of my tests I ran into the issue of having env var set when it shouldn't be - coming from the parent folder. Since then, I wasn't able to reproduce but it would be great in the future to have some tests for that. For the sake of completeness I am including the code I've used.

```python
import ayon_api
import os
from ayon_core.pipeline import get_current_context
from ayon_applications.manager import ApplicationManager


app = ApplicationManager()
context = get_current_context()
project_name = context["project_name"]
folder_path = context["folder_path"]
task_name = context["task_name"]
folder = ayon_api.get_folder_by_path(project_name, folder_path)
task = ayon_api.get_task_by_name(project_name, folder["id"], task_name)

folder_tools = folder["attrib"]["tools"]
task_tools = task["attrib"]["tools"]

print(f"Folder tools: {folder_tools}")
print(f"Task tools: {task_tools}")

folder_env = {}
task_env = {}

for tool in folder_tools:
    folder_env.update(app.tools.get(tool).environment)
    folder_env.update(app.tools.get(tool).group.environment)

for tool in task_tools:
    task_env.update(app.tools.get(tool).environment)
    task_env.update(app.tools.get(tool).group.environment)
    
print("-" * 40)
print("Should be set:")
for key, value in sorted(task_env.items()):
    print(f"{key}: {value}")

print(". " * 20)
print("Is set:")
for key, value in sorted(os.environ.items()):
    if key.lower().startswith("test"):
        print(f"{key}: {value}")
        
print("-" * 40)
```

Maybe we should rise application addon minor version for this? @iLLiCiTiT

moonyuet:
> > Maybe we should rise application addon minor version for this? @iLLiCiTiT
> 
> Not minor, but we should raise patch version.

already update the patch version.

LiborBatek:
heh...got merged during my testing. Even tho I werent able to force Ayon load the tools to my liking when in the ```Develop mode```....soo cumbersome

just nevermind tho

server_addon/applications/client/ayon_applications/utils.py
```diff
@@ -281,13 +281,20 @@ def prepare_app_environments(
...
     tool_by_group_name = collections.defaultdict(dict)
-    if folder_entity:
-        # Make sure each tool group can be added only once
-        for key in folder_entity["attrib"].get("tools") or []:
+    tools = None
+    if task_entity:
+        tools = task_entity["attrib"].get("tools")
+
+    if tools is None and folder_entity:
+        tools = folder_entity["attrib"].get("tools")
```
BigRoy, line 294: Could someone explain to me when `tools` would be `None` on the task entity? Shouldn't it either always inherit from the parent folder to begin with OR have explicit overrides? As such, the task attribute for `tools` would always be the actual needed value and there would be no reason to check the folder entity if tools is None?

As such this would more likely be:
```suggestion
    if task_entity:
       tools = task_entity["attrib"].get("tools")
    else:
       tools = folder_entity["attrib"].get("tools")
```

iLLiCiTiT: Folder entity might be optional too.
```python
tools = None
if task_entity:
    tools = task_entity["attrib"].get("tools")
elif folder_entity:
    tools = folder_entity["attrib"].get("tools")
```

BigRoy: In that case the current merged code is already containing a bug since it doesn't use folder entity as an optional value.

BigRoy: I think that's way more readable. 👍

moonyuet: Should I make the commit again?

PR #489, opened 2024-05-06
Blender: Fix deselect all function with context override

simonebarbieri:
## Changelog Description
Extraction of the review in Blender was failing in case of objects not in `OBJECT` mode because the `deselect_all` function didn't implement the context override.

## Testing notes:
1. Set any object in the scene as `EDIT` mode.
2. Try publish a review. It should publish normally.

### DIFF
--- a/client/ayon_core/hosts/blender/api/plugin.py
+++ b/client/ayon_core/hosts/blender/api/plugin.py
@@ -143,13 +143,19 @@ def deselect_all():
         if obj.mode != 'OBJECT':
             modes.append((obj, obj.mode))
             bpy.context.view_layer.objects.active = obj
-            bpy.ops.object.mode_set(mode='OBJECT')
+            context_override = create_blender_context(active=obj)
+            with bpy.context.temp_override(**context_override):
+                bpy.ops.object.mode_set(mode='OBJECT')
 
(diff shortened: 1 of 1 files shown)


client/ayon_core/hosts/blender/api/plugin.py
```diff
@@ -143,13 +143,19 @@ def deselect_all():
...
             bpy.context.view_layer.objects.active = obj
-            bpy.ops.object.mode_set(mode='OBJECT')
+            context_override = create_blender_context(active=obj)
+            with bpy.context.temp_override(**context_override):
+                bpy.ops.object.mode_set(mode='OBJECT')
 
-    bpy.ops.object.select_all(action='DESELECT')
+    context_override = create_blender_context()
+    with bpy.context.temp_override(**context_override):
+        bpy.ops.object.select_all(action='DESELECT')
```
BigRoy, line 152: Out of curiosity - wouldn't just `bpy.ops.object.select_all(action='DESELECT')` work?

moonyuet (approved):
Tested successfully in Blender. I have some additional remark that the frame range setting only works per folder level but not per task level, but it should be resolved in different PR.

PR #491, opened 2024-05-07
Loader: Don't change controller project on close

iLLiCiTiT:
## Changelog Description
Do not change controller's selected project on close of loader window which causes issues.

## Additional info
There is an issue with this approach. When the project change does not happen via project combobox then the combobox does not know that it should be deselected. In case there is no current project (e.g. loader used in tray action) it won't update the expected project and behave like there is no selected project, but it is. After some testing I found out that the change is not needed at all and can be discarted.

### Loader tray action
This is what this PR is fixing.
1. Open loader tool in tray
2. Select a project
3. Close loader tool
4. Open the loader tool again
5. The same project should be selected and it's folders are visible

### Loader in hosts
Validate that we didn't break expected behavior in hosts.
1. Launch host of your choice
2. Open loader tool
3. Change project to library project
4. Close loader tool
5. Open loader tool
6. Current project with current context should be selected


LiborBatek (approved):
I did perform both tests and all working as expected.

All working fine!

PR #492, opened 2024-05-07
Hiero: enhancing effect publishing categorization - AY-5146

tokejepsen:
## Changelog Description
This adds a setting and code for using track names when categorizing the soft effects in NukeStudio/Hiero.

## Testing notes:
1. In NukeStudio add effects to a track item and move those into separate video tracks.
2. In `ayon+settings://hiero/publish/CollectClipEffects/effect_tracks` add a category and track regex that match 1 or more track names.
3. Setup publishing of the track item and the effects should be categorized from the settings.
4. Publish and validate the effects are correct in Nuke.


server_addon/hiero/server/settings/publish_plugins.py
```diff
@@ -27,11 +27,21 @@ def validate_name(cls, value):
         return normalize_name(value)
 
 
+class CollectClipEffectsTracksModel(BaseSettingsModel):
+    _layout = "expanded"
+    name: str = SettingsField("", title="Name")
+    track_regex: str = SettingsField("", title="Track Regex")
```
jakubjezek001, line 33: this should be different type > list[str] 

There will be cases where we will need to hove node classess from multiple tracks to be one product.

jakubjezek001 (approved):
Lets process suggested changes also please include screenshot of settings with example testing configuration. You were using for your testing.

jakubjezek001 (changes requested):
Lets process suggested changes also please include screenshot of settings with example testing configuration. You were using for your testing.

server_addon/hiero/server/settings/publish_plugins.py
tokejepsen: We can target multiple tracks with a single regex. But if we are going down the route of multiple track names, then we might as well be explicit and not use regex, so something like `track_names` list where people can input `creative_look` and `invert_look` for example.

jakubjezek001: yeah lets  just be explicit, then. Important is that there will be multiple track inputs.

tokejepsen: Done

jakubjezek001 (changes requested):
it does not work for me. 

here is my testing scene with two layers. 
image

You can see that it is creating 3 instances but it should really create only 2.
image

jakubjezek001 (approved):
works like a charm

PR #493, opened 2024-05-08
Max: reset scene and clear undo buffer when starting new scene

moonyuet:
## Changelog Description
When user load assets and then start the new scene by `File -> New -> New All`, the scene isn't reset correctly. This PR is to fix this bug so that user can get nice and clean scene after starting new scene.

## Additional info
n/a

## Testing notes:
1. Launch Max
2. Load Asset
3. start the new scene by `File -> New -> New All`
4. Scene should be reset correctly


client/ayon_core/hosts/max/api/pipeline.py
```diff
@@ -52,11 +52,8 @@ def install(self):
 
         self._has_been_setup = True
 
-        def context_setting():
-            return lib.set_context_setting()
-
         rt.callbacks.addScript(rt.Name('systemPostNew'),
-                               context_setting)
+                               on_post_open)
```
BigRoy, line 56: This being used on `new` to me is confusing that it calls a function named `on_post_open`. Why isn't the function named `on_new` or `on_new_scene` or alike?

moonyuet: updated!

moonyuet:
> > When user load assets and then start the new scene by `File -> New -> New All`, the scene isn't reset correctly.
> 
> For sake of the PR description - could you add to the descriptioin what in particular "failed" to reset correctly? What was the actual issue? Etc. It's also unclear _why_ the changes here 'solve' this issue that lacks description currently.

You can try to replicate issue by the following steps in the develop branch.
1. Launch Max
2. Load Asset with AYON loader 
3. File -> New -> New All
4. Once you start the new scene, the viewport still shows up the loaded objects but obviously they are no longer at the scene.

BigRoy:
> 4\. Once you start the new scene, the viewport still shows up the loaded objects but obviously they are no longer at the scene.

Thanks - this line ☝️ should be in the PR description since it explains the actual issue at hand and also should be used in the testing notes to describe what should be checked for.

WHY was this bug there to begin with? I'd expect any DCC that a new scene should clear the scene and not bug out as oddly as this. This makes me curious as to what the difference is between a Max "new scene" and a "reset"? Is this really related to the AYON load logic? Or even if you just create a regular cube does the geometry persist somehow after "new scene"? I feel like we're now obfuscating some other bug.

moonyuet:
> > 4. Once you start the new scene, the viewport still shows up the loaded objects but obviously they are no longer at the scene.
> 
> Thanks - this line ☝️ should be in the PR description since it explains the actual issue at hand and also should be used in the testing notes to describe what should be checked for.
> 
> WHY was this bug there to begin with? I'd expect any DCC that a new scene should clear the scene and not bug out as oddly as this. This makes me curious as to what the difference is between a Max "new scene" and a "reset"? Is this really related to the AYON load logic? Or even if you just create a regular cube does the geometry persist somehow after "new scene"? I feel like we're now obfuscating some other bug.

I know it's weird, but this is somehow related to the part of the maxscript nature. 
I actually take a look at the maxscript docs for fixing this issue: https://help.autodesk.com/view/MAXDEV/2023/ENU/?guid=GUID-92E4150E-0AD4-430C-A852-783A45276D23

I have added the condition for checking unsaved change to avoid some possible errors encountered.

BigRoy:
> I know it's weird, but this is somehow related to the part of the maxscript nature.
> I actually take a look at the maxscript docs for fixing this issue: https://help.autodesk.com/view/MAXDEV/2023/ENU/?guid=GUID-92E4150E-0AD4-430C-A852-783A45276D23

But in your screen recording you're not using any scripting to perform the new file logic? You're literally using their menu entries. 🤯 Haha - I really know too little about 3DsMax.

moonyuet:
> > I know it's weird, but this is somehow related to the part of the maxscript nature.
> > I actually take a look at the maxscript docs for fixing this issue: https://help.autodesk.com/view/MAXDEV/2023/ENU/?guid=GUID-92E4150E-0AD4-430C-A852-783A45276D23
> 
> But in your screen recording you're not using any scripting to perform the new file logic? You're literally using their menu entries. 🤯 Haha - I really know too little about 3DsMax.

I just click to have new scene and it shows up the viewport issue. And it only happened when you have modified scene, and you want to start new scene without saving the modified one.

LiborBatek (approved):
Did test it and it correctly sets the ```New Clean Scene``` properly.

PR #495, opened 2024-05-09
Nuke: Baking write node inherit channel configuration 

moonyuet:
## Changelog Description
This PR is to make sure the color channel from the baking write node aligning with that from the create write node.

## Additional info
n/a

## Testing notes:
1. Launch Nuke
2. Create Write Render Instance
3. Publish
4. The color channel from the baking write node should be aligning with the channels showed in the create write render


jakubjezek001 (approved):
LGTM

PR #497, opened 2024-05-09
Blender: Use transform cache to handle alembic camera updates

simonebarbieri:
## Changelog Description
Use transform cache constraints to update alembic cameras.
Before, the camera was deleted and recreated. Now it updates the cache file.

## Additional info
This PR should be compatible with existing alembic cameras loaded in the scene. When updating, it will create a new transform cache, if it doesn't find one.
To assign the right object path, it has been necessary a workaround to force Blender to evaluate the cache file loaded. This is because the list of object paths in the cache file is not available immediately when the file has been loaded (https://devtalk.blender.org/t/alembic-cachefile-list-of-object-paths-not-available-at-creation/21573).

## Testing notes:
1. Try loading alembic cameras. They should have a Transform Cache in the object constraints.
2. Make changes to the camera (for instance, change field of view). The changes should be kept after updating the camera.
3. Try updating an alembic that has been loaded before this PR. When updating it, it should create the Transform Cache and keep all the changes.

### DIFF
--- a/client/ayon_core/hosts/blender/plugins/load/load_camera_abc.py
+++ b/client/ayon_core/hosts/blender/plugins/load/load_camera_abc.py
@@ -43,7 +43,10 @@ class AbcCameraLoader(plugin.AssetLoader):
     def _process(self, libpath, asset_group, group_name):
         plugin.deselect_all()
 
-        bpy.ops.wm.alembic_import(filepath=libpath)
+        # Force the creation of the transform cache even if the camera
+        # doesn't have an animation. We use the cache to update the camera.
+        bpy.ops.wm.alembic_import(
+            filepath=libpath, always_add_cache_reader=True)
 
         objects = lib.get_selection()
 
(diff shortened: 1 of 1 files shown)


moonyuet (changes requested):
When loading alembic camera, the transform cache modifiers added into the camera as expected.
But there is some issues on updating camera, of which the camera doesn't get updated correctly through scene inventory manager.

https://github.com/ynput/ayon-core/assets/64118225/fb1e9446-b05d-480a-93cd-bee483859122

simonebarbieri:
@moonyuet Could you check if there is any error in the console?

moonyuet (approved):
My bad, I loaded the published alembic from the folder where I performed delete old version before. (I realize after looking at the console)
```
Running callback: <bound method AssetLoader._load of []>
Alembic import of 'D:\sh_boss_proj\interior_lenny\lenny_bedroom\asset\publish\camera\cameraMain\v006\il_asset_cameraMain_v006.abc' took 0.6 ms
Done
Initialising WM_OT_avalon_manager...
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 279, in <lambda>
    lambda: self._show_version_dialog(items))
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 695, in _show_version_dialog
    self._update_containers(items, version)
  File "D:\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, in _update_containers
    update_container(item, item_version)
  File "D:\ayon-core\client\ayon_core\pipeline\load\utils.py", line 514, in update_container
    raise ValueError("Path {} doesn't exist".format(path))
ValueError: Path D:/sh_boss_proj\interior_lenny\lenny_bedroom\asset\publish\camera\cameraMain\v005\il_asset_cameraMain_v005.abc doesn't exist

```
I tested with the published alembic from other tasks, Looks good for me now!

simonebarbieri:
@moonyuet ah not a problem 😄 Thanks for the review!

PR #498, opened 2024-05-09
Chore: Be able to determine if is in AYON launcher process

iLLiCiTiT:
## Changelog Description
Implemented helper function `is_in_ayon_launcher_process` to determine if current process is AYON launcher.

## Additional info
Some functionality should be available only in AYON launcher process but there is not "unified" function that would help to determine that. This function is simplified and can be enhanced in future if we find out there may be other issues.

## Testing notes:
1. Function `is_in_ayon_launcher_process` returns `True` in AYON launcher process
2. Function `is_in_ayon_launcher_process` returns `False` in any other process (e.g. in maya)

```python
from ayon_core.lib import is_in_ayon_launcher_process

print(is_in_ayon_launcher_process())
```

### DIFF
--- a/client/ayon_core/lib/__init__.py
+++ b/client/ayon_core/lib/__init__.py
@@ -139,6 +139,7 @@ from .path_tools import (
 )
 
 from .ayon_info import (
+    is_in_ayon_launcher_process,
     is_running_from_build,
     is_using_ayon_console,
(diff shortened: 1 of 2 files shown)


kalisp (approved):
Tested in Console and Maya, works as advertised.

PR #502, opened 2024-05-10
After Effects: Fix workfiles on launch env var usage

BigRoy:
## Changelog Description

Fix after effects launch logic variable for `AVALON_PHOTOSHOP_WORKFILES_ON_LAUNCH` -> `AVALON_AFTEREFFECTS_WORKFILES_ON_LAUNCH`

## Additional info

See also [default settings](https://github.com/ynput/ayon-core/blob/4dd44b08b78f3a803c0e738841067096a5cea943/server_addon/applications/server/applications.json#L1028) for After Effects application.

Reported on [discord](https://discord.com/channels/517362899170230292/517382145552154634/1238412829145694298).

## Testing notes:

1. Toggling that variable in settings should adjust the behavior whether workfile tool starts on launch of After Effects or not.


iLLiCiTiT:
So many questions...
- Why we use env variable to change the behavior and not settings?
- Why we don't change the env to `AYON` prefix instead of `AVALON`?

BigRoy:
Haha, don't shoot the messenger. But yes, could be a better way for this I think - but in this case consider this just a 'hotfix'.

Whether there's feasibility in how this works to refactor to use settings instead I'll leave to @kalisp to answer who's deeper in this stuff. There could be two options:

- Just get the project settings in the launch logic and based on that decide whether or not to show. (simplest, might be just subtly slower since it requires querying the project settings an extra time)
- Or if there's a very good reason for that to _not_ access the project settings then instead set the environment variable from a prelaunch hook so that at least in settings it's a project toggle instead of an application env setting.

iLLiCiTiT (approved):

iLLiCiTiT:
Approving and merging as hotfix, but it would nice to change it @kalisp ...

BigRoy:
Just want to remark what came up on [Discord](https://discord.com/channels/517362899170230292/517382145552154634/1238422272260440084):

**Alain Xerri (Atomic):**
> I have found a setting, which name seems to imply it can be used instead of the env var (AVALON_PHOTOSHOP_WORKFILES_ON_LAUNCH), but it's just a guess :
>
> Web App-->Studio Settings-->Core-->Tools-->Workfiles-->Open workfile tool on launch-->Hosts

**Jakub Trllo:**
> Looks like that should be used instead...

PR #503, opened 2024-05-10
Chore: Add version dependency

iLLiCiTiT:
## Changelog Description
Updated versions in package.py .

## Additional info
Added AYON launcher version dependency to package.py and use ayon attribute to descrive requested server version.

## Testing notes:
1. Create package of ayon core
2. Upload the package to server
3. Create new bundle and select uploaded core version
4. If AYON launcher is not selected or is selected older version, it should complain that bundle contains incompatible versions

### DIFF
--- a/package.py
+++ b/package.py
@@ -6,7 +6,7 @@ client_dir = "ayon_core"
 
 plugin_for = ["ayon_server"]
 
-ayon_server_version = ">=1.0.3<2.0.0"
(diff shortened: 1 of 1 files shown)


64qam:
still allows me to create a bundle with launcher 1.0.0

64qam:
still not working (still able to use any launcher)

64qam (approved):

PR #505, opened 2024-05-10
Blender: Deadline asset dependencies

simonebarbieri:
## Changelog Description
Implemented deadline file dependencies for Blender. Images and alembic files are set as dependencies.

## Testing notes:
1. Create a render in Blender for a scene that had alembic files (with a Transform Cache modifier) and/or textures.
2. Publish the render to deadline.
3. Check if the job in deadline has the files as dependencies:
image


client/ayon_core/modules/deadline/plugins/publish/submit_blender_deadline.py
```diff
@@ -40,6 +40,17 @@ class BlenderSubmitDeadline(abstract_submit_deadline.AbstractSubmitDeadline,
...
+    @classmethod
+    def apply_settings(cls, project_settings):
+        settings = project_settings["deadline"]["publish"]["BlenderSubmitDeadline"]  # noqa
+
+        # Take some defaults from settings
+        cls.asset_dependencies = settings.get("asset_dependencies",
+                                              cls.asset_dependencies)
+        cls.use_published = settings.get("use_published", cls.use_published)
+        cls.priority = settings.get("priority", cls.priority)
+        cls.group = settings.get("group", cls.group)
```
BigRoy, line 52: This should be redundant and be the default behavior if `apply_settings` is not defined. Right @iLLiCiTiT ?

BigRoy: Also, does this mean that `chunk_size` does not come from settings since it's not defined here? :)

moonyuet: it should be working as expected as there is similar to use `apply_settings` in `ExtractReviewIntermediates` plugin

moonyuet (approved):
The scene has been successfully submitted to the deadline and rendered successfully.

client/ayon_core/modules/deadline/plugins/publish/submit_blender_deadline.py
BigRoy: I'm sure the `apply_settings` works, but by design if you do not implement the method it defaults automatically to what it seems to be doing here currently. As such, I believe it should be completely redundant.

ynbot: If that's the case the issue unrelated to Blender, please don't silently fix it in isolation.
Use the default implementation and create an issue with the problem you discovered / expected behavior.

When in doubt, ask @iLLiCiTiT for advice !

ynbot: Oops, sorry for commenting as Ynbot ^^
Guess who am I 😉

simonebarbieri: Well, from my pov it was more aligning Blender to Maya, not really a silent fix 😅 but I got the message, waiting for some answer from @iLLiCiTiT :)

iLLiCiTiT: Long story short you should not need to override the method.

BigRoy:
The Github listed changes for this PR are crazy - I think something went wrong with a recent merge?

client/ayon_core/modules/deadline/plugins/publish/submit_blender_deadline.py
BigRoy: This comment still applies I believe.

BigRoy (approved):
Didn't test but changes look good to me. I wonder if we should also be bumping the blender addon version since we're introducing a new publish plug-in for it.

simonebarbieri:
@BigRoy just asked and you're right! I was still back at when we needed to bump only when changing setting 😅 fixed that :)

PR #506, opened 2024-05-10
Houdini: Workfile creator fix call to `create_context_node` on create

BigRoy:
## Changelog Description

Fix call to `create_context_node` on create.

## Additional info

The method only exists on the houdini host - not on the plugin itself.

Note that it would actually never error currently because apparently this gets called just beforehand elsewhere:
https://github.com/ynput/ayon-core/blob/57d225009293a3123f02fd2b48e4e38948e0f711/client/ayon_core/hosts/houdini/api/pipeline.py#L176-L178

Which in itself, also generates the node. But at least this fixes the broken code were the issue to arise that the node did not exist at that point in time.

I'd actually want to refactor `HoudiniHost.get_context_data()` to not create a node unless it were 'updating' data. Because now even on querying data it creates a node - which to me feels like the wrong point in time to do so.

## Testing notes:

Publishing should work.


MustafaJafar:
A question that might be irrelevant to this PR: 
When should `create_context_node()` be called ? 
tbh, I thought it's part of `Installing AYON...` 

I add a stupid print statement inside it to find out when this method is triggered ? 
I found out that:
### When it's not called
- when opening a new file while `Skip Opening last workfile` is enabled
- when saving that file inside the same folder path.
- when opening that file again from launcher while Skip Opening last workfile` is disabled
- when switching contexts (even if the context node doesn't exist in the other workfile I'm opening)

### When it's called
- when opening a new file while `Skip Opening last workfile` is enabled and saving to a different folder path.
- when opening the creator the first time in a scene that doesn't have a context node
- before `Auto-creating workfile instance...`  in `get_context_data` if the context node is not found. so, it's always found before running `create_workfile.py`
- in `update_context_data` if the context node is not found. this happens when the node is destroyed after opening the publisher tool.


---
I think I can now understand your comment:
> Note that it would actually never error currently because apparently this gets called just beforehand elsewhere:
https://github.com/ynput/ayon-core/blob/57d225009293a3123f02fd2b48e4e38948e0f711/client/ayon_core/hosts/houdini/api/pipeline.py#L176-L178
>
> Which in itself, also generates the node. But at least this fixes the broken code were the issue to arise that the node did not exist at that point in time.


The only way to show the error is by adding 
```python
hou.node(CONTEXT_CONTAINER).destroy()
```
between these two lines
https://github.com/ynput/ayon-core/blob/0e4e845407d9f0c425b392d3f454dd5866b819b5/client/ayon_core/hosts/houdini/plugins/create/create_workfile.py#L97-L98

MustafaJafar (approved):
It works.

moonyuet (approved):
Make Sense as `create_context_node` function lives in `HoudiniHost` class

antirotor (approved):

PR #511, opened 2024-05-14
Removed deprecated function 'get_template_data_from_session'

iLLiCiTiT:
## Changelog Description
Removed function that used concept of "session" which is not available anymore in ayon core.

## Testing notes:
1. Validate the function is not used.


BigRoy (approved):
I don't see this used anywhere in the codebase. Removal looks good to me!

PR #516, opened 2024-05-14
Move houdini client code

MustafaJafar:
## Changelog Description
Move Houdini client code.. 
This is one more step towards splitting addons to their repos.

## Additional info
- Houdini addon in this PR works in dev mode. you'd need to set the addon directory to 
  ```
  <your-ayon-core-clone>/server_addon/houdini/client
  ```
- To create addon zip file, you can do any one of the following
- Run `python server_addon\create_ayon_addons.py --addon houdini`
- Copy `create_package.py` from https://github.com/ynput/example-studio-addon/pull/4 to `<your-ayon-core-clone>/server_addon/houdini` and Run it.

## Additional info 2
I don't think that settings are applied correctly in the publish plugins.
Any advice are much appreciated.

## Testing notes:
1. Update the Houdini addon on server. 
2. Launch Houdini via ayon launcher
3. Everything should work as usual 
   - Create / Publish
   - Manage / Load
   - Work Files tool
   - etc


server_addon/houdini/package.py
```diff
@@ -1,3 +1,10 @@
 name = "houdini"
 title = "Houdini"
-version = "0.2.14"
+version = "0.3.14"
+
+client_dir = "ayon_houdini"
+
+ayon_required_addons = {
+    "core": ">=0.3.0",
```
iLLiCiTiT, line 8:
```suggestion
    "core": ">0.3.1",
```

server_addon/houdini/package.py
```diff
@@ -1,3 +1,10 @@
 name = "houdini"
 title = "Houdini"
-version = "0.2.14"
+version = "0.3.14"
```
iLLiCiTiT, line 3:
```suggestion
version = "0.3.0"
```

MustafaJafar: https://github.com/ynput/ayon-core/pull/516/commits/6c75a2912bcf5e853f101c3e9ca0c7c1cbba9d3a

moonyuet (approved):
Looks good. Tested successfully in Houdini with dev mode and build mode.

MustafaJafar:
@iLLiCiTiT 
each plugins category like (load, collect, validate, publish) doesn't have a parent abstract class. 
So, I had to add `settings_category = "houdini"` to each plugin. which I find it tedious.

Also, I faced a problem with `ValidateContainers` how can I add the `settings_category` to it. 

are there other solutions ?

iLLiCiTiT:
> Also, I faced a problem with ValidateContainers how can I add the settings_category to it.

So if I understand it correctly, the plugin lives in core, but houdin settings are applied to it. In that case please make a copy of the plugin in houdini and remove `"houdini"` from the hosts list inside core.

BigRoy:
> So if I understand it correctly, the plugin lives in core, but houdin settings are applied to it. In that case please make a copy of the plugin in houdini and remove `"houdini"` from the hosts list inside core.

Note that this is also true for harmony, after effects, photoshop, maya - they all have settings for this validator it seems but no dedicated plug-in. I wonder if this isn't a case that might be better suited to use `profiles` instead?

server_addon/houdini/client/ayon_houdini/version.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+"""Package declaring Houdini addon version."""
+__version__ = "0.3.14"
```
iLLiCiTiT, line 3:
```suggestion
__version__ = "0.3.0"
```

server_addon/houdini/package.py
```diff
@@ -1,3 +1,10 @@
 name = "houdini"
 title = "Houdini"
-version = "0.2.15"
+version = "0.3.0"
+
```
iLLiCiTiT, line 4:
```suggestion
app_host_name = "houdini"

```

server_addon/houdini/client/ayon_houdini/version.py
MustafaJafar: I think I resolved the conflict when merging the develop branch in a wrong way!
Thanks.

MustafaJafar: fixed in https://github.com/ynput/ayon-core/pull/516/commits/a34647b17545bec61bab87ba7a2097c17b0b059f

MustafaJafar:
I made a PR  https://github.com/ynput/ayon-core/pull/536 to this PR to add settings category for all plugins by default.

server_addon/houdini/package.py
```diff
@@ -1,3 +1,12 @@
...
+app_host_name = "houdini"
+
```
iLLiCiTiT, line 6: My fault, it should be there, but `app_host_name` must be same for all addon versions. That changed in 1.1.1 server.
```suggestion
```

server_addon/houdini/package.py
```diff
@@ -1,3 +1,10 @@
 name = "houdini"
 title = "Houdini"
-version = "0.2.15"
+version = "0.3.0"
+
+client_dir = "ayon_houdini"
+
+ayon_required_addons = {
+    "core": ">0.3.1",
```
iLLiCiTiT, line 8:
```suggestion
    "core": ">0.3.2",
```

moonyuet:
@MustafaJafar can you resolve the conflict? And should we wait for other PR related to setting_category in order to test this?

MustafaJafar:
Hello @moonyuet , 
> can you resolve the conflict?
I've resolved the conflicts and merged latest develop. 

> should we wait for other PR related to setting_category in order to test this?
This PR should be working without other PRs as I added `setting_category = "houdini"` inside the publish plugins that have settings.
The other PR is discussing adding `setting_category`  in a generalized way by implementing base classes so we don't have to add `settings_category` in each plugin.

BigRoy:
> This PR should be working without other PRs as I added `setting_category = "houdini"` inside the publish plugins that have settings. The other PR is discussing adding `setting_category` in a generalized way by implementing base classes so we don't have to add `settings_category` in each plugin.

> I made a PR #536 to this PR to add settings category for all plugins by default. The main reason creating such a PR is that I find it's so tedious to add `settings_category = "houdini"` to all publish plugins.

@MustafaJafar let's get those plugins in here similar to what happened in the Maya addon code move - just so we align the code style a bit - see [the Maya PR](https://github.com/ynput/ayon-core/pull/546/files#diff-4144b58e6210be392766cdb86c00e3e3f9a4fadac21fd2ac44267103d9130c83R1030-R1054) and then do another test run.

moonyuet (approved):
Do a simple test and looks good. I also tested to see if there is any issue with extract fbx as the log above shows the possibility of plugin being "disappeared" as it can't be found in the setting category but the extractor is still there and it doesn't cause the error on `integrate.py`.

MustafaJafar:
Thank you for your review. 

> I also tested to see if there is any issue with extract fbx as the log above shows the possibility of plugin being "disappeared" as it can't be found in the setting category

Also, for reference. that log is just a warning about there are no settings for that plugin.
The warning is coming from these lines. In my opinion, it should be tweaked. 

https://github.com/ynput/ayon-core/blob/242bc8796b453e92eb866d0fb785d80256cd82a0/client/ayon_core/pipeline/publish/lib.py#L337-L350

antirotor (approved):
I did quick testst too, had a really quick look at the code (as quick as it can be with ~130 files changes).

PR #518, opened 2024-05-15
Deadline: replace env var with field from Settigns

kalisp:
## Changelog Description
Removes global environment variable and provide more control per Deadline server.
image

## Additional info
`verify` key is useful to skip ssl validation of self-signed or not present certificates on artist machine.

Requires installation of new addon, changed field name to be more descriptive.

## Testing notes:
1. still not sure, how to setup Deadline to use https and certificates, maybe this could be tested on customer environment?
2. on our side it would be cool to just test publishing from various DCC if it is not breaking anything


64qam:
Can test it as Launcher's Ayon_api version is 1.0.1 and the PR requires 1.0.5

64qam (approved):
Tested on DL repo without ssl validation (with or without ssl bypass in stettings), all works as usually.

PR #520, opened 2024-05-16
Maya: Remove validate animated reference for animation.fbx family

moonyuet:
## Changelog Description
This PR is to remove validate animated reference for checking animation contents before exporting as fbx. It is being removed because it doesn't allow to export contents such as joint, which are crucial for importing the skeleton/animation into Unreal

## Additional info
Make sure you put only the root joint into the `skeletonAnim_SET`

## Testing notes:
1. Use the scene with animation instance
2. Publish


LiborBatek (approved):
All works and no ```Validations``` fails if non ```transforms``` entities present in the out set.

image

PR #521, opened 2024-05-16
Unreal: fix Unreal Editor executable name on linux and macos

antirotor:
## Changelog Description
During the change from UE4 to UE5 Editor's executable was renamed from `UE4Editor` to `UnrealEditor`. We've reflected that change only on Windows, but left `UE4Editor` for Linux and macOS. This is fixing the executable name for all platforms.

## Testing notes:
Start Unreal on linux and macOS. It should build the project and start as expected.


iLLiCiTiT (approved):

simonebarbieri (approved):
Cannot test myself, but the code looks good!

PR #525, opened 2024-05-20
Chore: Validate Content plugin settings

iLLiCiTiT:
## Changelog Description
Added profile based settings for Validate Content publish plugin.

## Additional info
The settings will change behavior of the plugin based on host name. The plugin is disabled if host does not support load functionality, or there are no profiles. Removed settings of the plugin from other hosts.

We should do this for the move of hosts. The first encounter happened in https://github.com/ynput/ayon-core/pull/516 .

## Testing notes:
1. Create package of core and all changed host addons (AE, PS, Nuke, Maya, Houdini, Harmony).
2. Launch DCC of one of the host integrations.
3. Validate Content should be there and can be disabled (with default settings).
4. Based on changed settings the plugin should change it's state -> remove all profiles, remove the host from list, add new hosts, change optional state etc.

### DIFF
--- a/server/settings/publish_plugins.py
+++ b/server/settings/publish_plugins.py
@@ -81,7 +81,8 @@ class ValidateContainersModel(BaseSettingsModel):
 
     _isGroup = True
     plugin_state_profiles: list[ValidateContainersProfile] = SettingsField(
-        default_factory=list
+        default_factory=list,
+        title="Plugin enable state profiles",
     )
 
 
--- a/server/settings/publish_plugins.py
+++ b/server/settings/publish_plugins.py
@@ -67,9 +67,9 @@ class ValidateContainersProfile(BaseSettingsModel):
         title="Host names"
     )
     # Profile values
-    enabled: bool = SettingsField(True)
-    optional: bool = SettingsField(True)
-    active: bool = SettingsField(True)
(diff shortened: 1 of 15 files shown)


MustafaJafar (approved):
It works on my side. 
Also, I'm able to modify the settings.
image

--- 
I like the idea of this PR.
The way I see it: as long as the plugin exists in core, then its settings should be in core even it's meant to run in side different DCCs.

server/settings/publish_plugins.py
```diff
@@ -770,6 +797,10 @@ class PublishPuginsModel(BaseSettingsModel):
         default_factory=ValidateBaseModel,
         title="Validate Version"
     )
+    ValidateContainers: ValidateContainersModel = SettingsField(
+        default_factory=ValidateContainersModel,
+        title="Validate Containers"
```
BigRoy, line 802: Should we update the title to `Validate Outdated Containers` so it matches the label on the plug-in itself? 

With just **Validate Containers** it leaves me a bit guessing _what_ it validates about the containers?

iLLiCiTiT: Eh, why you wrote it after merge, now I have to create new PR...

BigRoy: Sorry, day off today - saw it a bit too late.

PR #526, opened 2024-05-20
Colorspace: Fix Anatomy object initialization

iLLiCiTiT:
## Changelog Description
Use kwargs to pass project entity to Anatomy initialization.

## Additional info
Project entity was passed in as site name instead of project entity.

## Testing notes:
- Make sure you have core with new imageio color settings (or create package from this PR and upload to server to be sure).
1. Enable colormanagement in core addon.
2. Launch host of your choice (should have ability to run python scripts)
3. Run this
```python
from ayon_core.pipeline.colorspace import get_current_context_imageio_config_preset

get_current_context_imageio_config_preset()
```
It should not crash.


MustafaJafar (approved):
works on my side.

PR #536, opened 2024-05-21
Houdini: move houdini client code - Add Settings Category

MustafaJafar:
## Changelog Description
Add settings category by implementing a base class.

## Additional info
I made a separate PR because I'm not sure if this solution is favored.

## Testing notes:
1. Launch Houdini
2. Settings should be applied to their corresponding plugins correctly.
- `ayon+settings://houdini/create` 
- `ayon+settings://houdini/publish`


MustafaJafar:
## Note
This solution worked perfectly with creator plugins because I add settings for each single creator plugin. 

Unlike with publish plugins. I faced a lot of warnings because many plugins don't have corresponding settings.
To name a few: 
```
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ValidatePrimitiveHierarchyPaths' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'SaveCurrentScene' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractOpenGL' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractAlembic' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractBGEO' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractComposite' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractHDA' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractMantraIFD' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractRender' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractUSD' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractUSDLayered' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractAss' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractFBX' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractRedshiftProxy' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractVDBCache' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'ExtractActiveViewThumbnail' settings under settings category 'houdini'  ] 
*** WRN: >>> { filter_pyblish_plugins }: [  Couldn't find plugin 'IncrementCurrentFile' settings under settings category 'houdini'  ] 
```

server_addon/houdini/client/ayon_houdini/api/plugin.py
```diff
@@ -349,3 +354,21 @@ def apply_settings(self, project_settings):
...
+class HoudiniInstancePlugin(pyblish.api.InstancePlugin):
+    """Base class for most of the Houdini instance publish plugins."""
+
+    settings_category = SETTINGS_CATEGORY
+
+
+class HoudiniContextPlugin(pyblish.api.ContextPlugin):
+    """Base class for most of the Houdini context publish plugins."""
+
+    settings_category = SETTINGS_CATEGORY
```
BigRoy, line 374: Admittedly the extra inheritance here to me isn't really warranted. If it were more than one attribute, and more custom logic I'd understand. In this case I'd avoid from refraining for more inheriting the custom plug-in and instead just add `settings_category = "houdini"` to each OR maybe set it using the `SETTINGS_CATEGORY` constant.

This hardly feels like anything that _should_ change for the Houdini integration over time and if it does, then batch changing the attributes is just as trivial code-wise.

server_addon/houdini/client/ayon_houdini/api/plugin.py
```diff
@@ -14,9 +14,14 @@
     CreatedInstance,
     AYON_INSTANCE_ID,
     AVALON_INSTANCE_ID,
+    load
 )
 from ayon_core.lib import BoolDef
 from .lib import imprint, read, lsattr, add_self_publish_button
+import pyblish.api
```
iLLiCiTiT, line 21: This import should be after `import hou`

server_addon/houdini/client/ayon_houdini/api/plugin.py
iLLiCiTiT: Both approaches have caviads. We've discussed it with @antirotor and we don't have strong opinion about it. Both options are viable.

During the discussion we've came to semi-conclusion that maybe having base classes in hosts that have a lot of plugins might be better as you don't have to think about adding `settings_category` when creating new plugin, just use the base plugin class.

server_addon/houdini/client/ayon_houdini/api/plugin.py
MustafaJafar: do you mean _shouldn't_ ? 
because it's already after `import hou`

iLLiCiTiT: It's about PEP ordering of imports.
1. dunder imports
2. built-in imports
3. 3rd party imports
4. from the package itself
5. relative imports

moonyuet (approved):
Did simple test in my side. Looks good.
image

server_addon/houdini/client/ayon_houdini/api/plugin.py
BigRoy: #### Add hosts?

If we're going for base classes here. Should we also add `hosts = ["houdini"]` for those for which it is relevant?

#### Docstrings

Also, the docstrings state "for most of the Houdini plugins" - when shouldn't it be used? Should we remove the "most of the" part of the docstrings and just make it for example """Base class for Houdini context publish plugins."""

server_addon/houdini/client/ayon_houdini/api/plugin.py
MustafaJafar: I fixed it and all other files too. Thanks.
https://github.com/ynput/ayon-core/pull/536/commits/a47c329273bff64b0761e1cb6b2aaae1ead8ad6c

server_addon/houdini/client/ayon_houdini/api/plugin.py
BigRoy: Let's fix up the docstrings as stated and add the `hosts = ["houdini"]` as well, similar [as done in the Maya](https://github.com/ynput/ayon-core/pull/546/files#diff-4144b58e6210be392766cdb86c00e3e3f9a4fadac21fd2ac44267103d9130c83R1030-R1054)

MustafaJafar: What about the Extractors ? 
Should I add a base class for extractors to be similar to Maya ?  Knowing that `publish.Extractor` is only an instance plugin with an importable function. 

https://github.com/ynput/ayon-core/blob/0594f49f4f03b2127db6224b8792bab078ff326e/client/ayon_core/pipeline/publish/publish_plugins.py#L271-L290

PR #537, opened 2024-05-21
Launcher: Make sure actions in Launcher are sorted

fabiaserra:
## Changelog Description
Sort action items by identifier so their order doesn't keep changing on every launch.

## Testing notes:
1. Open Launcher
2. Actions should be sorted

### DIFF
--- a/client/ayon_core/tools/launcher/models/actions.py
+++ b/client/ayon_core/tools/launcher/models/actions.py
@@ -332,7 +332,7 @@ class ActionsModel:
         selection = self._prepare_selection(project_name, folder_id, task_id)
         output = []
         action_items = self._get_action_items(project_name)
-        for identifier, action in self._get_action_objects().items():
+        for identifier, action in sorted(self._get_action_objects().items()):
             if not action.is_compatible(selection):
                 continue
 
--- a/client/ayon_core/tools/launcher/models/actions.py
+++ b/client/ayon_core/tools/launcher/models/actions.py
@@ -332,7 +332,7 @@ class ActionsModel:
         selection = self._prepare_selection(project_name, folder_id, task_id)
         output = []
         action_items = self._get_action_items(project_name)
-        for identifier, action in sorted(self._get_action_objects().items()):
+        for identifier, action in self._get_action_objects().items():
             if not action.is_compatible(selection):
                 continue
 
--- a/client/ayon_core/tools/launcher/ui/actions_widget.py
+++ b/client/ayon_core/tools/launcher/ui/actions_widget.py
@@ -290,6 +290,34 @@ class ActionDelegate(QtWidgets.QStyledItemDelegate):
         painter.drawPixmap(extender_x, extender_y, pix)
 
 
+class ActionsProxyModel(QtCore.QSortFilterProxyModel):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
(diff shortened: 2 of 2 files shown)


client/ayon_core/tools/launcher/models/actions.py
```diff
@@ -332,7 +332,7 @@ def get_action_items(self, project_name, folder_id, task_id):
         selection = self._prepare_selection(project_name, folder_id, task_id)
         output = []
         action_items = self._get_action_items(project_name)
-        for identifier, action in self._get_action_objects().items():
+        for identifier, action in sorted(self._get_action_objects().items()):
```
BigRoy, line 335: I think it would be better to sort these already inside `get_action_objects` so they are sorted in `self._action_items` and don't require the sorting on each call since `_get_action_objects` performs some caching.

Does this change influence only entries where the `order` is the same value? Because technically these should be sorted first by `order` and then by `identifier` second (but maybe the order sorting is already happening elsewhere?)

fabiaserra: sure, @iLLiCiTiT will be the best to answer this, I don't mind how it's implemented but the results should be consistent on each launch

iLLiCiTiT: As far as I know the sorting should happen in UI, there should be proxy model for that.

iLLiCiTiT: > why is there an ActionsQtModel (under ui and not models)

To separate UI logic from business logic.

> I think you are trying to do a sort here

This is sorting of items in grouped items.

> but refresh doesn't get called on initialize?

It does on init and on context change. But that is not the place where sorting should be.

I'll ask again, can I push to your branch, or I should create new PR?

iLLiCiTiT: > I'm aware of that practice but that was exactly my point, I'd have expected QStandardItemModel to be business logic and not UI

Well the issue with `QStandardItemModel` is that it is Qt related (you have to have available Qt binding to import it). 
Also with this approach you can technically show UI in separated process as the communcation between UI and backend controller can be serialized.
Another issue with having logic in Qt widgets is that it was hard to maintain the backend logic, and nobody actualy was able to know how it works. Whereas when you check the controller and backend model you can see the api + you you have single source of information: Controller instead of guessing.

BigRoy:
@iLLiCiTiT isn't it much easier to just turn the sort role data [here](https://github.com/BigRoy/ayon-core/blob/45611ea27e8e3f8106df6dc45075a02e2a1145fd/client/ayon_core/tools/launcher/ui/actions_widget.py#L146) to e.g.:
```python
            item.setData((action_item.order, action_item.identifier), ACTION_SORT_ROLE)
```
So that it sorts on the tuple of order, then identifier? Of course you can replace identifier there with e.g. nice display name if you prefer that.

iLLiCiTiT:
> isn't it much easier to just turn the sort role data [here](https://github.com/BigRoy/ayon-core/blob/45611ea27e8e3f8106df6dc45075a02e2a1145fd/client/ayon_core/tools/launcher/ui/actions_widget.py#L146) to e.g.:

My experience with using python objects (tuple of 2 values) in role data of model are not very good. It also does not handle TypeError.

fabiaserra:
works as expected, thank you @iLLiCiTiT !

PR #538, opened 2024-05-22
Chore: Add addon name to 'version.py' of server packages

iLLiCiTiT:
## Changelog Description
Change how `version.py` text looks like and use addon name in it.

## Additional info
There is not yet such addon which uses this functionality, but right now would be filled with "AYON core" which is not right.


kalisp (approved):
Mocked by providing version to `applications` addon, seems to work:
image

PR #545, opened 2024-05-23
Maya: Workfile Increment in one go 

moonyuet:
## Changelog Description
This PR is to add the option to increment and save workfile in Maya.

## Additional info
- Not working for untitled scene:)/new scene
- We need to think about the implementation of increment and save in other hosts
- Please make sure you updated the latest version of ayon core addon.

## Testing notes:
1. Enable the plugin in `ayon+settings://core/tools/Workfiles/ayon_menu/version_up_workfile`
image

2. Launch Maya
3. Go to AYON
4. Click `Version up Workfile`
image


LiborBatek (commented):
The tool does what it should also tested with custom ```_suffix``` present in the filename and all working fine.

There might be few questions still. 

First if the tool should be named as it is atm

image

or changed to someting like ```Workfile Increment Version``` or similar @antirotor @BigRoy  opinions on that?

And second if it should be somehow treated in relation to native maya ```CTRL+ALT+S``` Incremental save action...as now it is not working with conjunction with it and produce odd results on the active workfile version.... (by appending another version like ```myworkfilename_v0010.001``` )

Again something to consider imho.

BigRoy:
> or changed to someting like `Workfile Increment Version` or similar @antirotor @BigRoy opinions on that?

I agree that it looks odd in the list where it is. Note that we've had this in Maya all this time already but in our custom scripts menu.

For sake of reference, here is a screenshot:

image

What is nice about that one is that it's very close to the top since it's such a frequently clicked button - and preferably also NOT misclicked with "Set frame range" in the list where it is now.

As such I'm wondering whether it's worth thinking about moving it directly below the greyed out current context entry as just "Version up" or "Version up workfile".

> And second if it should be somehow treated in relation to native maya CTRL+ALT+S Incremental save action...as now it is not working with conjunction with it and produce odd results on the active workfile version.... (by appending another version like myworkfilename_v0010.001 )

It would be nice to have it be able to pick that up - yet at the same time we do want to keep it abstractly working also with other DCCs or with custom comments in the names like `_v001_my_comment_01`. Anyway, easiest maybe to make it 'version up' exactly like what the workfiles tool would propose as new filename, which I believe would just strip off the `.001` and go for `_v011`

LiborBatek (approved):
Works as expected and now the tool sits in ```Custom Menu```.

image


Still would like to hear opinions on that... @antirotor @m-u-r-p-h-y 

But can be merged as is imho.

BigRoy:
If we're going for custom menu - can we move it to the top by default? If we're not, then I'd still want to argue for moving it to the top of the AYON menu (which is what I initially meant).

antirotor:
My opinion is not to have it in custom menu - where we shouldn't add core ayon behavior and leave it for studio customizations and maybe some optional or very specialized scripts. And that would mean having it as high as possible in the main menu.

LiborBatek (approved):
Works nicely...

image

If something then Im not 100 percemt sure about the settings and its location in the ```Core addon```...its the very last setting and I would expect it more  up aka closer to ```Workfile``` settings

image

BigRoy:
> Works nicely...
> 
> image
> 
> If something then Im not 100 percemt sure about the settings and its location in the `Core addon`...its the very last setting and I would expect it more up aka closer to `Workfile` settings
> 
> image

Agreed - with that label on the setting I also wouldn't expect that to influence the menu entry. We should really label it something like "Add 'Version Up Workfile' to AYON menu" or something alike, and of course place it in a sensible place.

BigRoy:
> > Add 'Version Up Workfile' to AYON menu
> 
> I have added the description to explain what it is doing. image

I'd argue that this is one of those instances that I personally wouldn't even hover it since I'd expect the toggle to be something completely different. In this case the label itself is just plain non-informative unless it would be under a "AYON menu" header.

moonyuet:
@BigRoy I have put the version up to workfile inside AYON menu.

BigRoy:
> I have put the version up to workfile inside AYON menu.

>        image

I quite like that actually.

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +579,14 @@ def get_process_id():
...
+
+def version_up_workfile():
+    """Function to increment and save workfile
+    """
+    host = registered_host()
+    current_file = host.get_current_workfile()
+    if not current_file:
+        return None
+    filepath = version_up(current_file)
+    host.save_workfile(filepath)
```
BigRoy, line 592: The question now becomes - can we also make it save the workfile with a default name if it wasn't saved? Because it still seems this may not be using the exact same logic the workfiles tool would?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +579,14 @@ def get_process_id():
...
     return _process_id
+
+
+def version_up_workfile():
+    """Function to increment and save workfile
+    """
+    host = registered_host()
+    current_file = host.get_current_workfile()
+    if not current_file:
+        return None
```
BigRoy, line 590: Should this at least log a warning?

Another approach could be to use maya menu's `postMenuCommand` argument to trigger a callback whenever the AYON menu opens, and if no current scene name is set that _then_ we disable the `Version Up Workfile` button.

BigRoy (approved):
Didn't test it - but overall it looks good visually. Did have some notes however that I feel would be nice to haves.

client/ayon_core/pipeline/context_tools.py
moonyuet: I would like to know how to do it. @iLLiCiTiT do you have any idea?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -587,6 +587,7 @@ def version_up_workfile():
     host = registered_host()
     current_file = host.get_current_workfile()
     if not current_file:
+        log.debug("Current file not found.")
```
BigRoy, line 590: Current file not found seems quite warning for the user - did someone delete it? Also, Maya may have a filename set but the file not existing anymore - we're not checking that in `host.get_current_workfile`

I suppose the right wording is along the lines of:
```suggestion
        log.debug(
            "Current file is unsaved and can't be versioned up. "
            "Please save your file first.")
```

Also, we'll need to make sure the debug messages show up to the artists in one way or another that an artist may actually notice (I'm not sure if they do by default in all hosts, and in this case in Maya)?

LiborBatek (approved):
Works for me, I am fine with settings and also looks and works nicely within maya host.

Screenshot 2024-05-30 115048


Screenshot 2024-05-30 115255

Simply LGTM!

antirotor (approved):

client/ayon_core/pipeline/context_tools.py
antirotor: I guess to code for it is here https://github.com/ynput/ayon-core/blob/750ad82744145da0397100fef98c04d4e150bcee/client/ayon_core/tools/workfiles/models/workfiles.py#L381-L439

antirotor (changes requested):
Can we add the default name if scene not saved? I was thinking about adding it in additional PR, but at the end it is probably superuseful feature.

BigRoy:
Also be aware that this version up logic seems to version up to the next available free version?

Say there are files:
```
v001
v003
v004
```
When opening v001 and versioning up I suspect it may save to `v002` even though workfiles tool would default to `v005`.
I might be mistaken reading the code - but that may be an edge case we'd also want to check?

client/ayon_core/version.py
```diff
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 """Package declaring AYON core addon version."""
-__version__ = "0.3.3-dev.1"
+__version__ = "0.3.4-dev.1"
```
iLLiCiTiT, line 3: Please change the core version back.

package.py
```diff
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "0.3.3-dev.1"
+version = "0.3.4-dev.1"
```
iLLiCiTiT, line 3: Please change the core version back.

client/ayon_core/pipeline/context_tools.py
```diff
@@ -11,7 +11,7 @@
 
 from ayon_core import AYON_CORE_ROOT
 from ayon_core.host import HostBase
-from ayon_core.lib import is_in_tests, initialize_ayon_connection, emit_event
+from ayon_core.lib import is_in_tests, initialize_ayon_connection, emit_event, version_up
```
iLLiCiTiT, line 14:
```suggestion
from ayon_core.lib import (
    is_in_tests,
    initialize_ayon_connection,
    emit_event,
    version_up
)
```

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +579,17 @@ def get_process_id():
     if _process_id is None:
         _process_id = str(uuid.uuid4())
     return _process_id
+
+
+def version_up_workfile():
```
iLLiCiTiT, line 584:
```suggestion
def version_up_current_workfile():
```

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +579,17 @@ def get_process_id():
...
+    """Function to increment and save workfile
+    """
+    host = registered_host()
+    current_file = host.get_current_workfile()
+    if not current_file:
+        log.debug(
+            "Current file is unsaved and can't be versioned up. "
+            "Please save your file first.")
+        return None
+    filepath = version_up(current_file)
```
iLLiCiTiT, line 594: This will version up current file into next version, which might potentially override existing file?

client/ayon_core/pipeline/context_tools.py
iLLiCiTiT:
```suggestion
        return
```

server/settings/tools.py
```diff
@@ -118,7 +118,19 @@ class WorkfilesLockProfile(BaseSettingsModel):
...
+    version_up_workfile: bool = SettingsField(
+        False,
+        title="Version Up Workfile",
+        description="Add 'Version Up Workfile' to AYON menu"
+    )
+
+
 class WorkfilesToolModel(BaseSettingsModel):
+    ayon_menu: AyonMenuModel = SettingsField(
+        default_factory=AyonMenuModel,
```
iLLiCiTiT, line 131:
```suggestion
    ayon_menu: AYONMenuModel= SettingsField(
        default_factory=AYONMenuModel,
```

server/settings/tools.py
```diff
@@ -118,7 +118,19 @@ class WorkfilesLockProfile(BaseSettingsModel):
...
+class AyonMenuModel(BaseSettingsModel):
+    version_up_workfile: bool = SettingsField(
+        False,
+        title="Version Up Workfile",
+        description="Add 'Version Up Workfile' to AYON menu"
+    )
+
+
 class WorkfilesToolModel(BaseSettingsModel):
+    ayon_menu: AyonMenuModel = SettingsField(
```
iLLiCiTiT, line 130: This is me being stupid: Why it's called `ayon_menu` and lives in wokfiles tool settings?

If we would like to add another actions in future to `ayon_menu` that are not workfiles related we would still add it here?

BigRoy: Good question - likely best to be moved to a more standalone section.

client/ayon_core/pipeline/context_tools.py
BigRoy: > This will version up current file into next version, which might potentially override existing file?

Looking at the logic it actually checks for existing files and versions up beyond that?

server/settings/tools.py
moonyuet: any suggestion where it should be moved to?

client/ayon_core/version.py
BigRoy: Shouldn't this PR require both an update to maya version and core version since it relies on updated functions to be available in core?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +586,39 @@ def get_process_id():
...
     return _process_id
+
+
+def version_up_current_workfile():
+    """Function to increment and save workfile
+    """
+    host = registered_host()
+    project_name = get_current_project_name()
+    folder_path = get_current_folder_path()
+    task_name = get_current_task_name()
```
BigRoy, line 597: Should we rely on `host.get_current_context()` here? (which in the end by default does the same, but may differ per host if they wanted to for whatever reason?)

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +586,39 @@ def get_process_id():
...
+    # Define saving file extension
+    extensions = host.get_workfile_extensions()
+
+    work_root = work_template["directory"].format_strict(data)
+    file_template = work_template["file"].template
+    last_workfile_path = get_last_workfile(
+        work_root, file_template, data, extensions, True
+    )
+    current_workfile_path = version_up(last_workfile_path)
+    host.save_workfile(current_workfile_path)
```
BigRoy, line 624: Variable naming here is weird - the versioned up path from the latest workfile isn't the current workfile path?

```suggestion
    new_workfile_path = version_up(last_workfile_path)
    host.save_workfile(new_workfile_path )
```

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +586,39 @@ def get_process_id():
...
+    work_template = anatomy.get_template_item("work", template_key)
+
+    # Define saving file extension
+    extensions = host.get_workfile_extensions()
+
+    work_root = work_template["directory"].format_strict(data)
+    file_template = work_template["file"].template
+    last_workfile_path = get_last_workfile(
+        work_root, file_template, data, extensions, True
+    )
```
BigRoy, line 622: This seems to, in no way, rely on any current filename even if there were. We may need to somehow preserve the current file extension if the host DOES have a filename, so that if e.g. in maya the current scene is `.ma` it'll still be saved as `.ma` and if it is `.mb` the next save is also `.mb`.

Aside of that of course it's still a question that if the file is unsaved - what should it default to then? I suppose the same as what workfiles tool would?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +586,39 @@ def get_process_id():
...
+        folder_path,
+        task_name,
+        host_name,
+    )
+    anatomy = Anatomy(project_name)
+
+    data = get_template_data_with_names(
+        project_name, folder_path, task_name, host_name
+    )
+    data["root"] = anatomy.roots
```
BigRoy, line 611: This seems overly specific 'hack' that is likely copied from elsewhere where this code is based on. Copying that logic here isn't necessarily wrong but does point to something that we might want to expose in a more sensible manner for others to use.

This begs the question - why is it this involved or complex from an API standpoint to define what the next potential workfile name should be? Shouldn't we expose that in a better way for easier re-use (also if someone from another addon may want to figure that out?)

For example, if there were a `get_next_workfile_version_filepath(context)` function then this would be more easily accessible for re-use without it also automatically saving the file. I would want to avoid us needing to copy this exact logic around to define a workfile filename from anywhere.

May not have to be this PR - but just wanted to remark that we should definitely improve the API access here.

BigRoy (commented):
With maya client code separated I believe this PR should also bump the maya addon version.

client/ayon_core/pipeline/context_tools.py
moonyuet: These three functions are already done that.
https://github.com/ynput/ayon-core/blob/118b4c189f13741426f4ecabd47c4421ad22325b/client/ayon_core/pipeline/context_tools.py#L347-L372

client/ayon_core/pipeline/context_tools.py
moonyuet: I have added the conditions to check to see if there is current file. If there is one, it will follow the extension of the current workfile. If there isn't, it will use .ma extension.

client/ayon_core/version.py
iLLiCiTiT: PRs should not bump version of core addon.

server/settings/tools.py
```diff
@@ -268,6 +277,10 @@ class PublishToolModel(BaseSettingsModel):
 
 
 class GlobalToolsModel(BaseSettingsModel):
+    ayon_menu: AYONMenuModel = SettingsField(
+        default_factory=AYONMenuModel,
+        title="Tools"
```
BigRoy, line 282: Is this title wrong? :) I suppose it should be "Menu" or "AYON Menu"?

BigRoy (approved):
Didn't test - code overall looks good to me.

I did have one comment on title in settings - there's also still an open comment by @iLLiCiTiT to revert ayon core version change.

Also, [this comment](https://github.com/ynput/ayon-core/pull/545/files#r1624896875) should still be considered but for a separate issue.

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +586,46 @@ def get_process_id():
...
+    if current_file:
+        extensions = [os.path.splitext(current_file)[-1]]
+
+    work_root = work_template["directory"].format_strict(data)
+    file_template = work_template["file"].template
+    last_workfile_path = get_last_workfile(
+        work_root, file_template, data, extensions, True
+    )
+    new_workfile_path = version_up(last_workfile_path)
+    host.save_workfile(new_workfile_path)
```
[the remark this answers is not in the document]
moonyuet: Maybe just to be safe, we should check if the path exists and version up again if it does exists and save.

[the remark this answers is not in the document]
BigRoy: `version_up` function does that already as far as I know. However, it doesn't cross "gaps" like I described.

antirotor (approved):
Tested, it works nice. Thanks

BigRoy:
@antirotor could you test [this edge case](https://github.com/ynput/ayon-core/pull/545#discussion_r1627697790)?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -579,3 +586,47 @@ def get_process_id():
...
+    current_file = host.get_current_workfile()
+    if current_file:
+        extensions = [os.path.splitext(current_file)[-1]]
+
+    work_root = work_template["directory"].format_strict(data)
+    file_template = work_template["file"].template
+    workfile_path = get_last_workfile(
+        work_root, file_template, data, extensions, True
+    )
+    if os.path.exists(workfile_path):
```
BigRoy, line 630: As said before - this check does not fix things. See [here](https://github.com/ynput/ayon-core/pull/545#discussion_r1627782264) - it's redundant code.

However, the issue presented [here](https://github.com/ynput/ayon-core/pull/545#discussion_r1627697790) I assume is still present?

BigRoy (changes requested):

client/ayon_core/pipeline/context_tools.py
moonyuet: is there any way you can suggest fixing it?

PR #546, opened 2024-05-23
Chore: Move Maya addon code

antirotor:
## Changelog Description
Move client Maya addon code from AYON core to the folder along the server part. Refactoring imports to accommodate this move.

## Additional info
This is also adding rather hacked support of inline codespell ignores to ease development. You can now use syntax:
```python
foo = "some intentional tipo"  # codespell:ignore tipo
```
In next release of codespell (>2.2.6) it should be natively supported.

## Testing notes:
Build and install the addon
Test the integration


iLLiCiTiT:
It is also required to define `settings_category ="maya"` in the plugins so settings can be applied automatically.

LiborBatek (approved):
All seems normal and working ok.

Did bunch of tests for publishing etc. All good.

antirotor:
With the settings category, we need to retest loading and publishing. Should work as before.

All plugins have their own Maya specific base class defining settings category and host now.

server_addon/maya/client/ayon_maya/plugins/publish/extract_rig.py
```diff
@@ -2,17 +2,15 @@
 """Extract rig as Maya Scene."""
 import os
 
+from ayon_maya.api.lib import maintained_selection
+from ayon_maya.api.plugin import MayaExtractorPlugin
```
BigRoy, line 6: Doesn't directly importing a `pyblish plug-in` into this space mean that Pyblish logic will also pick up that class and thus basically will also load that `MayaExtractorPlugin` base class as a plug-in? Isn't that right @iLLiCiTiT ?

iLLiCiTiT: You are right.

BigRoy: In that case @antirotor all these imports need to change so that we load the module instead and inherit from the plug-in that way. Similar to how the code looked before.

antirotor: yeah, you're right. I forgot about this quirk. Hopefuly easy job for find&replace.

PR #548, opened 2024-05-24
Loader: Show statuses in loader

iLLiCiTiT:
## Changelog Description
Loader tool shows status name or status short name of version.

## Additional info
Added `status` to `VersionItem` which is a status name. Projects model now have method to retrieve list of `StatusItem` for a project. A short name of status is used if name does not fit the column. An icon is not displayed as that would require to implement MaterialIcon for Qt.

## Screenshot
image

## Testing notes:
1. Open loader tools (from tray or in DCC).
2. Select project and folder.
3. Products view should display status next to version.


BigRoy:
Love it! 

Any chance we can also get the status icons to match from the web frontend? :)

iLLiCiTiT:
> Any chance we can also get the status icons to match from the web frontend? :)

As description says: An icon is not displayed as that would require to implement MaterialIcon for Qt.

64qam (approved):

PR #501, opened 2024-05-09
Nuke: Publish intermediate files on demand 

tokejepsen:
## Changelog Description
Intermediate files from the baking of LUTs in Nuke can be published with this PR.

## Testing notes:
1. Enable the `Publish` toggle on the outputs; `ayon+settings://nuke/publish/ExtractReviewIntermediates/outputs/0/publish`
2. Publish a `render` in Nuke.
3. Validate the output of the intermediate files are in the Loader on the products representations.

PR #551, opened 2024-05-24
Publish representation with `isIntermediate` flag in data

jakubjezek001:
## Changelog Description
Publish intermediate file representation with `isIntermediate` flag in data

## Additional info
- Adjusted method parameters in ExporterReview class for better readability.
- Added colorspace parameter to get_representation_data method.

## Testing notes:
1. in Nuke activate `ayon+settings://nuke/publish/ExtractReviewIntermediates/outputs/0/publish`
2. in Nuke publish normal render product type and capture published representation id from `integrate.py` plugin log.
3. Open AYON console from Tray and use following snippet to check the data of published representation.
```
from pprint import pprint
from ayon_api import get_representation_by_id

repre = get_representation_by_id(
    [YOUR PROJECT NAME], [PUBLISHED_REPRE_ID],)

pprint(repre)
```
4. you should be able to find `isIntermediate` flag within the repre document data keys.

this is follower for #501


moonyuet (approved):
Tested successfully in nuke and our console.
image
image

kalisp:
Why won't we use `tag` for it? Or maybe I am misunderstanding use case.

client/ayon_core/hosts/nuke/api/plugin.py
```diff
@@ -584,14 +587,22 @@ def get_representation_data(
                                     Defaults to False.
             custom_tags (list[str], optional): user inputted custom tags.
                                                Defaults to None.
+            colorspace (str, optional): colorspace name.
+                                        Defaults to None.
         """
         add_tags = tags or []
```
iLLiCiTiT, line 593: Rather add the tag here instead of appending it in the dict.
```suggestion
        add_tags = tags or []
        add_tags.append(self.name.replace("_", "-"))
```

BigRoy:
> Why won't we use `tag` for it? Or maybe I am misunderstanding use case.

I actually had the same question in my mind as @kalisp had there. @jakubjezek001 could you elaborate a bit on that?

jakubjezek001:
Tags are used only when we publish, and they're not part of the database document. We need to fetch the flagged data from the database during delivery-related tasks.

BigRoy:
> Tags are used only when we publish, and they're not part of the database document. We need to fetch the flagged data from the database during delivery-related tasks.

Sorry - isn't it the case that we're able to add `data["tags"] = ["intermediate"]` even on the representation? That's what we're asking, right?

Also, I thought having _actual_ supported tags per representation on a lower level already so that in the database we can actually tag things as well (or does it only exist like that on folders?)

The question kind of is - couldn't this be a tag on the representation or whatever. Or does this mean that we might end up flooding with arbitrary `data` again on the representations that we do start to rely on elsewhere (e.g. in deliveries) that we do not want to be as fluid but prefer to make more strict, almost like attributes or tags, etc. over time?

PR #552, opened 2024-05-24
Workfiles tool: Author visible in UI

iLLiCiTiT:
## Changelog Description
Add author username to workfiles tool.

## Additional info
Workfiles tool shows author of workfile based on workfile info on server or author of published version.

It was discovered that workfile info does not fill `createdBy` and `updatedBy` automatically, which is fixed in AYON server 1.1.3, but for backwards compatibility they're both filled and auto-fixed when updating existing file.

The UI shows create user instead of update user, maybe that is not right? Also the UI is "not enough" wide to show all the information, maybe we should make the tool wider by default? Not sure if my workfile template is "production standard" but it does not fit.

## Screenshot
image

## Testing notes:
1. Launch a host of your choice.
2. Save workfile via workfiles tool.
3. Open workfile tool again.
4. The new workfile should have author filled with your username.
5. Published workfiles should show author of the representation.


m-u-r-p-h-y:
> The UI shows create user instead of update user, maybe that is not right? Also the UI is "not enough" wide to show all the information, maybe we should make the tool wider by default? Not sure if my workfile template is "production standard" but it does not fit.

It is a good question. I would rather see the last user who touched the file instead of the original author.  Would you agree @dee-ynput ?

MilaKudr (approved):

BigRoy:
> It is a good question. I would rather see the last user who touched the file instead of the original author. Would you agree @dee-ynput ?

Agreed - last modified author makes most sense as data to me - especially because it also lists date modified, not date created.

MilaKudr (approved):
Screenshot 2024-05-30 102140

PR #554, opened 2024-05-27
Chore: Validate Version plugin enabled based on profiles

iLLiCiTiT:
## Changelog Description
Plugin `ValidateVersion` is enabled, optional and active based on profiles in settings.

## Additional info
This is one of many PRs that tries to resolve plugins in core that have host name filtering.

## Testing notes:
1. Create core package.
2. Upload the package to server.
3. Make sure the AYON launcher is using new core client code.
4. Validate Version should work with default settings the same way as did before this PR.
5. Changing the settings should take effect in Publisher (e.g. change behavior for a host).


MilaKudr (approved):
With this PR, publishing and validating works the same way as before.

client/ayon_core/plugins/publish/validate_version.py
```diff
@@ -1,6 +1,10 @@
 import pyblish.api
+
+from ayon_core.lib import filter_profiles
 from ayon_core.pipeline.publish import (
-    PublishValidationError, OptionalPyblishPluginMixin
+    PublishValidationError,
+    OptionalPyblishPluginMixin,
+    get_current_host_name,
 )
```
BigRoy, line 8: This PR broken this validator. This import is wrong.
It is actually in `ayon_core.pipeline.context_tools`.

PR #558, opened 2024-05-28
Create logic: Fix parent property in PublishAttributeValues

iLLiCiTiT:
## Changelog Description
Property `parent` actually does return the parent object.

## Testing notes:
1. The property is not used so there is not much to test.


BigRoy (approved):
Didn't test but the change itself makes sense.

PR #560, opened 2024-05-28
Publisher: filter creators based on task

kalisp:
## Changelog Description
This provides a way how to limit list of provided Creators in Publisher based on current context. This should
help artists decision which creator they should use.

Creator labels can by `copy&pasted` from Publisher.

## Additional info
If you want to limit which creators should be shown, provide some profile, based on host name and task info.
Put creator identifier to `Allowed Creator Labels`.

## Testing notes:
1. configure `ayon+settings://core/tools/creator/filter_creator_profiles?project=ayon_test`
core_tools_filter_creators

3. check that configured profile was applied and appropriate creator is shown/not shown in middle column with list of creators inside of Publisher


client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1853,45 @@ def _collect_creator_items(self):
...
+        if host_name == "traypublisher":
+            # no real context known
+            return allowed_creator_identifiers
+        proj_settings = get_project_settings(project_name)
+        filter_creator_profiles = (
+            proj_settings
+            ["core"]
+            ["tools"]
+            ["creator"]
+            ["filter_creator_profiles"]
```
BigRoy, line 1878: Should we be careful with someone running an older server addon but a newer client potentially? Or is that more of an issue for e.g. `houdini` etc. because they are included within ayon core?
```suggestion
            .get("filter_creator_profiles", [])
```

client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1853,45 @@ def _collect_creator_items(self):
...
+    ):
+        """Provide configured creator identifier in this context
+
+        If no profile provided for current context, it shows all creators
+        """
+        allowed_creator_identifiers = []
+        if host_name == "traypublisher":
+            # no real context known
+            return allowed_creator_identifiers
+        proj_settings = get_project_settings(project_name)
```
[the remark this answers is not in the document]
kalisp: I was looking for it, couldn't find it. It is not on `CreateContext` (and Task entity or better task_type neither).

server/settings/tools.py
```diff
@@ -35,6 +35,22 @@ class ProductNameProfile(BaseSettingsModel):
...
+    hosts: list[str] = SettingsField(default_factory=list, title="Hosts")
+    task_types: list[str] = SettingsField(
+        default_factory=list,
+        title="Task types",
+        enum_resolver=task_types_enum
+    )
+    task_names: list[str] = SettingsField(default_factory=list,
+                                          title="Task names")
+    creator_identifiers: list[str] = SettingsField(
+        "", title="Allowed Creator Identifiers")
```
iLLiCiTiT, line 51:
```suggestion
        default_factory=list,
        title="Allowed Creator Identifiers")
```

server/settings/tools.py
```diff
@@ -35,6 +35,22 @@ class ProductNameProfile(BaseSettingsModel):
...
+
+    _layout = "expanded"
+    hosts: list[str] = SettingsField(default_factory=list, title="Hosts")
+    task_types: list[str] = SettingsField(
+        default_factory=list,
+        title="Task types",
+        enum_resolver=task_types_enum
+    )
+    task_names: list[str] = SettingsField(default_factory=list,
+                                          title="Task names")
```
iLLiCiTiT, line 49:
```suggestion
    task_names: list[str] = SettingsField(
        default_factory=list,
        title="Task names")
```

client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1853,45 @@ def _collect_creator_items(self):
...
         return output
 
+    def _get_allowed_creator_identifiers(
+            self,
+            project_name,
+            host_name,
+            task_name,
+            task_type,
+            log=None
+    ):
```
iLLiCiTiT, line 1863: This should return `None` if there are no filters instead of empty list.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1853,45 @@ def _collect_creator_items(self):
...
+        """Provide configured creator identifier in this context
+
+        If no profile provided for current context, it shows all creators
+        """
+        allowed_creator_identifiers = []
+        if host_name == "traypublisher":
+            # no real context known
+            return allowed_creator_identifiers
```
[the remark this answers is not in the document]
iLLiCiTiT: There should not be anything host specific here.

client/ayon_core/tools/publisher/control.py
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
        project_settings = get_project_settings(project_name)
```

client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1853,45 @@ def _collect_creator_items(self):
 
         return output
 
+    def _get_allowed_creator_identifiers(
+            self,
+            project_name,
+            host_name,
+            task_name,
+            task_type,
+            log=None
```
iLLiCiTiT, line 1862:
```suggestion
        self,
        project_name,
        host_name,
        task_name,
        task_type,
        log=None
```

client/ayon_core/tools/publisher/control.py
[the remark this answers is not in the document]
iLLiCiTiT: But I agree it should be taken from `CreateContext`. We should probably add `project_settings` property to `CreateContext` class.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1827,8 +1830,19 @@ def _reset_plugins(self):
     def _collect_creator_items(self):
         # TODO add crashed initialization of create plugins to report
         output = {}
+        allowed_creator_identifiers = self._get_allowed_creator_identifiers(
+            self.project_name,
+            self._create_context.host_name,
+            self.current_task_name,
+            get_current_task_entity()["taskType"],
```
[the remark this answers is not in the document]
iLLiCiTiT: Should be added to `CreateContext`, controller should use only host or create context object to get information, never "global" context.

server/settings/tools.py
```diff
@@ -35,6 +35,22 @@ class ProductNameProfile(BaseSettingsModel):
...
+    creator_identifiers: list[str] = SettingsField(
```
iLLiCiTiT, line 50: We should maybe use enum switcher to "Show all" vs. "Filter by identifier".
```python
def creator_profilte_filter_options():
    return [
        {"value": "creator_identifiers", "label": "Filter by identifier"},
        {"value": "show_all", "label": "Show all"},
    ]


class FilterCreatorProfile(BaseSettingsModel):
    """Provide list of allowed Creator identifiers for context"""

    _layout = "expanded"
    host_names: list[str] = SettingsField(
        default_factory=list, title="Host names"
    )
    task_types: list[str] = SettingsField(
        default_factory=list,
        title="Task types",
        enum_resolver=task_types_enum
    )
    task_names: list[str] = SettingsField(
        default_factory=list,
        title="Task names"
    )
    filter_variant: str = SettingsField(
        "creator_identifiers",
        title="Filter variant",
        enum_resolver=creator_profilte_filter_options,
        conditionalEnum = True,
    )
    creator_identifiers: list[str] = SettingsField(
        default_factory=list,
        title="Allowed Identifiers"
    )
```

server/settings/tools.py
```diff
@@ -35,6 +35,22 @@ class ProductNameProfile(BaseSettingsModel):
     template: str = SettingsField("", title="Template")
 
 
+class FilterCreatorProfile(BaseSettingsModel):
```
iLLiCiTiT, line 38: We should maybe use enum switcher to "Show all" vs. "Filter by identifier".
```python
def creator_profilte_filter_options():
    return [
        {"value": "creator_identifiers", "label": "Filter by identifier"},
        {"value": "show_all", "label": "Show all"},
    ]


class FilterCreatorProfile(BaseSettingsModel):
    """Provide list of allowed Creator identifiers for context"""

    _layout = "expanded"
    host_names: list[str] = SettingsField(
        default_factory=list, title="Host names"
    )
    task_types: list[str] = SettingsField(
        default_factory=list,
        title="Task types",
        enum_resolver=task_types_enum
    )
    task_names: list[str] = SettingsField(
        default_factory=list,
        title="Task names"
    )
    filter_variant: str = SettingsField(
        "creator_identifiers",
        title="Filter variant",
        enum_resolver=creator_profilte_filter_options,
        conditionalEnum = True,
    )
    creator_identifiers: list[str] = SettingsField(
        default_factory=list,
        title="Allowed Identifiers"
    )
```

client/ayon_core/tools/publisher/control.py
iLLiCiTiT: > I had seen a lot of fallbacks being applied in other PRs, but maybe that was because the client facing code was ayon_core being sepately updated from e.g. ayon_houdini server addon - so I guess you can ignore this, yes.

Yep, the fallbacks are in hosts and modules, which have settings in `server_addons` folder.

client/ayon_core/tools/publisher/control.py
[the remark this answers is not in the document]
kalisp: I changed querying of task_type to be more safe. Traypublisher doesnt have anything in `get_current_task_entity`

client/ayon_core/tools/publisher/control.py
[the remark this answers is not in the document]
kalisp: Added project_settings to CreateContext

MustafaJafar (approved):
Hey, 
I gave it a quick shot on my side in Houdini. it works. 

For reference here's how my settings looks like:
image

and here's a snap from Houdini in different tasks: 
<table>
<tr>
<td></td>
<td></td>
</tr>
</table>

---

I have one question though. Is it easy for pipeline admins to know the value of creator identifier ?

client/ayon_core/pipeline/create/context.py
```diff
@@ -1401,6 +1401,7 @@ def __init__(
         self._current_folder_path = None
         self._current_task_name = None
```
iLLiCiTiT, line 1402:
```suggestion
        self._current_task_name = None
        self._current_task_entity = None
```

client/ayon_core/pipeline/create/context.py
```diff
@@ -1401,6 +1401,7 @@ def __init__(
...
         self._current_workfile_path = None
+        self._current_project_settings = None
```
iLLiCiTiT, line 1404: Both `self._current_project_settings` and `self._current_task_entity` should be reset to `None` in `reset_current_context`.

client/ayon_core/pipeline/create/context.py
```diff
@@ -1592,6 +1604,12 @@ def get_current_project_anatomy(self):
                 self._current_project_name)
         return self._current_project_anatomy
 
+    def get_current_project_settings(self):
```
iLLiCiTiT, line 1607: This method should be used in `_reset_creator_plugins`.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1827,8 +1839,25 @@ def _reset_plugins(self):
     def _collect_creator_items(self):
         # TODO add crashed initialization of create plugins to report
         output = {}
+        task_type = None
+        current_task_entity = get_current_task_entity()
```
iLLiCiTiT, line 1843:
```suggestion
        current_task_entity = self._create_context.get_current_task_entity()
```

client/ayon_core/tools/publisher/control.py
```diff
@@ -1662,6 +1664,16 @@ def current_task_name(self):
 
         return self._create_context.get_current_task_name()
 
+    @property
+    def current_project_settings(self):
```
iLLiCiTiT, line 1668: I would make this private `_current_project_settings`.

client/ayon_core/pipeline/create/context.py
kalisp: Done

client/ayon_core/tools/publisher/control.py
kalisp: I was following other `current_folder_path`, ` current_task_name`. Not really sure why these are as properties and `current_project_settings` should be private.

iLLiCiTiT: Those properties are public api of controller, `current_project_settings` is used as private property of the controller.

server/settings/tools.py
```diff
@@ -50,9 +50,11 @@ class FilterCreatorProfile(BaseSettingsModel):
     task_names: list[str] = SettingsField(
         default_factory=list,
         title="Task names")
-    creator_identifiers: list[str] = SettingsField(
+    creator_labels: list[str] = SettingsField(
         default_factory=list,
-        title="Allowed Creator Identifiers")
+        title="Allowed Creator Labels",
+        description="Copy creator label from Publisher, regex supported."
+    )
```
BigRoy, line 57: Oh boy - is matching by label really the best choice here? I understand `identifiers` aren't necessarily easily accessible or whatever but at least they are supposed to be static. But labels feel way more 'fluid' than that.

How about we just allow to enable creators based on their `product_type`? Which may be even easier than the label, and at least also considered stable? (We might still need the additional filtering to 'identifier' for the specific cases where multiple creators exist for a single product type - which may happen in some cases).

server/settings/tools.py
```diff
@@ -70,7 +72,9 @@ class CreatorToolModel(BaseSettingsModel):
 
     filter_creator_profiles: list[FilterCreatorProfile] = SettingsField(
         default_factory=list,
-        title="Filter creator profiles"
+        title="Filter creator profiles",
+        description="White list of creator labels that will be only shown if "
```
BigRoy, line 76: _Whitelist_ is one word.
```suggestion
        description="Whitelist of creator labels that will be only shown if "
```

Deadline a few years back changed "whitelist" to "Allow list" due to connotation with with and black in whitelist and blacklist. Do we want to do the same here?
```suggestion
        description="Allowed list of creator labels that will be only shown if "
```

server/settings/tools.py
kalisp: That doesn't match planned aliasing of creators. If I understood it correctly, you might have `ImageHD`, `ImageLD` creators both producing `image` product type, but you would like still to show only `ImageHD` for one task and the other creator for different.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1885,8 +1887,16 @@ def _get_allowed_creator_identifiers(self):
...
-        return allowed_creator_identifiers
+            allowed_creator_labels = profile["creator_labels"]
+            self.log.debug(f"Only allowed `{allowed_creator_labels}` creators")
+        return allowed_creator_labels
+
+    def _label_matches_allowed(self, label, allowed_labels):
+        """Implement regex support for allowed labels."""
+        for allowed in allowed_labels:
+            if re.match(allowed, label):
+                return True
```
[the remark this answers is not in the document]
kalisp: Hm, I would expect that speed up would be negligent, but it is not too unreadable so ok.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1893,9 +1893,9 @@ def _get_allowed_creator_labels(self):
 
     def _label_matches_allowed(self, label, allowed_labels):
         """Implement regex support for allowed labels."""
-        for allowed in allowed_labels:
-            if re.match(allowed, label):
-                return True
+        allowed_patterns = re.compile("|".join(allowed_labels))
```
BigRoy, line 1896: Even better if we can take this join and compile outside of the `for loop` where this method is called.
Which has the benefit that this just turns into:
```python
return bool(pattern.match(label))
```

server/settings/tools.py
BigRoy: I see. So 'aliases' would still have the same identifier as well, just different labels?

Anyway, I might be nitpicky. But "labels" definitely don't seem like a constant 'name' to me we should be using as an identifier or key for a particular creator but maybe this use case warrants it.

iLLiCiTiT: You can't have create 2 plugins with same identifier.

server/settings/tools.py
iLLiCiTiT: BTW what about this comment, which adds option to not show any create plugins.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1827,8 +1838,16 @@ def _reset_plugins(self):
     def _collect_creator_items(self):
         # TODO add crashed initialization of create plugins to report
         output = {}
+        allowed_creator_labels = self._get_allowed_creator_labels()
         for identifier, creator in self._create_context.creators.items():
             try:
+                if (
+                    allowed_creator_labels is not None
```
iLLiCiTiT, line 1845: This condition could be moved to `_label_matches_allowed`.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1858,46 @@ def _collect_creator_items(self):
...
+            "task_names": self.current_task_name,
+            "task_types": task_type,
+            "host_names": self._create_context.host_name
+        }
+        profile = filter_profiles(
+            filter_creator_profiles,
+            filtering_criteria,
+            logger=self.log
+        )
+
```
iLLiCiTiT, line 1888:
```suggestion

allowed_creator_labels = None
```

MilaKudr (approved):

server/settings/tools.py
kalisp: Yeah, using labels as keys sounds really bad, but I don't think any lead would like to dive into code to get those, product type seems reasonable too, but you would still figure out if that product type is actually produced in DCC, with Labels it seems pretty straightforward to open DCC and Publisher, copy&paste from it directly into Settings with no fuss.

I dont think we are changing labels too often, they are pretty static after they are implemented, imho. It could bite us if we would like to do internationalization, but that would open completely new can of worms so this one would be the smallest issue.

client/ayon_core/tools/publisher/control.py
```diff
@@ -1827,8 +1838,13 @@ def _reset_plugins(self):
     def _collect_creator_items(self):
         # TODO add crashed initialization of create plugins to report
         output = {}
+        allowed_creator_labels = self._get_allowed_creator_labels()
         for identifier, creator in self._create_context.creators.items():
             try:
+                if (not self._is_label_allowed(
+                        creator.label, allowed_creator_labels)):
+                    self.log.debug(f"{creator.label} not allowed for context")
+                    continue
```
BigRoy, line 1847: You're still recompiling and joining the regexes in the for loop - which we don't want. Plus we're putting 'too much' logic in the try..except that may just fail without us wanting to ignore it.

We're better off preparing the pattern before the loop. Like this:
```suggestion
        allowed_creator_labels = self._get_allowed_creator_labels()
        allowed_creator_label_regex = re.compile("|".join(allowed_created_labels))
        for identifier, creator in self._create_context.creators.items():
            try:
                if not allowed_creator_label_regex.match(creator.label):
                    self.log.debug(f"{creator.label} not allowed for context")
                    continue
```
However, that wouldn't handle the no filters case - but similar to that we could make it `None`.  just make sure to `join` the list and pattern outside of the loop instead of doing this per creator.

kalisp: I think we are sacrificing encapsulation and readability to negligent speed improvement.

BigRoy: The joining on each iteration will be relatively costly - likely slower even than your previous code with the regex match per pattern due to that caching internally per frame.

Also, the comparison should not be in the try except.


I feel like here we're not suffering that much readability actually
```python
allowed_creators_pattern = self._get_allowed_creators_pattern()
for identifier, creator in self._create_context.creators.items():
    if (
        allowed_creators_pattern is not None
        and not allowed_creators_pattern.match(creator.label)
    ):
       continue

    try:
        output[identifier] = CreatorItem.from_creator(creator)
    except Exception:
        self.log.error(
            "Failed to create creator item for '%s'",
            identifier,
            exc_info=True
        )
```

where `_get_allowed_creators_pattern()` returns `Optional[re.Pattern]`

64qam (approved):

server/settings/tools.py
kalisp: That is already handled by creating profile without any allowed creators

client/ayon_core/tools/publisher/control.py
kalisp: Updated

client/ayon_core/tools/publisher/control.py
```diff
@@ -1839,6 +1855,60 @@ def _collect_creator_items(self):
...
+    def _is_label_allowed(self, label, allowed_label_regexes):
+        """Implement regex support for allowed labels.
+
+        Args:
+            label (str): Label of creator - shown in Publisher
+            allowed_label_regexes (str): compiled regular expression
+        """
+        if not allowed_label_regexes:
+            return True
+        return bool(allowed_label_regexes.match(label))
```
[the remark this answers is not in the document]
kalisp: Updated

BigRoy (approved):
Didn't test the code, but overall the code changes look good. Might need some testing by someone still if it hasn't been thoroughly tested yet.

PR #572, opened 2024-05-30
RoyalRender: enabled toggle is ignored in royal render

kalisp:
## Changelog Description
This PR uses On/Off state of RoyalRender in Studio Settings to enable it.

## Additional info
On/Off toggle should be on RR addon as we might allow to enable different render management system (eg RR, Deadline, OpenCue) per project. But that is not implemented currently. Now it is expected single RMS for all projects

## Testing notes:
1. keep RR AND DL in a bundle 
2. keep RR toggle disabled in Studio Settings
image
3. Try to publish to DL, it shouldnt complain about RR stuff


MilaKudr (approved):
Without this PR rendering to Deadline fails, after applying it the whole publish and render to DL works correctly.

iLLiCiTiT (dismissed):

iLLiCiTiT:
I think the `enabled` state should not turn off addon, but it's functionality. The checkbox is in studio and project settings, but if you would turn it off in studio settings then project settings value is not propagated.

iLLiCiTiT:
https://github.com/ynput/ayon-core/pull/572#issuecomment-2142249877

Should we at least change scope of `enabled` to only `studio`?

PR #575, opened 2024-05-31
Maya: check for invalid reference compatible with Maya 2022

antirotor:
## Changelog Description
Check if reference is valid is using API call missing from Maya 2022 and older. This is adding backward compatibility.

## Testing notes:
- Loading of references should work in both Maya 2022 and newer


server_addon/maya/client/ayon_maya/api/lib.py
```diff
@@ -1734,6 +1734,15 @@ def is_valid_reference_node(reference_node):
...
     sel.add(reference_node)
     depend_node = sel.getDependNode(0)
+    # maya 2022 is missing `isValidReference` so the check needs to be
+    # done in different way.
+    if cmds.about(version=True) < 2023:
+        try:
+            cmds.referenceQuery(reference_node, filename=True)
+            return True
+        except RuntimeError:
+            return False
```
BigRoy, line 1744: Technically this can move even above the `sel = OpenMaya.MSelectionList` line.
```suggestion
    # maya 2022 is missing `isValidReference` so the check needs to be
    # done in different way.
    if cmds.about(version=True) < 2023:
        try:
            cmds.referenceQuery(reference_node, filename=True)
            return True
        except RuntimeError:
            return False
    sel = OpenMaya.MSelectionList()
    sel.add(reference_node)
    depend_node = sel.getDependNode(0)
```

BigRoy (approved):
Looks good to me - might need a maya 2022 test-run. @LiborBatek 🔥 ?

LiborBatek (approved):
LGTM

PR #581, opened 2024-05-31
Unreal: Standalone loader for animation

simonebarbieri:
## Changelog Description
Implemented a standalone loader for animation that do not require a layout to be loaded first.

## Additional info
Loader for Animation in Unreal is heavily based on the original solution using Blender and Unreal. Animation published from Maya doesn't have that layout json and so loading such animation will fail.


antirotor:
when https://github.com/ynput/ayon-core/pull/614 is merged, we need to reopen it in https://github.com/ynput/ayon-unreal

antirotor:
PR re-opened here https://github.com/ynput/ayon-unreal/pull/3

PR #582, opened 2024-06-03
Chore: Move Blender client code

iLLiCiTiT:
## Changelog Description
Moved blender client code next to server codebase.

## Additional information
Renamed base plugin classes to contain `Blender` prefix. Added base classes for blender instance and context pyblish plugins having `settings_category`.

## Notes
Discovered that workfile create plugin inherits from both `AutoCreator` and `Creator`, which should not happen.

## Testing notes:
1. Create package of blender and core.
2. Upload both to server.
3. Make sure AYON launcher is using client code from this PR (e.g using dev paths).
4. Blender integration should work as before.


moonyuet (approved):
Tested successfully in blender.__

PR #587, opened 2024-06-03
Chore: Change log level for missing settings on publish plugin

iLLiCiTiT:
## Changelog Description
Log `debug` for missing settings of plugin instead of `warning`.

## Additional info
After movement we will have a lot of plugins that don't have plugin settings but have filled settings category.

## Testing notes:
1. When Publisher is opened e.g. in maya it should not log a lot of warnings (but lot of debugs)


BigRoy:
I wonder if it makes sense to still log a warning if the settings category itself does not exist? But only debug log if the plug-in key does not exist (or even ignore that case even more silently?)

iLLiCiTiT:
> wonder if it makes sense to still log a warning if the settings category itself does not exist? But only debug log if the plug-in key does not exist (or even ignore that case even more silently?)

You're probably right. Settings category should still warn, not sure about missing plugin tho...

BigRoy (approved):
Didn't test this but code changes look good to me.

PR #588, opened 2024-06-03
Expand folders when searching in the folder widgets (Loader + Launcher).

BigRoy:
## Changelog Description

Expand folders when searching in the folder widgets (Loader + Launcher).

## Additional info

I'm not entirely sure whether the change implemented here is 'correct' according to the model/view separation logic we want to have in tools, but it shows the behavior artists in our studio we like more (and is logic that behaved similarly in OpenPype actually).

Without this PR when searching in the folder widget it would not expand any folders:

image

With this PR it defaults to auto-expanding all:
image

The benefit to this is that when searching you automatically can see all the results and directly can see whether you found what  you're looking for or should filter more or not.

## Testing notes:

1. Use this branch in dev mode for ayon core
2. Launcher and loader tool should now auto-expand all folders when typing in the filter field.


BigRoy:
Let me know what you think about the code @iLLiCiTiT - pointers are welcome to restructure or move this logic elsewhere.

iLLiCiTiT (approved):
I don't think there is better solution without storing information on each index "did user expand this" 🙂 .

PR #590, opened 2024-06-04
Substance Painter: Add support for sync workfile version 

moonyuet:
## Changelog Description
This PR is to add `sync workfile version` option into substance painter host.

## Additional info
Build the latest version of core addon and substance painter addon

## Testing notes:
1. Enable Follow workfile version in either `ayon+settings://core/publish/CollectAnatomyInstanceData/follow_workfile_version` 
2. Publish
3. TextureSet and Workfile version should be aligned


package.py
```diff
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "0.3.3-dev.1"
+version = "0.3.4-dev.1"
```
iLLiCiTiT, line 3:
```suggestion
version = "0.3.3-dev.1"
```

LiborBatek (approved):
I did test both occasions when ```Follow workfile version``` been set in ```Core addon``` as seen here

Screenshot 2024-06-04 175718

Small note: there is no ```hint message``` for this option as with others...we should also incorporate it for this settings too.

when published already present ```texture set``` publish instance it works and publish the ```version``` following the ```workfile version```.

Screenshot 2024-06-04 181311

It also works when the ```Core addon``` settings being disabled, but the user sets it ```ON``` when creating new ```Publish instance``` also works fine!

Overall works nicely in both occasions...

LiborBatek:
One small comment:

@moonyuet  would it be possible to keep visible that option ```Follow workfile version``` on the ```Publish Instance``` so user could potentially switch the behaviour?? not sure if possible and also if good to have it exposed for ever...

BigRoy:
> @moonyuet would it be possible to keep visible that option `Follow workfile version` on the `Publish Instance` so user could potentially switch the behaviour?? not sure if possible and also if good to have it exposed for ever...

I wouldn't actually make that an artist's choice - that's a studio choice usually. Can you think of a use case where an artist really wants to take that control?

server_addon/substancepainter/client/ayon_substancepainter/plugins/create/create_textures.py
```diff
@@ -234,9 +244,15 @@ def get_instance_attr_defs(self):
...
             attr_defs.append(
                 BoolDef("use_selection", label="Use selection",
-                        tooltip="Select Layer Stack(s) for exporting")
+                        tooltip="Select Layer Stack(s) for exporting"),
+
             )
+
+        attr_defs.append(BoolDef("follow_workfile_version",
+                                 label="Follow Workfile Version",
+                                 default=self.follow_workfile_version))
```
BigRoy, line 256: Out of curiosity - why are we making this configurable by the artist? Usually the studio define whether it's one OR the other and not up to the artist to pick and choose.

server_addon/substancepainter/client/ayon_substancepainter/plugins/publish/collect_workfile_version.py
BigRoy, line 1: Why do we need this? I haven't seen similar in Maya, Houdini integrations which are also capable of syncing to the workfile version already. I feel like this would be handled by a global plug-in instead?

server_addon/substancepainter/client/ayon_substancepainter/plugins/publish/collect_context_data.py
```diff
@@ -1,17 +1,21 @@
...
     hosts = ["substancepainter"]
 
     def process(self, context):
         host = registered_host()
         path = host.get_current_workfile()
         context.data["currentFile"] = path
         self.log.debug(f"Current workfile: {path}")
+        version = int(get_version_from_path(path))
+        context.data["version"] = version
+        self.log.debug(f"Current Version: {version}")
```
BigRoy, line 21: Logic is basically the same as the `collect_workfile_version.py` file after.
```suggestion
```
Please revert the changes to this file.

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -313,8 +313,9 @@ def fill_anatomy_data(self, context):
 
             # Define version
             version_number = None
-            if self.follow_workfile_version:
-                version_number = context.data("version")
+            if self.follow_workfile_version or instance.data.get(
+                "follow_workfile_version", self.follow_workfile_version):
+                    version_number = context.data("version")
```
BigRoy, line 318: Just wanted to remark this logic seems very similar to what I implemented here: https://github.com/ynput/ayon-core/pull/295/files#diff-7df3c84c641d94efac636cfd13f276c9a0bea4e3a35aa36e62f74f370bee12ecR317-R323

But in that case there are certain instances that can't be matched to the workfile versions and have to version independently due to how they work - as such, it's still not artist choice but a pipeline implementation detail.

server_addon/substancepainter/client/ayon_substancepainter/plugins/publish/collect_context_data.py
moonyuet: > Logic is basically the same as the `collect_workfile_version.py` file after.
> 
> Please revert the changes to this file.

I tried to search this particular python script in ayon_core but there is zero result. When I search similar scripts like `collect_version.py`, they are only lasted per host. Should we implement it in the general core addon?
image

LiborBatek:
@BigRoy I get your point, makes sense

@moonyuet  But now we can control it only globally within ```Ayon Core Addon``` with this setting ?

Screenshot 2024-06-04 181806

I think we need to allow user to set it within ```Substance Painter addon``` imho....kind of local override so it doesnt follow global settings only imho?

@BigRoy opinion on this?

moonyuet:
> @moonyuet But now we can control it only globally within `Ayon Core Addon` with this setting ?

Yes, it is

> I think we need to allow user to set it within `Substance Painter addon` imho....kind of local override so it doesnt follow global settings only imho?

@BigRoy I think we can do this too by using `instance.data["followWorkfileVersion"]`

BigRoy:
> But now we can control it only globally within `Ayon Core Addon` with this setting ?
> Screenshot 2024-06-04 181806
> 
> I think we need to allow user to set it within `Substance Painter addon` imho....kind of local override so it doesnt follow global settings only imho?

Currently this is global to all I believe, and yes - that's how we use it. If we want to make that "profile" based so that we can go more complex there I'd make that a separate issue and separate PR but it's a valid request I suppose.

BigRoy:
> But now we can control it only globally within `Ayon Core Addon` with this setting ?
> Screenshot 2024-06-04 181806

Looking at the logic now it's still in the creator - and not "globally" like in other hosts. Is it correct that you still need to implement it so that it matches e.g. the maya/houdini, etc. logic?

moonyuet:
> > But now we can control it only globally within `Ayon Core Addon` with this setting ?
> > Screenshot 2024-06-04 181806
> > I think we need to allow user to set it within `Substance Painter addon` imho....kind of local override so it doesnt follow global settings only imho?
> 
> Currently this is global to all I believe, and yes - that's how we use it. If we want to make that "profile" based so that we can go more complex there I'd make that a separate issue and separate PR but it's a valid request I suppose.
> 
> > I think we can do this too by using instance.data["followWorkfileVersion"]
> 
> I think profiles filtering logic can be used for this - but I'd not merge that into this PR.

I actually implement the logic here for temporary use. And once the `profile` setting has been finished we can make changes on 
that.

BigRoy:
> I actually implement the logic here for temporary use. And once the profile setting has been finished we can make changes on
that.

I'd make the logic here match the other integrations - so it's aligned and once we move things over it works for all. Best not to deviate here, especially if it's "temporary".

moonyuet:
> > I actually implement the logic here for temporary use. And once the profile setting has been finished we can make changes on
> > that.
> 
> I'd make the logic here match the other integrations - so it's aligned and once we move things over it works for all. Best not to deviate here, especially if it's "temporary".

I already restored the creator logic and the sync workfile version can only sync through the global settings and we can integrate profile setting later.

LiborBatek (approved):
All works as supposed...when ```Follow workfile version``` being ```Enabled```

here I have produced ```version 021``` from the ```workfile v021``` and all matches.

Screenshot 2024-06-07 102138

There is also correct ```Validate``` when trying to produce invalid / existing or lower version than already present in the DB.


Screenshot 2024-06-07 102254

LGTM!

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -313,8 +313,15 @@ def fill_anatomy_data(self, context):
...
-                version_number = context.data("version")
+
+            # Allow an instance to force enable or disable the version
+            # following of the current context
+            use_context_version = self.follow_workfile_version
+            if "followWorkfileVersion" in instance.data:
+                use_context_version = instance.data["followWorkfileVersion"]
+
+            if use_context_version:
+                    version_number = context.data("version")
```
iLLiCiTiT, line 324:
```suggestion
                version_number = context.data("version")
```

antirotor (approved):

PR #592, opened 2024-06-04
Chore: Editorial dos not use asset naming

iLLiCiTiT:
## Changelog Description
Use key `'newHierarchyIntegration"` instead of `"newAssetPublishing"`.

## Additional info
The value stored under the key defines that a new folder and tasks may be created. In that case folder entity and task entity may not be available.

## Testing notes:
1. All editorial integrations should work.

### DIFF
--- a/server_addon/traypublisher/client/ayon_traypublisher/plugins/create/create_editorial.py
+++ b/server_addon/traypublisher/client/ayon_traypublisher/plugins/create/create_editorial.py
@@ -676,6 +676,8 @@ or updating already created. Publishing will create OTIO file.
             "shotName": shot_name,
             "variant": variant_name,
             "task": None,
+            "newHierarchyIntegration": True,
+            # Backwards compatible
             "newAssetPublishing": True,
             "trackStartFrame": track_start_frame,
             "timelineOffset": timeline_offset,
--- a/client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
+++ b/client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
@@ -391,7 +391,11 @@ class CollectAnatomyInstanceData(pyblish.api.ContextPlugin):
             anatomy_data.update(folder_data)
             return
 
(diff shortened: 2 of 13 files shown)


server_addon/traypublisher/client/ayon_traypublisher/plugins/create/create_editorial.py
```diff
@@ -676,6 +676,8 @@ def _get_base_instance_data(
             "shotName": shot_name,
             "variant": variant_name,
             "task": None,
+            "newHierarchyIntegration": True,
+            # Backwards compatible
             "newAssetPublishing": True,
```
BigRoy, line 681: Should we now start removing this older data key to ensure the new one does what we want it to do?

iLLiCiTiT: You can have new traypublisher addon and old ayon-core addon...

BigRoy: Should we add a `remove backwards compatibility by version x.y.z` label to the comment so we know when to remove it in the future?

fabiaserra: can you not make use of the package `ayon_required_addons` to force a minimum version?

iLLiCiTiT: > Should we add a remove backwards compatibility by version x.y.z label to the comment so we know when to remove it in the future? Or do we prepare an issue now to remind us in say 3 months?

At this moment specific version is too "wild" guess. But I guess we can add at least date when it "started" to be deprecated?

> can you not make use of the package ayon_required_addons to force a minimum version?

You can, but core cannot. And at this moment we don't know about all unofficial addons using this.

jakubjezek001 (approved):

PR #598, opened 2024-06-05
Loader tool: Safe containers discovery

iLLiCiTiT:
## Changelog Description
Wrapped `get_containers` into try except block to avoid broken UI because of broken function implementation.

## Additional info
Added because of bugfix in [PR](https://github.com/ynput/ayon-core/pull/595). This is 3rd ocassion when loader UI was not showing anything to load because of bug in `get_containers`. Scene inventory would not work, but loader can be at least used to load.

### DIFF
--- a/client/ayon_core/tools/loader/control.py
+++ b/client/ayon_core/tools/loader/control.py
@@ -348,10 +348,18 @@ class LoaderController(BackendLoaderController, FrontendLoaderController):
             return set()
 
         if not self._loaded_products_cache.is_valid:
-            if isinstance(self._host, ILoadHost):
-                containers = self._host.get_containers()
-            else:
-                containers = self._host.ls()
+            try:
+                if isinstance(self._host, ILoadHost):
+                    containers = self._host.get_containers()
+                else:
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/loader/control.py
```diff
@@ -348,10 +348,18 @@ def get_loaded_product_ids(self):
...
-                containers = self._host.ls()
+            try:
+                if isinstance(self._host, ILoadHost):
+                    containers = self._host.get_containers()
+                else:
+                    containers = self._host.ls()
+
+            except BaseException:
+                self.log.error(
+                    "Falied to collect loaded products.", exc_info=True
```
BigRoy, line 359: Typo:
```suggestion
                    "Failed to collect loaded products.", exc_info=True
```

I'm also wondering whether mentioning "containers" instead of "products" may be clearer here?
```suggestion
                    "Failed to collect loaded containers.", exc_info=True
```

Anyway, pick your juice.

BigRoy (approved):
Didn't test it - but code looks good aside from a typo in the logged message.

PR #600, opened 2024-06-05
Chore: Create server addon package enhancements

iLLiCiTiT:
## Changelog Description
Create package script does not copy files around unless it was asked for. So if zip should be created, only file which is created on disk is output package zip.

## Additional info
Do most of the logic only in memory without copying files around. This means that `--keep-sources` argument does not have usage as there are no sources. But `--skip-zip` still can be used.

Changed docstring in houdini `version.py` (the template changed while PR was opened).

## Testing notes:
1. Running `python ./create_ayon_addon.py` should output the same packages as before but a little bit faster.

### DIFF
--- a/server_addon/create_ayon_addons.py
+++ b/server_addon/create_ayon_addons.py
@@ -1,3 +1,4 @@
+import io
 import os
 import sys
 import re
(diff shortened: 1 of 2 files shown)


MustafaJafar (approved):
It works and it's faster. 
I didn't notice any difference.

PR #602, opened 2024-06-06
Substance Painter: add review attributes in texture creator

moonyuet:
## Changelog Description
This PR is to add review attributes support in texture creator so that the texture can be published as thumbnail for review.

## Additional info
n/a

## Testing notes:
1. Build the substance addon with this branch or use` {YOUR AYON_CORE REPO}\server_addon\substancepainter\client` in developement mode with this branch
2. Launch Substance Painter
3. There would be review options for users to set in both Create and Publish Tab
image
4. Publish


LiborBatek (approved):
All seems working as expected.

LGTM


Screenshot 2024-06-10 141751
Screenshot 2024-06-10 141751
Screenshot 2024-06-10 141751

PR #607, opened 2024-06-06
Chore: Remove Python 2 vendor

iLLiCiTiT:
## Changelog Description
Removed Python 2 modules from vendor.

## Additional info
Moved content of `client/ayon_core/vendor/python/common/` to `client/ayon_core/vendor/python/` as we don't need Python 2/3 separation.

## Testing notes:
All modules from vendor should be working.
1. Loaders using `qargparse` (Loader plugins with options e.g. `LoadImage` in nuke).
2. Custom menu in hosts using `scriptsmenu` are loaded in nuke, hiero and maya (not sure how exactly).


antirotor (approved):
quickly tested scriptsmenu inmaya, all seems to be fine

PR #609, opened 2024-06-07
After Effect: multiselection for file loading into the folder

moonyuet:
## Changelog Description
This PR is to fix the issue when loading multiple files into the selected folder and only the first loaded assets get into it.

## Additional info
n/a

## Testing notes:
1. Build the aftereffect addon with this branch or use {YOUR_AYON_CORE_REPO}\server_addon\aftereffects\client\ as addon directory
2. Install addon
3. Launch After Effects -> Load
4. Create New folder
5. Select the folder
6. load file from the loader
7. load another file from the loader again (or select all the assets and load at the same time)
8. Should be all added into the folder.


iLLiCiTiT:
Please bump aftereffects version.

64qam:
Loaded files still end up in root outside of folders. @LiborBatek could you please also test it?

64qam (approved):
Working now. Problem between keyboard and chair

server_addon/aftereffects/client/ayon_aftereffects/plugins/load/load_file.py
```diff
@@ -51,7 +55,7 @@ def load(self, context, name=None, namespace=None, data=None):
 
         comp = stub.import_file(path, stub.LOADED_ICON + comp_name,
                                 import_options)
-
+        print("comp", comp)
```
moonyuet, line 58:
```suggestion
```

PR #611, opened 2024-06-07
Chore: move job queue client code

iLLiCiTiT:
## Changelog Description
Moved JobQueue code next to server codebase.

## Additional information
Removed `aiohttp_json_rpc` dependency from ayon-core client dependencies as core is not using it anywhere.

## Testing notes:
The addon is probably not used in production. The setup to test the changes is too complicated.


iLLiCiTiT:
Merging without approval because the addon does not have production value at this moment.

PR #615, opened 2024-06-07
Pass status when integrating version

fabiaserra:
## Changelog Description
This passes the `status` ~~and `author`~~(now resolved at https://github.com/ynput/ayon-core/commit/43787fd9c39f16bc8bf75d401b038a5a2fb56988) args to the creation of new version on integrate. Without passing `author` when integrating in Deadline it was defaulting to the `service` user

~~I also changed the `collect_current_ayon_user` plugin to this, although looking back now I'm not sure if that was necessary or it was just a matter of exposing the `author`.~~
```python
        user = os.getenv("AYON_USERNAME")
        if not user:
            user = get_ayon_username()
```

## Additional Information
Needless to say that currently the `instance.data["status"]` isn't being set by any of the collect plugins in AYON but it would be as simple as adding that in order for this to take any effect. In my situation, I'm setting that over the integrate_shotgrid_version of the `ayon-shotgrid` as there's some other data that I need from the SG integration to also add to the AYON version entity. This is the code that I'm running in my `ayon-shotgrid:integrate_shotgrid_version`:
```
        # Update AYON version entity with updated data
        update_data = {
            "attrib": {
                "shotgridId": str(version_id),
                "shotgridType": "Version"
            },
        }

        # We need to map the SG status short name to the long name as AYON
        # expects that longer name
        project_statuses = ayon_api.get_project(project_name)["statuses"]
        for proj_status in project_statuses:
            if proj_status["shortName"] == sg_status:
                update_data["status"] = proj_status["name"]
                break

        op_session = OperationsSession()
        version_entity = instance.data["versionEntity"]
        op_session.update_entity(
            instance.context.data["projectName"],
            "version",
            version_entity["id"],
            update_data
        )
        op_session.commit()
        self.log.info("Updated AYON version with %s", update_data)
```
## Testing notes:
1. Publish in Deadline
2. Check the author


client/ayon_core/plugins/publish/integrate.py
```diff
@@ -380,29 +380,29 @@ def prepare_product(self, instance, op_session, project_name):
...
             )
             if product_group is not None:
-                attribibutes["productGroup"] = product_group
+                attributes["productGroup"] = product_group
 
         product_id = None
         if existing_product_entity:
             product_id = existing_product_entity["id"]
-
+
```
iLLiCiTiT, line 399:
```suggestion

```

iLLiCiTiT:
I think we should rather use ayon api functionality to behave as other user during farm publishing, rather than changing coller user functionality. Can you remove the author from kwargs in this PR, we can create PR that will resolve that.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -464,6 +463,8 @@ def prepare_version(
             version_number,
             product_entity["id"],
             task_id=task_id,
+            author=instance.context.data.get("user"),
```
fabiaserra, line 466: @iLLiCiTiT we still need this to be passed as arg or will the user be collected implicitly (again, because `collect_current_ayon_user.py` does it already)?

iLLiCiTiT: It should not be necessary. When service works as other user, server should auto-fill the value correctly.

fabiaserra: I'd prefer to be explicit tbh

iLLiCiTiT: > I'd prefer to be explicit tbh

Well, version has `author` but product, representation and thumbnail has `createdBy` which is almost equivalent. I would rather prefer to not use the value as it might hide unresolved issues at other places.

fabiaserra: done at https://github.com/ynput/ayon-core/pull/615/commits/f6e3dcb4742e083664744de016476d8c134f5ef0

iLLiCiTiT (approved):

client/ayon_core/plugins/publish/integrate.py
iLLiCiTiT: BTW have you tested it?

fabiaserra: not your changes yet no, I'm still relying on my explicit `collect_current_ayon_user` of my branch, I can test it later this afternoon

PR #620, opened 2024-06-10
Chore: Transcode color spaces plugin does not remove all representations

iLLiCiTiT:
## Changelog Description
Plugin `Transcode color spaces` does remove only representations that were processed by the plugin.

## Testing notes:
1. Not having profiles for specific context in `Transcode color spaces` should not cause that all reviews are skipped


antirotor (approved):

LiborBatek (commented):
Already merged but worked for me too...

PR #621, opened 2024-06-10
Chore: Applications addon decouple of logic

iLLiCiTiT:
## Changelog Description
Moved some amount of applications logic to applications addon. Core still has to know some logic of applications addon but not all of it.

## Additional info
The biggest disadvantage is that core addon has to know how to create action classes, their icons, labels if are enabled/disabled tc. which should be in charge of the applications addon. Applications addon is also providing the icons.

Selection object in launcher has support to get project settings.

Applications addon did not yet change the logic, only has it available, so old ayon core will still work, and new ayon-core can work with both new and old applications addon versions.

Because we're planning how applications addon work this is "must have" PR.

## Testing notes:
1. Applications actions show icons in launcher tool.
2. Works with new core and new applications addon.
3. Works with old core and new applications addon.
4. Works with new core and old applications addon.
5. Works with old core and old applications addon (behavior before this PR).


64qam (approved):
Looks good, but someone else should also test it.

antirotor (approved):
seems to work fine, not tested it in all combinations though

PR #631, opened 2024-06-11
Max: fix the incorrect fps value when loading point cache

moonyuet:
## Changelog Description
This PR is to fix the incorrect fps value when loading point cache in Max.

## Additional info
Note: With the change in this PR whenever loading a pointcache the workfile's FPS will also be set to the FPS of the loaded pointcache. As such, this affects the FPS of your workfile.

## Testing notes:
1. Launch Max 
2. Load Point Cache
3. Frame Range should be correct


server_addon/max/client/ayon_max/api/lib.py
```diff
@@ -284,6 +282,15 @@ def reset_frame_range(fps: bool = True):
         frame_range["frameStartHandle"], frame_range["frameEndHandle"])
 
 
+def set_fps():
+    """Set fps to current folder
+    """
+    task_entity = get_current_task_entity()
+    task_attributes = task_entity["attrib"]
```
iLLiCiTiT, line 289:
```suggestion
def set_fps():
    """Set fps based on task."""
    task_entity = get_current_task_entity()
    if not task_entity:
        return
    task_attributes = task_entity["attrib"]
```

server_addon/max/client/ayon_max/plugins/load/load_pointcache.py
```diff
@@ -31,6 +31,8 @@ def load(self, context, name=None, namespace=None, data=None):
         file_path = self.filepath_from_context(context)
         file_path = os.path.normpath(file_path)
 
+        set_fps()
```
BigRoy, line 34: Wait a second - this doesn't seem like the right fix? You don't want to force change the FPS of your scene on loading something?

server_addon/max/client/ayon_max/api/lib.py
```diff
@@ -284,6 +282,16 @@ def reset_frame_range(fps: bool = True):
         frame_range["frameStartHandle"], frame_range["frameEndHandle"])
 
 
+def set_fps():
```
BigRoy, line 285: I'm personally quite confused that a function `set_fps` does not take arguments. `set_timeline` does, `set_render_frame_range` does - why not here?

In maya for example the equivalent to this in AYON is:
```python
set_scene_fps(get_fps_for_current_context())
```
See [here](https://github.com/ynput/ayon-core/blob/a3aed0d4b7db251fd1db56f3d89e1eaa87200a76/server_addon/maya/client/ayon_maya/api/lib.py#L2533).

server_addon/max/client/ayon_max/plugins/load/load_pointcache.py
moonyuet: But the fps value does change the custom frame range of the alembic object. let's say you have exported the alembic with 1001 to 1010 as animation range. And when you load back the point cache into the scene, it will be loaded into frame range between 1201 to 1210, which are 200 frames more than the original export frame range. 
And wIth the fix here, the point cache loaded in the current animation range. 
Would be great if you can suggest any better solution to fix it.

BigRoy: Interesting - so where does the shift in range come from? Is it solely the fact that your scene's FPS is different? (Since that's all you're setting?)

A difference from 1001 to 1201 start frame - is that a 25 FPS / 30 FPS difference issue?

Could you make a screen recording of the issue itself? Unfortunately I know to little of the issue and 3DsMax in general to provide good pointers in this area.

I'd say that if the issue is only on LOAD then we should instead use the FPS of the loaded product (not that of the current task) and then maybe set it temporarily only during the load:

In the simplest form something along the lines of:
```python
product_fps = float(context["version"]["attrib"].get("fps")
if product_fps is None:
    # Just stick to current scene FPS
    product_fps = current_fps()

with scene_fps(product_fps):
    load_file(alembic)
```

LiborBatek (approved):
Been tested and LGTM

server_addon/max/client/ayon_max/api/lib.py
```diff
@@ -284,6 +285,39 @@ def reset_frame_range(fps: bool = True):
...
+        fps = folder_entity.get("attrib", {}).get("fps")
+        if not fps:
+            project_entity = ayon_api.get_project(
+                project_name, fields=["attrib.fps"]
+            ) or {}
+            fps = project_entity.get("attrib", {}).get("fps")
+
+            if not fps:
+                fps = 25
+    return fps
```
BigRoy, line 318: As far as I know in AYON task entities should **always** have the FPS attrib so there should never be a need for a fallback? I've asked this a few times and I believe @iLLiCiTiT confirmed that before.

But if that's the case - this would just be:
```python
task_entity = get_current_task_entity(fields={"attrib"})
return task_entity["attrib"]["fps"]
```

server_addon/max/client/ayon_max/plugins/load/load_pointcache.py
```diff
@@ -31,45 +31,50 @@ def load(self, context, name=None, namespace=None, data=None):
...
-            abc_objects.append(abc_object)
-        # rename the abc container with namespace
-        abc_container_name = f"{namespace}:{name}"
-        abc_container.name = abc_container_name
-        abc_objects.append(abc_container)
+        product_fps = float(context["version"]["attrib"].get("fps"))
+        if product_fps is None:
+            # Just stick to current scene FPS
+            product_fps = float(rt.frameRate)
+        with scene_fps(product_fps):
```
BigRoy, line 39: Preferably with this fix you wrap only the "smallest" amount of code that's needed for this fix.

In this case I'd expect you'd only need to do:

```python
with scene_fps(product_fps):
    rt.importFile(file_path, 
                         rt.name("noPrompt"), 
                         using=rt.AlembicImport)
```

server_addon/max/client/ayon_max/api/lib.py
```diff
@@ -573,3 +607,17 @@ def suspended_refresh():
...
+    """Set scene fps during context
+
+    Args:
+        fps (float): fps value
+    """
+    try:
+        rt.frameRate = float(get_fps_for_current_context())
+        yield
+    finally:
+        rt.frameRate = fps
```
BigRoy, line 623: Logic here is wrong.

Should be something like:

```suggestion
    fps_before = rt.frameRate
    try:
        rt.frameRate = float(fps)
        yield
    finally:
        rt.frameRate = fps_before
```

LiborBatek (changes requested):
Hmm doesnt work for me atm...not sure why but when Im having workfile with ```NTSC``` fps aka ```30``` and my ABC ```pointcache``` being ```25``` it still offsets to ```1201-xxx```  instead of ```1001```

server_addon/max/client/ayon_max/api/lib.py
iLLiCiTiT: Task will always have `"fps"`.

server_addon/max/client/ayon_max/plugins/load/load_pointcache.py
```diff
@@ -31,13 +31,21 @@ def load(self, context, name=None, namespace=None, data=None):
         file_path = self.filepath_from_context(context)
         file_path = os.path.normpath(file_path)
 
+        rt.frameRate = get_fps_for_current_context()
```
BigRoy, line 34: Pretty sure this does nothing currently. And if we're fixing this in a different way later this should also not be here - we don't want to enforce the frame rate of the scene to current task on loading/updating something into the scene.

server_addon/max/client/ayon_max/plugins/load/load_pointcache.py
```diff
@@ -7,7 +7,7 @@
 import os
 from ayon_core.pipeline import load, get_representation_path
 from ayon_max.api import lib, maintained_selection
-from ayon_max.api.lib import unique_namespace
+from ayon_max.api.lib import unique_namespace, get_fps_for_current_context
```
BigRoy, line 10:
```suggestion
from ayon_max.api.lib import unique_namespace
```

BigRoy:
Can we update the description of the PR so that it explains HOW this temp fix fixes it, by adding for example:

**Note:** With the change in this PR whenever loading a pointcache the workfile's FPS will also be set to the FPS of the loaded pointcache. As such, this affects the FPS of your workfile.

moonyuet:
I made another update on this PR as I find out that somehow the loaded asset still has incorrect frame range with merely setting up fps value with context. And the frame range would be accurate when the frame range and fps value have been reset. 

For the next stage, we need to implement a check on the fps and frame range during launching 3dsmax

LiborBatek (approved):
Now working well...my ```pointcache``` having correct ```frame range``` even when FPS mismatched in active workfile.

image

PR #633, opened 2024-06-11
Tray: Cleanup PYTHONPATH from existing addons on update

iLLiCiTiT:
## Changelog Description
Tray is cleaning up PYTHONPATH on update and restart.

## Additional info
It looks like current `ayon-core` stays in PYTHONPATH on update & restart.

## Testing notes:
NOTE: The fix will work from this version up.
1. Create package of ayon core from this PR and upload it to server (**Package 1**).
2. Bump ayon core to higher version and create package again and upload the package to server too (**Package 2**).
3. Create production bundle with the **Package 1**.
4. Run tray in production bundle.
5. Create new bundle with **Package 2** and set it to production.
6. Wait for tray to ask for update and restart.
7. Do update and restart.
8. Open admin console to find out path from where is ayon core used. It should point to **Package 2** code.
```python
print(AYON_CORE_ROOT)
```


jakubjezek001 (approved):
LGTM!

PR #639, opened 2024-06-11
Traypublisher: Actually enable Override Colorspace collector when it should be enabled

BigRoy:
## Changelog Description

Fix CollectColorspace plug-in enabled state for tray publisher. Without this change it never shows since `enabled` is always False.

## Additional info

With this fix it shows again and seems to work:
image

## Testing notes:

1. Enable color management
2. Publish image or plate via traypublisher
3. The instance should have an "override colorspace" option
4. The overridden colorspace should work to set the interpreted published colorspace of the file.


iLLiCiTiT:
Please bump version.

BigRoy:
> Please bump version.

Done.

iLLiCiTiT (approved):

PR #640, opened 2024-06-12
Workfiles tool: Artists can see users

iLLiCiTiT:
## Changelog Description
Fixed implementation of users model in workfiles tool.

## Additional info
Only admin and manager users can see all users, other users can see only other users assigned to same projects. It is necessary to add project name filtering to users GraphQl, which requires fix in ayon-python-api, for now the code from ayon api was copied next to users model.

## Testing notes:
1. Create artist user and assign him to a project.
2. Start AYON launcher as the user.
3. Use host integration of your choice and open workfiles tool.
4. It should not crash, and should show user full name.

### DIFF
--- a/client/ayon_core/tools/common_models/users.py
+++ b/client/ayon_core/tools/common_models/users.py
@@ -143,7 +143,7 @@ class UsersModel:
             Union[UserItem, None]: User item or None if not found.
 
         """
-        self._invalidate_cache()
+        self._invalidate_cache(project_name)
         for user_item in self.get_user_items(project_name):
             if user_item.username == username:
                 return user_item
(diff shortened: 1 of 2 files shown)


fabiaserra:
I can confirm this now works, thank you @iLLiCiTiT !

client/ayon_core/tools/common_models/users.py
```diff
@@ -70,15 +144,21 @@ def get_user_item_by_username(self, username):
 
         """
         self._invalidate_cache()
```
fabiaserra, line 146:
```suggestion
        self._invalidate_cache(project_name)
```

jakubjezek001 (approved):
works well in Nuke

PR #641, opened 2024-06-12
Maya: Pointcache create plugins fix

iLLiCiTiT:
## Changelog Description
Pointcache does not append attr definitions to base list but to new list.

## Testing notes:
1. Create pointcache in maya.
2. Workfile instance should not suddenly have attributes.


BigRoy (approved):
Yes, thank you very much! Works for me.

I have a feeling that this bug may have been hiding some worse issues so it's good to find the root of it and have it solved.

PR #642, opened 2024-06-12
Loader tool: Hero version is smaller than standard version

iLLiCiTiT:
## Changelog Description
Change order of hero version in versions combobox so hero version is under standard version.

## Additional info
When hero version is higher then hero version is first option to use for load which is unconventional.

## Testing notes:
1. Loader tool should preselect last standard version even if there is hero version with same value.


LiborBatek (approved):
Been tested and all works fine. The ```Hero version``` doesnt take precedence of ordinary ```versions```.

Works well.

image

PR #643, opened 2024-06-12
Scene Inventory: remove unused container attribute

antirotor:
## Changes

`name` isn't used anywhere but it is hard required, causing crashes in hosts that are not having it. Also adding broad try/except to catch any invalid containers so they won't crash the whole list.

## Testing
Scene inventory should list loaded stuff even without `name` - currently this affects Unreal as far as I know.

### DIFF
--- a/client/ayon_core/tools/sceneinventory/models/containers.py
+++ b/client/ayon_core/tools/sceneinventory/models/containers.py
@@ -90,7 +90,6 @@ class ContainerItem:
         representation_id,
         loader_name,
         namespace,
-        name,
         object_name,
         item_id
     ):
@@ -98,7 +97,6 @@ class ContainerItem:
         self.loader_name = loader_name
         self.object_name = object_name
         self.namespace = namespace
-        self.name = name
         self.item_id = item_id
 
     @classmethod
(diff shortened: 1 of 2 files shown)


client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -329,14 +326,23 @@ def _update_cache(self):
...
+                item = ContainerItem.from_container_data(container)
+                containers_by_id[item.item_id] = container
+                container_items_by_id[item.item_id] = item
+                container_items.append(item)
+            except Exception as e:
+                # skip item if required data are missing
+                self._controller.log_error(
+                    f"Failed to create item: {e}"
+                )
+                continue
```
iLLiCiTiT, line 344: Or this
```suggestion
            try:
                item = ContainerItem.from_container_data(container)
            except Exception as e:
                # skip item if required data are missing
                self._controller.log_error(
                    f"Failed to create item: {e}"
                )
                continue

            containers_by_id[item.item_id] = container
            container_items_by_id[item.item_id] = item
            container_items.append(item)
```

iLLiCiTiT (approved):

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -329,15 +326,25 @@ def _update_cache(self):
...
+
             containers_by_id[item.item_id] = container
             container_items_by_id[item.item_id] = item
             container_items.append(item)
 
+
```
iLLiCiTiT, line 347:
```suggestion
```

PR #646, opened 2024-06-12
Chore: Burnin script shell execution

iLLiCiTiT:
## Changelog Description
Try to auto-fix usage of `shell=True` in burnin script.

## Additional info
In some cases the shell set to `True` might cause that the subprocess don't have access to the executable. Ideal case would be to not use `shell=True` at all, that would require more modifications. To avoid it we can run "testing" ffmpeg command and change it to `False` if it crashes.

### DIFF
--- a/client/ayon_core/scripts/otio_burnin.py
+++ b/client/ayon_core/scripts/otio_burnin.py
@@ -14,9 +14,10 @@ from ayon_core.lib import (
     convert_ffprobe_fps_value,
 )
 
+FFMPEG_EXE_COMMAND = subprocess.list2cmdline(get_ffmpeg_tool_args("ffmpeg"))
 FFMPEG = (
     '{}%(input_args)s -i "%(input)s" %(filters)s %(args)s%(output)s'
-).format(subprocess.list2cmdline(get_ffmpeg_tool_args("ffmpeg")))
+).format(FFMPEG_EXE_COMMAND)
 
 DRAWTEXT = (
     "drawtext@'%(label)s'=fontfile='%(font)s':text=\\'%(text)s\\':"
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Works at client site.

client/ayon_core/scripts/otio_burnin.py
```diff
@@ -482,10 +483,19 @@ def render(self, output, args=None, overwrite=False, **kwargs):
...
+            )
+            test_proc.wait()
+        except BaseException:
+            use_shell = False
+
         kwargs = {
             "stdout": subprocess.PIPE,
             "stderr": subprocess.PIPE,
-            "shell": True,
+            "shell": use_shell,
```
BigRoy, line 498: Don't these subprocess calls technically mean they may show up a command prompt dialog on Windows? Shouldn't these have that "hide console window" special flag?

Also, it's really annoying we have to do this because these subprocess calls startup always take longer than I'd like. I mean it's better than failing.

Maybe we could cache the "should we run in shell" result for the running process? (Does that even make sense since this would be a script that would maybe already itself be running in a subprocess maybe?)

iLLiCiTiT: > Also, it's really annoying we have to do this because these subprocess calls startup always take longer than I'd like. I mean it's better than failing.
>
> Maybe we could cache the "should we run in shell" result for the running process? (Does that even make sense since this would be a script that would maybe already itself be running in a subprocess maybe?)

This is called only once during the process lifetime (single extract burning process).

Ideal fix is to rewrite the logic so output is list of arguments, in that case we can get rid of `shell=True`.

BigRoy: Yup - thanks for confirming.

PR #647, opened 2024-06-13
Workfiles tool: Context selection is properly propagated to workarea

iLLiCiTiT:
## Changelog Description
Selection of context in workfiles tools does correctly resolve workfile entity from AYON and new workfile can save note.

## Additional info
When context is changed from a folder 1 to folder 1 and they have the same tasks then there is a chance that workfile entity is not loaded for correct task. And when new workfile is created, it was not possible to change note because we didn't have access to workfile entity id.

### New workfile
1. Create new workfile using workfiles tools.
2. Set note on the workfile and hit save.
3. It should save the note.

### Context change
1. Make sure you have 2 folders with same tasks.
2. Open last workfile on one of the tasks on folder 1.
3. Open workfiles tool.
4. Set some note on the workfile (this is how we find out it works).
5. Change folder to folder 2 (The task should stay selected).
6. Change folder to folder 1 (The task should stay selected).
7. The last workfile should stay selected and the same note from step 4 should be visible in side panel.

### DIFF
--- a/client/ayon_core/tools/workfiles/models/workfiles.py
+++ b/client/ayon_core/tools/workfiles/models/workfiles.py
@@ -173,7 +173,7 @@ class WorkareaModel:
         folder_mapping[task_id] = workdir
         return workdir
 
-    def get_file_items(self, folder_id, task_id):
+    def get_file_items(self, folder_id, task_id, task_name):
         items = []
         if not folder_id or not task_id:
             return items
@@ -192,7 +192,7 @@ class WorkareaModel:
                 continue
 
             workfile_info = self._controller.get_workfile_info(
-                folder_id, task_id, filepath
+                folder_id, task_name, filepath
             )
             modified = os.path.getmtime(filepath)
             items.append(FileItem(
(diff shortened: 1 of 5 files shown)


client/ayon_core/tools/workfiles/models/workfiles.py
```diff
@@ -587,6 +588,7 @@ def _create_workfile_info_entity(self, task_id, rootless_path, note):
 
         username = self._get_current_username()
         workfile_info = {
+            "id": uuid.uuid4().hex,
```
BigRoy, line 591: What does this resolve? Wouldn't the server solve this for us? Or do we need the `id` in our stored workfile info somewhere so we can update it easily without requiring to query again?

iLLiCiTiT: > Or do we need the id in our stored workfile info somewhere so we can update it easily without requiring to query again?


Yep, so we can update.

BigRoy (approved):
No more errors - seems to work as intended - lovely!

However, I've had some cases now when clicking on another folder path where it doesn't show me its tasks... had to reclick it before it would show. As if it just doesn't find the tasks in-time on the folder switch. Other than that it worked perfectly!

Also wanted to mention a caveat that I'm not too sure of. When changing a Note on an existing workfile it also updates the Author to the current user.... whereas I actually have not created the file. I wonder, when doing that whether we should avoid ALSO updating the author of the file but only update the note which is what I'd expect to happen instead. So that might be worth a separate issue.

64qam (approved):

PR #649, opened 2024-06-14
Scene Inventory tool: Set version dialog can change version for whole selection

iLLiCiTiT:
## Changelog Description
Scene inventory can change version for whole selection, but does not show statuses next to version if are from different products.

## Additional info
Sometimes it might be helpfull to change version for multiple loaded products at the same time. But we're not able to show status in that case, lazy solution is to hide the statuses to avoid confusion.

## Testing notes:
1. It is possible to change version of multiple containers from different products in scene manager tool.


LiborBatek (approved):
Works like charm!

Before:
Manage_ImageTex

After this PR (fixed):
Manage_ImageTex_Fixed

PR #651, opened 2024-06-14
Maya: Implementation of Ornatrix family 

moonyuet:
## Changelog Description
Implementation of Ornatrix Exports in Maya
Ornatrix Rig: Publishing Ornatrix or any Ornatrix related objects. If there are any painted map(s) by user, they are published too and the published maps would be located in the `{your publish folder}/resources`
Ornatrix Cache: Publishing Ornatrix fur stimulation as ornatrix alembic cache. It can be loaded as Arnold StandIn, Vray Proxy and Alembic reference while it can be loaded as the file cache in the Ornatrix-related GuideFromMeshNode. 

## Additional info
Although Ornatrix cache loader can add HairfromGuideNode into the operator stack, it can only be showed once into the stack editor(due to some api limitation". Users need to select the objects they want to load the file cache in order to add the cached nodes "safely" into the operator stack.

- [x] Load Ornatrix Groom
- [x] Validate Ornatrix content
- [x] Connect Ornatrix Groom

## Testing notes:
Ornatrix Rig
1. Create Ornatrix Rig with your geometry which is assigned with ornatrix fur
2. Publish
3. Texture maps and ma scene would be published along with rigsettings which stores all the ornatrix related parameters.
4. Start new scene and load the reference scene of the published ornatrix rig
5. Once you start publisher, the ornatrix cache will be automatically created with the references of the ornatrix rig.
6. If you load the ornatrix cache, it will create group with a mesh shape which includes the GuideFromMeshNode storing the abc file cache.
7. You can also load the ornatrix cache with Arnold StandIn, Vray Proxy and Alembic reference


client/ayon_core/plugins/publish/collect_resources_path.py
```diff
@@ -64,7 +64,8 @@ class CollectResourcesPath(pyblish.api.InstancePlugin):
                 "skeletalMesh",
                 "xgen",
                 "yeticacheUE",
-                "tycache"
+                "tycache",
+                "OxRig"
```
BigRoy, line 68: No other families seem to start with a a capital letter? Should this also start with lowercase?

server_addon/maya/client/ayon_maya/plugins/create/create_ornatrix_cache.py
```diff
@@ -0,0 +1,75 @@
...
+    lib,
+    plugin
+)
+from ayon_core.lib import BoolDef, NumberDef
+
+
+class CreateOxCache(plugin.MayaCreator):
+    """Output for procedural plugin nodes of Ornatrix """
+
+    identifier = "io.openpype.creators.maya.OxCache"
```
BigRoy, line 11: This creator will need to have matching settings to allow being disabled/enabled per studio - likely even disabled by default since I don't think Ornatrix comes with Maya by default?

server_addon/maya/client/ayon_maya/plugins/create/create_ornatrix_rig.py
```diff
@@ -0,0 +1,27 @@
...
+from ayon_maya.api import (
+    lib,
+    plugin
+)
+
+
+class CreateOxRig(plugin.MayaCreator):
+    """Output for Ornatrix nodes"""
+
+    identifier = "io.openpype.creators.maya.OxRig"
```
BigRoy, line 12: Should have studio settings to allow enable/disable this creator.

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_rig.py
```diff
@@ -0,0 +1,121 @@
...
+
+        with lib.maintained_selection():
+            cmds.select(ox_geo_nodes + [ox_input], noExpand=True)
+            create_context.create(
+                creator_identifier=creator_identifier,
+                variant=variant,
+                pre_create_data={"use_selection": True}
+            )
+
+
```
BigRoy, line 102: Please check code style for your files.
```suggestion
```

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_rig.py
```diff
@@ -0,0 +1,121 @@
...
+        Args:
+            namespace (str): namespace
+            path (str): published filepath
+        """
+        _, maya_extension = os.path.splitext(path)
+        settings_path = path.replace(maya_extension, ".rigsettings")
+        with open(settings_path, "r") as fp:
+            image_attributes = json.load(fp)
+            fp.close()
```
BigRoy, line 114: Closing is not needed because the file will be closed at the end of the `with:` statement.
```suggestion
```

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_rig.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def use_resources_textures(self, namespace, path):
+        """Use texture maps from resources directories
+
+        Args:
+            namespace (str): namespace
+            path (str): published filepath
+        """
+        _, maya_extension = os.path.splitext(path)
+        settings_path = path.replace(maya_extension, ".rigsettings")
```
BigRoy, line 111: This seems like an odd way to go about this - why "replace" it if you already have the filepath without extension?

```suggestion
        path_no_ext, _ = os.path.splitext(path)
        settings_path = f"{path_no_ext}.rigsettings"
```

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_rig.py
```diff
@@ -0,0 +1,121 @@
...
+        with open(settings_path, "r") as fp:
+            image_attributes = json.load(fp)
+            fp.close()
+        if image_attributes:
+            for image_attribute in image_attributes:
+                texture_attribute = "{}:{}".format(
+                    namespace, image_attribute["texture_attribute"])
+                cmds.setAttr(texture_attribute,
+                             image_attribute["destination_file"],
+                             type="string")
```
BigRoy, line 121:
```suggestion
        if not image_attributes:
            return
        for image_attribute in image_attributes:
            texture_attribute = "{}:{}".format(
                namespace, image_attribute["texture_attribute"])
            cmds.setAttr(texture_attribute,
                         image_attribute["destination_file"],
                         type="string")
```

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_cache.py
```diff
@@ -0,0 +1,38 @@
+import pyblish.api
+from ayon_maya.api import lib
+from ayon_maya.api import plugin
+from maya import cmds
+
+
+class CollectOxCache(plugin.MayaInstancePlugin):
+    """Collect all information of the Ornatrix caches
+
+    """
```
BigRoy, line 10:
```suggestion
    """Collect all information of the Ornatrix caches"""
```

server_addon/maya/client/ayon_maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -15,6 +15,12 @@
             "cbId"}
 
 
+ORNATRIX_NODES = {
+    "HairFromGuidesNode", "GuidesFromMeshNode",
+    "MeshFromStrandsNode", "SurfaceCombNode"
+}
```
BigRoy, line 21: It's quite weird that now suddenly we have Ornatrix logic in a file related to Yeti? Seems wrong?

server_addon/maya/client/ayon_maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -304,3 +310,84 @@ def _replace_tokens(self, strings):
...
+            return []
+
+        ox_nodes = [
+            ox_node for ox_node in cmds.listConnections(node_shape, destination=True)
+                    if cmds.nodeType(ox_node) in ORNATRIX_NODES
+        ]
+        ox_imageFile = [
+            ox_img for ox_img in cmds.listConnections(ox_nodes, destination=False)
+            if cmds.nodeType(ox_img) == "file"
+        ]
```
BigRoy, line 360: Instead of checking nodetype one by one you're better of just doing a `cmds.ls(type=ORNATRIX_NODES)` query. UNLESS that node type does not exist (e.g. plug-in is not available?) In that case you may first need to check also if the plug-in is loaded, etc.

server_addon/maya/client/ayon_maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -304,3 +310,84 @@ def _replace_tokens(self, strings):
...
+        resources = []
+        node_shape = cmds.listRelatives(node, shapes=True)
+        if not node_shape:
+            return []
+
+        ox_nodes = [
+            ox_node for ox_node in cmds.listConnections(node_shape, destination=True)
+                    if cmds.nodeType(ox_node) in ORNATRIX_NODES
+        ]
+        ox_imageFile = [
```
BigRoy, line 357: Variable name doesn't seem to adhere to `PEP8`?

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_rig.py
```diff
@@ -0,0 +1,113 @@
...
+
+import os
+import json
+
+from ayon_maya.api import lib
+from ayon_maya.api import plugin
+from maya import cmds
+
+
+
```
BigRoy, line 12: Check your code linter ;)
```suggestion
```

server_addon/maya/client/ayon_maya/plugins/publish/collect_yeti_rig.py
moonyuet: The texture logics from yeti is also applicable for ornatrix too. maybe we can somehow implement the relevant logic in `lib.py` and use it here. (We can separate into two collectors).

BigRoy: Yup - either make it abstract and reusable - or if it's still very specific to these two plug-ins maybe change just the filename. Up to you. :)

server_addon/maya/client/ayon_maya/plugins/publish/collect_yeti_rig.py
moonyuet: i have tried with making the single list with cmds.ls, but it can find all the ornatrix-related nodes instead of being those who connected to the geometry which needs to be published.

moonyuet: > Collaborator

The second one works but the first one would error out as the flag only accepts string value.

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_cache.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Returns:
+            _type_: _description_
+        """
+        nodes = []
+        orig_guide_name = node_settings["name"]
+        orig_shape_name = node_settings["shape"]["name"]
+        mesh_shape_name = "{}:{}".format(namespace, orig_shape_name)
+        guide_name = "{}:{}".format(namespace, orig_guide_name)
+        mesh_shape_node = cmds.ls(guide_name, type="mesh")
```
BigRoy, line 137: This would be a `List[str]` which seems odd with the variable name, bug?

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_cache.py
```diff
@@ -0,0 +1,164 @@
...
+        if not mesh_shape_node:
+            mesh_shape_node = cmds.createNode("mesh", name=guide_name)
+        hair_guide_node = cmds.ls(guide_name, type="HairFromGuidesNode")
```
BigRoy, line 140: This would be a `List[str]` which seems odd with the variable name, bug?
Also, the usage below with `nodes.append(hair_guide_node)` then seems wrong.

You might be better of just doing:
```python
hair_guide_node = guide_name
if not cmds.objExists(hair_guide_node):
    hair_guide_node = cmds.createNode('mesh", name=guide_name)
```

Also it feels wrong that we're looking for a guide name AND a shape name, but then if it does not exist create the _shape_ with the guide name? Are we mixing up things here by accident?

Note that if you create a `mesh` type node in Maya that it will ALWAYS also create a parent transform automatically if you have not named it or supplied a parent so this seems like you're not taking the full control you actually want?

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_rig.py
```diff
@@ -0,0 +1,128 @@
...
+            )
+
+        color = plugin.get_load_color_for_product_type("oxrig")
+        if color is not None:
+            red, green, blue = color
+            cmds.setAttr(group_name + ".useOutlinerColor", 1)
+            cmds.setAttr(
+                group_name + ".outlinerColor", red, green, blue
+            )
+        self.use_resources_textures(namespace, path)
```
BigRoy, line 64: Note that this would make "reference overrides" - preferably these textures are remapped before publishing on the extracted output file (similar to Look publish) so that you don't need 'fixing' on load.

server_addon/maya/client/ayon_maya/plugins/load/load_reference.py
```diff
@@ -124,6 +124,7 @@ class ReferenceLoader(plugin.ReferenceLoader):
         "skeletalMesh",
         "mvLook",
         "matchmove",
+        "OxCache",
```
BigRoy, line 127: Should still turn lowercase?
```suggestion
        "oxcache",
```

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_cache.py
```diff
@@ -0,0 +1,36 @@
...
+    def process(self, instance):
+
+        settings = {"nodes": []}
+        ox_shapes = cmds.ls(instance[:], shapes=True, long=True)
+        for ox_shape in ox_shapes:
+            # Get transform data
+            parent = cmds.listRelatives(ox_shape, parent=True)[0]
+            mesh_shape_data = {"name": parent, "cbId": lib.get_id(parent)}
+            ox_cache_nodes = cmds.listConnections(
+                ox_shape, destination=True, type="HairFromGuidesNode") or []
```
BigRoy, line 23: Do we also want to include shapes that happen to be in the instance but are not related to hair from guides nodes? Because we don't seem to be skipping those?

BigRoy, line 23: Warning: this may return `None` if it has no destination HairFromGuidesNode which will raise an error down below trying to iterate over the variable for `cache_file_attribute`

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_cache.py
```diff
@@ -0,0 +1,36 @@
...
+            # transfer cache file
+            shape_data = {
+                "shape": mesh_shape_data,
+                "name": ox_shapes,
+                "cbId": lib.get_id(ox_shape),
+                "ox_nodes": ox_cache_nodes,
+                "cache_file_attribute": ["{}.cacheFilePath".format(ox_node)
+                                        for ox_node in ox_cache_nodes]
```
BigRoy, line 31: It seems wrong that this key name is singular but contains a list. Should this be `cache_file_attributes`?

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_rig.py
```diff
@@ -0,0 +1,89 @@
...
+        # avoid duplicate dictionary data
+        instance.data["resources"] = [
+            i for n, i in enumerate(ornatrix_resources)
+            if i not in ornatrix_resources[n + 1:]
+        ]
+        self.log.debug("{}".format(instance.data["resources"]))
+        start = cmds.playbackOptions(query=True, animationStartTime=True)
+        for key in ["frameStart", "frameEnd",
+                    "frameStartHandle", "frameEndHandle"]:
+            instance.data[key] = start
```
BigRoy, line 45: Why are we forcing the time slider's start frame? Don't recall we're doing that anywhere else in Maya?

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_rig.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def get_texture_resources(self, node):
```
BigRoy, line 47: This may benefit from type hints?
```suggestion
    def get_texture_resources(self, node: str) -> List[Dict[str, Any]]:
```

E.g. it seems to expect that the `node` is a dag node that has child shapes. So it expects a transform node? Might be nice in the docstring.

server_addon/maya/client/ayon_maya/plugins/publish/collect_yeti_rig.py
```diff
@@ -34,7 +33,7 @@ def process(self, instance):
         yeti_nodes = cmds.ls(instance[:], type="pgYetiMaya", long=True)
         for node in yeti_nodes:
             # Get Yeti resources (textures)
-            resources = self.get_yeti_resources(node)
+            resources = self.get_texture_resources(node)
```
BigRoy, line 36: Seems like this may need to be reverted because the method is still called `get_yeti_resources`?
```suggestion
            resources = self.get_yeti_resources(node)
```

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_cache.py
```diff
@@ -0,0 +1,88 @@
...
+    families = ["oxrig", "oxcache"]
+
+    def process(self, instance):
+        cmds.loadPlugin("Ornatrix.mll", quiet=True)
+        # Define extract output file path
+        ox_nodes = cmds.ls(instance[:], long=True)
+        dirname = self.staging_dir(instance)
+        attr_values = instance.data["creator_attributes"]
+        # Start writing the files for snap shot
+        # <NAME> will be replace by the Yeti node name
```
BigRoy, line 24: This doesn't seem to be the case for ornatrix and comment is copied from yeti logic?
```suggestion
```

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_rig.py
```diff
@@ -0,0 +1,113 @@
...
+import os
+import json
+
+from ayon_maya.api import lib
+from ayon_maya.api import plugin
+from maya import cmds
+
+
+class ExtractxRig(plugin.MayaExtractorPlugin):
```
BigRoy, line 12: Typo?
```suggestion
class ExtractOxRig(plugin.MayaExtractorPlugin):
```

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_rig.py
```diff
@@ -0,0 +1,113 @@
...
+                'name': "rigsettings",
+                'ext': "rigsettings",
+                'files': os.path.basename(settings_path),
+                'stagingDir': dirname
+            }
+        )
+
+        self.log.debug("Extracted {} to {}".format(instance, dirname))
+
+        cmds.select(clear=True)
```
BigRoy, line 113: Why is this here?
```suggestion
```

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_rig.py
```diff
@@ -0,0 +1,113 @@
...
+        dirname = self.staging_dir(instance)
+        settings_path = os.path.join(dirname, "ornatrix.rigsettings")
+        image_search_path = instance.data["resourcesDir"]
+
+        # add textures to transfers
+        if 'transfers' not in instance.data:
+            instance.data['transfers'] = []
+
+        resources = instance.data.get("resources", [])
+        for resource in instance.data.get('resources', []):
```
BigRoy, line 50: Seems like variable could be used?
```suggestion
        resources = instance.data.get("resources", [])
        for resource in resources:
```

server_addon/maya/server/settings/creators.py
```diff
@@ -373,6 +381,18 @@ class CreatorsModel(BaseSettingsModel):
...
+        "default_variants": [
+            "Main"
+        ]
+    },
+    "CreateOxCache": {
+        "enabled": True,
+        "default_variants": [
+            "Main"
+        ]
+    },
```
BigRoy, line 395: Should these default to `False` so that only studios using Ornatrix could enable them?

server_addon/maya/client/ayon_maya/api/lib.py
```diff
@@ -4241,3 +4241,77 @@ def get_node_index_under_parent(node: str) -> int:
...
+    Supports negative frame ranges like -001, 0000, 0001 and -0001,
+    0000, 0001.
+
+    Arguments:
+        filepath (str): The full path to filename containing the given
+        pattern.
+        pattern (str): The pattern to swap with the variable frame number.
+
+    Returns:
+        list: file sequence.
```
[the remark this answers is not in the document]
moonyuet: We can do it with some other pull request for cleaning up.

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_cache.py
```diff
@@ -0,0 +1,164 @@
...
+    product_types = {"oxcache", "oxrig"}
+    representations = {"abc"}
+
+    label = "Load Ornatrix Cache"
+    order = -9
+    icon = "code-fork"
+    color = "orange"
+
+    def load(self, context, name=None, namespace=None, data=None):
+        """Loads a .fursettings file defining how to load .abc into
```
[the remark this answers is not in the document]
moonyuet: It should be `cachesettings`. and yes it supports for both oxcache and oxrig product type

server_addon/maya/client/ayon_maya/plugins/create/create_ornatrix_cache.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Add animation data without step and handles
+        remove = {"handleStart", "handleEnd"}
+        defs = [attr_def for attr_def in lib.collect_animation_defs()
+                if attr_def.key not in remove]
+        defs.extend(
+            [
+                BoolDef("renderVersion",
+                        label="Use Render Version",
+                        default=True),
```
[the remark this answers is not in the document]
moonyuet: they have an option of `Use Render Version` in their setting
I am thinking if we should support the format type too
image

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_cache.py
moonyuet: > Note that if you create a `mesh` type node in Maya that it will ALWAYS also create a parent transform automatically if you have not named it or supplied a parent so this seems like you're not taking the full control you actually want?

It is something challenging for loading the cached abc in the guide node as it is connected to the shape node of the connection. And the Operator stack requires you to connecting these two together in order to show up the guide node (and the shape node) in the stack editor. Would be great if you can suggest solution for this case.

server_addon/maya/client/ayon_maya/api/lib.py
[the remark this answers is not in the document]
BigRoy: Yup - agree. The CollectLook fix would be separate PR - but type hint fix should be here.

server_addon/maya/client/ayon_maya/plugins/create/create_ornatrix_cache.py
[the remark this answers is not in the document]
BigRoy: Alembic exporter also has the option for changing format type but I've never seen a need to change that in any production I've worked on (Ogawa has now been the long-standing latest standard). If we _were_ to expose that I have a big feeling that'll then also need to be set up in settings to become an optional toggle, etc. which becomes way more complicated looking at the Pointcache logic in Maya which has that.

I'd personally avoid it unless requested.

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_rig.py
moonyuet: I implemented the contextmanager function of `attribute_values` inside the `extract_ornatrix_rig`

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_cache.py
moonyuet: > Could share the Node graph of what it looks like in Maya when the connections are correct (and e.g. this is currently loaded?) Like some screenshots.

already DM you in discord, it's a clip

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_cache.py
```diff
@@ -0,0 +1,37 @@
...
+                "shape": mesh_shape_data,
+                "name": ox_shapes,
+                "cbId": lib.get_id(ox_shape),
+                "ox_nodes": ox_cache_nodes,
+                "cache_file_attributes": ["{}.cacheFilePath".format(ox_node)
+                                          for ox_node in ox_cache_nodes]
+            }
+            if shape_data["cbId"]:
+                settings["nodes"].append(shape_data)
+        instance.data["cachesettings"] = settings
```
BigRoy, line 37: Would this be easier to just use the `nodes` list directly and just make the settings `dict` at the end?
```suggestion
        nodes = []
        ox_shapes = cmds.ls(instance[:], shapes=True, long=True)
        for ox_shape in ox_shapes:
            # Get transform data
            parent = cmds.listRelatives(ox_shape, parent=True)[0]
            mesh_shape_data = {"name": parent, "cbId": lib.get_id(parent)}
            ox_cache_nodes = cmds.listConnections(
                ox_shape, destination=True, type="HairFromGuidesNode") or []
            if not ox_cache_nodes:
                continue
            # transfer cache file
            shape_data = {
                "shape": mesh_shape_data,
                "name": ox_shapes,
                "cbId": lib.get_id(ox_shape),
                "ox_nodes": ox_cache_nodes,
                "cache_file_attributes": ["{}.cacheFilePath".format(ox_node)
                                          for ox_node in ox_cache_nodes]
            }
            if shape_data["cbId"]:
                nodes.append(shape_data)
        instance.data["cachesettings"] = {"nodes": nodes}
```

BigRoy: Also, we may want to just skip much earlier if there's no `cbId` since we ignore it then anyway.
Like:
```suggestion
        nodes = []
        ox_shapes = cmds.ls(instance[:], shapes=True, long=True)
        for ox_shape in ox_shapes:
            ox_shape_id = lib.get_id(ox_shape)
            if not ox_shape_id:
                continue
            # Get transform data
            parent = cmds.listRelatives(ox_shape, parent=True)[0]
            mesh_shape_data = {"name": parent, "cbId": lib.get_id(parent)}
            ox_cache_nodes = cmds.listConnections(
                ox_shape, destination=True, type="HairFromGuidesNode") or []
            if not ox_cache_nodes:
                continue
            # transfer cache file
            nodes.append({
                "shape": mesh_shape_data,
                "name": ox_shapes,
                "cbId": ox_shape_id,
                "ox_nodes": ox_cache_nodes,
                "cache_file_attributes": ["{}.cacheFilePath".format(ox_node)
                                          for ox_node in ox_cache_nodes]
            })
        instance.data["cachesettings"] = {"nodes": nodes}
```

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_rig.py
```diff
@@ -0,0 +1,85 @@
...
+    def get_texture_resources(self, node: str) -> List[Dict[str, Any]]:
+        resources = []
+        node_shape = cmds.listRelatives(node, shapes=True)
+        if not node_shape:
+            return []
+
+        ox_nodes = [
+            ox_node for ox_node in cmds.listConnections(node_shape, destination=True)
+                    if cmds.nodeType(ox_node) in ORNATRIX_NODES
+        ] or []
```
BigRoy, line 52: This should be the same as:
```suggestion
        ox_nodes = cmds.listConnections(
            node_shape, destination=True, type=ORNATRIX_NODES) or []
```

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_rig.py
```diff
@@ -0,0 +1,85 @@
...
+                self.log.debug("Texture is absolute path, ignoring "
+                               "image search paths for: %s" % texture)
+                files = lib.search_textures(texture)
+            else:
+                root = os.environ["AYON_WORKDIR"]
+                filepath = os.path.join(root, texture)
+                files = lib.search_textures(filepath)
+                if files:
+                    # Break out on first match in search paths..
+                    break
```
BigRoy, line 70: You don't want to _break_ out here? Since now you're skipping all next file nodes. I think you may have copied them from elsewhere that had another extra for loop?

server_addon/maya/client/ayon_maya/api/lib.py
```diff
@@ -1733,7 +1733,7 @@ def is_valid_reference_node(reference_node):
     """
     # maya 2022 is missing `isValidReference` so the check needs to be
     # done in different way.
-    if cmds.about(version=True) < 2023:
+    if int(cmds.about(version=True)) < 2023:
```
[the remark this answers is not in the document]
BigRoy: Nevermind, seems to get implemented here already: https://github.com/ynput/ayon-core/pull/686

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_rig.py
moonyuet: > You don't want to _break_ out here? Since now you're skipping all next file nodes. I think you may have copied them from elsewhere that had another extra for loop?

it's being part of the collect yeti rig before it is split into one collector(you can notice the commit).
I will do continue instead of break.

server_addon/maya/client/ayon_maya/plugins/publish/collect_ornatrix_rig.py
moonyuet: got the error below if I choose to changing this:
```   
TypeError: Invalid arguments for flag 'type'.  Expected string, got [ str, str, str, str ]
```

BigRoy: Interesting. Oh, funny maya. In that case:
```suggestion
        ox_nodes = cmds.ls(cmds.listConnections(
            node_shape, destination=True) or [], type=ORNATRIX_NODES)
```

moonyuet: this one works and I committed the changes in [fe2d422](https://github.com/ynput/ayon-core/pull/651/commits/fe2d42299b94eb685a09573e9349418d7fc2638b)

server_addon/maya/client/ayon_maya/plugins/load/load_ornatrix_cache.py
```diff
@@ -29,14 +29,16 @@ def load(self, context, name=None, namespace=None, data=None):
         and allow published looks to also work for Ornatrix rigs and its caches.
 
         """
+        # Ensure Ornatrix is loaded
+        if not cmds.pluginInfo("Ornatrix.mll", query=True, loaded=True):
+            cmds.loadPlugin("Ornatrix.mll", quiet=True)
```
BigRoy, line 34: Is there a reason this was changed from:
```suggestion
        cmds.loadPlugin("Ornatrix.mll", quiet=True)
```

server_addon/maya/client/ayon_maya/api/lib.py
```diff
@@ -4241,3 +4241,78 @@ def get_node_index_under_parent(node: str) -> int:
...
+
+    This also parses to full sequences for those with dynamic patterns
+    like <UDIM> and %04d in the filename.
+
+    Args:
+        filepath (str): The full path to the file, including any
+            dynamic patterns like <UDIM> or %04d
+
+    Returns:
+        list: The files found on disk
```
BigRoy, line 4257: I wonder if we should update this return type to `list[str | clique.Collection]` since `get_sequence` returns a list of collections - which may be unexpected?
```suggestion
        list[str | clique.Collection]: The files found on disk
```

server_addon/maya/client/ayon_maya/api/lib.py
```diff
@@ -4241,3 +4241,78 @@ def get_node_index_under_parent(node: str) -> int:
...
+    Supports negative frame ranges like -001, 0000, 0001 and -0001,
+    0000, 0001.
+
+    Arguments:
+        filepath (str): The full path to filename containing the given
+        pattern.
+        pattern (str): The pattern to swap with the variable frame number.
+
+    Returns:
+        list: clique.Collection.
```
BigRoy, line 4302:
```suggestion
        list[clique.Collection]: List of sequence frame collections.
```

I wonder if it's ever confusing that this actually returns a list of collections?

server_addon/maya/client/ayon_maya/api/lib.py
```diff
@@ -4241,3 +4241,78 @@ def get_node_index_under_parent(node: str) -> int:
...
+
+    source_dir = os.path.dirname(filepath)
+    files = [f for f in os.listdir(source_dir)
+                if re.match(re_pattern, f)]
+
+    pattern = [clique.PATTERNS["frames"]]
+    collection, remainder = clique.assemble(
+        files, patterns=pattern, minimum_items=1)
+
+    return collection
```
BigRoy, line 4318: I believe `clique.assemble` actually returns a list of collections, not a single one.

```suggestion
    collections, remainder = clique.assemble(
        files, patterns=pattern, minimum_items=1)

    return collections
```

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_rig.py
```diff
@@ -0,0 +1,117 @@
...
+class ExtractOxRig(plugin.MayaExtractorPlugin):
+    """Extract the Ornatrix rig to a Maya Scene and write the Ornatrix rig data."""
+
+    label = "Extract Ornatrix Rig"
+    families = ["oxrig"]
+    scene_type = "ma"
+
+    def process(self, instance):
+        """Plugin entry point."""
+        cmds.loadPlugin("Ornatrix.mll", quiet=True)
```
BigRoy, line 21: Just thought of this now, but I believe the usage of the `.mll` suffix here makes it Windows only since the plug-in files are `.so` on linux/mac? You should be able to load these without the `.mll` suffix to be cross-platform?

BigRoy (approved):
Didn't test it but overall the code looks good. I did have some small last remarks that may need to be considered, but with that fixed this should be good to go (and have one last test run I suppose?)

server_addon/maya/client/ayon_maya/plugins/publish/extract_ornatrix_rig.py
moonyuet: maybe just doing `cmds.loadPlugin("Ornatrix", quiet=True)`

BigRoy: Yes, exactly. 👍

LiborBatek:
I was able to produce ```oxrigMain``` product for Ornatrix Rig successfully as a first step

image

with resource files (ornatrix maps) too

image

LiborBatek:
When loaded into empty workfile as ```oxrigMain``` 

image

This is probably not the best as it gets full character hierarchy with also char rig....

We should decide how to approach it (if similarly like with yeti integration or by using Ornatrix Grooms as presets....)  not sure whats best atm tho...need to investigate it more

image

LiborBatek:
I have briefly tried to use Ornatrix ```Groom``` asset as it saves all the setup including the resources as seen here:

image

Here is the workflow in nutshell...

https://github.com/ynput/ayon-core/assets/112623825/5a601df8-b031-4cfa-853b-81de6ad747de


The issue is that when I have tried to apply such groom setup on something else than the character rig it got broken if not in the reference pose when connecting...aka when applied on lets say shot anim ```pointcache``` so it needs some thinking still how to approach it.

@BigRoy  @moonyuet what do you think? I suppose we should follow regular workflow for Ornatrix rather then adopting our Yeti like one (by duplicating and connecting transforms etc on groom meshes and publish those instead for OrnatrixRig product)

We should first think how we approach it and if there any other limitations speaking of the native "preset" of Ornatrix using ```Groom``` aka preset as in the video above...

moonyuet:
> @BigRoy @moonyuet what do you think? I suppose we should follow regular workflow for Ornatrix rather then adopting our Yeti like one (by duplicating and connecting transforms etc on groom meshes and publish those instead for OrnatrixRig product)
> 

When I was working on the ornatrix cache loader, there is a function to connect the guide node with alembic cache to the selected mesh and it would be showed in the operator stack editor. But there is some issue when there is no selected mesh, where the operator stack editor can't add the guide node,
I am not saying this is not possible, but we need some more time to work on creating nodes and loading the data (which has been stored in all the ornatrix-related_node) for Ornatrix. Or even we no need to have two product type(rig and cache), and just ornatrix. And we can create nodes and loading the data(textures, cached alembic) all at once.

BigRoy:
Thanks for tagging me - admittedly going based off of the video and the comments I can get an idea of what the issue is but knowing nothing of ornatrix I don't have any tips on how to solve this best.

It sounds like:

1. We want to export an Ornatrix "Rig" which is an ornatrix groom that we want to re-apply to e.g. an animated character.
2. The Groom requires a static rest pose as it gets applied to the mesh so that it can work for deformed meshes.
3. The ornatrix operator stack allows to include a shape node as the rest pose (which can be stored within an Ornatrix preset).

Meaning there's basically an Ornatrix built-in way to solve number 2 OR we decide to make a full rig that itself 'includes' meshes of which one is static and one is maybe the output mesh - which is then somewhat similar to how the yeti rig setups are designed (or at least recommended).

What is the best choice depends hugely on those who want to use it. But I'd say the Yeti way has been very flexible because you can build many hacks into a rig that maybe just a ornatrix preset may not. Yet I've also heard others felt it was "technical to set up". If we could rely fully on Ornatrix "presets" then all we'd need to export is just that preset maybe and hence no maya exports of our own?

But as Kayla stated she did face some issues from the API using the Ornatrix methods - I can't help much with that without having access to Ornatrix with the same setup and debug from there to see where it really fails.

moonyuet:
> Thanks for tagging me - admittedly going based off of the video and the comments I can get an idea of what the issue is but knowing nothing of ornatrix I don't have any tips on how to solve this best.
> 
> It sounds like:
> 
> 1. We want to export an Ornatrix "Rig" which is an ornatrix groom that we want to re-apply to e.g. an animated character.
> 2. The Groom requires a static rest pose as it gets applied to the mesh so that it can work for deformed meshes.
> 3. The ornatrix operator stack allows to include a shape node as the rest pose (which can be stored within an Ornatrix preset).
> 
> Meaning there's basically an Ornatrix built-in way to solve number 2 OR we decide to make a full rig that itself 'includes' meshes of which one is static and one is maybe the output mesh - which is then somewhat similar to how the yeti rig setups are designed (or at least recommended).
> 
> What is the best choice depends hugely on those who want to use it. But I'd say the Yeti way has been very flexible because you can build many hacks into a rig that maybe just a ornatrix preset may not. Yet I've also heard others felt it was "technical to set up". If we could rely fully on Ornatrix "presets" then all we'd need to export is just that preset maybe and hence no maya exports of our own?
> 
> But as Kayla stated she did face some issues from the API using the Ornatrix methods - I can't help much with that without having access to Ornatrix with the same setup and debug from there to see where it really fails.


The workflow I can think of now is:
1. Create Instances with the target mesh assigned with the Ornatrix-related nodes, has an validator to check on the Ornatrix's content, export the Ornatrix presets along with other resources(e.g. textures, file cache) 
2. Create Instances with target mesh assigned with the Ornatrix-related nodes, has an validator to check on the Ornatrix's content, export maya scene with other resources(e.g. texture, file cache). 
3. Create Instances with target mesh assigned with the Ornatrix-related nodes, has an validator to check on the Ornatrix's content, export maya scene with other texture resource. There is still cache family for exporting the alembic cache
4. Create Instances with target mesh assigned with the Ornatrix-related nodes, has an validator to check on the Ornatrix's content, export json file which stores all attributes data of Ornatrix-related nodes along with other resources. When users loaded back to the scene, they need to select the target mesh to load the ornatrix data so that maya can find the target mesh and create the Ornatrix-related nodes connected to it(it also ensures the ornatrix nodes connected to operator stack editor). The attributes of nodes would be synced with the attribute data from the published json

And 1 causes the possible rig broke as we can notice in Libor's testing.

LiborBatek:
@moonyuet haha Im completely lost in your suggested workflow sry (no offense just me being dumb :) ...I would need practical example.

I can think of diving more into Ornatrix and best practises first (myself) and find out how it should be done when transfering groom from char to char (aka between same char but different scenes)

moonyuet:
> @moonyuet haha Im completely lost in your suggested workflow sry (no offense just me being dumb :) ...I would need practical example.
> 
> I can think of diving more into Ornatrix and best practises first (myself) and find out how it should be done when transfering groom from char to char (aka between same char but different scenes)

I agree that we need to do more research on the Ornatrix grooms, like doing more manual test before tweaking on the code.

LiborBatek:
I have made some more investigation and self education regarding Ornatrix best practices and found out few things...

Speaking of our new publish family ```oxRig``` ...we should def. use the native ```Groom``` presets as they cover 95 % of all usecases (there is just one unsupported type of setup ```hair from mesh strips``` which we do not need to take care of as its not very common - seems more like for converting game ready mesh type haircards to actual hair strands setup)

I have also found out that its even simpler speaking of the native ```Groom``` preset as if you apply it without any selection it will bring also the groom geometry!! so it does internally exactly what we do manually speaking of our Yeti rig workflow!

And last reason to use such approach being that ```Groom``` preset can be used accross any supported DCC so its ideal for transfering the hair rig inbetween lets say ```maya``` and ```max``` or ```c4d``` ....it should be 100% compatible.

excerpt from user Ornatrix docs and the cmd for using it

```
OxSaveGroom [-path <string>] [-thumbnail <string>] [-optional]

OxLoadGroom [-path <string>] [-scale <double>] [-detail <double>]
```


I will attach vid soon to demonstrate how easily it works....

Speaking of  our ```oxCache``` product it should follow standard yeti workflow we have but using ```Ornatrix Alembic``` format it seems there is native support for rendering via ```Vray``` as proxy and I expect that the same goes with ```Arnold``` etc.

LiborBatek:
Here is the process...

and last step might be to make connection of ```outmesh``` and ```inmesh``` to connect those ```ShapeNodes``` and it could also be
done similarly to our ```xgen``` and ```yeti``` workflows via right click ```Action``` as ```connect geometry``` too...

It would fit the rest of hair addons and would follow the same practice.

Only question now if there is a way to use such ```Groom``` like referenced file or just locally created one or how to approach versioning of such product aka ```oxRig``` any ideas @moonyuet @BigRoy 


https://github.com/ynput/ayon-core/assets/112623825/10dc3719-d0f6-4440-a60d-c4e15240c1af

LiborBatek:
Mockup of the possible way...


https://github.com/ynput/ayon-core/assets/112623825/0742388b-7356-4877-a77f-e8034b864d04



"Publishing" of the already created hair setup originating from the native Ornatrix Groom preset and saving it as a ```maya scene```

moonyuet:
> Mockup of the possible way...
> 
>  Ornatrix_Publish.mp4 
> "Publishing" of the already created hair setup originating from the native Ornatrix Groom preset and saving it as a `maya scene`

So for the improvement on the Ornatrix Rig product type. (Just the note for myself.)
The workflow for Ornatrix Rig is:
1. Create Ornatrix Rig Instance with selected geometry which needed for the fur stimulation
2. Publish the Ornatrix Rig with maya scene and ornatrix groom presets to the `publish` folder
3. When loading Ornatrix Rig, user can connect the presets by using the `action -> connect ornatrix rig` with `inMesh` and `outMesh`

iLLiCiTiT:
Please move PR to https://github.com/ynput/ayon-maya

PR #652, opened 2024-06-14
Houdini: update frames in latest version

MustafaJafar:
## Changelog Description
Rendering is expensive, sometimes it is helpful only to re-render changed frames and reuse existing.
Todo list:
- [ ] Add profiles to `CollectFramesFixDef` settings
- [x] **ExtractLastPublished**: A generic extractor to copy the files of the latest published version `last_version_published_files` .
It only copies files that do not need to be fixed.
- **ExtractRender**
    - [ ] local render: overwrite them by rendering the frames to fix.. By looping over the frame ranges and setting the parms, rendering and finally returning it as it was.  
    - [ ] Farm render: update the job_info.Frames. luckily it expects the same format as frames_to_fix. 

## Additional info
This PR is the Houdini equivalent for https://github.com/ynput/OpenPype/pull/4362

## Additional info 2
Unrelated to this PR. but, I'll just save some notes for the future.
- this [line](https://github.com/ynput/ayon-core/blob/0490ab315c2250db4c5914f828eab9f575cf09b5/server_addon/nuke/client/ayon_nuke/plugins/publish/extract_render_local.py#L72) in nuke/extrack_render_local  can be replaced by `frames_to_render = intspan(frames_to_fix).ranges()` and remove the definition of `_get_frames_to_render` 

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/server_addon/houdini/client/ayon_houdini/plugins/publish/Extract_last_published.py
+++ b/server_addon/houdini/client/ayon_houdini/plugins/publish/Extract_last_published.py
@@ -1,0 +1,93 @@
+import os
+import shutil
+import hou
+from intspan import intspan
+import pyblish.api
+from ayon_core.lib import collect_frames
+from ayon_houdini.api import plugin, lib
+
+
+class ExtractLastPublished(plugin.HoudiniExtractorPlugin):
+    """
+    Generic Extractor that copies files from last published
+    to staging directory.
+    It works only if instance data includes "last_version_published_files"
+    and there are frames to fix.
+
+    The files from last published are base of files which will be extended/fixed for specific
+    frames.
+    """
+
+    order = pyblish.api.ExtractorOrder - 0.1
+    label = "Extract Last Published"
+    families = ["*"]
+
+    def process(self, instance):
+        frames_to_fix = instance.data.get("frames_to_fix")
+        last_published = instance.data.get("last_version_published_files")
+        if not (frames_to_fix or last_published):
+            self.log.debug("Nothing to copy.")
+            return
+
+        # Get a list of expected filenames
+        node = hou.node(instance.data.get("instance_node", ""))
+        if not node:
+            self.log.debug("Skipping Collector. Instance has no instance_node")
+            return
+        output_parm = lib.get_output_parameter(node)
+        if not output_parm:
+            self.log.debug("ROP node type '{}' is not supported for cleaning up."
+                           .format(node.type().name()))
+            return
+        filepath = output_parm.eval()
+        if not filepath:
+            self.log.warning("No filepath value to collect.")
+            return
+
+        expected_filenames = []
+        staging_dir, _ = os.path.split(filepath)
+        expectedFiles = instance.data.get("expectedFiles", [])
+
+        # 'expectedFiles' are preferred over 'frames'
+        if expectedFiles:
+            # Products with expected files
+            # This can be Render products or submitted cache to farm.
+            for expected in expectedFiles:
+                # expected.values() is a list of lists
+                expected_filenames.extend(sum(expected.values(), []))
+        else:
+            # Products with frames or single file.
+            frames = instance.data.get("frames", "")
+            if isinstance(frames, str):
+                # single file.
+                expected_filenames.append(filepath)
+            else:
+                # list of frame.
+                expected_filenames.extend(
+                    ["{}/{}".format(staging_dir, f) for f in frames]
+                )
+
+        if not os.path.exists(staging_dir):
+            os.makedirs(staging_dir)
+
+        anatomy = instance.context.data["anatomy"]
+        last_published_and_frames = collect_frames(last_published)
+
+        expected_and_frames = collect_frames(expected_filenames)
+        frames_and_expected = {v: k for k, v in expected_and_frames.items()}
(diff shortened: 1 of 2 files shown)


BigRoy:
Oh yeah baby - as soon as we'll have a steady "workflow" for this I'll also be porting this to Maya and Fusion integration.
However, few questions:

### Avoid `intspan`?

Does it really warrant introducing a new dependency `intspan`? By the way, it seems the [corresponding bitbucket homepage](https://bitbucket.org/jeunice/intspan) for `intspan` does not exist anymore and [the latest release is from "Sep 24, 2019"](https://pypi.org/project/intspan/) which may indicate that it's dead? (Which doesn't necessarily mean it's not useful - just not maintained anymore?). In essence it's also a single file and could be vendorized? Or better, maybe we can already use `clique` for this?
```python
import clique
collection = clique.parse("1,3-5,10", "{ranges}")
print(list(collection.indexes))
# [1, 3, 4, 5, 10] 
```

### Is the logic ignoring aovs?

I have a feeling the logic is assuming a single instance in houdini turns into one published file - but I believe, when e.g. submitting to the farm - there may be a single render instance+job that ends up publishing multiple separate AOVs/products? Is this logic allowing that?

### Publishing into existing version?

I have a feeling that publishing _into_ the previous version is the wrong and unsafe approach. Instead, I wonder whether we should instead publish a new version but just happen to re-use frames from the previous publish. So that the old (shorter) publish still exists as previous version and our new version just happens to be extended?

Note that similar logic was actually originally in OpenPype as well for Maya called "extend frames" where the new render frame range that you're publishing/submitting would be appended/replaced to the previous publish generating a new (potentially longer) version.

What I like about the current concept here however is that you can specify the extract "new" frames you want to generate or replace. So that e.g. you could say, add three frames to the end but also replace frames 15-20. Do note however, that whatever render logic we'd end up with that we ALSO avoid rendering the actual frames we DO NOT want to re-render but want to preserve from the previous render. I have a feeling that this PR Is lacking the logic that makes it skip existing frames?

### Extend frames / render new base logic

With Nuke having something in this area, apparently - and now Houdini also trying to have it. And Maya originally having "extend frames" logic as well (inheriting from `avalon`) I really feel it's worth to abstract this way into "base publishing logic" that hopefully works for all publish types that write out per-frame files.

It might be worth laying out/proposing a framework that we could use _that way_ so that hopefully to support rendering only a subset of frames or extending an existing publish becomes trivial across each host integration without much effort on the development side. I feel like even this PR seems to be doing a lot of 'logic' that should be abstracted away and joined together with any feature in another integration also supporting it.

In essence it is:

```python
previous_frames = get_last_publish_frames()
current_frames = get_current_frames()
publish frames = previous_frames.update(current_frames)
```

However it gets more complicated upon merging when deciding what to do with 'handles', etc. or what if new frame range is 1020-1030 and the previous was 1000-1010. What do we do with the gap? Currently nothing in AYON expects gaps in published frame ranges.

server_addon/houdini/client/ayon_houdini/plugins/publish/Extract_last_published.py
```diff
@@ -0,0 +1,93 @@
...
+            return
+        output_parm = lib.get_output_parameter(node)
+        if not output_parm:
+            self.log.debug("ROP node type '{}' is not supported for cleaning up."
+                           .format(node.type().name()))
+            return
+        filepath = output_parm.eval()
+        if not filepath:
+            self.log.warning("No filepath value to collect.")
+            return
```
BigRoy, line 45: Logged messages here all seem incorrect/unrelated.

MustafaJafar:
@BigRoy @moonyuet @iLLiCiTiT 
This PR may take some time also, it affects ayon-core. 
should I close it and re-open after ayon-houdini craetion ?

BigRoy:
Yes please.

iLLiCiTiT:
> This PR may take some time also, it affects ayon-core.

You mean modification of dependencies? That dependency should be added to ayon-houdini, not to ayon-core.

server_addon/houdini/client/ayon_houdini/plugins/publish/Extract_last_published.py
```diff
@@ -0,0 +1,93 @@
...
+                # single file.
+                expected_filenames.append(filepath)
+            else:
+                # list of frame.
+                expected_filenames.extend(
+                    ["{}/{}".format(staging_dir, f) for f in frames]
+                )
+
+        if not os.path.exists(staging_dir):
+            os.makedirs(staging_dir)
```
iLLiCiTiT, line 71:
```suggestion
        os.makedirs(staging_dir, exist_ok=True)
```

server_addon/houdini/client/ayon_houdini/plugins/publish/Extract_last_published.py
BigRoy, line 1: Filename doesn't match naming conventions of the others. Should have lowercase `E`.

MustafaJafar: good catch. 
I'll rework this PR today.

MustafaJafar:
Re-created in ayon-houdini 
https://github.com/ynput/ayon-houdini/pull/45

PR #654, opened 2024-06-17
adding deprecation flag on OCIO config override settings

jakubjezek001:
## Changelog Description
Deprecating OCIO config overrides at individual addon.

## Testing notes:
1. all should work as it is


jakubjezek001:
I am not sure if version on individual add-ons should be bumped too?

iLLiCiTiT (approved):

PR #656, opened 2024-06-17
Maya: Implement backward compatibility for version up workfile

moonyuet:
## Changelog Description
Implement backward compatibility for version up workfile in maya menu for just in case the related setting being turned off or the user has been in the lower addon version.

## Additional info

## Testing notes:
1. Build the maya addon with this branch
2. Install addon
3. Launch Maya
4. AYON menu should show all the plugins as expected.


antirotor (commented):
maybe using [EAFP](https://docs.python.org/3/glossary.html#term-EAFP) here is more efficient:

```python
try:
    ...
except KeyError:
   # old settings
   pass
```

try/except is usually faster then if/else, the only exception is if there is actually exception :) But since this is not operation running on huge data sets it is just a matter of style I suppose.

server_addon/maya/client/ayon_maya/api/menu.py
```diff
@@ -75,13 +75,17 @@ def add_menu():
...
 
-        if project_settings["core"]["tools"]["ayon_menu"].get(
-            "version_up_current_workfile"):
-                cmds.menuItem(divider=True)
-                cmds.menuItem(
-                    "Version Up Workfile",
-                    command=lambda *args: version_up_current_workfile()
-                )
+        try:
+            if project_settings["core"]["tools"]["ayon_menu"].get(
```
antirotor, line 79: well, you could even ditch this if with the try/except :)

moonyuet: But `version_up_workfile` is a toggled button in the core setting.  Users can choose to turn on/off in the setting.

antirotor: Ah yeah, sorry. I was over-optimizing :)

antirotor (approved):

PR #657, opened 2024-06-17
Workfiles tool: Folder change update tasks correctly

iLLiCiTiT:
## Changelog Description
Fix edge cases when change of folder does not update tasks.

## Additional info
Tasks model is waiting until the tasks for the folder are fetched if are already fetched.

## Testing notes:
NOTE: This is edge case, usually happens on first open of workfile, and selecting differnet folder in different level.
1. Open workfiles tool.
2. Change of folder should propagate correct tasks in tasks view.


BigRoy (approved):
Works in Fusion 18 + can't reproduce the issue where tasks didn't refresh correctly. So seems to work as intended!

Very smooth.

LiborBatek (approved):
All working fine!

PR #658, opened 2024-06-17
Blender: Use task level attributes instead of asset

simonebarbieri:
## Changelog Description
Improved creators to set up scene properties instead of using asset-level attributes.

## Testing notes:
The creators and the extractors that have been updated should be tested to check if they're using the information set in the creator.


server_addon/blender/client/ayon_blender/plugins/create/create_model.py
```diff
@@ -29,3 +30,11 @@ def create(
...
 
         return asset_group
+
+    def get_instance_attr_defs(self):
+        return [
+            NumberDef("unitScale",
+                      label="Unit Scale (FBX)",
+                      default=1.0,
+                      tooltip="Scale of the model, valid only for FBX export.")
+        ]
```
BigRoy, line 40: I wonder if these are better implemented as attribute definitions on the Extractors instead since it seems so tied to that exporter?

server_addon/blender/client/ayon_blender/plugins/publish/extract_abc.py
```diff
@@ -44,6 +44,20 @@ def process(self, instance):
...
+        scene.frame_end = instance.data.get("frameEnd", frame_end)
+        scene.frame_step = instance.data.get("frameStep", frame_step)
+        inst_fps = instance.data.get("fps")
+        if inst_fps:
+            scene.render.fps = inst_fps
+            scene.render.fps_base = 1
+
         with bpy.context.temp_override(**context):
             # We export the abc
             bpy.ops.wm.alembic_export(
```
BigRoy, line 63: The [alembic export command](https://docs.blender.org/api/current/bpy.ops.wm.html#bpy.ops.wm.alembic_export) can take a `start` and `end`? Why do we need to change the scene?

Additionally, I would avoid exposing the FPS on the instance for this if it's this tied to the scene. Coming from maya where often changing the FPS has so much influence on the scene that it hardly makes a reproducable output file. But maybe that's different in Blender?

If we do need to change the scene to export the way we want - then we should use a context manager!

Also, if the logic is this involved and we're duplicating it this many times we should make it into a library function.

server_addon/blender/client/ayon_blender/plugins/publish/extract_fbx.py
```diff
@@ -55,8 +55,9 @@ def process(self, instance):
                 new_materials.append(mat)
                 new_materials_objs.append(obj)
 
-        scale_length = bpy.context.scene.unit_settings.scale_length
-        bpy.context.scene.unit_settings.scale_length = 0.01
+        scene = bpy.context.scene
+        scale_length = scene.unit_settings.scale_length
+        scene.unit_settings.scale_length = instance.data.get("unitScale")
```
BigRoy, line 60: Shouldn't this just use the `global_scale` argument of the export function? [see here](https://docs.blender.org/api/current/bpy.ops.export_scene.html#module-bpy.ops.export_scene) or why is changing the scene better?

If we are to change the scene - then this should be done via a context manager!

BigRoy:
For what it's worth - looking at the code I see nothing related to "task level attributes" versus asset level attributes. Is the PR title and description wrong maybe?

server_addon/blender/client/ayon_blender/plugins/create/create_model.py
moonyuet: It is either that we add the collector which is specific to collect the unit scale for fbx exporter. Adding the definition into the extractor seems to have the same sort of confusion with the definition being implemented here.

server_addon/blender/client/ayon_blender/plugins/publish/extract_fbx.py
moonyuet: agreed. This should be done with the contextmanager + context.exitstack()
You can get some inspirations here: https://github.com/ynput/ayon-core/pull/674#discussion_r1649068696

server_addon/blender/client/ayon_blender/plugins/publish/extract_camera_abc.py
```diff
@@ -53,6 +67,12 @@ def process(self, instance):
                 flatten=True
             )
 
+        scene.frame_start = frame_start
+        scene.frame_end = frame_end
+        scene.frame_step = frame_step
+        scene.render.fps = fps
+        scene.render.fps_base = fps_base
```
moonyuet, line 74: Maybe it is a good idea to implement the function in the library to set all these frame_range related settings?

LiborBatek (changes requested):
Neither ```frame range``` nor ```image resolution``` is taken into account when changed on the ```task level```.

It just validates ok with different values compared to those on the task level...
Task level entries:
Screenshot 2024-06-25 142650

and wrong but validated values during publishing as they match those set on ```asset level```:


Screenshot 2024-06-25 142708
 the values above got validated as they follow the asset level attribs (not the underlined highlighted task level ones)

image

moonyuet:
Regarding to the comment above, we need to also adjust the `get_folder_attributes` function in `pipeline.py`. And there should be possible change of the function name.
Instead of merely using folder entity, we use task entity and we can use the folder entity as backward compatibility(maybe or maybe not). 
The function which can get the related data from task entity(take below as example)
```python
from ayon_core.pipeline.context_tools import get_current_task_entity
task_attributes = get_current_task_entity(fields={"attrib"})["attrib"]

iLLiCiTiT:
PR moved to ayon-blender repository https://github.com/ynput/ayon-blender/pull/1

PR #659, opened 2024-06-17
Traypublisher: CSV ingest supports relative and absolute paths

iLLiCiTiT:
## Changelog Description
CSV ingest can work with both relative and absolute paths.

## Additional info
Refactored the code to wrap the code into more readable structure using objects instead of dictionaries. Collection of representations was a little bit confusing and might require more validation of correct behavior.

## Testing notes:
1. Use CSV ingest in traypublisher.
2. Validate it works.


jakubjezek001 (approved):
working as it was

PR #660, opened 2024-06-17
Fix: aftereffects review not displayed in ftrack

kalisp:
## Changelog Description
Solved bug caused review not getting created.

## Testing notes:
1. render locally (single frame, multiframe)
2. render via Deadline
3. check that review is present on DL.


64qam (approved):

server_addon/deadline/client/ayon_deadline/plugins/publish/submit_aftereffects_deadline.py
```diff
@@ -57,7 +57,7 @@ def get_job_info(self):
         dln_job_info.Plugin = "AfterEffects"
         dln_job_info.UserName = context.data.get(
             "deadlineUser", getpass.getuser())
-        if self._instance.data["frameEnd"] > self._instance.data["frameStart"]:
+        if self._instance.data["frameEnd"] >= self._instance.data["frameStart"]:  # noqa
```
iLLiCiTiT, line 60: Wait, so it must be filled even if you render single frame? In that case why is this condition here?

server_addon/aftereffects/client/ayon_aftereffects/plugins/publish/collect_render.py
```diff
@@ -173,24 +181,25 @@ def get_expected_files(self, render_instance):
...
                         render_instance.productName,
                         version_str,
                         str(frame).zfill(self.padding_width),
                         ext
-                    ))
-                    expected_files.append(path)
+                    )
+
+                file_path = os.path.join(base_dir, file_name)
+                expected_files.append(file_path)
```
iLLiCiTiT, line 202: I guess this should be indented one more to add all frames to `expected_files`.
```suggestion
                    file_path = os.path.join(base_dir, file_name)
                    expected_files.append(file_path)
```

kalisp: Shhh...fixed

server_addon/deadline/client/ayon_deadline/plugins/publish/submit_aftereffects_deadline.py
kalisp: Probably protection to flipped frameStart and frameEnd, but in that case it would failed later, so this is not necessary.

iLLiCiTiT (approved):

PR #662, opened 2024-06-18
Tool UIs: Handle missing folder short name

iLLiCiTiT:
## Changelog Description
It looks like `"shortName"` does not have to be filled on project so safe access to the key was added.

## Additional info
Also added base handling of failed thread in folders widget to at least print exception.

## Testing notes:
1. Projects without `"shortName"` in folder type should show folders in UIs.

### DIFF
--- a/client/ayon_core/tools/utils/folders_widget.py
+++ b/client/ayon_core/tools/utils/folders_widget.py
@@ -192,7 +192,11 @@ class FoldersQtModel(QtGui.QStandardItemModel):
             or thread_id != self._current_refresh_thread.id
         ):
             return
-        folder_items, folder_type_items = thread.get_result()
+        if thread.failed:
+            # TODO visualize that refresh failed
+            folder_items, folder_type_items = {}, {}
+        else:
+            folder_items, folder_type_items = thread.get_result()
         self._fill_items(folder_items, folder_type_items)
         self._current_refresh_thread = None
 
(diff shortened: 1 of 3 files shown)


client/ayon_core/tools/utils/folders_widget.py
```diff
@@ -192,7 +192,11 @@ def _on_refresh_thread(self, thread_id):
             or thread_id != self._current_refresh_thread.id
         ):
             return
-        folder_items, folder_type_items = thread.get_result()
+        if thread.failed:
+            # TODO visualize that refresh failed
+            folder_items, folder_type_items = {}, {}
```
BigRoy, line 197: Can we do anything to reproduce this particular issue when the thread would fail?

BigRoy (commented):
Didn't test, but changes make sense for the described issue.

I did notice my projects have these as default values:
image

And I've not faced an issue before this PR with anything. The question then becomes, how do I get my project into such a state that it _is_ an actual issue?

MilaKudr (approved):

64qam (approved):

client/ayon_core/tools/utils/folders_widget.py
iLLiCiTiT: It may wary, could be because server failed, or like reason of this PR, missing key in data.

PR #669, opened 2024-06-19
Maya: Fix the scriptmenu error found during launch maya 2025

moonyuet:
## Changelog Description
This PR is to fix the bug found during the launch of Maya 2025.
```
 Error: TypeError: file C:\Users\murph\AppData\Local\Ynput\AYON\addons\core_0.4.0\ayon_core\vendor\python\scriptsmenu\launchformaya.py line 133: int() argument must be a string, a bytes-like object or a real number, not 'KeyboardModifier'
```
## Additional info
n/a

## Testing notes:
1. Launch Maya with different versions
2. Check Script Editor
3. the scriptmenu related error should be gone


BigRoy (approved):
Not entirely sure about the issue - but the change makes sense because with `PySide6` you're required to pass the modifiers directly instead of the `int`

m-u-r-p-h-y (approved):
tested with
- Maya2025
- Maya2023

works as expected, above error is gone

PR #674, opened 2024-06-19
Maya: add option to stripe shader assignment for setdress family

moonyuet:
## Changelog Description
This PR is to add support to exclude shaders assignment when publishing the scene with setdress family

## Additional info
Update the maya addon to 0.2.5
You can set the setdress settings in ayon setting 
image

## Testing notes:
1. Create setdress and families such as mayascene and layout
2. In setdress, you can choose whether you want to include shader 
image

3. Publish 
4. Should be published successfully
5. Check published setdress scene, the scene should be without shader assignment if the shader option is turned off.
6. Check the published scene from mayascene/layout family, the scene should be published with shader assignment


server_addon/maya/client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -96,7 +96,7 @@ def process(self, instance):
                           "preserve_references"
                       ],
                       constructionHistory=True,
-                      shader=True,
+                      shader=True if instance.data.get("shader", True) else False,
```
BigRoy, line 99: Could this just be like this?
```suggestion
                      shader=instance.data.get("shader", True),
```

moonyuet: Agreed.

BigRoy (approved):
Didn't test it, but changes look good to me.

It's good to note that whenever you load a maya scene that was exported with `shader=False` I vaguely recall that maya will make them green by default because they will have no shader assigned at all instead of defaulting to just getting a default shader assigned - but that may be different these days in Maya 2024+ era.

LiborBatek:
> Didn't test it, but changes look good to me.
> 
> It's good to note that whenever you load a maya scene that was exported with `shader=False` I vaguely recall that maya will make them green by default because they will have no shader assigned at all instead of defaulting to just getting a default shader assigned - but that may be different these days in Maya 2024+ era.

Can we do anything about it? I would def rather having ```standardSurface1``` assigned when shaders stripped...

What you think guys?

moonyuet:
> > Didn't test it, but changes look good to me.
> > It's good to note that whenever you load a maya scene that was exported with `shader=False` I vaguely recall that maya will make them green by default because they will have no shader assigned at all instead of defaulting to just getting a default shader assigned - but that may be different these days in Maya 2024+ era.
> 
> Can we do anything about it? I would def rather having `standardSurface1` assigned when shaders stripped...
> 
> What you think guys?

we can do it either during loading scene with the setdress family (with an option) or somehow"editing" the published scene to assign the default shader. @BigRoy and @antirotor any thought?

BigRoy:
> we can do it either during loading scene with the setdress family (with an option) or somehow"editing" the published scene to assign the default shader. @BigRoy and @antirotor any thought?

The `model` publish does this as well on extract to temporarily assign the default shader. See [here](https://github.com/ynput/ayon-core/blob/1132da882fb0145141fc591995686d0afcb81882/server_addon/maya/client/ayon_maya/plugins/publish/extract_model.py#L79-L80). That would suffice.

But as said, please first confirm the `shaders=False` flag actually still acting like that when loading in the meshes. Do they turn green still? This may have been an issue back in Maya 8.5 or whatever which has (potentially?) been solved along the way?

BigRoy:
By the way, I noticed this is "sponsored" which means this is likely a client request. If they ALWAYS want no shaders then it may also be worth it to expose the default value to settings so the default value can be customized for that studio?

moonyuet:
> But as said, please first confirm the `shaders=False` flag actually still acting like that when loading in the meshes. Do they turn green still? This may have been an issue back in Maya 8.5 or whatever which has (potentially?) been solved along the way?

I tested with 2025 and 2023, they dont turn green anymore and they are assigned by default shader(lambert1).
Not sure about the version elder than 2023.
image

BigRoy:
> I tested with 2025 and 2023, they dont turn green anymore and they are assigned by default shader(lambert1).
> Not sure about the version elder than 2023.

I gave this a quick go. But it unfortunately still does - when the original shader that was applied was NOT the default shader.

image

I tested with `file -exportSelected -shader 0 -type mayaAscii "E:/test.ma"`

It seems that `shader=False` means, do not include the shader in the export.

moonyuet:
> > I tested with 2025 and 2023, they dont turn green anymore and they are assigned by default shader(lambert1).
> > Not sure about the version elder than 2023.
> 
> I gave this a quick go. But it unfortunately still does - when the original shader that was applied was NOT the default shader.
> 
> image
> 
> I tested with `file -exportSelected -shader 0 -type mayaAscii "E:/test.ma"`
> 
> It seems that `shader=False` means, do not include the shader in the export. But if it contains the default shader assignment - it still assigns it:
> 
> ```
> connectAttr "pCubeShape1.iog" ":initialShadingGroup.dsm" -na;
> ```
> 
> But if it had a different shader assign on export, that line is gone and thus on load it doesn't understand it should receive a shader - turning the default maya "I do not have a shader"-green.

Which means we need to assign the default one first by using contextlib and then `cmds.file`.

BigRoy:
> Does it mean we need to assign the default one first by using contextlib and then `cmds.file`?

The workaround for it then would still be doing what we're doing in the model extractor [here](https://github.com/ynput/ayon-core/blob/1132da882fb0145141fc591995686d0afcb81882/server_addon/maya/client/ayon_maya/plugins/publish/extract_model.py#L79-L80), yes.

server_addon/maya/client/ayon_maya/plugins/publish/extract_maya_scene_raw.py
```diff
@@ -88,17 +88,30 @@ def process(self, instance):
...
+                          force=True,
+                          typ="mayaAscii" if self.scene_type == "ma" else "mayaBinary",  # noqa: E501
+                          exportSelected=True,
+                          preserveReferences=attribute_values[
+                              "preserve_references"
+                              ],
+                          constructionHistory=True,
+                          shader=True,
+                          constraints=True,
+                          expressions=True)
```
BigRoy, line 114: Preferably we don't need to duplicate all these lines of code.

How about using `ExitStack`?

```python
with contextlib.ExitStack() as stack:
    if not instance.data.get("shader", True):
        # Fix bug where export without shader may import the geometry 'green' 
        # due to the lack of any shader on import.
        stack.enter_context(shader(selection, shadingEngine="initialShadingGroup"))

    cmds.file(path,
              force=True,
              typ="mayaAscii" if self.scene_type == "ma" else "mayaBinary",
              exportSelected=True,
              preserveReferences=attribute_values["preserve_references"],
              constructionHistory=True,
              shader=instance.data.get("shader", True),
              constraints=True,
              expressions=True)
```

BigRoy (approved):
Didn't do a test run but the code looks good to me. Nice work. @LiborBatek up for breaking this thing?

LiborBatek (approved):
All seems working as expected.

Screenshot 2024-06-24 102532

The option does what it should and setting got propagated when performing Publishing of ```Setdress``` product. Also still correct when using other types like ```layout``` or ```mayaScene``` on these it doesnt have any effect.

PR #675, opened 2024-06-19
Houdini: Fix single frame publishes

### LINKED ISSUE #672: Houdini: Some extractors fail when publishing single frame

- [X] I have searched the existing issues

> [!NOTE] 
> I didn't make PR because I didn't check all the product types. 
> Currently, the problem was reported in `Composite` and `Mantra IFD` products.
> FYI,  `Mantra IFD` should be removed but that's another issue.

### Current Behavior:

Extractor plugins of some product types fail when publishing a single frame.

### Expected Behavior:

no failing.

### Version

1.0.0

Windows

### Steps To Reproduce:

1. publish a single frame in Composting product.

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

_No response_

### Additional context:

_No response_

[cuID:[AY-5903](https://app.clickup.com/t/86bz9d15e)]

BigRoy:
## Changelog Description

The [`instance.data["frames"]` in the CollectFrames](https://github.com/ynput/ayon-core/blob/0bec3f0d2e2616471102b905785cc739724ee031/server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py#L11) plug-in stores a `str` if it's a single frame but a `list` if it's multiple files, similar to how the Integrator currently likes it.

However, there was some scattered logic that seemed to assume that `instance.data["frames"]` was always a `List[str]` which isn't expected anymore now.

## Additional info

Fixes #672 

I feel like it'd still be worth it to start moving into logic where there's a single `ExtractROP` plug-in that works the same for all ROP nodes because the shared logic across them is 99% the same, and where it currently isn't it _should_ become the same. So I may follow up with another PR after this to start working to a single ExtractROP plug-in.

## Testing notes:

1. All Houdini publishes types should allow to publish single frame and multi-frame publishes without issue.

### DIFF
--- a/server_addon/houdini/client/ayon_houdini/api/plugin.py
+++ b/server_addon/houdini/client/ayon_houdini/api/plugin.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 """Houdini specific Avalon/Pyblish plugin definitions."""
+import os
 import sys
 from abc import (
     ABCMeta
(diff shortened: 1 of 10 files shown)


BigRoy:
Closing this in favor of #676

PR #679, opened 2024-06-20
Add simple Create Project Structure launcher action

BigRoy:
Imp for [AY-5076](https://app.clickup.com/t/6658547/AY-5076)

## Changelog Description

Add simple Create Project Structure launcher action.

image

Which creates the folder structure as defined in project settings: `ayon+settings://core/project_folder_structure`

image

## Additional info

It currently only shows when you're in a project in the launcher but have NO folder or task selected to avoid clutter in launcher.

The label unfortunately is a bit too long to be nicely displayed by default, see screenshot. **Any ideas for better icon and shorter label that doesn't get cut off?**

---

This feature request came up a few times:
- [Ynput Community Forum: How set up custom folder creation when using the launcher](https://community.ynput.io/t/how-to-set-up-custom-folder-creation-when-using-the-launcher-or-ftrack-actions/452/10?u=bigroy)
- [Ynput Community Forum: Default Project Hierarchy](https://community.ynput.io/t/default-project-hierarchy-made-of-folder-type-folder/871?u=bigroy) mentions the setting but seemed to describe a different feature.
- [Discord](https://discord.com/channels/517362899170230292/563751989075378201/1253392662036480142)

## Testing notes:

1. Update core addon
2. Run the launcher action when you enter a project.
3. Folders should now be generated as per settings.


client/ayon_core/plugins/actions/create_project_folder_structure.py
```diff
@@ -0,0 +1,19 @@
...
+    name = "create_project_structure"
+    label = "Create Project Structure"
+    icon = "sitemap"
+    color = "#e0e1e1"
+    order = 1000
+
+    def is_compatible(self, selection) -> bool:
+        return (
+            selection.is_project_selected and
+            not selection.is_folder_selected
```
iLLiCiTiT, line 15:
```suggestion
            selection.is_project_selected
            and not selection.is_folder_selected
```

alainxi:
> Any ideas for better icon and shorter label that doesn't get cut off?

Your icon choice is already very good.

Shorter labels :  The "disk" concept is important, to avoid confusion with the project's folder structure in database.
- Add disk proj struct
- Add proj struct on disk
- Create Project Structure (the current one, but with a different icon :   mdi.harddisk-plus  or  mdi.harddisk )

LiborBatek:
I can think of these names:

```Create Dir```
```Create PRJ```
```Project Dir```

**Comment by BigRoy** [comment]
I think maybe it's also worth discussing whether it might be nice to shrink the font sizes just a tad there to ensure there's less cut off. I often see it on other applications as well - which may fix the chosen names here as well?

@iLLiCiTiT what do you think?

**Review by LiborBatek — APPROVED** [review]
Took me a while to realize its not pre-creating the folders from the Project browser/editor but by Ayon Core definition (as mentioned in testing steps:) but after that small confusion all working fine...LGTM

image
they obviously been matching

**Comment by LiborBatek** [comment]
@BigRoy btw what happens if user trigger ```Create Project...``` if those already existing?? I hope it will just skip the creation :) ?

**Comment by BigRoy** [comment]
> @BigRoy btw what happens if user trigger `Create Project...` if those already existing?? I hope it will just skip the creation :) ?

Not sure - since I didn't implement that code myself. But I suspect it would just generate the folders on disk that currently do not exist yet - so if you've adjusted your studio settings to have new folders in the template - then those get added.

Are you afraid to click the button? If so, what is making you scared and what could we do to make it feel less scary? A tooltip?

**Comment by LiborBatek** [comment]
> > @BigRoy btw what happens if user trigger `Create Project...` if those already existing?? I hope it will just skip the creation :) ?
> 
> Not sure - since I didn't implement that code myself. But I suspect it would just generate the folders on disk that currently do not exist yet - so if you've adjusted your studio settings to have new folders in the template - then those get added.
> 
> Are you afraid to click the button? If so, what is making you scared and what could we do to make it feel less scary? A tooltip? A pop-up before the creation? (Like a confirmation? Which might make it feel more scary even?)

yeah, some popup would do! informing user whats gonna happen...

**Comment by LichiMan** [comment]
Thanks a lot for this @BigRoy , this is one of the things I was missing from our current pipeline while testing Ayon.

In our current pipeline we always have a folder called `00_MATERIALS` where we save a lot of stuff in different subfolders like `audio`, `references` (with more subfolders), `logos`, `client_material`...

The production team is who usually save there all the material and they need to have the folder structure created to start copying files.

Now my question. As far as I know you can setup the folder structure in `ayon+settings://core/project_folder_structure` but there's no way to create these folders automatically before this PR, but if you setup a folder structure in that Ayon setting it's because you want it. So, shouldn't this folder structure be created automatically when you create a new project with that Bundle instead of have to go the the Launcher and click on this icon?

Of course, this is just for that folder structure setting and not for the whole folders/tasks you define inside the `Project Settings`.

**Comment by BigRoy** [comment]
> Thanks a lot for this @BigRoy , this is one of the things I was missing from our current pipeline while testing Ayon.
> 
> In our current pipeline we always have a folder called `00_MATERIALS` where we save a lot of stuff in different subfolders like `audio`, `references` (with more subfolders), `logos`, `client_material`...
> 
> The production team is who usually save there all the material and they need to have the folder structure created to start copying files.
> 
> Now my question. As far as I know you can setup the folder structure in `ayon+settings://core/project_folder_structure` but there's no way to create these folders automatically before this PR, but if you setup a folder structure in that Ayon setting it's because you want it. So, shouldn't this folder structure be created automatically when you create a new project with that Bundle instead of have to go the the Launcher and click on this icon?
> 
> Of course, this is just for that folder structure setting and not for the whole folders/tasks you define inside the `Project Settings`.

It's a good question - but admittedly with site sync a project may be accessible from multiple locations and thus may have already been existing before that 'creation'. Also the project creation may happen in any variety of ways, yes currently most likely is the ayon-frontend triggering this but it may even be automated by the API or e.g. occur due to syncing from kitsu, ftrack or shotgrid/flow (which you may not want to automatically flood a local directory with project folders or not). Also, what if a producer just happens to create a project using the web frontend without access to the physical storage | what would create the project folder and files where?

Anyway, nothing is holding anyone from triggering e.g. this logic anytime the AYON Tray Launcher starts up or whatever - I'd just say that from my perspective I don't think I'd recommend that.

Do note that potentially soon using Web Actions from the front-end it may be possible to trigger an action like this from the web front-end, not requiring to do it via the Tray Launcher UI. 

---

Additionally, having it accessible as a callable action does allow you to update the template in settings, and re-run it again if you happened to have changed the template.

**Comment by LiborBatek** [comment]
I have done one more test run to see what happens if those folders are already present and filled with some content...

The good is, even when re-triggering this ```Create Proj``` action nothing happens aka no deleting or similar destructive action.

However still missing that popup info message what are you about to do.... @BigRoy any chance you incorporate it? :)

**Comment by LichiMan** [comment]
> It's a good question - but admittedly with site sync a project may be accessible from multiple locations and thus may have already been existing before that 'creation'. Also the project creation may happen in any variety of ways, yes currently most likely is the ayon-frontend triggering this but it may even be automated by the API or e.g. occur due to syncing from kitsu, ftrack or shotgrid/flow (which you may not want to automatically flood a local directory with project folders or not). Also, what if a producer just happens to create a project using the web frontend without access to the physical storage | what would create the project folder and files where?
> 
> Anyway, nothing is holding anyone from triggering e.g. this logic anytime the AYON Tray Launcher starts up or whatever - I'd just say that from my perspective I don't think I'd recommend that.
> 
> Do note that potentially soon using Web Actions from the front-end it may be possible to trigger an action like this from the web front-end, not requiring to do it via the Tray Launcher UI.
> 
> Additionally, having it accessible as a callable action does allow you to update the template in settings, and re-run it again if you happened to have changed the template.

Yes, it makes sense.

I'm still don't know how Web Actions will work but it sounds like a nice way for doing this, better than the Tray Launcher as everything stay in the web front-end, IMHO.

**Comment by BigRoy** [comment]
@iLLiCiTiT any preference where to go from here aside of the icon and label cosmetics?

**Comment by iLLiCiTiT** [comment]
> any preference where to go from here aside of the icon and label cosmetics?

Not from my point of view. Code is ok.

**Comment by krishna8008** [comment]
> The question is, if it is "pipeline ok". Not sure if any artist should be able to run the action?

Or should the option be like, select the folders/shots on the left pane and choose the option!?
even creating all the folders by artists is also okay, there is no harm I believe, anyway production already carefully adds the shots in ayon web. We are waiting for this option, everytime when we wanted to paste something inside shots its manual, I think this helps a lot

**Comment by MustafaJafar** [comment]
For the first glance when I read the PR's description (sorry if it was mentioned before):
What will happen if the  `ayon+settings://core/project_folder_structure` doesn't match my actual project structure in the project editor ? 
What should be the source of truth for the folder creation ?

**Comment by alainxi** [comment]
> For the first glance when I read the PR's description (sorry if it was mentioned before) I had these thoughts:
> 
> * What will happen if the `ayon+settings://core/project_folder_structure` doesn't match my actual project structure in the project editor ?
> * What should be the source of truth for the folder creation ?

Logical structure (in Editor) and Disk structure (on file server) do not need to be the same.
Look at Jakub's message on that matter, here :
https://discord.com/channels/517362899170230292/890894521817260052/1147160515576082552

**Comment by BigRoy** [comment]
> For the first glance when I read the PR's description (sorry if it was mentioned before) I had these thoughts:
> 
>     * What will happen if the `ayon+settings://core/project_folder_structure` doesn't match my actual project structure in the project editor ?
> 
>     * What should be the source of truth for the folder creation ?

AFAIK this is not at all about creating folders for "folder" entities or tasks, etc. It's to create generic folders for your project that you may want to populate for "a project". Like e.g. you may - out of the pipeline - have a static "resources" folder, or a static "documents" folder where producers maybe put in some stuff. Etc.

So it's more of a "Please just create me some standard folders for the project" that are not related to actual assets or shots.

**Review by MustafaJafar — APPROVED** [review]
This works on my side. 

But, there are few things to keep in mind: 
1. I can only find it if I've nothing selected in the launcher (which happens when selecting a project) but once I select a folder, I won't be able to unselect it. so I have to return and reselect my project. 
2. I wish if I can see the folders created reflected in the launcher view. and be able to use button `Explore here`. 
3. (irrelevant to this PR) when using the button `Explore here` on intermediate folders, it keep telling me `Folder does not exist yet.`
  image
    ```python
    *** WRN: >>> { ActionsModel }: [  Action trigger FAILED.  ] 
    ==========================
    Folder does not exist yet.
    ==========================
      File "E:\Ynput\ayon-core\client\ayon_core\tools\launcher\models\actions.py", line 405, in trigger_action
        action.process(selection)
      File "E:\Ynput\ayon-core\client\ayon_core\plugins\actions\open_file_explorer.py", line 26, in process
        path = self._get_workdir(selection)
      File "E:\Ynput\ayon-core\client\ayon_core\plugins\actions\open_file_explorer.py", line 94, in _get_workdir
        raise AssertionERROR("Folder does not exist yet.")
    AssertionERROR: Folder does not exist yet.
    ```

Anyways, I think this feature/enhancement may come lead to more suggestions/requests/complaints like mine above.

**Comment by LiborBatek** [comment]
@krishnaavril this action is not intended as a substitution for creation of AYON driven project folders (assets, shots etc) but for additional folders predefined in the ```ayon-core``` settings ```ayon+settings://core/project_folder_structure``` (for resources like plates, footage, offline  and other items production might need and to save those in a standard places)

@BigRoy as the confusion already happening I think we definetely should use different naming than ```Create Project Folders``` but refer more to ```Resources folders``` or similar.

**Comment by alainxi** [comment]
@LiborBatek 

I agree we should use different naming than `Create Project Folders`, to avoid confusion.
For that, the "disk" concept is important.
Here are alternatives (very short ones, to not clutter GUI) :
- Add disk proj struct
- Add proj struct on disk

**Comment by BigRoy** [comment]
@dee-ynput could you answer the remaining questions:

- What should be the label on the action? It must be short because as can be seen from the screenshot - it gets cut off. These have come up:
```
    Add disk proj struct
    Add proj struct on disk
    Create Project Structure (will be cut off after "Project...")
    Create Project Folders (will be cut off after "Project...")
```
- What's a sensible icon to use. I think you can pick one from: https://fontawesome.com/v4/icons/
- Should it pop up a confirmation dialog (maybe even explaining which folders would get created; it could even report if they already exist too) to avoid accidental clicks?

**Comment by dee-ynput** [comment]
Thanks for the summary @BigRoy!

> * What should be the label on the action? It must be short because as can be seen from the screenshot - it gets cut off. These have come up:

Let's match the setting name "Project folder structure", so: `Create Project Folders`.
The name being cut off indeed sucks, but that's something we'll fix at UI level (this limitation is a bummer for many other situations)

> * What's a sensible icon to use. I think you can pick one from: [fontawesome.com/v4/icons](https://fontawesome.com/v4/icons/)
The sitemap icon kinda works. I could consider a simple [folder icon](https://fontawesome.com/v4/icon/folder-o) or the [hdd icon](https://fontawesome.com/v4/icon/hdd-o)...

> * Should it pop up a confirmation dialog (maybe even explaining which folders would get created; it could even report if they already exist too) to avoid accidental clicks?

Let's add this later if requested 👍 
(I do like it)

> * Should it appear for ALL users? Or just certain roles? (If so, which?)
I need to check with CS team, but worst case would be to have a setting to state if it's managers only or for artists too.

> * Should it have a setting to enable/disable the action completely?

Having the setting value empty should disable it.
It must be stated in the settings inline help / tooltip.

**Comment by dee-ynput** [comment]
@BigRoy 
After reviewing the CS request:

> * Should it appear for ALL users? Or just certain roles? (If so, which?)

Let's do it for all users. 
The restriction will be implemented if/when we move this to a proper web action 👍

**Comment by BigRoy** [comment]
> Let's match the setting name "Project folder structure", so: Create Project Folders.
The name being cut off indeed sucks, but that's something we'll fix at UI level (this limitation is a bummer for many other situations)

Done.

Unfortunately reads like this:

image

> Having the setting value empty should disable it.
> It must be stated in the settings inline help / tooltip.

Done.

image

> The sitemap icon kinda works. I could consider a simple [folder icon](https://fontawesome.com/v4/icon/folder-o) or the [hdd icon](https://fontawesome.com/v4/icon/hdd-o)...

I'll keep sitemap for now.

File: server/settings/main.py [hunk 2]
```diff
@@ -301,6 +301,11 @@ class CoreSettings(BaseSettingsModel):
         "{}",
         widget="textarea",
         title="Project folder structure",
+        description=(
+            "Defines project folders to create on 'Create project folders'."
+            " When empty, the action will be hidden from users in the"
+            " launcher because there is nothing to create."
+        ),
```
MustafaJafar, line 308: I understand it will create the project structure in the setting. 
but, could we be clear that it is an additional structure that is not necessarily related to the project structure in the project editor?

BigRoy: Do you mean like adding "This does not create folder and task entities, but creates project folders on disk unrelated to any entity"? It seems quite long and verbose and still unclear too. Any ideas on how to explain it with less words and more correct?

iLLiCiTiT: BTW it is still used for actions that actually do use this setting to create it for project folder structure.

iLLiCiTiT: > Yes - so you're saying it will not do anything there when empty but it will still be visible/shown as action (it does not get hidden there). Correct?

No, I'm saying that this setting is used for 2 features:
1. Create folder structure in project.
2. Create folder structure on disk.

BigRoy: > Create folder structure in project.

Wait - what!? Where? How? So this **does** create entities?

BigRoy: Ah found it - it's here `ayon-ftrack`: https://github.com/BigRoy/ayon-ftrack/blob/288b7c6609df78f2ac283cffa55f1efed2905779/client/ayon_ftrack/event_handlers_user/action_create_project_structure.py#L88-L90

Hmm.

iLLiCiTiT: Like I wrote, that was original purpose of these settings https://github.com/ynput/ayon-ftrack/blob/develop/client/ayon_ftrack/event_handlers_user/action_create_project_structure.py . It was then used to also create folders on disk.

alainxi: That's why the "disk" word is important for the button name.
With "Create Project Folders", people will think it creates folder entities.

BigRoy: So this: https://github.com/ynput/ayon-core/pull/679/commits/5065e0e8dbb4da9e4a05918abced984b2f077e5d

image

No idea how it formats it like this haha 🤦‍♂️

BigRoy: > That's why the "disk" word is important for the button name.
> With "Create Project Folders", people will think it creates folder entities, and they will be surprised to discover it created disk structure.

@dee-ynput ? :)

MustafaJafar: > Do you mean like adding "This does not create folder and task entities, but creates project folders on disk unrelated to any entity"? It seems quite long and verbose and still unclear too. Any ideas on how to explain it with less words and more correct?

I mean something like this. 
imo, we may not add all of the that to the tool tip in favor of adding them to the documentation. 

```
Defines project folders to be created when selecting 'Create project folders'. 

- If you use the exact same structure in the project editor, it will create all the folders.
- If you use a different structure, it will add them as additional directories.
- If the setting is empty, the 'Create project directories' action will be hidden from users in the launcher.

This setting is also used in ftrack addon to create both the folders on disk and the ftrack entities.
```

BigRoy: > - If you use the exact same structure in the project editor, it will create all the folders.
> - If you use a different structure, it will add them as additional directories.

I personally don't understand these? Same structure to what? :) 

Anyway, make a decision and I'll be happy to adjust accordingly - so we don't keep this lingering much longer.

MustafaJafar: I think the new tool tip is much better.

MustafaJafar (commented):
I think I've found a bug. 
Can we use lists in `ayon+settings://core/project_folder_structure` setting?

I made this 
```json
{
    "__project_root__": {
        "Assets": {
            "Characters": {
                "NewRobo": {
                    "work": [
                        "cfx",
                        "concept",
                        "modeling",
                        "walkcycle"
                    ]
                },
                "RoboPoro": {
                    "work": [
                        "cfx",
                        "concept",
                        "modeling",
                        "walkcycle"
                    ]
                },
                "RoboMad": {
                    "work": [
                        "cfx",
                        "concept",
                        "modeling",
                        "walkcycle"
                    ]
                }
            }
        }
    }
}
```
and I got this error: 
```
*** WRN: >>> { ActionsModel }: [  Action trigger FAILED.  ] 
==============================
'list' object has no attribute 'items'
==============================
  File "E:\Ynput\ayon-core\client\ayon_core\tools\launcher\models\actions.py", line 405, in trigger_action
    action.process(selection)
  File "E:\Ynput\ayon-core\client\ayon_core\plugins\actions\create_project_folders.py", line 29, in process
    project_folders.create_project_folders(selection.project_name)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\project_folders.py", line 66, in create_project_folders
    basic_paths = get_project_basic_paths(project_name)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\project_folders.py", line 112, in get_project_basic_paths
    return _list_path_items(folder_structure)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\project_folders.py", line 90, in _list_path_items
    paths = _list_path_items(value)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\project_folders.py", line 90, in _list_path_items
    paths = _list_path_items(value)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\project_folders.py", line 90, in _list_path_items
    paths = _list_path_items(value)
  [ Previous line repeated 2 more times ]
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\project_folders.py", line 85, in _list_path_items
    for key, value in folder_structure.items():
AttributeERROR: 'list' object has no attribute 'items'
```

alainxi (approved):
I have tested with this, which works perfectly :

```
{
    "{root[work]}/{project[name]}": {
        "DELIVERIESbis": {
            "ASSETS_FOR_CLIENTS": {},
            "ICON": {
                "OLD": {}
            },
            "MOF_AND_PUB": {
                "OLD": {}
            }        },
        "PREPRODbis": {
            "ASSETS_FROM_ARCHIVE": {
                "NOM_DU_PROJET": {}
            },
            "ASSETS_FROM_CLIENT": {
                "GUIDELINES_AND_STORYBOARD": {
                    "OLD": {}
                },
                "VIDEOS_AND_SOUNDS": {
                    "OLD": {}
                }
            },
            "MOCKUP_AND_STORYBOARD": {
                "EXPORT": {
                    "OLD": {}
                },
                "WIP": {}
            }
        }
    }
}
```

I used  {root[work]}/{project[name]}  instead of  `__project_root__`  , because the later is marked as legacy in the code (and it tries to create the same folder tree on all Anatomy roots, a behaviour that we don't want).

philippe-ynput:
@BigRoy Is this still something we want ?

BigRoy:
Yes, last I heard this is still a valid request but we did not want it inside the launcher, but instead wanted it to become a web action - so this was left here dangling.

LiborBatek:
@BigRoy  I guess theres no way to show the action only for certain user role atm, right?

BigRoy:
> @BigRoy I guess theres no way to show the action only for certain user role atm, right?

It's possible - not sure how fast the exact query would be, but it should be possible. Do we still want it in launcher, or are you asking web actions?

LiborBatek:
Well its not about my preference but make this PR happen I cant tell we should rather ask e.g. @iLLiCiTiT  but

I think it could be still in the Launcher but visible only for ```admin user``` or so...not making possible harm by artists

alainxi:
If it's only available for some people (managers, admins, etc...), may I ask to have an option to allow simple users to also use it ?

For us, everybody can create the default folder tree, as there is no risk of deleting anything (it only adds folders that don't already exist), and managers don't have time for this creation (they consider all project creation should automatically create the default folder tree).

BigRoy:
> simple users

🚚 🔥 🔥 🔥 

Would it be problematic for you @alainxi if it does move to "web actions"? or do you rather prefer it in the launcher itself?

alainxi:
> Would it be problematic for you @alainxi if it does move to "web actions"? or do you rather prefer it in the launcher itself?


On the contrary, I would prefer it to be a web action  🔥

BigRoy:
> I guess theres no way to show the action only for certain user role atm, right?

In web actions this doesn't seem to be possible yet for as far as I could tell. This might need changes on backend/frontend to handle that additional check. @martastain may know where an issue for that in particular may be most appropriate.

In tray action I could actually implement such filter currently though.

MustafaJafar:
closing this PR in favor of https://github.com/ynput/ayon-core/pull/1193

PR #684, opened 2024-06-24
Maya: Remove the legacy creator from before new publisher UI

BigRoy:
## Changelog Description

Remove legacy creator. This should not interfere with the 'detection' of legacy instances nor the conversion/update to new publisher. Instead it would only disallow creation in the old Creator UI.

## Additional info

Note that the refactor to new publisher for Houdini was back in July, 2023: https://github.com/ynput/OpenPype/pull/4388

## Testing notes:

1. [x] Maya Creators and publishing should still work
2. [x] Conversion of legacy instances should still work.


moonyuet (approved):
Tested successfully in Maya 2024, 2025.

server_addon/maya/client/ayon_maya/api/plugin.py
```diff
@@ -35,6 +34,9 @@
 from .lib import imprint, read
 from .pipeline import containerise
 
+# Backwards compatibility
+NewCreator = Creator
```
iLLiCiTiT, line 38: I think we can remove this. It's not imported anywhere.
```suggestion
```

BigRoy: Sweet. I was just preparing for a studio maybe inheriting from it for whatever reason, but it's odd maybe if they were relying on importing this anyway.

PR #686, opened 2024-06-24
Maya: Convert version string to integer

iLLiCiTiT:
## Changelog Description
Fix version comparison in Maya lib.

## Additional info
The version comparison breks assign look UI.

## Testing notes:
1. Assign look UI should not crash


BigRoy (approved):
Seems like a fix that was in another PR as well - so likely has its place. Didn't test but change makes sense. Thanks.

PR #687, opened 2024-06-24
Houdini: Fix Active View extraction error when no scene viewer is an active tab

BigRoy:
## Changelog Description

This fixes an issue where Houdini would error on extraction (at the very end after all the other long running extractors, during integrator!) due to a failure to write the active view's image.

## Additional info

n/a

## Testing notes:

1. [x] Active view thumbnail should work fine with opened scene viewer pane.
2. [x] This should also still work if the only active view is in a floating panel.
3. [x] This should also work (but skip active view extraction) when there is no visible scene viewer.


moonyuet (approved):
Tested all scanerios which mentioned(With/without scene view, and floating scene view)

PR #689, opened 2024-06-24
Publisher: Separate backend into models

iLLiCiTiT:
## Changelog Description
Separate publisher tool logic into smaller models.

## Additional info
Content of `control.py` was huge mess with ton of public methods and properties, it was hard to find out what is used where and how. Hopefully this helps a little bit to manage the logic. Some of the changes may cause issues with custom modifications of the tool (e.g. traypublisher). Some methods/properties were kept to keep backwards compatibility.

This is first PR that is "refactoring" the logic. There will be more. This is already huge.

## Testing notes:
Publisher tool should work as usual in host integrations.
1. Validate any feature of publisher.
- create
- publish
- pause publishing
- see validation errors
- see errors
- change attributes
- change context
- save should keep values
- invalid instance context should not cause crash (which did in develop)
- reset should do what did reset do
- buttons are enabled/disabled when they should be
- publish report can be loaded in publish report viewer


BigRoy:
Just wanted to remark that I'm not sure what this separation helps with aside of maybe from a software-design purism - to me it's still just a lot of code. (According to the "changed lines" it's even more code, hehe) Not something we can fix I think but anything with a "cleanup" or "organize it more" effort I feel is a good goal so from that perspective I'd approve. But do keep in mind that at first sight to an 'outsider' this PR doesn't necessarily solve it. I likely still need to ask you where something is.

Side-question. Does this PR also tend to solve that e.g. the Publish Report isn't accessible from an API standpoint? Because my preference would be to refactor it so that the core functionality is API-driven first (no GUIs, no controllers; just a plain publishing iterator or whatever?) and then the controller just uses that.

Sorry if these are dumb questions.

client/ayon_core/tools/publisher/models/create.py
```diff
@@ -0,0 +1,758 @@
...
+        if profile:
+            allowed_creator_labels = {
+                label
+                for label in profile["creator_labels"]
+                if label
+            }
+            self.log.debug(f"Only allowed `{allowed_creator_labels}` creators")
+            allowed_creator_pattern = (
+                re.compile("|".join(allowed_creator_labels)))
+        return allowed_creator_pattern
```
BigRoy, line 743: Shouldn't this be in a controller as opposed to the model?

64qam (approved):

client/ayon_core/tools/publisher/models/create.py
iLLiCiTiT: It should not be controller, but hard to tell if it should be done the way it is done now. Like stated in description, this is not "everything is ideal", but "first rough refactor".

BigRoy: Cool - not saying it has to be perfect. Just checking what the design goals are so I can review accordingly.

MilaKudr (approved):

PR #690, opened 2024-06-24
Scene Inventory: Action to update to latest approved version

iLLiCiTiT:
## Changelog Description
Scene inventory tool now can update to latest approved version.

## Additional info
This is a proposal how it could be implemented, we still need to modify validator which checks versions of loaded containers. The latest approved version is determined by Status "state" which has 4 options `"not_started"`, `"in_progress"`, `"done"` and `"blocked`. We use `"done"`.

~~Function `get_outdated_containers` now does not mark last approved as outdated > not sure if this is an issue or enhancement?~~ Reverted

## Testing notes:
1. Scene inventory should be able to load latest approved version.

### DIFF
--- a/client/ayon_core/tools/common_models/projects.py
+++ b/client/ayon_core/tools/common_models/projects.py
@@ -1,8 +1,7 @@
 import contextlib
-from abc import ABCMeta, abstractmethod
+from abc import ABC, abstractmethod
 
 import ayon_api
-import six
 
 from ayon_core.style import get_default_entity_icon_color
 from ayon_core.lib import CacheItem, NestedCacheItem
@@ -10,8 +9,7 @@ from ayon_core.lib import CacheItem, NestedCacheItem
 PROJECTS_MODEL_SENDER = "projects.model"
 
 
-@six.add_metaclass(ABCMeta)
-class AbstractHierarchyController:
+class AbstractHierarchyController(ABC):
     @abstractmethod
     def emit_event(self, topic, data, source):
         pass
--- a/client/ayon_core/tools/common_models/projects.py
+++ b/client/ayon_core/tools/common_models/projects.py
@@ -1,6 +1,6 @@
 import contextlib
 from abc import ABC, abstractmethod
-from typing import Literal, Dict, Any
+from typing import Dict, Any
 
 import ayon_api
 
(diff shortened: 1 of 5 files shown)


client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -149,26 +151,35 @@ def new_invalid(cls):
...
     @classmethod
-    def from_entity(cls, version_entity, is_latest):
+    def from_entity(cls, version_entity, is_latest, is_last_approved):
         return cls(
             version_id=version_entity["id"],
             product_id=version_entity["productId"],
             version=version_entity["version"],
             status=version_entity["status"],
             is_latest=is_latest,
+            is_last_approved=is_last_approved,
```
BigRoy, line 182: Is it odd that we call the state `done` but here refer to it as `approved`. Should we rename this `is_last_done`?
Admittedly, that's a confusing name - but at least it's referring to the same thing. Or, we refactor the status name from `done` to `approved` so it matches with this?

client/ayon_core/tools/common_models/projects.py
```diff
@@ -1,17 +1,24 @@
...
+from typing import Literal, Dict, Any
 
 import ayon_api
-import six
 
 from ayon_core.style import get_default_entity_icon_color
 from ayon_core.lib import CacheItem, NestedCacheItem
 
 PROJECTS_MODEL_SENDER = "projects.model"
+StatusStatesType = Literal["not_started", "in_progress", "done", "blocked"]
```
BigRoy, line 11: It's good to note that adding `Literal` typing makes this support Py3.8+ only. [It's added in Py3.8](https://docs.python.org/3/library/typing.html#typing.Literal). Is that a problem? 

I believe Maya 2022 is Py3.7 still maybe?

client/ayon_core/tools/sceneinventory/models/containers.py
iLLiCiTiT: It should really care about last approved, how the value is filled might change in future.

NOTE: We had a long discussion about how to define "last approved" and came to conclusion that it might change in future with more specific settings.

client/ayon_core/tools/common_models/projects.py
iLLiCiTiT: Hmm, might be.

m-u-r-p-h-y (approved):
image

correctly offers Updated to last approved only for outdated items and loads the highest approved version

client/ayon_core/tools/common_models/projects.py
iLLiCiTiT: Changed to use `str`.

BigRoy:
I like it - but now I start to definitely want to have a "set default status on publish" (settings) and "choose publish status in publisher" (publisher attribute per instance) implementations 🗡️

PR #693, opened 2024-06-24
Houdini: Remove unused "remote publish" workflow for Houdini

BigRoy:
## Changelog Description

Remove the "Remote Publish" workflow that lives within the `ayon_houdini` codebase but is unused. Some farm submissions for rendering and caching do work, with render targets allow local, farm rendering and using existing frames (for some) but none of these seem to rely on this Remote Publish logic that essentially was unused and likely would not work at all.

## Additional info

This likely was once a 'straight' copy from `colorbleed` codebase back in the Avalon-era.

I also detected one other plug-in that seemed to have `deadline` as registered target for the Pyblish plug-in [here](https://github.com/ynput/ayon-core/blob/9b258490148671e9f37413ad9e483e30c09ab544/server_addon/deadline/client/ayon_deadline/plugins/publish/validate_expected_and_rendered_files.py#L16) [(reported on discord)](https://discord.com/channels/517362899170230292/517382145552154634/1254902546452054158). Which likely tells me that this plug-in is currently never triggering since we never seem to be registering that target. We only ever seem to register `local`, `farm`, `remote` and `automated` according to the codebase.

## Testing notes:

1. Houdini integration should still work
2. Houdini publishing should still work
  - [x] Local publishing
  - [x] Farm render publishing
  - [x] Farm cache publishing


iLLiCiTiT (approved):

moonyuet (approved):
Tested successfully in Houdini.
Farm cache submission hit some issue but it is not related to this PR.
```
  File "D:\ayon-core\client\ayon_core\pipeline\farm\pyblish_functions.py", line 942, in create_instances_for_cache
    raise ValueError("Found multiple non related files "
ValueError: Found multiple non related files to render, don't know what to do with them.
```

PR #699, opened 2024-06-25
Maya: Publishing fbx in model family

moonyuet:
## Changelog Description
This PR is to add support to publish fbx in model product type in Maya

## Additional info
Update maya addon to 0.2.8

## Testing notes:
1. Go to your AYON server, make sure `Collect Model for FBX export` enabled
image

2. Launch Maya
3. Create Model Instance
4. You can choose to turn on/off `Collect Model for FBX Export`
image
5. If you turn on, it will publish fbx with model product type.

### DIFF
--- a/server_addon/maya/client/ayon_maya/plugins/publish/collect_fbx_model.py
+++ b/server_addon/maya/client/ayon_maya/plugins/publish/collect_fbx_model.py
@@ -9,7 +9,7 @@ class CollectFbxModel(plugin.MayaInstancePlugin,
     """Collect Camera for FBX export."""
 
     order = pyblish.api.CollectorOrder + 0.2
-    label = "Collect Model for FBX export"
+    label = "Collect Fbx Model"
     families = ["model"]
     optional = True
 
--- a/server_addon/maya/client/ayon_maya/plugins/publish/collect_fbx_model.py
+++ b/server_addon/maya/client/ayon_maya/plugins/publish/collect_fbx_model.py
@@ -1,0 +1,29 @@
+import pyblish.api
+from ayon_core.pipeline import OptionalPyblishPluginMixin
+from ayon_maya.api import plugin
+
+
+
+class CollectFbxModel(plugin.MayaInstancePlugin,
+                      OptionalPyblishPluginMixin):
+    """Collect Camera for FBX export."""
+
+    order = pyblish.api.CollectorOrder + 0.2
+    label = "Collect Model for FBX export"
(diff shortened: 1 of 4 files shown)


server_addon/maya/client/ayon_maya/plugins/publish/collect_fbx_model.py
```diff
@@ -0,0 +1,29 @@
...
+from ayon_maya.api import plugin
+
+
+
+class CollectFbxModel(plugin.MayaInstancePlugin,
+                      OptionalPyblishPluginMixin):
+    """Collect Camera for FBX export."""
+
+    order = pyblish.api.CollectorOrder + 0.2
+    label = "Collect Model for FBX export"
```
BigRoy, line 12: This would be quite an odd label in publisher UI for the checkbox to toggle the FBX export on/off for a model family? If this were just "Export FBX (model)" or alike then it'd be I think a better checkbox in the publisher UI but maybe an odd label for the Collector itself.

---

Also, this highlights again the issues with how the publishing logic works that it requires this odd family "ingesting" from a collector to support the FBX export. Still feels odd we need to do it this way but can't think of anything easier within current design. "Product aliases" and other things have been mentioned in the past and also a lot of "yes yes - should be improved" has come up often. Would still be very interested in seeing proposals in this area or having momentum of pushing that forward. The longer we are where we are the bigger this codebase will just grow with these workarounds that feel out-of-place yet unfortunately are required for now.

server_addon/maya/server/settings/publishers.py
```diff
@@ -1047,6 +1056,9 @@ class PublishersModel(BaseSettingsModel):
     "CollectFbxCamera": {
         "enabled": False
     },
+    "CollectFbxModel": {
+        "enabled": False
+    },
```
BigRoy, line 1061: Should we also expose the optional state so that a studio could enforce the FBX generation if they want to?

BigRoy (approved):
**Didn't test**

I understand why it's needed in the current publishing design from product type -> family and how validators apply or not based on this family getting ingested into it. Nonetheless I feel we should really push back to the drawing board and think a bit again about the API design we actually want to just pick and choose whether a particular product export would or would not get certain output representations, and/or validations, etc.

LiborBatek (approved):
All works as expected...

I was able to produce ```model``` in the ```FBX``` repre and also tested using ```Loader```...

Screenshot 2024-06-26 111911

PR #701, opened 2024-06-25
Houdini: Support collect frames for files with dot in filename (and other special characters)

BigRoy:
## Changelog Description

Fixes detection of output sequences where the filename had an extra dot in it, like e.g. `$HIP/vdbcache.fixed.1001.vdb`.

## Additional info

The previous regex match of `\w` in the filename disallowed matching `.` or other special characters - it now allows any character before the expected end.

Also adds some small debug logs to make debugging easier in the future.

## Testing notes:

1. A publish should work with a `vdbcache` publish with e.g. output path set to `$HIP/vdbcache.fixed.1001.vdb`

### DIFF
--- a/server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
+++ b/server_addon/houdini/client/ayon_houdini/plugins/publish/collect_frames.py
@@ -48,16 +48,18 @@ class CollectFrames(plugin.HoudiniInstancePlugin):
         match = re.match(pattern, file_name)
 
         if match and start_frame is not None:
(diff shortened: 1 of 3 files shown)


moonyuet (approved):
image
Published successfully in my side(Houdini 19.5)

MustafaJafar:
I was going to suggest using `clique` like in https://github.com/ynput/ayon-core/pull/9
But, then I found out that it was already implemented there.

BigRoy:
Closing this in favor of changes in #9

PR #723, opened 2024-06-27
Publisher: Remove Literal typehint

iLLiCiTiT:
## Changelog Description
Remove used `Literal` typehint.

## Additional info
Typehint `Literal` is breaking older DCCs with Python older than 3.8.

## Testing notes:
1. Publisher works in those DCCs.


BigRoy (approved):
Change makes sense :) the fun of supporting many versions, yay!

64qam (approved):

Issue #703, opened 2024-06-25
Scene Inventory: Do not fail with invalid representation ids on containers


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

As we are transitioning from OpenPype (or even converting scenes from all the way back from Avalon) we're facing issues with the scenes after the conversion if there's any trace of a detected container in the scene with an OpenPype or Avalon representation id.

When this occurs the Scene Inventory will remain empty and show no logged errors whatsoever. 

There seems to be an invalid id check in the code [here](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/tools/sceneinventory/models/containers.py#L226-L230) - but apparently it doesn't do what it should?

Note that the issue itself goes beyond just the Scene Inventory in certain validation plug-ins or alike where a query is made to the `ayon_api` with these invalid ids where it fails on the Graph QL queries because it disallows the invalid ids. An approach we could take is to have Ayon API filter those out with the uuid check.

### Expected Behavior:

It should allow to list invalid containers and if any issue arises like this it should provide a logged stack trace as well.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

To reproduce:

- Load a published product into a host integration
- On the container manipulate the data and remove some characters from representation id.
- Load another published product
- Open the scene inventory
- Nothing shows, both the broken and the normal container are not listed.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Additional context:

I wish there was some log or error output to share - but there's literally no output whatsoever when this fails.

PR #725, opened 2024-06-27
SceneInventory: Ignore containers with invalid UUID in representation

iLLiCiTiT:
## Changelog Description
Fake representation id of containers with invalid representation id.

## Additional info
This is proposal of possible issue with using scenes created in OpenPype. Proposed solution does show the container in UI but as invalid so user can see it. API functions are silently ignoring containers with invalid representation id.

Another solution would be to completelly ignore it even in UI to avoid other possible issues.

## Testing notes:
1. Have scene with referenced content from OpenPype.
2. Open scene inventory.
3. It should not crash.
4. ValidateContainers should not crash either.

Resolves https://github.com/ynput/ayon-core/issues/703


BigRoy:
And here are the details when logging the exception explicitly:
```
  File "E:\dev\ayon-core\client\ayon_core\tools\sceneinventory\model.py", line 153, in refresh
    version_items_by_product_id = self._controller.get_version_items(
  File "E:\dev\ayon-core\client\ayon_core\tools\sceneinventory\control.py", line 114, in get_version_items
    return self._containers_model.get_version_items(product_ids)
  File "E:\dev\ayon-core\client\ayon_core\tools\sceneinventory\models\containers.py", line 312, in get_version_items
    version_entities = list(ayon_api.get_versions(
  File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\server_api.py", line 4564, in get_versions
    for parsed_data in query.continuous_query(self):
  File "C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\graphql.py", line 380, in continuous_query
    raise GraphQlQueryFailed(
ayon_api.exceptions.GraphQlQueryFailed: GraphQl query Failed: Variable '$productIds' got invalid value None at 'productIds[2]'; Expected non-nullable type 'String!' not to be None. (Line 1 Column 43)A
```
It seems that any error in the `refresh` method doesn't raise any actual error or log it anywhere. This may be due to how the Qt UI processes it (it's maybe in a callback or whatever?)

BigRoy:
Still an empty scene inventory and **no errors** visible in console.

With some changes the stack trace now is:
```python
  File "E:\dev\ayon-core\client\ayon_core\tools\sceneinventory\model.py", line 129, in refresh
    self.__refresh(selected)
  File "E:\dev\ayon-core\client\ayon_core\tools\sceneinventory\model.py", line 240, in __refresh
    repre_info.representation_name
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
Note that the line numbers might be off because I had to put the whole `refresh` in a try except to get any error.
I basically added the lines:
```python
    def refresh(self, selected=None):
        """Refresh the model"""
        # for debugging or testing, injecting items from outside
        try:
            self.__refresh(selected)
        except Exception:
            self.log.error("Error occurred during refresh.", exc_info=True)

    def __refresh(self, selected=None):
        # old refresh method
```

Somehow due to how refresh gets called via Qt it seems to report no error or stack trace?

BigRoy (approved):
**Lovely!**

Works in Fusion:
image

Works in Houdini 20.0.724
image

Works in Maya 2024:
image

Here you can also see that faulty containers with same representation id value are grouped together.

---

Not entirely related to this PR, but starting Houdini as I tried to break some things... it had a custom HDA that I was lacking in this branch and on startup I'd then get this error:
```python
> E:\dev\ayon-core\server_addon\houdini\client\ayon_houdini\api\pipeline.py
  File "E:\dev\ayon-core\client\ayon_core\lib\events.py", line 339, in process_event
    callback()
  File "E:\dev\ayon-core\server_addon\houdini\client\ayon_houdini\api\pipeline.py", line 356, in on_open
    if any_outdated_containers():
  File "E:\dev\ayon-core\client\ayon_core\pipeline\load\utils.py", line 797, in any_outdated_containers
    if get_outdated_containers(host, project_name):
  File "E:\dev\ayon-core\client\ayon_core\pipeline\load\utils.py", line 824, in get_outdated_containers
    return filter_containers(containers, project_name).outdated
  File "E:\dev\ayon-core\client\ayon_core\pipeline\load\utils.py", line 869, in filter_containers
    repre_ids = {
  File "E:\dev\ayon-core\client\ayon_core\pipeline\load\utils.py", line 872, in <setcomp>
    if _is_valid_representation_id(container["representation"])
KeyError: 'representation'
```
Not something we need to fix here - just seemed somewhat related and I wanted to mention it somewhere

iLLiCiTiT:
> Not something we need to fix here - just seemed somewhat related and I wanted to mention it somewhere

It is not handled on multiple places, definetelly should be. But indeed different PR.

PR #728, opened 2024-07-01
Fix: dirmap wasn't using local site values

kalisp:
## Changelog Description
Some DCC(Maya, Nuke) offer dirmap functionality, eg. remapping file paths from source to destination. If dirmap is enabled and artist is using local site, it should use as a source root value from `Site Settings` (which might be different for each artist) rather than static mapping from `ayon+settings://nuke/dirmap?project=ayon_test` for example.

## Testing notes:
1. enable dirmap in Nuke in `ayon+settings://nuke/dirmap?project=ayon_test`
2. you might keep mapping there empty, values from `Local Settings` should be used
3. open Nuke and look for 
image
where values in `active roots overrides` should match what artist has in `Site Settings`
image


LiborBatek (approved):
I have followed testing steps and all looking as supposed to...so all good!

My settings:
Screenshot 2024-07-02 115103

And console after launching showing dir remap in use...


Screenshot 2024-07-02 115119

iLLiCiTiT (approved):

PR #200, opened 2024-03-19
Deadline: webservice password

kalisp:
## Changelog Description
This PR provides support for Deadline Webservice authentication.

## Additional info
Settings are rudimentary and will need to be changed as we currently don't have `password` type of `SettingsField` which would mask and hash values. AFAIK I cannot use `Secrets` as they are not per artist.

Currently each artist needs to provide credentials via Site Settings, I didn't provide some default values as I assume if client expects  authentication they will not use shared account. 

## Testing notes:
1. Enable authentication in Deadline Monitor `Tools>Configure Repository Options>Web Service Settings>Require Authentication`
2. Enable authentication requirement for the project
image
3. Create user in DL and provide password
image
4. Use these credentials in Site Settings
image

PR #729, opened 2024-07-01
Deadline: provide default user credential fields in studio settings

kalisp:
## Changelog Description
Some studios might have only single user/password credentials for communication with Deadline webservice. This PR provides new fields in `Studio Settings` which could be used as default values. These could be overridden by values provided by artists in `Site Settings` if necessary.

image

## Additional info
AYON secrets cannot be used as artist don't have access to those (and this runs in Tray). I am expecting pushback from customer though.

## Testing notes:
1. enable on Deadline > `Tools > Configure Repository Options > Web Service Settings > Require Authentication`
2. set passwor in Deadline > `Tools > Manage Users > Webservice password`
3. build and install Deadline addon (remove existing unzipped from `AppData`)
4. provide credentials in `ayon+settings://deadline/deadline_urls` (could be same as from https://github.com/ynput/ayon-core/pull/200 - just delete them first from `Site Settings` (on Studio page)
5. publish something to DL, check output from `Collect Deadline User Credentials`


MilaKudr (approved):

kalisp:
Hmm, was worried about that, so reverted to plain string field. 
We need password Setting key and probably fix `Since settings are sent to clients, secrets provide a way to hide sensitive data from users.`  in https://ayon.ynput.io/docs/admin_server_provisioning/#secrets as it points that secrets could be used (but actually cannot, just for use in services).

iLLiCiTiT:
Well artists don't have access to settings.

MilaKudr (approved):

PR #731, opened 2024-07-02
Chore: Removed Resolve addon

iLLiCiTiT:
## Changelog Description
Removed Maya addon from ayon-core repository as is moved to https://github.com/ynput/ayon-resolve .


BigRoy (approved):
I was able to save/load workfiles + load and manage products. I was unable to publish due to some errors but they seem errors not related to the move of the addon, so can be ignored. I've reported them on `ayon-resolve`.

PR #734, opened 2024-07-03
Chore: Add collect resource path for texture outputs from Ornatrix implementation 

moonyuet:
## Changelog Description
Add the feature of collecting resources path from Ornatrix implementation in this PR https://github.com/ynput/ayon-maya/pull/1 

## Additional info
n/a

## Testing notes:
1. Build the addon with this branch 
2. Test with the built addon from this PR https://github.com/ynput/ayon-maya/pull/1


client/ayon_core/plugins/publish/collect_resources_path.py
BigRoy, line 1: This seems to target a massive amount of families - and as the separation of addon showed [here](https://github.com/ynput/ayon-maya/pull/1#issuecomment-2205316866) I have a feeling we'll be doing ourselves a favor if we could look into whether this could just target all families instead.

iLLiCiTiT: I even didn't know we have such plugin. Strangely it expects that default publish template is used all the time...

BigRoy: I believe it's used for Maya Look publishing, Houdini USD Look publishing, Yeti Publishing and Ornatrix publishing to know before publishing the destination publish folder and remap paths to the new texture locations in the `{publish_dir}/resources` folder or alike.

iLLiCiTiT: Ok, but most of these product types are using different template than `"default"`.

BigRoy: That may very well be case. I'd say, separate issue to track that then?

iLLiCiTiT: In that case this whole conversation can be separate issue? 🙂 We can merge this PR and resolve the source issue with different PR.

BigRoy (approved):

client/ayon_core/plugins/publish/collect_resources_path.py
BigRoy: Done: https://github.com/ynput/ayon-core/issues/757

PR #746, opened 2024-07-03
Global: dirmap for non local nor studio sites

kalisp:
## Changelog Description
Site Sync addon throws ValueError for no local/studio sites as it doesn't make sense to override roots for sftp or gdrive. This handles gracefully this exception.

## Testing notes:
1. create GDrive or SFTP site in Studio Settings and choose it Site Settings as remote site
2. without this fix and enabled dirmap Maya won't show AYON in the menu
3. for local-studio tests nothing should change


client/ayon_core/host/dirmap.py
```diff
@@ -220,3 +218,19 @@ def _get_local_sync_dirmap(self):
 
             self.log.debug("local sync mapping:: {}".format(mapping))
         return mapping
+
+    def _get_site_root_overrides(
+            self, sitesync_addon, project_name, site_name):
```
iLLiCiTiT, line 223:
```suggestion
    def _get_site_root_overrides(
        self, sitesync_addon, project_name, site_name
    ):
```

iLLiCiTiT:
> Site Sync addon throws ValueError for no local/studio sites as it doesn't make sense to override roots for sftp or gdrive. This handles gracefully this exception.

But, shouldn't that be handled by sitesync addon? Asking foe root overrides should return empty dict inm that case, or?

client/ayon_core/host/dirmap.py
```diff
@@ -220,3 +218,22 @@ def _get_local_sync_dirmap(self):
...
+
+    def _get_site_root_overrides(
+        self, sitesync_addon, project_name, site_name
+    ):
+        """Safely handle root overrides.
+
+        SiteSync raises ValueError for non local or studio sites.
+        TODO: could be removed when `get_site_root_overrides` is not raising
+        an Error but just returns {}
+        """
```
iLLiCiTiT, line 230:
```suggestion
        """
        # TODO: could be removed when `get_site_root_overrides` is not raising
        #     an Error but just returns {}
```

iLLiCiTiT (approved):

PR #752, opened 2024-07-05
Loader tool: Filter by statuses

iLLiCiTiT:
## Changelog Description
Added combobox for filtering by statuses in loader tool.

## Additional info
Added multiselection combobox for statuses that can be used to filter versions in loader tool. Selection in combobox does show stauts icons in the combobox field. Combobox view also has 3 action items: Select all, Deselect all and Swap which swaps selection. These might be removed if it's too much.

We might add similar combobox for product types, but there would have to be added better implementation of paint for the combobox field as icons are not enough.

### Screenshots
image

image

## Testing notes:
1. Use loader from tray or use loader tool in host. It would be maybe better to test multiple hosts with Qt Nuke, Maya, Houdini etc. so we validate it looks and works there.

- [x] Maya
- [x] Blender 
- [ ] Nuke
- [ ] Houdini

3. Select folder with products.
4. Change statuses filter > versions or whole products should start to disappear.

EDITED:
I changed behavior of the combobox, when nothing is selected it behaves as everything is selected. Everything is deselected by default, so user is actually checking statuses he wants to see instead of deselecting them.
Maybe with this new behavior we don't need the helper actions to de/select all and swap selection.


LiborBatek (approved):
The status filters and all the options work well...been tested in Maya and Blender so far without any issues found.

There is small idea for functionality, as ```Disabled All``` it can appear as completely empty ```Loader``` aka no assets for loading as seen here when applied such filter ```None```

image

It is correct however if ```Loader``` closed and re-opened user can be easily confused as no product visible for loading! I would propose to loose any filter when Loader session ends and when re-opened...so this doesnt occur and no confusion might not happen.

LiborBatek (changes requested):
I have did another test round and now it behaves much more predictably - when loosing all filters the Loader shows content again (there cant happen situation with empty loader if no filter applied which is desirable - so working well now!)

However I have found issue that ```status``` diminishes as you re-open or refresh ```Loader``` as seen on my scrn capture below...

Loader_Statuses

iLLiCiTiT:
> owever I have found issue that status diminishes as you re-open or refresh Loader as seen on my scrn capture below...

Good catch, should be fixed.

LiborBatek:
It definetely is fixed now but here is another issue when adding some more ```statuses``` into filter...seems like new issue or undiscovered one too

Loader_Statuses_02

iLLiCiTiT:
I couldn't replicate the issue, please DM me when you're available.

LiborBatek (approved):
Now all statuses get properly refreshed...didnt spot any issues!

All seems working smooth and nicely!

image

BigRoy:
This is huge guys, great stuff!

PR #761, opened 2024-07-08
Publishing collectors: Ignore invalid representation ids

BigRoy:
## Changelog Description

Ignore invalid representation ids (e.g. instances from OpenPype/Avalon or broken containers or other special cases) so that publishing does not error on them.

## Additional info

Fixes https://github.com/ynput/ayon-houdini/issues/31

The `is_valid_uuid` function now exists twice in `ayon-core`, once in `ayon-maya` and once in `ayon-houdini`. I wonder whether it'd be worth it to somehow expose a function for it in ayon core lib?

## Testing notes:

1. Publishing with containers in scene that have invalid ids should be fine.


client/ayon_core/plugins/publish/collect_input_representations_to_versions.py
```diff
@@ -1,7 +1,18 @@
+import uuid
+
 import ayon_api
 import pyblish.api
 
 
+def is_valid_uuid(value) -> bool:
```
iLLiCiTiT, line 7: We might use `convert_entity_id` from `ayon_api`.

BigRoy: I like that! :) Implemented with https://github.com/ynput/ayon-core/pull/761/commits/a782ede959d3689e27d06e6cc991d5d31b1a2741

iLLiCiTiT (approved):

PR #764, opened 2024-07-09
Chore: Make sure udim in representation is not a list

moonyuet:
## Changelog Description
Udim in representation context is not list of integers but only single integer.

## Additional info
This PR is to fix the bug where the scene inventory cannot update the filepath with udim extension, which uncovered that key udim in representation context data is stored as list of integers. That was changed to store first used udim value. For old representations was added auto-fix in get representation path functions to automatically use first item from udim list.

## Testing notes:
Build and install the addon with this branch.
1. Build addon with this branch
2. Launch Substance Painter
3. Publish anything with udim
4. Launch Maya
5. load image texture
6. update/set version with scene inventory


iLLiCiTiT:
What is difference from default template?

moonyuet:
> What is difference from default template? And why we have default setting for it if we don't have that template in default anatomy?

very small difference:<_{udim}> and <.{udim}>

BigRoy:
> very small difference:<_{udim}> and <.{udim}>, any way to add the template into default anatomy?

Aren't we better of just updating the default then?
And also, why does Maya fail with the first one - isn't that technically valid as well? (And hence, may just need fixing om some loading part in maya instead?)

moonyuet:
>Aren't we better of just updating the default then?
>And also, why does Maya fail with the first one - isn't that technically valid as well? (And hence, may just need fixing om some loading part in maya instead?)

It doesn't fail in loading but updating assets, and it is not failing in maya loading but the scene inventory.

BigRoy:
> I am trying to find the json data which stores the data in `template.json` in OP. But I can't

Ah, good point - those are **anatomy settings** I believe and are actually in the backend as presets. I noticed that recently when doing this draft PR: https://github.com/ynput/ayon-backend/pull/252

As such they should be somewhere here: https://github.com/ynput/ayon-backend/blob/develop/ayon_server/settings/anatomy/templates.py

> It doesn't fail in loading but updating assets, and it is not failing in maya loading but the scene inventory.

Again, seems like we should fix that instead. This sounds like a Maya loader issue - not a template issue.

iLLiCiTiT:
NOTE: The bug that happened on update IS NOT related to template.

BigRoy:
> NOTE: The bug that happened on update IS NOT related to template. If this is meant as bugfix of that bug then this won't help.

I had to figure out what the issue was - I reported it for now as this: https://github.com/ynput/ayon-core/issues/765

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -765,7 +765,7 @@ def prepare_representation(
             # Manage anatomy template data
             template_data.pop("frame", None)
             if is_udim:
-                template_data["udim"] = repre["udim"][0]
+                template_data["udim"] = repre["udim"]
```
BigRoy, line 768: This is wrong and will break the templating for single UDIM files. Please revert.

iLLiCiTiT:
This change breaks backwards compatibility, does not resolve existing representations and to be honest I don't think is correct fix for this.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -792,7 +792,7 @@ def prepare_representation(
         # Explicitly store the full list even though template data might
         # have a different value because it uses just a single udim tile
         if repre.get("udim"):
-            repre_context["udim"] = repre.get("udim")  # store list
+            repre_context["udim"] = repre.get("udim")[0]  # store list
```
BigRoy, line 795: The comment above this scares me so much - why did I do this? :)

Anyway, I believe that technically we should be able to just take the value from the `template_data` because that would've always been the `udim` individual tile that formatted the path. Which is what this value should've been to begin with.

Only thing somewhat related I can find is:
- this comment discussion: https://github.com/ynput/OpenPype/pull/2898#discussion_r829199778
- https://github.com/ynput/OpenPype/pull/2898/commits/b128e0addffc77991e5ff25f2d219d8ed8613136
- https://github.com/ynput/OpenPype/pull/2898#issuecomment-1092124350

BigRoy: Having a look at the code again I think unfortunately, how odd it is - this may be the correct fix. I can't find any other usage of the `udim` key in which this seems to become problematic at first sight.

Only other thing that I thought of is that we could also just remove these lines completely because the `udim` key should already be in `repre_context["udim"]` due to them being used values in the template. The only thing keeping this line around with your changes does is make sure the `udim` is also in the context if it were not used by the template. Whether that'd be correct or not behavior @iLLiCiTiT will know - going based of the links in my above comment he had some opinions on that, but looking back now and re-reading that I'm still not sure I fully understood what he concluded it at.

iLLiCiTiT:
Modified PR. Adding explicitly `"udim"` to representation context is skipped as it is there automatically within `used_values` on filled template. Added auto-fix for old representations to use first udim during get representation path.

client/ayon_core/pipeline/load/utils.py
```diff
@@ -639,6 +639,14 @@ def get_representation_path_with_anatomy(repre_entity, anatomy):
     try:
         context = repre_entity["context"]
         context["root"] = anatomy.roots
+
+        # Auto-fix 'udim' being list of integers
+        # - This is a legacy issue for old representation entities,
+        #   added 24/07/10
+        udim = context.get("udim")
+        if isinstance(udim, list):
+            context["udim"] = udim[0]
```
BigRoy, line 648: Can we make one `_fix_representation_context_compatibility` function that is called in both these places that does this logic? That way, if we ever run into another issue for this, we can just expand that function? May be nicer than having it duplicated now?

Kind of [like this in avalon](https://github.com/getavalon/core/blob/master/avalon/pipeline.py#L1215)?

BigRoy (approved):
Changes look good to me. Would be great if @moonyuet and/or @LiborBatek can confirm this works with existing substance painter products and whether new products also work AFTER publishing with this update from substance painter.

That publish from substance painter will need to be 'investigated manually' because the "backwards compatibility" fix will hide it if currently the new published data is still wrong.

iLLiCiTiT:
> That publish from substance painter will need to be 'investigated manually' because the "backwards compatibility" fix will hide it if currently the new published data is still wrong.

What's wrong with substance painter? Did I miss something?

LiborBatek (approved):
It works well now!

It even does work when ```Manage > Update to latest version``` and each ```version``` of ```image``` product have different publishing template used both with ```_{udim}``` and ```.{udim}``` which is even more impressive to me!

I guess can be merged then...I have also test to produce new ```image``` product in ```Substance Painter``` host to be sure nothing is broken by this PR. All good and working...

@iLLiCiTiT small note tho, probably for other PR?? when performing ```Manage > Set Version``` it works just on single ```image``` product not for multiple selected image products (even tho other actions like the aformentioned ```update to latest``` works for multiselection of images)

here is the traceback for you:
```
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 366, in <lambda>
#     lambda: self._show_version_dialog(item_ids, active_repre_id)
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 820, in _show_version_dialog
#     version_option = SelectVersionDialog.ask_for_version(
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\sceneinventory\select_version_dialog.py", line 255, in ask_for_version
#     dialog.set_versions(version_options)
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\sceneinventory\select_version_dialog.py", line 242, in set_versions
#     self._versions_combobox.set_versions(version_options)
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\sceneinventory\select_version_dialog.py", line 175, in set_versions
#     icon = get_qt_icon({
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\utils\lib.py", line 578, in get_qt_icon
#     return _IconsCache.get_icon(icon_def)
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\utils\lib.py", line 496, in get_icon
#     cache_key = cls._get_cache_key(icon_def)
#   File "C:\Work\REPO\ayon-core\client\ayon_core\tools\utils\lib.py", line 489, in _get_cache_key
#     return "|".join(parts)
# TypeError: sequence item 1: expected str instance, NoneType found
```
I think this one was fixed recently but maybe with ```UDIMs``` its not fixing it...

LiborBatek:
@iLLiCiTiT for better info here is the working occasion vs mutliselection when trying ```Manage > Set Version```

Again, this could be for other PR maybe??

iLLiCiTiT:
>  small note tho, probably for other PR?? when performing Manage > Set Version it works just on single image product not for multiple selected image products (even tho other actions like the aformentioned update to latest works for multiselection of images)

Looking into it. Merging this one tho.

BigRoy:
> What's wrong with substance painter? Did I miss something?

Sorry @iLLiCiTiT what I meant was - we should double check whether the new UDIM publish now actually does integrate with the value to NOT be a list anymore in the database and that it results in a value we expected in `representation_entity["context"]`

iLLiCiTiT:
@LiborBatek did you try to publish new texture with udim? Could you quickly try with develop if not?

iLLiCiTiT:
Libor confirmed he tried to publish and update new textures.

BigRoy:
> Libor confirmed he tried to publish and update new textures.

Haha, sorry for going back to this again. But my point is that the updating itself would work regardless, even if the new publish data would still be storing it as a `list` for whatever reason because we have the conversion logic there now as well.

To be entirely sure the data is now fixed we may need to investigate the actual representation entity's data to ensure `representation_entity["context"]` is now a `str` like `1001` instead of `list` like `["1001"]`.

Looking at the code I suspect it should be fixed - but the code is also complex in other areas and we might be missing another subtle bug.

iLLiCiTiT:
> To be entirely sure the data is now fixed we may need to investigate the actual representation entity's data to ensure representation_entity["context"] is now a str like 1001 instead of list like ["1001"].
>
> Looking at the code I suspect it should be fixed - but the code is also complex in other areas and we might be missing another subtle bug.

Aaah, now I get it. Will validate with Libor tomorrow.

PR #766, opened 2024-07-09
Loader plugin: Filtering by version status

iLLiCiTiT:
## Changelog Description
Added simplified filtering by version status into loader plugins.

## Additional info
Added another simplified filter to loader plugin similar to product types and representation extensions. The filtering is not used anywhere now.

## Testing notes:
1. Validate the code changes.

### DIFF
--- a/client/ayon_core/pipeline/load/plugins.py
+++ b/client/ayon_core/pipeline/load/plugins.py
@@ -1,5 +1,6 @@
 import os
 import logging
+from typing import Any, Dict
 
 from ayon_core.settings import get_project_settings
 from ayon_core.pipeline.plugin_discover import (
@@ -25,6 +26,7 @@ class LoaderPlugin(list):
 
     product_types = set()
     representations = set()
+    statuses = set()
     extensions = {"*"}
     order = 0
     is_multiple_contexts_compatible = False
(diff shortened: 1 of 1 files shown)


BigRoy:
I have a feeling this is getting a bit into the realm of over designing the system.

Why couldn't the loader just expose a 'is_compatible' method so that the loader can decide on the full context whether it should show or not where the default behavior may be product types check, but can be overridden for other specialized behavior, like filtering on a status?

Also, having not seen this before - a question for context. What type of loader actions would maybe filter based in status? In what situations does it make sense to completely disallow a loader action to run over other status even if the user wanted to? Like, what is the use case?

The only thing I can think of is maybe the delivery action? But even then it feels overly strict maybe.

iLLiCiTiT:
I don't know, it's true I've created the task for a client based on mouth of word so I might hear wrong, or it was misinterpreted.

m-u-r-p-h-y:
Isn't it a simple case where the studio wants only to show approved versions in the loader?

BigRoy:
> Isn't it a simple case where the studio wants only to show approved versions in the loader?

That would just mean that the Loader UI would filter versions which is this PR: https://github.com/ynput/ayon-core/pull/752

This PR is about actually making the Loader actions not show at all when certain statuses are set - which I think will become very tricky with e.g. scene inventory and version up behavior - because you may be unable to version up then just because the newer version may not have a matching status and then suddenly it will start telling you the loader is 'incompatible' with the newer version? Anyway, this spawns off potentially a host of other problems and I think https://github.com/ynput/ayon-core/pull/752 does solve the majority of the use cases whilst remaining generic functionally in the loader plug-ins themselves.

iLLiCiTiT:
Closing for irrelevant change.

PR #769, opened 2024-07-10
UI Icons: Cache key is safely generated

iLLiCiTiT:
## Changelog Description
Cache key expect values to be `None` and `QColor`.

## Additional info
Issue discovered in https://github.com/ynput/ayon-core/pull/764#pullrequestreview-2169128475 causing that `Set version` on multiselection does not work as it request icon and color set as `None` . That already happened few times, always causing crashes of UI, so maybe we should expect broken values and show empty icons rather than crashing tools.

## Testing notes:
1. Set version in scene inventory does work on multiselection of multiple different products.


LiborBatek (approved):
All working great...

Multiselection, mutlitselection of mixed up products all good for managing versions...

Good to go!

PR #772, opened 2024-07-12
Loader: Change placeholder text to "Version status filter..."

BigRoy:
## Changelog Description

Change placeholder text from "Statuses..." to "Version status filter..." to align more with the filter fields to left of it.

Before:

before

After:

image

## Additional info

For context - there's also a "Folder name filter..." a bit further left.

image

## Testing notes:

1. Loader should display the new "Version status filter..." placeholder text


jakubjezek001 (approved):
LGTM 😍

PR #773, opened 2024-07-12
Delete old versions: Add confirm prompt to avoid accidental delete

BigRoy:
## Changelog Description

Add informative confirm prompt for delete versions to avoid accidental clicks deleting versions and files.

## Additional info

image

With "Show details..." clicked:

image

## Testing notes:

1. Open Loader from Tray
2. Use Delete Old Versions action
3. Deletion should be confirmed (and do nothing when cancelled!)

### DIFF
--- a/client/ayon_core/plugins/load/delete_old_versions.py
+++ b/client/ayon_core/plugins/load/delete_old_versions.py
@@ -1,7 +1,7 @@
 import collections
 import os
 import uuid
-from typing import Optional, List, Dict, Any
+from typing import List, Dict, Any
 
 import clique
 import ayon_api
--- a/client/ayon_core/plugins/load/delete_old_versions.py
+++ b/client/ayon_core/plugins/load/delete_old_versions.py
@@ -48,7 +48,7 @@ class DeleteOldVersions(load.ProductLoaderPlugin):
         size = 0
 
         for dir_path in dir_paths:
-            # Delete all files and fodlers in dir path
+            # Delete all files and folders in dir path
             for root, dirs, files in os.walk(dir_path, topdown=False):
                 for name in files:
                     file_path = os.path.join(root, name)
--- a/client/ayon_core/plugins/load/delete_old_versions.py
+++ b/client/ayon_core/plugins/load/delete_old_versions.py
@@ -220,10 +220,8 @@ class DeleteOldVersions(load.ProductLoaderPlugin):
         messagebox.setIcon(QtWidgets.QMessageBox.Warning)
         messagebox.setWindowTitle("Delete Old Versions")
         messagebox.setText(text)
-        if informative_text:
-            messagebox.setInformativeText(informative_text)
-        if detailed_text:
-            messagebox.setDetailedText(detailed_text)
+        messagebox.setInformativeText(informative_text)
+        messagebox.setDetailedText(detailed_text)
         messagebox.setStandardButtons(
             QtWidgets.QMessageBox.Yes
             | QtWidgets.QMessageBox.Cancel
--- a/client/ayon_core/plugins/load/delete_old_versions.py
+++ b/client/ayon_core/plugins/load/delete_old_versions.py
@@ -1,7 +1,7 @@
 import collections
 import os
 import uuid
-from typing import Optional
+from typing import Optional, List, Dict, Any
 
 import clique
 import ayon_api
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
When I click on delete old version:
it shows the popup dialog as expected
image
if I hit Cancel, it won't error out
If I hit Ok, it would keep the latest 45 versions as expected.
image

I also tested with Calculate Old Version, it also works smoothly without error.

BigRoy:
@iLLiCiTiT could you provide this a quick glance and merge if you're ok with it?

iLLiCiTiT:
I would rename `Yes` to `Delete` and moved `Show details` to most left.

client/ayon_core/plugins/load/delete_old_versions.py
```diff
@@ -408,6 +458,8 @@ class CalculateOldVersions(DeleteOldVersions):
         )
     ]
 
+    requires_confirmation = False
```
iLLiCiTiT, line 461: Is this used?

client/ayon_core/plugins/load/delete_old_versions.py
```diff
@@ -365,19 +394,40 @@ def main(self, project_name, data, remove_publish_folder):
...
+        # Because we do not want this run by accident we will add an extra
+        # user confirmation
+        if self.requires_confirmation:
+            contexts_list = "\n".join(sorted(
+                "- {folder[name]} > {product[name]}".format_map(context)
+                for context in contexts
+            ))
+            num_contexts = len(contexts)
+            s = "s" if num_contexts > 1 else ""
+            if not self.confirm(
```
iLLiCiTiT, line 418: I would change `confirm` to private, and the method can expect only `contexts` and the rest of text would be filled there?

BigRoy:
> I would rename `Yes` to `Delete` (if possible you can set object name of delete button to `#DeleteButton`), and moved `Show details` to most left.

This would 'adhere' to the rules of the current OS - and hence it may look different on Mac OS or Linux if the "Show details" behavior is different there because it's default Qt message box behavior. I'd rather not touch it.

client/ayon_core/plugins/load/delete_old_versions.py
BigRoy: Yes - because it inherits the `load` implementation of the class above it and we don't want to prompt if we're just calculating versions.

client/ayon_core/plugins/load/delete_old_versions.py
BigRoy: I could do that - I just made it this way so that the `confirm` is generic and could potentially be reused down the line for other confirmations. Just felt the most sensible to have the confirm box be a generic function.

Happy to refactor it though if you still want me to. Let me know if you want that.

iLLiCiTiT: Still `confirm` should be private method. I don't think there will be re-use of the method, but that still can be achieved by adding one other method `_confirm_delete` which would expect only contexts and prepare texts for `_confirm`.

iLLiCiTiT:
> This would 'adhere' to the rules of the current OS - and hence it may look different on Mac OS or Linux if the "Show details" behavior is different there because it's default Qt message box behavior. I'd rather not touch it.

I'm not good at UX, but this doesn't look nice. I had to read the text 2 times to understand what would the buttons do, and was confused why is the biggest button showing details between `Yes` and `Cancel`, and still don't understand why it's there (not as a developer but as user).

client/ayon_core/plugins/load/delete_old_versions.py
BigRoy: Done with https://github.com/ynput/ayon-core/pull/773/commits/1ed42d601a27b349e155ac541b84761b8e72799d

BigRoy:
> I'm not good at UX, but this doesn't look nice. I had to read the text 2 times to understand what would the buttons do, and was confused why is the biggest button showing details between `Yes` and `Cancel`, and still don't understand why it's there (not as a developer but as user).

I see what you mean - yet at the same time - it does allow us to not having to write dedicated widgets and rely on what apparently are more OS native designs for this.

I could also just remove the details information - but I felt since it was now simple using this feature that it's nice 'extra check' someone can do to decide whether it's deleting what they wanted?

@m-u-r-p-h-y from artist perspective? Would you like me to redesign the pop-up dialog? See screenshots in PR description.

m-u-r-p-h-y (approved):
Delete dialogue is working fine and prevents from accidental deletion of files.

The UI itself needs a lot of love though. 

Details page is not taking the `Remove Publish Folder` into account 

image

BigRoy:
> Details page is not taking the `Remove Publish Folder` into account

Would you like the details tab to just report that it's also removing the publish folder - or what kind of output did you expect additionally there?

PR #774, opened 2024-07-13
Workfile Template Builder: Allow Create Placeholder to define the 'active' state

BigRoy:
## Changelog Description

Workfile Template Builder: Allow Create Placeholder to define the 'active' state of the created instance.

## Additional info

Example in Houdini using https://github.com/ynput/ayon-houdini/pull/36

image

The template:

image

After template build:

image

## Testing notes:

> would be nice to test in Nuke and Maya if it does not break anything.
>    Use old template where are placeholders without active.
>    Use new template and try if it creates activated/deactivated instances based on the placeholder data.

1. Test whether this works when creating create placeholders in other hosts, like:
    - [ ] Nuke 
    - [x] Houdini (requires https://github.com/ynput/ayon-houdini/pull/36
 and bugfix https://github.com/ynput/ayon-houdini/pull/41 )
    - [ ] Maya (does not have "Create" placeholders - but may be good to test whether it still works)
    - [ ] (not sure which other hosts support Create Placeholders?)
2. Existing create placeholders in an existing template should still work and default to active.

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -2043,7 +2043,8 @@ class CreateContext:
         variant,
         folder_entity=None,
         task_entity=None,
-        pre_create_data=None
+        pre_create_data=None,
+        active=True
     ):
         """Trigger create of plugins with standartized arguments.
 
@@ -2061,6 +2062,8 @@ class CreateContext:
                 of creation (possible context of created instance/s).
             task_entity (Dict[str, Any]): Task entity.
             pre_create_data (Dict[str, Any]): Pre-create attribute values.
+            active (Optional[bool]): Whether the created instance defaults
+                to be active or not. Defaults to True.
 
         Returns:
             Any: Output of triggered creator's 'create' method.
@@ -2124,7 +2127,8 @@ class CreateContext:
             "folderPath": folder_entity["path"],
             "task": task_entity["name"] if task_entity else None,
             "productType": creator.product_type,
-            "variant": variant
+            "variant": variant,
+            "active": bool(active)
         }
         return creator.create(
             product_name,
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -2130,7 +2130,12 @@ class CreateContext:
             "variant": variant
         }
         if active is not None:
-            instance_data["active"] = bool(active)
+            if not isinstance(active, bool):
+                self.log.warning(
+                    "CreateContext.create 'active' argument is not a bool. "
+                    f"Converting {active} {type(active)} to bool.")
+                active = bool(active)
+            instance_data["active"] = active
 
         return creator.create(
             product_name,
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -2044,7 +2044,7 @@ class CreateContext:
         folder_entity=None,
         task_entity=None,
         pre_create_data=None,
-        active=True
+        active=None
     ):
         """Trigger create of plugins with standartized arguments.
 
@@ -2063,7 +2063,7 @@ class CreateContext:
             task_entity (Dict[str, Any]): Task entity.
             pre_create_data (Dict[str, Any]): Pre-create attribute values.
             active (Optional[bool]): Whether the created instance defaults
-                to be active or not. Defaults to True.
+                to be active or not.
 
         Returns:
             Any: Output of triggered creator's 'create' method.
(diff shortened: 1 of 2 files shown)


client/ayon_core/pipeline/create/context.py
```diff
@@ -2043,7 +2043,8 @@ def create(
         variant,
         folder_entity=None,
         task_entity=None,
-        pre_create_data=None
+        pre_create_data=None,
+        active=True
```
iLLiCiTiT, line 2047: I don't like that `create` method changed signature. Shoulnd't that be part of `pre_create_data`? Or deactivate the instance after creation?

BigRoy: Done with https://github.com/ynput/ayon-core/pull/774/commits/849ab8b624586db321b3615e5ee6e71aa94b54df

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1826,6 +1836,7 @@ def populate_create_placeholder(self, placeholder, pre_create_data=None):
         legacy_create = self.builder.use_legacy_creators
         creator_name = placeholder.data["creator"]
         create_variant = placeholder.data["create_variant"]
+        active = placeholder.data.get("active", True)
```
iLLiCiTiT, line 1839:
```suggestion
        active = placeholder.data.get("active")
```

BigRoy: Thanks, missed that one.

iLLiCiTiT (approved):
Code looks good, would be nice to test in Nuke and Maya if it does not break anything.

client/ayon_core/pipeline/create/context.py
```diff
@@ -2126,6 +2129,9 @@ def create(
             "productType": creator.product_type,
             "variant": variant
         }
+        if active is not None:
+            instance_data["active"] = bool(active)
```
iLLiCiTiT, line 2133: Why is needed specific cast to `bool`?

BigRoy: Good question. Without it, it gave an issue in Houdini due to the fact that the HoudinI Placeholder node for the workfile template builder would return 1 or 0 instead of a boolean value - due to that being Houdini's return type for whatever reason for a Toggle parameter.

This meant, when creating the instance using the workfile template builder that the active `parm` would be "imprinted" on creation as an integer field instead of a boolean field.

I'd say this is also somewhat a separate bug or something we look into on how we're reading these toggle parms from Houdini to maybe ensure we're getting them as booleans instead. But since this was the safest way, and at least ensures the value is a boolean across all hosts (which it should be) I felt it was the easiest and best solution moving forward.

So basically:
- The Houdini placeholder node would have an `active` toggle
- It returns the value as `0` or `1`
- Which means `placeholder.data["active"]` is `0` or `1`
- Which means suddenly we're passing integer here instead of bool.
- Which influences how the value is imprinted on the created instance due to it being based on the type of the value in the instance data.

Hope that clarifies it.

@MustafaJafar - we might want to track this as a separate (non-urgent) issue for ayon-houdini.

BigRoy:
> Code looks good, would be nice to test in Nuke and Maya if it does not break anything.
> 
>     * Use old template where are placeholders without `active`.
> 
>     * Use new template and try if it creates activated/deactivated instances based on the placeholder data.

Note: Maya does not have support for "Create" placeholders. Nonetheless, I added these testing notes to the PR description. 👍

MustafaJafar (approved):
It works in Houdini.

client/ayon_core/pipeline/create/context.py
iLLiCiTiT: Well but that is related to houdini, I would say that in core it should expect `bool`, and it's job of houdini to convert it.

BigRoy: I mean - it's similarly hard to just pass on `"0"` then and debug that. We could also have it explicitly error if it's not a boolean so we at least spot the mistakes directly and solve them. It does mean we'd need additional fixes on the Houdini side of things, which we should do anyway.

Shall I just to a type check if it is not None? Something like:
```python
if active is not None and not isinstance(active, bool):
    raise TypeError(f"Active value must be boolean, got: {active} ({type(active)})")
```

BigRoy: After discussion in DM on discord we came to the conclusion to do the automatic conversion, but log a warning if the type is not a bool yet.

BigRoy: Added warning: https://github.com/ynput/ayon-core/pull/774/commits/05f6b0f2ce000910004c3ce9913e17ec528bc04b

PR #775, opened 2024-07-15
Enable Validate Outdated Containers by default for Fusion

BigRoy:
## Changelog Description

Also enabled the Validate Outdated Containers validator in Fusion by default.

## Additional info

It's often missed, but this being a **ContextPlugin** means that the "optional" toggle appears on the **Context** entry in the Publisher UI. See:

ayon_contextplugin_optional_plugin_state_in_publisher_ui

## Testing notes:

1. Ensure the default settings are updated
2. The plug-in should be listed in Fusion and run accordinly
3. The optional state should work.


BigRoy:
@EmberLightVFX any chance you want to test-run this or provide comments on whether this may or may not be problematic to enable by default?

jakubjezek001 (approved):
LGTM! Thanks ;)

PR #778, opened 2024-07-15
Chore: Add 'parents' to folder template data

iLLiCiTiT:
## Changelog Description
Add `"parents"` subkey to folder template data.

## Additional info
This does allow to use indexing of parent names in templates. I'm not sure if this is correct approach, but it is the best I could think of. This change does not consider that someone would like to use e.g. labels instead of names in future, hopefully that won't happen.

There is no safeguard for using the list, so if the parent index is not available it will just crash. That might cause issues in specific logic which is filling template paths and does not count on this.

## Testing notes:
1. You can use e.g. `{folder[parents[0]]}` in templates.
2. Test it with an optional template key <{folder[parents[0]}> works from a ROOT folder (does not have any parent folders).

Resolves https://github.com/ynput/ayon-core/issues/867


client/ayon_core/pipeline/template_data.py
```diff
@@ -87,14 +87,13 @@ def get_folder_template_data(folder_entity, project_name):
...
-    hierarchy_parts.pop(0)
+    # Remove empty string from the beginning and split by '/'
+    parents = path.lstrip("/").split("/")
     # Remove last part which is folder name
-    folder_name = hierarchy_parts.pop(-1)
-    hierarchy = "/".join(hierarchy_parts)
-    if hierarchy_parts:
-        parent_name = hierarchy_parts[-1]
+    folder_name = parents.pop(-1)
+    hierarchy = "/".join(parents)
```
BigRoy, line 94: To be backwards compatible I think `hierarchy` should get an extra `/` prefixed to it, no?
Whether that's sensible I'm not sure?

BigRoy:
> This does allow to use indexing of parent names in templates. I'm not sure if this is correct approach, but it is the best I could think of. This change does not consider that someone would like to use e.g. labels instead of names in future, hopefully that won't happen.
> 
> There is no safeguard for using the list, so if the parent index is not available it will just crash. That might cause issues in specific logic which is filling template paths and does not count on this.

Hmm - yeah, hmm.. this is bound to give issues. If anyone would use e.g. `parents[-1]` in a template and they are working in an asset that has no parents the template formatting would already fail. This seems very prone to production issues.

Can we re-iterate why `parents` is needed as a key and what the admin/user is expected to do with it?

jakubjezek001:
> asset that has no parents the template formatting would already fail. This seems very prone to production issues.

I wonder if we could use for that reason optional decoration `...}</{folder[parents[-1]]}>/...`

jakubjezek001:
BTW this seems to be very good case for unittesting.

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
```diff
@@ -407,8 +407,10 @@ def _fill_folder_data(self, instance, project_entity, anatomy_data):
             anatomy_data["hierarchy"] = hierarchy
 
             parent_name = project_entity["name"]
+            parents = []
```
jakubjezek001, line 410: It seems valid to always have parents filled with at least project, right? Since we are converting parent_name from project name in case there is no parent at all it would be fair to also add it. In this case `{folder[parent[-1]]}`n would always work.

iLLiCiTiT: I really don't think project should be there. But the use-case is that they would like to be able to use `{folder[parents][0]}` and `{folder[parents][1]}` not really `-1`.

BigRoy:
> > asset that has no parents the template formatting would already fail. This seems very prone to production issues.
> 
> I wonder if we could use for that reason optional decoration `...}</{folder[parents[-1]]}>/...`

Ah - _if_ that works I must admit that's fine then I suppose.

client/ayon_core/plugins/publish/collect_anatomy_instance_data.py
antirotor: I don't think project name should be there too - the reason why is that you should be explicit enough if you want to have project name there (and you can access it with another key).

So if you have asset folder directly under project, `{folder[parents][0]}` will not work if I am not mistaken.

iLLiCiTiT: > So if you have asset folder directly under project, {folder[parents][0]} will not work if I am not mistaken.

Yes, it won't.

BigRoy:
So we should add to the testing notes:

- Test it with an optional template key `<{folder[parents[0]}>` and make sure it works from a ROOT folder (no parent folders)?

LiborBatek (commented):
I have tried to incorporate it into template for render products as this:

Screenshot 2024-07-31 150417

but constantly getting errors when publish job on DL takes place after succesful render job....

see DL publish job log below:


[Job_2024-07-31_15-02-09_66aa35d12c5b90b0d2104221.txt](https://github.com/user-attachments/files/16442202/Job_2024-07-31_15-02-09_66aa35d12c5b90b0d2104221.txt)

iLLiCiTiT:
Something I didn't think of is output of used values which right now in `StringTemplates` always outputs dictionary, that has to be changed to support list.

Another issue might be negative index `{folder[parents][-1]}` would not work either as python formatting resolves `-1` as string instead of number.

iLLiCiTiT:
Now it does work as expected. It is possible to use lists in values.

But negative indexes on list can cause issues when the tempalte is used out of `StringTemplate` so the missing piece is to change the template in result to use modified template in output `{parents[-1]}` should be in result template as `{parents[0]}` (or different index based on used values).

antirotor:
> Or not? Not sure, at the end if anyone is willing to use the template afterwards he still have to use `StringTemplate` because of optional keys.

I don't actually think we state anywhere that templates are compatible with python formatting. So if we say that `-1` isn't possible, then all is fine by me. 

Auto-converting it to proper index is ofc the way too.

If I would have to choose, I'd say -> If we have interactive validator for templates in the Settings, then not supporting `-1` is perfectly fine (because you'll know right away). But since we don't, conversion is better?

dan-of-dan:
I've just used this change in my anatomy template for the default file name: using {folder[parents][1]}. 

On disk everything is correct but I'm now getting an error when publishing out of resolve at the Integrate Asset stage:

`DEBUG: Anatomy template name: default
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\plugins\publish\integrate.py", line 158, in process
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\plugins\publish\integrate.py", line 246, in register
    prepared = self.prepare_representation(
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\plugins\publish\integrate.py", line 777, in prepare_representation
    template_filled = path_template_obj.format_strict(template_data)
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\lib\path_templates.py", line 147, in format_strict
    result = self.format(data)
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\pipeline\anatomy\templates.py", line 96, in format
    result = StringTemplate.format(self, data)
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\lib\path_templates.py", line 125, in format
    part.format(data, result)
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\lib\path_templates.py", line 709, in format
    part.format(data, new_result)
  File "C:\Users\danwn\AppData\Local\Ynput\AYON\addons\core_1.0.14\ayon_core\lib\path_templates.py", line 601, in format
    value = value[sub_key]
IndexError: list index out of range`

iLLiCiTiT:
It could be because folder used for instance does not have available index you require by template. So if template requires `{folder[parents][1]}` but used folder path is `/assets/Bob` then `parents = ["assets"]` so index `1` is out of bounds. You can use optional keys `<{folder[parents][1]_>` in the template, which means that the key is skipped if the key is not available.

dan-of-dan:
> It could be because folder used for instance does not have available index you require by template. So if template requires `{folder[parents][1]}` but used folder path is `/assets/Bob` then `parents = ["assets"]` so index `1` is out of bounds. You can use optional keys `<{folder[parents][1]_>` in the template, which means that the key is skipped if the key is not available.

I did try those optional keys but I still get the same error. My structure is <project_root>/shots/101/100/010

shots = folder
101 = episode (what I'm fetching with {folder[parents][1]) 
100 - sequence
010 = shot

iLLiCiTiT:
Hmm, discovered a bug. Will be fixed with https://github.com/ynput/ayon-core/pull/1121 .

dan-of-dan:
> Hmm, discovered a bug. Will be fixed with #1121 .

Thanks for confirming, will keep an eye on the ticket

PR #779, opened 2024-07-16
Chore: Python 2 deprecation in lib

iLLiCiTiT:
## Changelog Description
Remove Python 2 compatibility from functions in `ayon_core.lib`.

## Additional info
Classes do no inherit from `object`. Simplified doubled imports for Py2 and Py3 compatibility of `configparser` and `lru_cache`. Use `WeakMethod` directly from `weakref` and warn on import of `python_2_comp.py`.

## Testing notes:
1. Go through code and check possible errors I made.
2. Validate everything is working as expected.


BigRoy (approved):
- [x] Tray works
- [x] Tray loader works
- [x] Launcher works
- [x] Maya 2024 works, incl. workfiles, publishing, loading and managing
- [x] Fusion 18 works, incl. workfiles, publishing, loading and managing

Code seems fine. 


### Remove `python_2_comp.py` file now

I have my doubts about whether it's really worth it to have backwards compatibility for `python_2_comp`. I can't find any usage across the codebases and I wonder if anyone would've ever touched it at all. And even if they would - their logic wouldn't be valid for Py3 anyway. So I'd personally opt for removing `python_2_comp.py` completely.

@iLLiCiTiT

iLLiCiTiT:
> Remove python_2_comp.py file now

We have to be somewhat careful about breaking backwards compatibility. I can't know if there is an addon that is using it, having it there with warning for some time won't hurt.

> their logic wouldn't be valid for Py3 anyway

It would be Py3 valid. The content was Py2 and Py3 compatible, you could import it to support both.

Note, start filename with underscore if should not be imported publicly.

PR #781, opened 2024-07-17
Addons: Preparation steps for child processes

iLLiCiTiT:
## Changelog Description
Give addons option to initialize their environment before a pipeline process starts.

## Additional info
The preparation can be used to make sure user is logged into some service and has set environment variables that are necessary. It is achived using simple mechanism by implementing `ensure_is_process_ready` method which expects `ProcessContext` as an argument.

Right now process context contains information if is headless, which addon name and version triggered the request and which project name is used. All of the mentioned information are optional (maybe addon name and version should be required?). Using object gives option to improve what can be specified (we might also add version to `ProcessContext` for forwards/backwards compatibility?).

All addon has to do is to call `ensure_addons_are_process_ready` with object of `ProcessContext` to trigger the preparation. Addons can raise `ProcessPreparationError` with human readable message, in that case user does not have option to see/copy detail (detail contains stdout/stderr with traceback).

## Testing notes:
At this moment the functionality is implemented only in ftrack addon and used by applications addon. For that is also required to use different AYON launcher.
- AYON launcher with https://github.com/ynput/ayon-launcher/pull/112
- AYON launcher with https://github.com/ynput/ayon-launcher/pull/135 (PYTHONPATH issues)
- AYON ftrack https://github.com/ynput/ayon-ftrack/pull/109
- AYON applications https://github.com/ynput/ayon-applications/pull/1
- AYON server at least 1.3.0
1. Make sure to create packages of all of the addons.
2. It is necessary to build and install AYON launcher with the PR to register scheme uri handler.
3. Launch tray and logout from ftrack. You can close tray afterwards, but keeping it open will also test ftrack PR.
4. Start application from AYON web UI.
5. It should ask you to login to ftrack, if that fails it stops launching, otherwise it should launch the application. If you kept tray running, ftrack should update credentials.


client/ayon_core/addon/base.py
```diff
@@ -67,6 +68,56 @@
 }
 
 
+class ProcessPreparationError(Exception):
+    """Exception that can be used when process preparation failed.
+
+    The message is showed to user (either as UI dialog or printed). If
+        different error is raised a "generic" error message is showed to user
+        with option to copy error message to clipboard.
```
BigRoy, line 76:
```suggestion
    The message is shown to user (either as UI dialog or printed). If
        different error is raised a "generic" error message is shown to user
        with option to copy error message to clipboard.
```

client/ayon_core/addon/base.py
```diff
@@ -588,7 +639,29 @@ def connect_with_addons(self, enabled_addons):
...
+        This method is called when some action makes sure that addon has set
+        necessary data. For example if user should be logged in
+        and filled credentials in environment variables this method should
+        ask user for credentials.
+
+        Implementation of this method is optional.
+
+        Note:
+            The logic can be similar to logic in tray, but in tray not require
+                to be logged in.
```
BigRoy, line 658:
```suggestion
            The logic can be similar to logic in tray, but tray does not require
                to be logged in.
```

client/ayon_core/addon/base.py
iLLiCiTiT: I'll wait until you go through it so I can do one big batch commit 🙂 .

client/ayon_core/lib/execute.py
```diff
@@ -209,6 +209,14 @@ def run_ayon_launcher_process(*args, **kwargs):
         # - fill more if you find more
         env = clean_envs_for_ayon_process(os.environ)
 
+    if add_sys_paths:
+        new_pythonpath = list(sys.path)
+        for path in env.get("PYTHONPATH", "").split(os.pathsep):
+            if not path or path in new_pythonpath:
+                continue
```
BigRoy, line 216: Both sys.path and `PYTHONPATH` may be large lists, so we're now doing a `O1` lookup in `new_pythonpath` which may take some time due to that.

It may be better if this could be an ordered set or use a separate mechanism on the side to track whether a path was already included. That may yield performance benefits.

client/ayon_core/addon/base.py
BigRoy: Done.

client/ayon_core/lib/execute.py
iLLiCiTiT: It's at about 0.002 seconds. Do you really think it should be optimized more?

dee-ynput: Let's not optimize before it's used :)

BigRoy: > It's at about 0.002 seconds. Do you really think it should be optimized more?

Depends on the amount of entries on your PYTHONPATH I'd say? It'd be exponential and we must admit have MANY on our PYTHONPATH. And I dare say that anyone using `rez` will cry because there ever single dependency is its own entry in pythonpath.

At 1000 entries it's at 0.006 seconds.
At 10.000 entries it's at 0.6 seconds.
At 100.000 entries it's at 64 seconds.

Whereas just by making it a `set` as a lookup takes the 100.000 entries down to 0.01 seconds.

```python
num = 10000
values = list(range(num))
import time

s = time.time()
for x in values:
    if x in values:
        pass
e = time.time()
print(e-s)

values_lookup = set(values)
import time

s = time.time()
for x in values:
    if x in values_lookup:
        pass
e = time.time()
print(e-s)
```

BigRoy: Haha - man, you should check out some of the `rez` issues. There were literally cases where they were going beyond the limits of what OS'es allowed for these paths and hence stuff getting dismissed accidentally. So they swapped over the technique to specialized Python files you can drop on Pythonpath and alike pointing to many paths. Which interestingly enough also optimized startup times by insane amounts (for whatever reason).

Anyway, will mark resolved.

BigRoy:
> Cannot be tested without an addon that triggers the functionality. When the PR is ready for review there will be added example addon that can be used for the functionality test.

It's already marked "ready for review" no? So how to test? :)

iLLiCiTiT:
> It's already marked "ready for review" no? So how to test? :)

Ou, I did? Converting back.

64qam (approved):

client/ayon_core/lib/execute.py
```diff
@@ -209,6 +209,15 @@ def run_ayon_launcher_process(*args, **kwargs):
         # - fill more if you find more
         env = clean_envs_for_ayon_process(os.environ)
 
+    if add_sys_paths:
+        pp_set = frozenset(sys.path)
+        new_pythonpath = list(pp_set)
+        pythonpath = env.get("PYTHONPATH") or ""
+        for path in frozenset(pythonpath.split(os.pathsep)):
```
BigRoy, line 216: This is wrong... the `frozenset` will lose ordering hence the paths you are appending will be in wrong order.

client/ayon_core/lib/execute.py
```diff
@@ -209,6 +209,15 @@ def run_ayon_launcher_process(*args, **kwargs):
...
+            if path and path not in pp_set:
+                new_pythonpath.append(path)
```
BigRoy, line 218:
```suggestion
        paths_lookup = set(sys.path)
        paths = list(sys.path)
        pythonpath = env.get("PYTHONPATH") or ""
        for path in pythonpath.split(os.pathsep):
            if path and path not in paths_lookup:
                paths_lookup.add(path)
                paths.append(path)
```

`paths` in this example is the `new_pythonpath`.

Issue #782, opened 2024-07-17
Loader: "Enable grouping" toggle fails due to invalid access to statuses


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When clicking the "Enable Grouping" checkbox to disable product groups, nothing happens. An error occurs:

```python
Traceback (most recent call last):
  File "E:\dev\ayon-core\client\ayon_core\tools\loader\ui\window.py", line 417, in _on_product_group_change
    self._products_widget.set_enable_grouping(
  File "E:\dev\ayon-core\client\ayon_core\tools\loader\ui\products_widget.py", line 271, in set_enable_grouping
    self._products_model.set_enable_grouping(enable_grouping)
  File "E:\dev\ayon-core\client\ayon_core\tools\loader\ui\products_model.py", line 168, in set_enable_grouping
    self.refresh(self._last_project_name, self._last_folder_ids)
TypeError: refresh() missing 1 required positional argument: 'status_names'
```

![image](https://github.com/user-attachments/assets/dc7739db-adaf-4324-b014-8aa5c915d8ea)

### Expected Behavior:

The toggle should work.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Open Loader UI
2. Go to folder with products in product groups
3. Click Enable Grouping toggle

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
Traceback (most recent call last):
  File "E:\dev\ayon-core\client\ayon_core\tools\loader\ui\window.py", line 417, in _on_product_group_change
    self._products_widget.set_enable_grouping(
  File "E:\dev\ayon-core\client\ayon_core\tools\loader\ui\products_widget.py", line 271, in set_enable_grouping
    self._products_model.set_enable_grouping(enable_grouping)
  File "E:\dev\ayon-core\client\ayon_core\tools\loader\ui\products_model.py", line 168, in set_enable_grouping
    self.refresh(self._last_project_name, self._last_folder_ids)
TypeError: refresh() missing 1 required positional argument: 'status_names'
```

PR #783, opened 2024-07-17
Loader tool: Fix filtering

iLLiCiTiT:
## Changelog Description
Loader tool can enable/disable grouping of products and filtering by product types and status names should work.

## Additional info
~~This was broken with new statuses filter which required to pass status names on model refresh, which was mid-solution that is not necessary at the end. So changed the logic of Products model refresh back to previous integration to not require status names on refresh.~~
PR changed during it's lifetime as was discovered that the whole filtering was not working. Product type filtering was completelly broken and statuses filtering did break UI for multiselection of folders. Also products under group item did not propagate version change from combobox in specific cases.

## Testing notes:
- First of all grouping can be enabled/disabled.
- To validate we didn't break the source of the previous implementation:

1. Open loader tool.
2. Add some status filters (that will change version to lower version than latest).
3. Refresh or enable/disable grouping.
4. The version should stay at same version number and should show correct status and rest of version related information.

Resolves https://github.com/ynput/ayon-core/issues/782

### DIFF
--- a/client/ayon_core/tools/loader/ui/products_delegates.py
+++ b/client/ayon_core/tools/loader/ui/products_delegates.py
@@ -140,12 +140,6 @@ class VersionComboBox(QtWidgets.QComboBox):
         self.value_changed.emit(self._product_id, value)
 
 
-class EditorInfo:
-    def __init__(self, widget):
-        self.widget = widget
-        self.added = False
-
-
 class VersionDelegate(QtWidgets.QStyledItemDelegate):
     """A delegate that display version integer formatted as version string."""
 
@@ -154,7 +148,7 @@ class VersionDelegate(QtWidgets.QStyledItemDelegate):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-        self._editor_by_id: Dict[str, EditorInfo] = {}
+        self._editor_by_id: Dict[str, VersionComboBox] = {}
         self._statuses_filter = None
 
     def displayText(self, value, locale):
@@ -164,8 +158,8 @@ class VersionDelegate(QtWidgets.QStyledItemDelegate):
 
     def set_statuses_filter(self, status_names):
         self._statuses_filter = set(status_names)
-        for info in self._editor_by_id.values():
-            info.widget.set_statuses_filter(status_names)
+        for widget in self._editor_by_id.values():
+            widget.set_statuses_filter(status_names)
 
     def paint(self, painter, option, index):
         fg_color = index.data(QtCore.Qt.ForegroundRole)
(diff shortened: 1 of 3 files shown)


BigRoy:
It's only filtering to the "First status" inside a Group... not the individual statuses of each entry in the group.

image
image

Note how these are displaying statuses that are not selected. I suspect this is not a **new** issue due to this PR though?

---

Similarly the filtering gets confused it seems with Multiselections:

image

I think in this case the filter is correct, but...
image

It's just not displaying the status correctly maybe?

---

### Status filter finds 'last matching version' instead of filtering current list

Here I have no filters visible, sort by status... there are three approved entries:

image

I set the filter to "Approved" statuses only and suddenly get:

image

Is the behavior _correct_ that when filtering to a particular status that it shows the _last matching version_  with that status. So if e.g. an version was approved it'd then display that instead?

The same occurs when picking e.g. Omitted:

image

But if I don't set that filter it lists much newer versions that are not omitted at all.

Is this correct behavior or confusing? :)
I was confused at first.

---

This isn't a critical issue but it does look confusing. Say I have a filter:


When disabling the filter without going out of the dropdown menu the "version" of the product changes (likely due to the filtering described above). However, the status and the rest of the row does not update until I go out of the dropdown menu.

image
image
image

iLLiCiTiT:
Didn't read fully (looong text), but can confirm it is not intentional and should be fixed, could you try if the issue with groups (multiselection of folders) also occurs in develop?

BigRoy:
> Didn't read fully (looong text), but can confirm it is not intentional and should be fixed, could you try if the issue with groups (multiselection of folders) also occurs in develop?

Yes. Using develop.

Single folder select, status filtering fails in groups:
image

**BUT** I think it's actually correctly filtering to the last "Approved" version of that product however it fails to display its status correctly. Because if I manually go to v029 for workfileModeling I get:

image

Hence, the filtering may just be correct (it is actually approved).

---

Here with multiselect folder - similar issue:

image

It's the same product even! :)

BigRoy (approved):
I can't seem to break the filtering now. 👍 

---

I do still think that the filter by status finding the "latest version" that matches of a product instead of actually filtering the list is a bit confusing behavior. I wouldn't have expected it and I wonder whether we need a tooltip explaining that AND/OR even have a toggle to disable that. It disallows me to find the "approved" versions that are _currently listed_ (those that are the latest version).

I'll create another issue for that.

Also, it only dawned on me now that it ALSO filters the version selection boxes of the products 🤯 - it's actually quite nice, but still unexpected for me.

---

And a minor cosmetic one is that if it does filter out e.g. two products when multiselecting three folders that have a product by that name is that the "header" group of the multiselection will still list the number `(3)` at the end even though only one is listed.

image

This is hardly an issue - but just wanted to point that out.

PR #786, opened 2024-07-18
Better validation report when publishing from unsaved workfile

BigRoy:
## Changelog Description

The user now gets presented with a better validation report if a workfile is not saved.

## Additional info

Previously the user would get to a "This is not your fault" error on Collect, but often newcomers failed to identify why it failed.

After this PR that's more informative:
**Maya**
image

**Fusion**
image

## Testing notes:

1. Publish from host without saved workfile/filepath
2. Validator should catch it.


client/ayon_core/plugins/publish/validate_file_saved.py
```diff
@@ -1,17 +1,55 @@
...
+    """Save Workfile."""
+    label = "Show Workfiles Tool..."
+    on = "failed"
+    icon = "files-o"
+
+    def process(self, context, plugin):
+        show_workfiles()
 
 
 class ValidateCurrentSaveFile(pyblish.api.ContextPlugin):
```
moonyuet, line 30: Should we add this into some other hosts which they need to have saved workfile for publish? (e.g. Substance Painter, 3dsMax etc.)

BigRoy: This is in `ayon-core`. If it's valid for those hosts then we can add the relevant host to the `hosts = [x, y, z]` entries on this plug-in. Correct.

MustafaJafar (commented):
I gave it a shot in Houdini, where I opened a scene and did some changes and triggered validation. and nothing happened. 
I added some log in `validate_file_saved` and it was cool 😅 

image

I gave it another shot with an untitled scene. but the publisher errored telling me it's not my fault. 
image
image

So, how does it work ?

BigRoy:
> So, how does it work ?

VERY GOOD! :) I hadn't tested specifically Houdini, but Houdini is an interesting outlier here. When a scene is unsaved, the scene is _actually_ `untitled.hip` instead of `None` or `""`. So in a way Houdini DOES report as if it has a saved filename, and that filename indeed fails to collect scene version from.

So in a way - it's a different bug. Probably one we can best tackle on `ayon-houdini` to report a better file if 'unsaved' like that.

client/ayon_core/plugins/publish/validate_file_saved.py
BigRoy: Should I resolve for now - or do I need to add more hosts? Any chance you can do a test run with Substance Painter and Max @moonyuet ?

moonyuet:
Max
image
image

SP(Guess you need to set some condition for `workfile representation`)
image

BigRoy:
Seems Max has its own "validator" which we should indeed bring to ayon-core instead so it automatically has the nice description and easy access to the fix actions, etc.

1. Add max to families here
2. Remove validator from max

---

Substance Painter would need some additional fixes indeed - but then again. The error itself isn't WORSE than what it was since it'd crash publisher UI in a similar manner. We can set up a PR to fix Substance Painter specifically and improve that later!

1. Add substancepainter to families here
2. Fix other Collectors in substancepainter integration

client/ayon_core/plugins/publish/validate_file_saved.py
BigRoy: 1. Added substancepainter and max. 
2. Made a matching substancepainter PR to improve: https://github.com/ynput/ayon-substance-painter/pull/7
3. For Max we may want to remove Max's own "saved" validation in some time to allow this core one to be the nice and shiny one. For now, with this PR you might get two invalidations for unsaved files - one nice one, one legacy one from max itself.

BigRoy:
@MustafaJafar I've made a Houdini PR to improve Houdini's validation: https://github.com/ynput/ayon-houdini/pull/81

@moonyuet I've made a Substance Painter PR https://github.com/ynput/ayon-substance-painter/pull/7 to improve there. For Max we may want to follow up with removing Max's own one, but for now it may display two invalidations with this PR (the nice one from this PR, and the legacy one existing in Max).

Please double check the PRs in combination and approve if this works! 👍

BigRoy:
> Question? Are we moving the error to validation to show the message to artist "nicer"?

Yes.

iLLiCiTiT:
Then I might have a better solution, new PR in ayon-core (comming soon).

iLLiCiTiT:
> As we discussed in DM your 'coming up' solution may not allow to run actions to allow the user with easy to access actions if an invalidation occurs during Collect. Hence, this PR is still the recommended approach for the time being.

Just confirming that PR https://github.com/ynput/ayon-core/pull/861 would only show the message, but would not allow to trigger actions.

MustafaJafar (approved):
Maya
image
Houdini _using https://github.com/ynput/ayon-houdini/pull/81_
image

PR #790, opened 2024-07-19
Review: Integrate reviewables to AYON

iLLiCiTiT:
## Changelog Description
Integrate plugin does integrate reviewables to AYON server.

## Additional info
Upload reviewables to AYON server during integration. The method is not yet available in ayon-python-api so it had to be implemented there. Also modified default ExtractReview arguments to more suite needs of server, but previous defaults should work too.

I did have issues with ExtractBurnins, that could be because of my dev environment, please validate if is working for you.

There is one possible issue that should be resolved in future, which is that if there is only review representation without representation to integrate, it is actually not uploaded to server because the version is not integrated (as would not have any representations). This topic needs more thinking and usually happens only in edge cases...

## Testing notes:
1. Make sure you have AYON server 1.3.0 or higher.
2. Create package and upload it to server.
3. Make sure you have added tag `"webreview"` in ExtractReview settings.
4. Start host of your choice.
5. Run publishing with some review.
6. The review should be uploaded to AYON server.

### DIFF
--- a/client/ayon_core/plugins/publish/integrate.py
+++ b/client/ayon_core/plugins/publish/integrate.py
@@ -11,6 +11,7 @@ from ayon_api import (
     get_product_by_name,
     get_version_by_name,
     get_representations,
+    RequestTypes,
 )
 from ayon_api.operations import (
     OperationsSession,
@@ -1028,7 +1029,7 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
                 query = f"?label={label}"
 
             endpoint = (
-                f"/api/projects/{project_name}"
+                f"/projects/{project_name}"
                 f"/versions/{version_id}/reviewables{query}"
             )
             # Make sure label is unique
@@ -1049,7 +1050,8 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
             ayon_con.upload_file(
                 endpoint,
                 repre_path,
-                headers=headers
+                headers=headers,
+                request_type=RequestTypes.post,
             )
 
     def _validate_path_in_project_roots(self, anatomy, file_path):
--- a/client/ayon_core/plugins/publish/integrate.py
+++ b/client/ayon_core/plugins/publish/integrate.py
@@ -20,7 +20,7 @@ from ayon_api.operations import (
 )
 from ayon_api.utils import create_entity_id
 
-from ayon_core.lib import source_hash
+from ayon_core.lib import source_hash, get_media_mime_type
 from ayon_core.lib.file_transaction import (
     FileTransaction,
     DuplicateDestinationError
@@ -990,12 +990,10 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
         }
 
     def _upload_reviewable(self, project_name, version_id, instance):
-        uploaded_labels = set()
         ayon_con = get_server_api_connection()
         base_headers = ayon_con.get_headers()
-        endpoint = (
-            f"/api/projects/{project_name}/versions/{version_id}/reviewables"
-        )
+
+        uploaded_labels = set()
         for repre in instance.data["representations"]:
             repre_tags = repre.get("tags") or []
             # Ignore representations that are not reviewable
@@ -1014,13 +1012,25 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
             repre_path = get_publish_repre_path(
                 instance, repre, False
             )
-            repre_ext = os.path.splitext(repre_path)[-1].lower()
+            if not repre_path or not os.path.exists(repre_path):
+                # TODO log skipper path
+                continue
 
+            content_type = get_media_mime_type(repre_path)
+            if not content_type:
+                self.log.warning("Could not determine Content-Type")
+                continue
+
+            # Use output name as label if available
             label = repre.get("outputName")
-            if not label:
-                # TODO how to label the reviewable if there is no output name?
-                label = "Main"
+            query = ""
+            if label:
+                query = f"?label={label}"
 
+            endpoint = (
+                f"/api/projects/{project_name}"
+                f"/versions/{version_id}/reviewables{query}"
+            )
             # Make sure label is unique
             orig_label = label
             idx = 0
@@ -1033,12 +1043,14 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
             # Upload the reviewable
             self.log.info(f"Uploading reviewable '{label}' ...")
             headers = copy.deepcopy(base_headers)
+            headers["Content-Type"] = content_type
+            headers["x-file-name"] = os.path.basename(repre_path)
+            self.log.info(f"Uploading reviewable {repre_path}")
             ayon_con.upload_file(
-                f"/api/projects/{project_name}/versions/{version_id}/reviewables",
+                endpoint,
                 repre_path,
                 headers=headers
             )
-
 
     def _validate_path_in_project_roots(self, anatomy, file_path):
         """Checks if 'file_path' starts with any of the roots.
--- a/client/ayon_core/plugins/publish/integrate.py
+++ b/client/ayon_core/plugins/publish/integrate.py
@@ -992,6 +992,14 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
 
     def _upload_reviewable(self, project_name, version_id, instance):
         ayon_con = get_server_api_connection()
+        major, minor, _, _, _ = ayon_con.get_server_version_tuple()
+        if (major, minor) < (1, 3):
+            self.log.info(
+                "Skipping reviewable upload, supported from server 1.3.x."
+                f" User server version {ayon_con.get_server_version()}"
+            )
+            return
+
         base_headers = ayon_con.get_headers()
 
         uploaded_labels = set()
--- a/client/ayon_core/plugins/publish/integrate.py
+++ b/client/ayon_core/plugins/publish/integrate.py
@@ -6,6 +6,7 @@ import copy
 import clique
 import pyblish.api
 from ayon_api import (
+    get_server_api_connection,
     get_attributes_for_type,
     get_product_by_name,
     get_version_by_name,
@@ -25,6 +26,7 @@ from ayon_core.lib.file_transaction import (
     DuplicateDestinationError
 )
 from ayon_core.pipeline.publish import (
+    get_publish_repre_path,
     KnownPublishError,
     get_publish_template_name,
 )
@@ -347,6 +349,8 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
 
         self.log.debug("{}".format(op_session.to_data()))
         op_session.commit()
+
+        self._upload_reviewable(project_name, version_entity["id"], instance)
 
         # Backwards compatibility used in hero integration.
         # todo: can we avoid the need to store this?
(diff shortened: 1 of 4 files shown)


BigRoy (commented):
### Requires updated AYON API!

The integrator plug-in didn't work for me initially:
```python
Filepath:
E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py

Traceback:
File "E:\dev\ayon-core\client\ayon_core\pipeline\publish\lib.py", line 257, in publish_plugins_discover
module = import_filepath(abspath, mod_name)
File "E:\dev\ayon-core\client\ayon_core\lib\python_module_tools.py", line 38, in import_filepath
module_loader.exec_module(module)
File "", line 883, in exec_module
File "", line 241, in _call_with_frames_removed
File "E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 8, in 
from ayon_api import (
ImportError: cannot import name 'RequestTypes' from 'ayon_api' (C:\Program Files\Ynput\AYON 1.0.2\dependencies\ayon_api\__init__.py)
```

Which was fixed by picking a new launcher version 1.0.3 and the dependency package - but I didn't get any compatibility issues with 1.0.2 and the older dependency package when setting the bundle?

Might be nice if in frontend we can streamline this so the error is caught earlier?

Especially because with this plug-in failing to load the publish APPEARS to succeed (no errors) but there's just no integration... so no actual publish into AYON database at all.

**Side note:** If you set up the dev mode and are running against the 1.0.2 console using e.g. `"C:\Program Files\Ynput\AYON 1.0.2\ayon_console.exe"` but with `set AYON_USE_DEV=1` I saw NO error or warning whatsoever that it was actually still running 1.0.2 even though my full bundle was set to 1.0.3. It was still using 1.0.2 dependencies package, etc.

---

Once that was fixed:

# Works ✅ 

I was able to publish a reviewable from DCCs:

- [x] review from Maya, 
- [ ] review from Blender **didn't work but it was because for whatever reason extract playblast crashes my blender**
- [x] review from Houdini 
- [x] local render from Fusion.
- [x] plate publish from Tray Publisher

image

I didn't test any Deadline render setups though because I don't have a good test setup for that currently.

## Should the logic move into a Publish plug-in of its own?

Due to the reviewable upload the integration process does become a bit slower - which makes sense.
I just wonder whether it makes sense to move the logic into a plug-in of its own.

- It can be easily taken out/swapped out for other logic.
- It's clear it runs in isolation then.
- It could be easily offloaded to e.g. a farm job?
- The publisher UI would show the plug-in name for what it's doing, e.g. show "Uploading Reviewables" bottom left.
- If the plug-in were then made optional, the user could easily enable/disable the push to AYON for certain reviewables?

---

Reviewables being uploaded 🥳 
image

BigRoy:
**Question with regards to Hero versions**

Currently hero versions can also be played as reviewable on the web frontend.
But publishing currently, does not add a reviewable to a hero version. Should it?
Also, if it were to add it - what happens if we publish again a new version. Should it replace the old reviewable for it?
Does it even make technically sense to have HERO version be reviewable? Or?

Tagging @mkolar @Innders

m-u-r-p-h-y (commented):
I can confirm sucessful Blender 4.2.0 LTS review upload to Ayon. 
image

I have ftrack enabled at the same time though and the review was not uploaded to ftrack, nor version created. Only status has been changed
image

krishna8008:
Just testing this code with Nuke, It's working! Can see the video review in ayon
ClipWindowsGIF

iLLiCiTiT:
> Requires updated AYON API!

Not anymore.

> I saw NO error or warning whatsoever that it was actually still running 1.0.2 even though my full bundle was set to 1.0.3. It was still using 1.0.2 dependencies package, etc.

AYON python api is in AYON launcher, not in dependency package.

> Should the logic move into a Publish plug-in of its own?
> Due to the reviewable upload the integration process does become a bit slower - which makes sense.
> I just wonder whether it makes sense to move the logic into a plug-in of its own.
>
> It can be easily taken out/swapped out for other logic.
> It's clear it runs in isolation then.
> It could be easily offloaded to e.g. a farm job?
> The publisher UI would show the plug-in name for what it's doing, e.g. show "Uploading Reviewables" bottom left.
> If the plug-in were then made optional, the user could easily enable/disable the push to AYON for certain reviewables?

Not saying we can't move it to separate addon, but not because of mentioned reasons.
- Reviewables are usually marked for delete, so they are not integrated as representations and are deleted at the end of publish process. So you can't really off-load it for certain workflows. You would still need some overall logic how to depart certain plugins to different process, which we don't have now. (Bigger topic).
- You can turn off the logic by not adding `webreview` tag.
- Can't tell about having this feature as optional? Should artist tell if review should be uploaded?

BigRoy:
> AYON python api is in AYON launcher, not in dependency package.

well, the error was still there - so even though my AYON dev was set to "launcher 1.0.3" I got no error anywhere that it was actually running 1.0.2 in dev mode because I happened to launch through 1.0.2 console... Might need an obvious error log there?

> Reviewables are usually marked for delete, so they are not integrated as representations and are deleted at the end of publish process. So you can't really off-load it for certain workflows. You would still need some overall logic how to depart certain plugins to different process, which we don't have now. (Bigger topic. In brainstorm stage.)

👍 

In a way I suppose the same goes for uploads to Shotgrid/Ftrack (no idea whether those run after integration or before?) bit I guess that means each of those should also be able to 'control' whether it has processed the reviewable and should still need integrating or not?

Anyway - all are good topics and questions - maybe not critical for this 0.1 version of this feature, but definitely worth spawning off separate issues if we're merging this.

iLLiCiTiT:
> Could this be because the reviewables are indeed "deleted" @iLLiCiTiT and hence @krishna8008 was also expecting them integrated but they were now considered reviewables and thus they are not integrated anymore.

Could be, but not related to this PR. It just added option to upload reviewable to AYON.

iLLiCiTiT:
> Why are "review" representations not actual representations? Does this mean that reviewables won't be available to be loaded in the different DCCs as a representation?

They're not all the time, but you can add (I think it is) `"delete"` tag to it to prevend integration. It is handy if you want to upload reviewables to AYON/ftrack/shotgrid and not really integrating them as representation.

fabiaserra:
> They're not all the time, but you can add (I think it is) `"delete"` tag to it to prevend integration. It is handy if you want to upload reviewables to AYON/ftrack/shotgrid and not really integrating them as representation.

In our case what we upload to Flow (and what we would similarly upload to AYON potentially) we always want them as representations as users validate the review from Nuke constantly as well, not just from the production tracking tool

iLLiCiTiT:
> In our case what we upload to Flow (and what we would similarly upload to AYON potentially) we always want them as representations as users validate the review from Nuke constantly as well, not just from the production tracking tool

Ok, that is valid, but we also do allow to not integrate them.

dee-ynput:
> I have ftrack enabled at the same time though and the review was not uploaded to ftrack, nor version created. Only status has been changed

Please be sure to test with active production trackers (at least ftrack).
Unless this comment from @m-u-r-p-h-y is fixed, we cannot release.

m-u-r-p-h-y:
> > I have ftrack enabled at the same time though and the review was not uploaded to ftrack, nor version created. Only status has been changed
> 
> Please be sure to test with active production trackers (at least ftrack). Unless this comment from @m-u-r-p-h-y is fixed, we cannot release.


it seems the issue is missing defaults for Blender collect family in Ftrack. Will re-test again

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,98 @@
...
+        uploaded_labels = set()
+        for repre in instance.data["representations"]:
+            repre_tags = repre.get("tags") or []
+            # Ignore representations that are not reviewable
+            if "webreview" not in repre_tags:
+                continue
+
+            # exclude representations with are going to be published on farm
+            if "publish_on_farm" in repre_tags:
+                continue
```
BigRoy, line 49: In some cases the targeting to the farm is stored on `instance.data["farm"] = True` (no idea whether those have any representation data to begin with, but I've never seen the "publish_on_farm" tag on representations so I'm unsure whether there may be cases where `instance.data["farm"] = True` but the `representations` data may still be defined?

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,98 @@
...
+
+            # Skip thumbnails
+            if repre.get("thumbnail") or "thumbnail" in repre_tags:
+                continue
+
+            # include only thumbnail representations
```
BigRoy, line 55: Is this comment unrelated to the next lines of code?

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,98 @@
...
+            repre_path = get_publish_repre_path(
+                instance, repre, False
+            )
+            if not repre_path or not os.path.exists(repre_path):
+                # TODO log skipper path
+                continue
+
+            content_type = get_media_mime_type(repre_path)
+            if not content_type:
+                self.log.warning("Could not determine Content-Type")
```
BigRoy, line 65: Would it be worth logging the `repre_path` here as well so the user is aware of what was checked?
```suggestion
                self.log.warning(
                    "Could not determine Content-Type for: %s", repre_path)
```

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,98 @@
...
+            endpoint = (
+                f"/projects/{project_name}"
+                f"/versions/{version_id}/reviewables{query}"
+            )
+            # Make sure label is unique
+            orig_label = label
+            idx = 0
+            while label in uploaded_labels:
+                idx += 1
+                label = f"{orig_label}_{idx}"
```
BigRoy, line 83: Would this fail if `repre["outputName"]` was not defined? Seems we're lacking a fallback label?
Something tells me we're now defaulting to info logging "Uploading reviewable 'None' ..." which may be quite confusing?

client/ayon_core/plugins/publish/integrate_review.py
iLLiCiTiT: I don't know. I've copied ftrack review logic and changed `ftrackreview` to `webreview`.

client/ayon_core/plugins/publish/integrate_review.py
iLLiCiTiT:
```suggestion
```

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,102 @@
...
+            if label:
+                query = f"?label={label}"
+
+            endpoint = (
+                f"/projects/{project_name}"
+                f"/versions/{version_id}/reviewables{query}"
+            )
+
+            # Upload the reviewable
+            self.log.info(f"Uploading reviewable '{label}' ...")
```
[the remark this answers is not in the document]
iLLiCiTiT: This is not intentional.

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,102 @@
...
+import ayon_api
+from ayon_api.server_api import RequestTypes
+
+from ayon_core.lib import get_media_mime_type
+from ayon_core.pipeline.publish import get_publish_repre_path
+
+
+class IntegrateAYONReview(pyblish.api.InstancePlugin):
+    label = "Integrate AYON Review"
+    # Must happen after IntegrateNew
```
BigRoy, line 13: I'm pretty sure IntegrateNew does not exist anymore. :)
```suggestion
    # Must happen after IntegrateAsset
```

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -0,0 +1,102 @@
...
+            version_id = version_entity["id"]
+            self._upload_reviewable(project_name, version_id, instance)
+
+    def _upload_reviewable(self, project_name, version_id, instance):
+        ayon_con = ayon_api.get_server_api_connection()
+        major, minor, _, _, _ = ayon_con.get_server_version_tuple()
+        if (major, minor) < (1, 3):
+            self.log.info(
+                "Skipping reviewable upload, supported from server 1.3.x."
+                f" User server version {ayon_con.get_server_version()}"
```
BigRoy, line 36: It being referred to as the "User" server version seems a bit odd?
```suggestion
                f" Current server version {ayon_con.get_server_version()}"
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -348,6 +349,8 @@ def register(self, instance, file_transactions, filtered_repres):
         self.log.debug("{}".format(op_session.to_data()))
         op_session.commit()
 
+        self._upload_reviewable(project_name, version_entity["id"], instance)
+
```
BigRoy, line 353:
```suggestion
```

Solve:
```python
  File "E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 158, in process
  File "E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 352, in register
    self._upload_reviewable(project_name, version_entity["id"], instance)
AttributeError: 'IntegrateAsset' object has no attribute '_upload_reviewable'

  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 170, in process
    raise exc
  File "E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 158, in process
  File "E:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 352, in register
    self._upload_reviewable(project_name, version_entity["id"], instance)
AttributeError: 'IntegrateAsset' object has no attribute '_upload_reviewable'
```

m-u-r-p-h-y (approved):
uploading the review from blender works as expected.

we may need to investigate the ftrack in another issue/PR

Issue #680, opened 2024-06-20
Loader: Fix support for product_type_filter_profiles


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

The ayon core addon has settings for `ayon+settings://core/tools/loader/product_type_filter_profiles` which allows to define the default "filter" state for product types when showing the Loader.

It seems however since some recent changes in the UIs and refactoring this does nothing anymore. It seems the `product_type_filter_profiles` setting is currently completely unused.

### Expected Behavior:

Fix the feature again so we provide default "filters" set for the loader based on hosts, etc. (So actually use `ayon+settings://core/tools/loader/product_type_filter_profiles`)

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Configure setting `ayon+settings://core/tools/loader/product_type_filter_profiles`, for example:
```json
[
  {
    "hosts": [
      "maya"
    ],
    "task_types": [],
    "is_include": true,
    "filter_product_types": [
      "animation",
      "camera",
      "layout",
      "look",
      "model",
      "plate",
      "pointcache",
      "reference",
      "rig",
      "usd",
      "vdbcache",
      "yetiRig",
      "yeticache"
    ]
  },
  {
    "hosts": [
      "houdini"
    ],
    "task_types": [],
    "is_include": true,
    "filter_product_types": [
      "animation",
      "camera",
      "model",
      "plate",
      "pointcache",
      "usd",
      "vdbcache"
    ]
  },
  {
    "hosts": [
      "fusion"
    ],
    "task_types": [],
    "is_include": true,
    "filter_product_types": [
      "camera",
      "image",
      "matchmove",
      "plate",
      "prerender",
      "render"
    ]
  },
  {
    "hosts": [
      "resolve"
    ],
    "task_types": [],
    "is_include": true,
    "filter_product_types": [
      "editorial",
      "image",
      "matchmove",
      "plate",
      "prerender",
      "render",
      "review"
    ]
  }
]
```
2. When showing the loader in the specific profiles the default "product type" filters on the bottom should be enabled as per the matching profile.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Additional context:

Any pointers on how or where to implement this with all the controller/UI separation logic involved currently @iLLiCiTiT and I can take a look (hopefully understanding what you mean, hehe)

iLLiCiTiT:
[...]

PR #792, opened 2024-07-22
Bugfix: loader support product type filter profiles

BigRoy:
## Changelog Description

Adds back in support for having default Loader product type filters to be specified from settings from `ayon+settings://core/tools/loader/product_type_filter_profiles`

Fixes https://github.com/ynput/ayon-core/issues/680

## Additional info

- This feature existed in OpenPype, but was missing in AYON since some UI refactoring.
- Had to remove the default settings value because the default was actually set to for all profiles include nothing. (not sure how this will behave for studios updating?)
- Updated some available product types in settings that can be set.

### Example

With for example a config like `ayon+settings://core/tools/loader/product_type_filter_profiles`:

image

```json
[
  {
    "hosts": [
      "fusion"
    ],
    "task_types": [],
    "is_include": true,
    "filter_product_types": [
      "render",
      "plate",
      "prerender",
      "image"
    ]
  }
]
```

Then for Fusion host by default the loader will have only those product types enabled in the filter.
This provides the artist a much smaller list to look through where maybe most other products they do not care about anyway:

image

## Testing notes:

Update the core addon (so default settings are ok)

1. Loader tool should work
2. Toggling filters on/off manually should behave as expected
3. The `ayon+settings://core/tools/loader/product_type_filter_profiles` setting should work in the Loader tool.
4. The host and task type filtering should work.
5. By default, without task type filtering set all product types should be shown


client/ayon_core/tools/loader/control.py
```diff
@@ -425,3 +432,28 @@ def _create_event_system(self):
 
     def _emit_event(self, topic, data=None):
         self._event_system.emit(topic, data or {}, "controller")
+
+    def get_current_context_product_types_filter(self):
```
iLLiCiTiT, line 436: This method should expect project name. When you have selected library project, the product types are not the same.

client/ayon_core/tools/loader/control.py
```diff
@@ -425,3 +432,28 @@ def _create_event_system(self):
...
+        context = get_current_context()
```
iLLiCiTiT, line 437:
```suggestion
        context = self.get_current_context()
```

client/ayon_core/tools/loader/control.py
```diff
@@ -425,3 +432,28 @@ def _create_event_system(self):
...
+                    return ProductTypesFilter(
+                        is_include=profile["is_include"],
+                        product_types=profile["filter_product_types"]
+                    )
+
+        # Default to all as allowed
+        return ProductTypesFilter(
+            is_include=False,
+            product_types=[]
+        )
```
iLLiCiTiT, line 459:
```suggestion
    def get_product_types_filter(self, project_name):
        default = ProductTypesFilter(
            is_include=False,
            product_types=[]
        )
        # Without host is not determined context
        if self._host is None:
            return default

        context = self.get_current_context()
        if (
            not all(context.values())
            or context["project_name"] != project_name
        ):
            return default
        settings = get_current_project_settings()
        profiles =(
            settings
            ["core"]
            ["tools"]
            ["loader"]
            ["product_type_filter_profiles"]
        )
        if not profiles:
            return default
        task_entity = get_current_task_entity(fields={"taskType"})
        host_name = getattr(self._host, "name", get_current_host_name())
        profile = filter_profiles(
            profiles,
            {
                "hosts": host_name,
                "task_types": (task_entity or {}).get("taskType")
            }
        )
        if profile:
            return ProductTypesFilter(
                is_include=profile["is_include"],
                product_types=profile["filter_product_types"]
            )
        return default
```

client/ayon_core/tools/loader/control.py
BigRoy: Is `self.get_current_context()` the context the user browsed to in the UI or the context the UI was launched with? :)

iLLiCiTiT: Current context of host used in controller.

client/ayon_core/tools/loader/ui/product_types_widget.py
```diff
@@ -162,7 +163,26 @@ def _on_project_change(self, event):
...
         self._product_types_model.refresh(project_name)
 
+    def showEvent(self, event):
+        self._refresh_product_types_filter = True
+        super().showEvent(event)
+
     def _on_refresh_finished(self):
+
+        # Apply product types filter
+        if self._refresh_product_types_filter:
```
iLLiCiTiT, line 173: BTW I would say this should happen in the model instead.

BigRoy: How would that occur - if it's the viewer that needs to update the selection model?

BigRoy: Hmm - it can just be moved to the `product_types_model` since we're using functions of that anyway.
Should I hence make a `update_product_type_filters_for_project` to the model so the view can still trigger it on the show event?

iLLiCiTiT: Well, what is the point of using view? To refresh the settings on showing it? If yes then just add method `reset_product_type_filter` to the model to refresh it on next `refresh`?

BigRoy: Done with https://github.com/ynput/ayon-core/pull/792/commits/88959b2c54a5a9569d9e426a29587bd829c3b72f

client/ayon_core/tools/loader/ui/product_types_widget.py
```diff
@@ -71,6 +71,21 @@ def refresh(self, project_name):
         self._refreshing = False
         self.refreshed.emit()
 
+    def reset_product_types_filter(self):
+
```
iLLiCiTiT, line 75:
```suggestion
```

client/ayon_core/tools/loader/ui/product_types_widget.py
```diff
@@ -162,7 +178,16 @@ def _on_project_change(self, event):
         project_name = event["project_name"]
         self._product_types_model.refresh(project_name)
 
+    def showEvent(self, event):
+        self._refresh_product_types_filter = True
+        super().showEvent(event)
+
     def _on_refresh_finished(self):
+
```
iLLiCiTiT, line 186:
```suggestion
```

server/settings/tools.py
```diff
@@ -222,6 +222,13 @@ def _product_types_enum():
     ]
 
 
+def filter_type_enum():
+    return [
+        {"value": "is_allow_list", "label": "Allow list"},
+        {"value": "is_deny_list", "label": "Deny list"},
```
BigRoy, line 228: Could this have `value` True and False and have the type be `bool` whilst still being an Enum (for niceness)?

client/ayon_core/tools/loader/abstract.py
```diff
@@ -1006,3 +1014,13 @@ def get_representations_sync_status(
         """
 
         pass
+
+    @abstractmethod
+    def get_product_types_filter(self):
+        """Return product type filter for project name (and current context).
```
BigRoy, line 1020:
```suggestion
        """Return product type filter for current context.
```

server/settings/tools.py
iLLiCiTiT: Will try out...

iLLiCiTiT: Nope, it will become checkbox.

BigRoy: Ugh. :( I just really really dislike the usage of strings for this. It feels so darn nasty just so that we can stick a label on it for the user in the settings frontend.

I wonder if we shouldn't just keep it a checkbox for now so that the data is as we want it to be, yet open a frontend/backend issue to track and allow better options for checkboxes in that regard? (Like e.g. enum resolver with two options)

Thoughts?

iLLiCiTiT: Fine for me, I just wanted to change Exclude/Include to something that is understandable un human language 😄

BigRoy: Hmm keep it as is I suppose. It looks fine in setings, we'll figure out better settings design in the future maybe...

BigRoy:
@iLLiCiTiT PR works for me is. Build new addon, configured settings all work great.

- No profiles: works
- Allow list profile works
- Deny list profile works

✅ 🚀 
You may need to approve review yourself @iLLiCiTiT and then merge.

iLLiCiTiT (approved):

PR #794, opened 2024-07-23
Tray: Singleton tray

iLLiCiTiT:
## Changelog Description
Tray is now singleton process. Only one tray can be running for AYON url and settings variant (`production`, `staging` or dev variants).

## Additional info
Moved webserver module to tray tool and make it as base functionality of the tool. The webserver is also used to determine if tray is running or not. Tray is storing metadata file containing webserver url, pid and information if initialization finished which is used to determine if is already running.

Added some api functions to be able receive some of these information. The most important are `is_tray_running` and `get_tray_server_url`. The `get_tray_server_url` should be used instead of `os.environ["AYON_WEBSERVER_URL"]`.

## Notes
More features will come in future PRs, based on what we'll need.

## Testing notes:
1. Create ayon-core package and upload to AYON server.
2. Use it in a bundle.
3. Start tray.
4. Try to start tray again, it should not start second one.


64qam:
Still creates multiple publishers.

64qam (approved):

client/ayon_core/tools/tray/addons_manager.py
```diff
@@ -0,0 +1,247 @@
...
+        statisc_url = f"{webserver_url}/res"
+
+        # TODO stop using these env variables
+        # - function 'get_tray_server_url' should be used instead
+        os.environ[self.webserver_url_env] = webserver_url
+        os.environ["AYON_STATICS_SERVER"] = statisc_url
+
+        # Deprecated
+        os.environ["OPENPYPE_WEBSERVER_URL"] = webserver_url
+        os.environ["OPENPYPE_STATICS_SERVER"] = statisc_url
```
BigRoy, line 247:
```suggestion
        statics_url = f"{webserver_url}/res"

        # TODO stop using these env variables
        # - function 'get_tray_server_url' should be used instead
        os.environ[self.webserver_url_env] = webserver_url
        os.environ["AYON_STATICS_SERVER"] = statics_url

        # Deprecated
        os.environ["OPENPYPE_WEBSERVER_URL"] = webserver_url
        os.environ["OPENPYPE_STATICS_SERVER"] = statics_url
```

client/ayon_core/tools/tray/lib.py
```diff
@@ -0,0 +1,303 @@
...
+        time.sleep(5)
+        state = get_tray_state()
+        if state == TrayState.RUNNING:
+            return
+        if state == TrayState.STARTING:
+            file_info = get_tray_file_info() or {}
+            pid = file_info.get("pid")
+            if pid is not None:
+                os.kill(pid, signal.SIGTERM)
+            remove_tray_server_url()
```
BigRoy, line 300: This code would never run with the `return` above?

BigRoy (commented):
Overall this seemed to work.

1. I could not usually launch another tray to the same server
2. I was able to launch a dev tray to my dev server, and simultaneously launch one for my production server (both in dev mode)

---

I did notice that if I run tray very rapidly after each other (e.g. run twice almost at the same time) it gets a bit further, doesn't close down but reaches this error (in console):

```
  File "E:\dev\ayon-core\client\ayon_core\tools\tray\ui\tray.py", line 625, in _on_start_timer
    self._tray_widget.initialize_addons()
  File "E:\dev\ayon-core\client\ayon_core\tools\tray\ui\tray.py", line 540, in initialize_addons
    self._tray_manager.initialize_addons()
  File "E:\dev\ayon-core\client\ayon_core\tools\tray\ui\tray.py", line 187, in initialize_addons
    set_tray_server_url(
  File "E:\dev\ayon-core\client\ayon_core\tools\tray\lib.py", line 182, in set_tray_server_url
    raise TrayIsRunningError("Tray is already running.")
ayon_core.tools.tray.lib.TrayIsRunningError: Tray is already running.
```
Aside of that it visually appears to "keep running" from the UI perspective.

---

Aside of me seeing from the console that it's shutting down because it already had a tray running I can imagine it's very hard to miss by an artist why it may be launching (if they even see that at all). As such, I wonder whether we should add a pop-up that says something like "Shutting down because another Tray process is already running") Or - maybe better yet, have it shut down but send a message to the running tray to show a "tray message/popup" with that message (so it brings to attention the one that is actively running with a tray message - and when the message is clicked (and not dismissed) opens the Launcher for example. So that for the artists running it again it doesn't look like nothing happens, instead they get focused at the one that is already running?

client/ayon_core/tools/tray/addons_manager.py
```diff
@@ -0,0 +1,247 @@
...
+        "launcher_tool",
+        "clockify",
+    )
+
+    def __init__(self, tray_manager):
+        super().__init__(initialize=False)
+
+        self._tray_manager = tray_manager
+
+        self._webserver_manager = WebServerManager(find_free_port(), None)
```
BigRoy, line 29: Shouldn't this somehow pass it the WEBSERVER URL that's configured (e.g. through the env var?)

iLLiCiTiT: Not sure I understand?

BigRoy: Correct - but none of those seem passed to the `WebServerManager`?

iLLiCiTiT: Because `WebServerManager` is source of the value...

iLLiCiTiT:
> I did notice that if I run tray very rapidly after each other (e.g. run twice almost at the same time) it gets a bit further, doesn't close down but reaches this error (in console):

Should be resolved now.

BigRoy (approved):
> > I did notice that if I run tray very rapidly after each other (e.g. run twice almost at the same time) it gets a bit further, doesn't close down but reaches this error (in console):
> 
> Should be resolved now.

Correct, works now.

---

I still think this comment is valid:

> Aside of me seeing from the console that it's shutting down because it already had a tray running I can imagine it's very easy to miss by an artist why it may fail to launch (if they even see that at all). As such, I wonder whether we should add a pop-up that says something like "Shutting down because another Tray process is already running") Or - maybe better yet, have it shut down but send a message to the running tray to show a "tray message/popup" with that message (so it brings to attention the one that is actively running with a tray message - and when the message is clicked (and not dismissed) opens the Launcher for example. So that for the artists running it again it doesn't look like nothing happens, instead they get focused at the one that is already running?

**But** we can leave that to a separate issue and PR.

iLLiCiTiT:
> But we can leave that to a separate issue and PR. 

100% agree, 100% different PR.

PR #795, opened 2024-07-23
Chore: Faster cli initalization

iLLiCiTiT:
## Changelog Description
Create AddonsManager only once for cli processing.

## Additional info
Create only AddonsManager object only single time for whole main cli processing which should make the startup a little bit faster as it does not need to re-fetch all the data.

## Testing notes:
1. All command line commands are working as expected.
2. The most affected commands are `extractenvironments` and `publish` which now do not create addons manager bu expect it as argument. Both are used on farm.

### DIFF
--- a/client/ayon_core/cli.py
+++ b/client/ayon_core/cli.py
@@ -271,7 +271,7 @@ def _set_addons_environments(addons_manager):
 
 def _add_addons(addons_manager):
     """Modules/Addons can add their cli commands dynamically."""
-    log = Logger.get_logger("CLI-AddModules")
+    log = Logger.get_logger("CLI-AddAddons")
     for addon_obj in addons_manager.addons:
         try:
             addon_obj.cli(addon)
--- a/client/ayon_core/cli.py
+++ b/client/ayon_core/cli.py
@@ -120,7 +120,7 @@ def publish(path, targets, gui):
     """Start CLI publishing.
 
     Publish collects json from path provided as an argument.
-S
+
     """
     Commands.publish(path, targets, gui)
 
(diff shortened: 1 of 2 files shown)


client/ayon_core/cli_commands.py
```diff
@@ -20,28 +20,12 @@ def launch_tray():
...
-                )
-        return click_func
-
-    @staticmethod
-    def publish(path: str, targets: list=None, gui:bool=False) -> None:
+    def publish(
+        path: str,
+        targets: list = None,
+        gui: bool = False,
+        addons_manager=None,
```
BigRoy, line 27: Should this get a type hint? Also, argument is lacking in the docstring

client/ayon_core/cli.py
```diff
@@ -258,6 +269,21 @@ def _set_addons_environments():
         os.environ.update(env)
 
 
+def _add_addons(addons_manager):
+    """Modules/Addons can add their cli commands dynamically."""
+    log = Logger.get_logger("CLI-AddModules")
```
BigRoy, line 274: The logger name feels a bit weird?

It's adding CLIs, in a function named `_add_addons` yet the logger name references to Modules? Whih is it? :)

Should we drop naming this "modules" and just refer to them as addons?

iLLiCiTiT: Why you just don't add suggestion?
```suggestion
    log = Logger.get_logger("CLI-AddAddons")
```

client/ayon_core/cli_commands.py
iLLiCiTiT: Wanted to move the function to different place in following PR.

iLLiCiTiT: Life is short, added.

BigRoy (approved):
Can't say much about performance differences since I did not test that - but it did work for me over Deadline render (fusion) and publish where the environments did get applied from AYON, etc.

PR #796, opened 2024-07-23
USD plugin: Don't crash plugin file because of missing dependency

iLLiCiTiT:
## Changelog Description
USD publish plugin files do not crash on missing dependency.

## Additional info
Added validator to check missing dependencies.

This is first shot, any modifications welcommed, the point is that the file should not crash.

## Testing notes:
1. Open publisher (e.g. in traypublisher).
2. In Details tab should not be any crashed plugin paths (at least there should not be `extract_usd_layer_contributions.py`).


BigRoy (approved):
I started fiddling around with other ways that would completely "deregister" the plug-ins etc. and issue a one shot warning using `warnings` package... but it just felt less nice for artists because they'd still get a warning to the console.

This is much more hidden away until it hits a production case where it should have been working - so fine with me.

Worked in Maya 2024 with USD exports, and tray publisher + maya did indeed ignore it as crashed plug-ins.

PR #797, opened 2024-07-24
Tray: Add username to info

iLLiCiTiT:
## Changelog Description
Tray info received with `/tray` endpoint does return username.

## Additional info
Username will be important check for running webactions, which can be available from the start of tray process. Cache the username as user change requires restart.

## Testing notes:
1. You can get tray information to validate that `"username"` is there.


BigRoy:
Would love to test - but how do I "get" this tray info? How do I request it? Do you have an example for dummies?

iLLiCiTiT:
> Would love to test - but how do I "get" this tray info? How do I request it? Do you have an example for dummies?

Oh, when you start a tray, you can open Admin > Console and do
```python
import json
import requests
from ayon_core.tools.tray import get_tray_server_url

response = requests.get(f"{get_tray_server_url()}/tray")
print(json.dumps(response.json(), indent=4))
```

64qam (approved):

BigRoy:
By the way, also works when running the code not from console in tray but e.g. from a DCC (to showcase the point that it does get it via the request)

MilaKudr (approved):

PR #798, opened 2024-07-24
Tray: Force start tray

iLLiCiTiT:
## Changelog Description
Added option to force start of tray.

## Additional info
Using `--force` will kill existing tray process and start a new one. Can be used when tray is for some reason stuct and artist should be able to continue.

## Testing notes:
1. Create package and upload to server, or use dev path (server part is not needed).
2. Start a tray in one terminal.
3. Start a tray in second terminal > this should tell you that tray is already running.
4. Start a tray in second terminal with `--force` argument > previous tray should stop and new should start.

### DIFF
--- a/client/ayon_core/cli.py
+++ b/client/ayon_core/cli.py
@@ -59,13 +59,20 @@ def main_cli(ctx):
 
 
 @main_cli.command()
-def tray():
+@click.option(
+    "--force",
+    is_flag=True,
+    help="Force to start tray and close any existing one.")
+def tray(force):
     """Launch AYON tray.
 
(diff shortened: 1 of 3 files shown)


64qam:
Ayon is still running from the first terminal and the second always stops with "Tray is already running". Unable to stop the first one.

iLLiCiTiT:
> Ayon is still running from the first terminal and the second always stops with "Tray is already running". Unable to stop the first one.

Please try again, there was some issue with encoding.

64qam:
Works only if ayon is lauched with `tray --force`

64qam (approved):

PR #799, opened 2024-07-24
Tray: Make starting tray check faster

iLLiCiTiT:
## Changelog Description
Check if the PID defined in tray info file is alive to make validation of starting tray faster.

## Additional info
It happened to me that if tray is brute-force shutdown it does not cleanup the file, so next tray startup takes 10 seconds. Added check if the PID is still alive before any other check.

## Testing notes:
Not sure how to exactly replicate the issue. I've started tray in terminal and killed the terminal tab when splash showed up. With that done, the next startup takes at about 10 seconds. With this PR it should be faster.


64qam (approved):
well... maybe... I don't know...

PR #801, opened 2024-07-25
Tray: Send messages to tray

iLLiCiTiT:
## Changelog Description
Added option to send request to show message in tray and use it when tray is already running.

## Additional info
This may not work if OS does not support tray messages. Not sure what are the cases, I think it can happen only if it does not support even sys tray, which is bigger issue.

## Testing notes:
1. Start tray.
1. Try to start tray again.
2. A tray message should show up that tray is already running.


64qam (approved):

client/ayon_core/tools/tray/lib.py
```diff
@@ -356,6 +356,38 @@ def is_tray_running(
...
 
+def show_message_in_tray(
+    title, message, icon=None, msecs=None, tray_url=None
+):
+    """Show message in tray.
+
+    Args:
+        title (str): Message title.
+        message (str): Message content.
+        icon (Optional[str]): Icon for the message.
```
BigRoy, line 367: can we specify the available names in the docstring here?
And is there a type hint possibility where it forces a set of specific strings?

iLLiCiTiT: >  And is there a type hint possibility where it forces a set of specific strings?

There is Literal (not available in all pythons). But can be used in docstring.

iLLiCiTiT:
> There does seem to be the edge case if I run twice rapidly... one does close but (of course?) seems unable to emit the show message to the other.

Yeah I know, I would call it feature :D

PR #833, opened 2024-07-30
Update opencolorio version to ^2.3.2 in pyproject.toml

jakubjezek001:
## Changelog Description
Unlocking freezed opencolorio run time dependency version to 2.3.2 and higher


iLLiCiTiT (approved):
We've used older version for python 2 support I guess. As Python 2 was dropped and it is in runtime dependencies we probably don't have to care.

PR #838, opened 2024-08-07
Tray: Logout closes tray

iLLiCiTiT:
## Changelog Description
Logout does not keep tray running.

## Additional info
The issue was that `get_default_settings_variant` from `ayon_api` is actually using global connection, which can't log in if we just logged out.

## Testing notes:
1. Create package, upload to AYON server and use in bundle.
2. Start AYON tray with console (so you can see if it is closed).
3. Click on tray icon > Login.
4. In opened dialog click on Logour and confirm with Ok.
5. You should be logged out and tray should fully close.

### DIFF
--- a/client/ayon_core/tools/tray/lib.py
+++ b/client/ayon_core/tools/tray/lib.py
@@ -10,8 +10,8 @@ import signal
 import locale
 from typing import Optional, Dict, Tuple, Any
 
-import ayon_api
 import requests
+from ayon_api.utils import get_default_settings_variant
 
 from ayon_core.lib import (
     Logger,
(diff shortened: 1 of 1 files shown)


64qam (approved):

BigRoy:
Is this really nice UX? Should it restart maybe? I wouldn't expect the app to shutdown when clicking logout, I'd expect a new login prompt. But maybe that's just me?

iLLiCiTiT:
> Is this really nice UX? Should it restart maybe? I wouldn't expect the app to shutdown when clicking logout, I'd expect a new login prompt. But maybe that's just me?

There is option to change login in the UI. The login UI itself deserves better UX, in backlog,

NOTE: This PR did not change the logic, just fixed that there is not phantom tray process if you do use logout.

PR #840, opened 2024-08-08
Publish report viewer: Don't validate extension of dropped file

iLLiCiTiT:
## Changelog Description
Allow any extension to be dropped into the reports view. If it contains json, it will be loaded.

## Testing notes:
1. Open publish report viewer.
2. You can drop in .txt file with report.


m-u-r-p-h-y (approved):
works as expected 
image

PR #841, opened 2024-08-08
Use product name for render templates

antirotor:
## Changelog Description
This is fixing the issue where product names for render products were hardcoded. Old behavior is enabled by default, but deprecation warning is issued. New behavior (where product name templates are used) is controlled by option in the settings:

`ayon+settings://core/tools/creator/product_name_profiles/use_legacy_for_renders`

> [!NOTE]
> This is continuation and refactor of #447 - look there for previous discussions.

## Additional info

Old behavior was bypassing any template set in `ayon+settings://core/tools/creator/product_name_profiles`. With this change, one can set template like this:

`{product[type]}<{Renderlayer}><_{Aov}>`

resulting in:

`renderMainlayer_normal`

It is up to TD to set correctly such template to make sure, product name will be stable enough. As a side-effect, optional keys can be used now in product name templates adding flexibility (and perhaps some danger of misuse).

> [!WARNING]
To make use of this feature, `renderlayer` need to be set on the instance along with `aov` and is then collected to anatomy data. This is not in the scope of this PR (apart of the collection) and it must be implemented if following PRs to individual host addons.

It can be used in template formatting so you can use template for renders like:

```
{root[work]}/{project[name]}/{hierarchy}/{asset}/publish/{family}/{subset}/{@version}/<{renderlayer}/>
```
> [!NOTE]
> Already collected anatomy data (when available) could be passed to `get_product_name()` to allow even more keys to be used, but is there any real use case for it?

Also, as a point for discussion - there is `variant` that is now ignored in this case. It could be reasonable to use it exactly for this and move templating to "variant templates" but maybe it is breaking settings to even smaller unmanagable bits.

## Testing notes:
1. Set product name template to use `renderlayer` or `aov` keys (can be optional using`<{key}>` syntax)
2. Publish renders from DCC supporting setting of these keys (like Maya and Houdini)
3. Use `{renderlayer}` or  `{aov}` in anatomy templates

### Main points:

a) It should behave as before without any changes.
b) When option is set in the settings, new behavior should kick-in and renders should start to use product name templates.

### Testing steps:

- To test a) simply publish any render to farm. Product names of the result should stay the same as before.
- To test b), you'll need yet non-existent PRs in host addons. Stay tuned! :) - **correction:** with Maya it should already work.


antirotor:
This needs to solve `group_name` still

antirotor:
So I've tested it in Houdini and it works. There is one specific question about the **product group names** - strings that are used to group related publishes together in the UI.

So far the product group names were hardcoded - basically original hardcoded product name. This didn't include AOV names as they are added when render job is published so it correctly grouped together all renderpass AOVs. But with this PR, you can define product names with templates and so the group name itself cannot be that static. In this PR, I've made decision to explicitly "drop' aov name from the template data. So if you have product name template like so:

```
{product[type]}{Task[name]}<{Renderlayer}><_{Aov}>
```

and you have render layer called `Main` with **beauty**, **diffuse** and **z** AOVs, your resulting product names wil basically follow the previous behaviour (when you work on the task Lighting):

```
renderLightingMain_beauty
renderLightingMain_diffuse
renderLightingMain_z
```

and all of them will be grouped under `renderLightingMain` (because `aov` is not in template data and in template it is set as optional token).

But at the end, we should probably have something more robust, like templates for the group name itself. This is topic for another potential PR.

LiborBatek (approved):
I have tested both the ```legacy``` workflow and the ```templated``` workflow within ```maya``` and ```arnold``` renderer.

All seems working fine...my rendered files and folders follows the ```template``` and also using tokens ```renderlayer``` and ```aov``` and also for ```product name```.

Here is my output with using those tokens within folderpath:

Screenshot 2024-08-12 173515

client/ayon_core/pipeline/create/product_name.py
```diff
@@ -183,7 +182,10 @@ def get_product_name(
             fill_pairs[key] = value
 
     try:
-        return template.format(**prepare_template_data(fill_pairs))
+        return StringTemplate.format_template(
```
iLLiCiTiT, line 185:
```suggestion
        return StringTemplate.format_strict_template(
```

iLLiCiTiT: It should also probably be explicitly converted to `str` because this does return `TemplateResult` which might cause issues in some DCCs.

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -1,20 +1,17 @@
...
-
-from ayon_core.pipeline import (
-    get_current_project_name,
-    get_representation_path,
-)
 from ayon_core.lib import Logger
-from ayon_core.pipeline.publish import KnownPublishError
+from ayon_core.pipeline import get_current_project_name, get_representation_path
+from ayon_core.pipeline.create import get_product_name
 from ayon_core.pipeline.farm.patterning import match_aov_pattern
```
iLLiCiTiT, line 13: Could this be relative import?

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -515,6 +517,132 @@ def create_instances_for_aov(instance, skeleton, aov_filter,
...
+
+    return resulting_product_name, resulting_group_name
+
+
+def get_product_name_and_group_from_template(
+        task_entity,
+        project_name,
+        host_name,
+        product_type,
+        variant,
```
iLLiCiTiT, line 592:
```suggestion
        product_type,
        variant,
        host_name,
```

iLLiCiTiT, line 592: It should probably match the arguments order from `get_product_name`.
```suggestion
        project_name,
        task_entity,
        product_type,
        variant,
        host_name,
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -515,6 +517,132 @@ def create_instances_for_aov(instance, skeleton, aov_filter,
...
+
+    Returns:
+        tuple: product name and group name.
+
+    """
+    # remove 'aov' from data used to format group. See todo comment above
+    # for possible solution.
+    _dynamic_data = deepcopy(dynamic_data) or {}
+    if _dynamic_data["aov"]:
+        del _dynamic_data["aov"]
```
iLLiCiTiT, line 623:
```suggestion
    _dynamic_data.pop("aov", None)
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -708,6 +833,36 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
+            raise ValueError("Found multiple non related files "
+                             "to render, don't know what to do "
+                             "with them.")
+        return rem[0]
+    else:
+        # but we really expect only one collection.
+        # Nothing else make sense.
+        if len(cols) != 1:
+            raise ValueError("Only one image sequence type is expected.")  # noqa: E501
+        return list(cols[0])
```
iLLiCiTiT, line 863: Unnecesasry else statement.
```suggestion
    # but we really expect only one collection.
    # Nothing else make sense.
    if len(cols) != 1:
        raise ValueError("Only one image sequence type is expected.")  # noqa: E501
    return list(cols[0])
```

server/settings/tools.py
```diff
@@ -65,6 +66,14 @@ class CreatorToolModel(BaseSettingsModel):
             title="Create Smart Select"
         )
     )
+    # TODO: change to False in next releases
+    use_legacy_product_names_for_renders: bool = SettingsField(
+        True, title="Use legacy product names for renders",
```
iLLiCiTiT, line 71:
```suggestion
        True,
        title="Use legacy product names for renders",
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -539,68 +667,69 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
-        if isinstance(col, (list, tuple)):
-            staging = os.path.dirname(col[0])
         else:
-            staging = os.path.dirname(col)
+            product_name, group_name = get_product_name_and_group_from_template(
+                task_entity=instance.data["taskEntity"],
+                project_name=instance.context.data["projectName"],
+                host_name=instance.context.data["hostName"],
+                product_type=skeleton["productType"],
+                variant=instance.data.get('variant', source_product_name),
```
iLLiCiTiT, line 728:
```suggestion
                variant=instance.data.get("variant", source_product_name),
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -539,68 +667,69 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
-                    product_name = '{}_{}_{}'.format(group_name, cam, aov)
-                else:
-                    product_name = "{}_{}".format(group_name, aov)
-        else:
-            if aov:
-                product_name = '{}_{}'.format(group_name, aov)
-            else:
-                product_name = '{}'.format(group_name)
+        expected_filepath = collected_files[0] \
+            if isinstance(collected_files, (list, tuple)) else collected_files
```
iLLiCiTiT, line 683: Only because I had to read it 3 times to find out what it does...
```suggestion
        expected_filepath = collected_files
        if isinstance(collected_files, (list, tuple)):
            expected_filepath = collected_files[0]
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -515,6 +517,132 @@ def create_instances_for_aov(instance, skeleton, aov_filter,
...
+    # create product name `<product type><Task><Product name>`
+    if not source_product_name.startswith(product_type):
+        resulting_group_name = '{}{}{}{}{}'.format(
+            product_type,
+            task_name[0].upper(), task_name[1:],
+            source_product_name[0].upper(), source_product_name[1:])
+    else:
+        resulting_group_name = source_product_name
+
+    resulting_product_name = '{}'.format(resulting_group_name)
```
iLLiCiTiT, line 566:
```suggestion
    resulting_product_name = resulting_group_name
```

client/ayon_core/pipeline/farm/pyblish_functions.py
antirotor: PEP-8 state that:
> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)

Relative are perfectly valid in cases, where it can (in complex scenarios) improve readability. While these imports are definitely heading that way, I don't think that they are there yet :)

antirotor:
@LiborBatek could you please test it quickly for one last time? Thanks

LiborBatek (approved):
All good. Tested in both ```legacy``` mode and with publishing ```profiles```.

Here is my output folder with published renders following my publish template and also from the legacy mode.
image

Product names also follow these, so all good!

server/settings/tools.py
```diff
@@ -65,6 +66,15 @@ class CreatorToolModel(BaseSettingsModel):
...
         )
     )
+    # TODO: change to False in next releases
+    use_legacy_product_names_for_renders: bool = SettingsField(
+        True,
+        title="Use legacy product names for renders",
+        description="Use product naming templates for renders. "
+                    "This is for backwards compatibility enabled by default."
+                    "When enabled, it will ignore any templates for renders "
+                    "that are set in the product name profiles.")
```
BigRoy, line 76: This description seems a bit awkward. First it states "use product naming templates for renders." but if this is enabled, it does not do that, but it does the opposite... 

So the description is actually "Ignore product naming templates for renders but use a hardcoded template. This is for backwards compatibility..." etc. ?

client/ayon_core/plugins/publish/help/validate_unique_subsets.xml
```diff
@@ -11,7 +11,11 @@ Multiples instances from your scene are set to publish into the same folder > pr
 
 ### How to repair?
 
-Remove the offending instances or rename to have a unique name.
+Remove the offending instances or rename to have a unique name. Also, please
+    check your product name templates to ensure that resolved names are
+    sufficiently unique. You can find that settings:
+
+    ayon+settings://core/tools/creator/product_name_profiles
```
BigRoy, line 18: It would be **epic** if somehow we can make this a hyperlink in the UI that actually links to the AYON web frontend at those settings? @dee-ynput ?

client/ayon_core/plugins/publish/help/validate_unique_subsets.xml
```diff
@@ -11,7 +11,11 @@ Multiples instances from your scene are set to publish into the same folder > pr
 
 ### How to repair?
 
-Remove the offending instances or rename to have a unique name.
+Remove the offending instances or rename to have a unique name. Also, please
+    check your product name templates to ensure that resolved names are
+    sufficiently unique. You can find that settings:
```
BigRoy, line 16: _Admittedly_ as a user I wouldn't know 'what to do' at those settings to ensure the resolved uniqueness. Nothing critical - but if anyone gets this invalidation message I have a feeling we might start getting questions about this. Let's see what comes up.

PR #842, opened 2024-08-09
Chore: Storage and Local directories

iLLiCiTiT:
## Changelog Description
Added 2 new functions `get_launcher_local_dir` and `get_launcher_storage_dir`. One is used to get storage directory where addons and dependency packages are stored and the other where user/machine files are stored.

## Additional info
The PR is related to changes in AYON launcher (PR https://github.com/ynput/ayon-launcher/pull/144 ). The values are based on environment variable set during AYON launcher bootstrap, if are not set then previous implementation is used. This PR reduces usage of `appdirs` in ayon-core as those new functions should be used. Implementation is backwards compatible and new AYON launcher is not necessary.

### Deprecation
- Function `get_ayon_appdirs` was marked as deprecated.
- Removed `_create_local_site_id` function which is not used anywhere and is dangling there for more than 4 months. The functionality happen in AYON launcher and it does not make sense to use it in ayon-core (it would be too late).

## Testing notes:
1. When using AYON launcher without the PR above it should still load addons.
2. It should load addons even if you change addons directory with `AYON_ADDONS_DIR` to other place.
3. When using AYON launcher with the PR above it should correctly use addons and dependency packages in bounds of related changes. One has to read that PR description to know them, for example changing `AYON_LAUNCHER_STORAGE_DIR`.
4. More important part for ayon-core is that change of `AYON_LAUNCHER_LOCAL_DIR` will actually start to store user/machine specific files to different location, like console interpreter tabs or running tray information.


64qam (approved):
lgtm, but anyone else should also test it.

PR #845, opened 2024-08-14
Provided unique file prefix in Extract OTIO review

kalisp:
## Changelog Description
Limits overwrites from multiple instances when creating intermediate files for burnins creation.

## Additional info
Previously it resulted in `tempFile_burnin` file which got overwritten. Now it would be containing folder path and name, eg.
`shots_mk020_mk020_0010_mk020_0010_h264.mp4`.

This file is temporary and will be correctly renamed in integration, but if multiple instances would be creating review, they would be overwriting same temporary file.

## Testing notes:
1. publish from Hiero
2. check output of `ExtractBurnin`, look for `full_input_path`, it shouldnt contain `tempFile` but folder path and name
image


MilaKudr (approved):
Before 
Screenshot 2024-09-03 153108_2
After
Screenshot 2024-09-03 153851_2

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -62,6 +61,8 @@ def process(self, instance):
             make_sequence_collection
         )
 
+        self.temp_file_head = self._get_unique_file_prefix(instance)
```
BigRoy, line 64: Honestly - we shouldn't really be setting this on the class instance because the plugin instance is shared between instances. Sure, we're always resetting it here - but it may produce odd side effects if someone else doesn't know about that and starts tweaking the code for this?

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -491,3 +492,19 @@ def _get_ffmpeg_output(self):
...
+        unique prefix for intermediate file for burnins.
+        """
+        folder_path = instance.data["folderPath"]
+        folder_name = folder_path.split("/")[-1]
+        folder_path = folder_path.replace("/", "_").lstrip("_")
+
+        file_prefix = f"{folder_path}_{folder_name}."
+        self.log.debug(f"file_prefix::{file_prefix}")
+
+        return file_prefix
```
BigRoy, line 509: Does this _really_ enforce uniqueness? Is there no way one folder comes up more than once but with multiple media?

Can't we just use some "temp" file name but with a prefix that we want, like whatever Python `tempfile` would generate as path. If you provide a prefix it'll still attach some actual temp identifier after it?

jakubjezek001 (approved):
This is good to be merged. It would be good to address @BigRoy 's comments, but they are cosmetic and could be done in other PR - in case client waits for the fix now. Those comments are not blockers but definitelly would be improving workflow.

client/ayon_core/plugins/publish/extract_otio_review.py
kalisp: That would need much deeper refactoring as there are many other occurrences of other variables.

kalisp:
Renamed used method to be more descriptive, added details to docstring and TODO for future refactoring.

PR #847, opened 2024-08-19
Publisher: Added default iterator on publisher initialization

iLLiCiTiT:
## Changelog Description
Fix possible issue with running main iterator before reset.

## Additional info
The issue happens in flame randomly, we could not replicate the issue, so we don't know the "real" reason and it is possible that this won't fix it, but it definetelly won't hurt.

The logic actually changed in https://github.com/ynput/ayon-core/pull/689 , before that PR `_main_thread_iter` was set to `None` in `__init__` which was validated afterwards, looks like that might be the cause, but I don't know how.

### Traceback
  File "../ayon-core/client/ayon_core/tools/publisher/control_qt.py", line 52, in _execute
    item.process()
  File "../ayon-core/client/ayon_core/tools/publisher/control_qt.py", line 20, in process
    self.callback(*self.args, **self.kwargs)
  File "../ayon-core/client/ayon_core/tools/publisher/control_qt.py", line 95, in _next_publish_item_process
    func = self._publish_model.get_next_process_func()
  File "../ayon-core/client/ayon_core/tools/publisher/models/publish.py", line 918, in get_next_process_func
    item = next(self._main_thread_iter)
StopIteration

## Testing notes:
1. Flame does not complain about empty iterator.

### DIFF
--- a/client/ayon_core/tools/publisher/models/publish.py
+++ b/client/ayon_core/tools/publisher/models/publish.py
@@ -829,7 +829,7 @@ class PublishModel:
         )
 
         # Plugin iterator
-        self._main_thread_iter: Iterable[partial] = []
+        self._main_thread_iter: Iterable[partial] = self._default_iterator()
 
     def reset(self):
         create_context = self._controller.get_create_context()
@@ -900,10 +900,9 @@ class PublishModel:
         #   only in specific cases (e.g. when it happens for a first time)
 
         if (
-            self._main_thread_iter is None
             # There are validation errors and validation is passed
             # - can't do any progree
-            or (
+            (
                 self._publish_has_validated
                 and self._publish_has_validation_errors
             )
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
The issue of this error:
```python
  File "E:\dev\ayon-core\client\ayon_core\tools\publisher\control_qt.py", line 52, in _execute
    item.process()
  File "E:\dev\ayon-core\client\ayon_core\tools\publisher\control_qt.py", line 20, in process
    self.callback(*self.args, **self.kwargs)
  File "E:\dev\ayon-core\client\ayon_core\tools\publisher\control_qt.py", line 95, in _next_publish_item_process
    func = self._publish_model.get_next_process_func()
  File "E:\dev\ayon-core\client\ayon_core\tools\publisher\models\publish.py", line 930, in get_next_process_func
    item = next(self._main_thread_iter)
StopIteration
```
Upon validate, and then after continuing the publish is gone. I was unable to reproduce that error so this seems to be completely functional and fix the issue.

I tested in Maya 2024, Blender 4.2, Houdini 20.0.724 and Fusion 18. All good.
Resetting publisher, and republishing also seems to work.

---

I did a quick read over the code but it was a bit hard to grasp how and why this fixes it - so I couldn't say much about that. Sorry! ⚠️

PR #849, opened 2024-08-20
Prepare process: Enhancements

iLLiCiTiT:
## Changelog Description
Functions `ensure_addons_are_process_ready` and `ensure_addons_are_process_context_ready` returns True/False if failed or not.
Object of `ProcessContext` expects addon name and version and are not optional.
Added new function `is_headless_mode_enabled` to `ayon_core.lib`.

## Testing notes:
The functionality is not used anywhere at this moment so we don't have to worry about backwards compatibility.
1. Validate the changes made make sense.
2. Only addon using this function is `applications` addon in this PR https://github.com/ynput/ayon-applications/pull/1 which needs these changes to be able handle errors.


client/ayon_core/addon/utils.py
```diff
@@ -72,7 +72,7 @@ def ensure_addons_are_process_context_ready(
     process_context: ProcessContext,
     addons_manager: Optional[AddonsManager] = None,
     exit_on_failure: bool = True,
-) -> Optional[Exception]:
+):
```
BigRoy, line 75: The return type should also be updated in the docstring.

client/ayon_core/addon/utils.py
```diff
@@ -150,35 +148,52 @@ def ensure_addons_are_process_context_ready(
...
+    addon_name: Optional[str] = None,
+    addon_version: Optional[str] = None,
+    project_name: Optional[str] = None,
+    headless: Optional[bool] = None,
     addons_manager: Optional[AddonsManager] = None,
     exit_on_failure: bool = True,
     **kwargs,
-) -> Optional[Exception]:
+) -> ProcessContext:
     """Ensure all enabled addons are ready to be used in the given context.
```
BigRoy, line 164: So with the arguments now exposed (which I like) there is something now that strikes me as odd. Why does something that tells me whether "all addons are ready" take an addon name and version as input argument?

Is this doing it for all addons, or just the one? With the arguments now exposed it sounds more like this is more a `ensure_addon_is_process_ready` function?

Or should we improve the docstring to elaborate the use case? Why does it take an input addon if it checks for ALL addons to be ready?

iLLiCiTiT: https://github.com/ynput/ayon-core/pull/849/files#diff-6f20bc5b8745075bff6ad36db99c83009e00f864cfff5cefee4f56905c16d846R171

```
addon_name (Optional[str]): Addon name which triggered process.
addon_version (Optional[str]): Addon version which triggered process.
```

BigRoy: I see - so one addon, requests other addons?

iLLiCiTiT: Well, technically one addon requests ayon-core to prepare stuff, and one of steps in ayon-core is that it asks addons to prepare them (if they have the logic implemented).

client/ayon_core/addon/utils.py
```diff
@@ -150,35 +148,52 @@ def ensure_addons_are_process_context_ready(
...
+        addon_name,
+        addon_version,
+        project_name,
+        headless,
+        **kwargs
+    )
+    ensure_addons_are_process_context_ready(
         context, addons_manager, exit_on_failure
     )
+    return context
```
BigRoy, line 199: Does whoever receive this object have any meaningful reason to set the exception and or set the prepared state - I assume not? I have a feeling if maybe we should just be returning a (static?) data object instead to avoid confusion? We're really just returning the prepared state and the exception - that's it?

iLLiCiTiT: For my usecase the context has more value as it fills `headless` based on logic I should not know about at that specific case. I did want to be able to get exception to be able to handle the error in custom way.

For me `ensure_addons_are_process_ready` is just helper wrapper for exactly the content of the function when I don't want to encapsulate the logic to e.g. UI splash screen, but only run preparation and then get context information that was filled automatically.

BigRoy: Hmm - I see. I suppose the change is made to fit your use case and there are reasons to do it as such.

BigRoy (approved):
Didn't test - but the code changes visually make sense

client/ayon_core/addon/utils.py
iLLiCiTiT: Unresolving. Those are relevant comments and questions. This is not yet in any release and I would rather not touch it in future, so criticism is welcome.

Full context:
The main reason this functionality is necessary is to give addons option to prepare them -> what would happen in tray if you would use it, like fill ftrack credentials, fill environment variables that are filled only in `tray_start` etc. They can stop the process if their need is not matched. Main usage is for webactions, they can be started without ever running tray.

This cannot happen automatically, addon that might need it has to request it specifically.

Function `ensure_addons_are_process_ready` was added to simplify
```python
from ayon_core.addon import ProcessContext, ensure_addons_are_process_context_ready

def some_func():
    context = ProcessContext(
        "myaddon",
        "1.1.1",
        "myProject"
    )
    ensure_addons_are_process_context_ready(context)
    if context.headless:
        _headless_stuff()
    else:
        _ui_stuff()
```


```python
from ayon_core.addon import ensure_addons_are_process_ready

def some_func():
    context = ensure_addons_are_process_ready(
        "myaddon",
        "1.1.1",
        "myProject"
    )
    if context.headless:
        _headless_stuff()
    else:
        _ui_stuff()
```

BigRoy: I'm still not entirely sure about the exact use case. I think _if_ we're simplifying it makes more sense to simplify the need to pass the addon version (isn't that e.g. based on the current active bundle for your context?) Why do we need to explicitly pass it the version?

The fact that you need access to any exceptions or whether it's headless or not - I suppose makes sense. I'm also not entirely confident about the need for the `prepared` state.

I'd expect a `ensure_addons_are_process_ready` to just return something along the lines of a "boolean" to tell me whether it succeeded or not, potentially with the exception if it's not (which could also be an actual raised exception? Since we're requesting it to _ensure_ it?)

Again, take my comment with a grain of salt since I feel I'm quite on a different track than you and still not 100% sure.

Just here proposing an API that may seem more reasonable:
```python
context = ProcessContext(project_name="myProject")
result = context.initialize_for_addon("myaddon")
if not result:
    print(result.exception)

if context.is_headless:
    _headless_stuff()
else:
    _ui_stuff()
```

But maybe it's the docstring of `ProcessContext` that needs improving - because it doesn't seem to be 'generic context' but a process context specific to a particular addon? So it's more of a `AddonProcessContext`?

iLLiCiTiT: First idea was to expect `addon_name`, `addon_version` and `project_name` in `ensure_addons_are_process_ready`. Then I realized that we will in future need to add more data and changing signature of a function is PITE, thus `ProcessContext` was created, meant as "future-proof data holder".

client/ayon_core/addon/base.py
```diff
@@ -80,36 +81,41 @@ class ProcessPreparationError(Exception):
...
+        headless (Optional[bool]): Is process running in headless mode. Value
+            is filled with value based on state set in AYON launcher.
 
     """
     def __init__(
         self,
-        addon_name: Optional[str] = None,
-        addon_version: Optional[str] = None,
+        addon_name: str = None,
+        addon_version: str = None,
```
BigRoy, line 110: These should now not be optional according to the type hints?
```suggestion
        addon_name: str,
        addon_version: str,
```

PR #851, opened 2024-08-21
Widgets: Line edit with hints

iLLiCiTiT:
## Changelog Description
Implemented helper widget `HintedLineEdit` that can be used universally. Use the widget in Publisher for variant.

## Screenshot
image
image

## Testing notes:
Developers, look at the code please.
1. Open publisher in DCC (as many as possible).
2. It should look like before.

- the button is next to input (no gap inbetween)
- they have always same height (button and line edit)
- colors are propagated based on existence of product name
- hints of variant are shown and shows also hints base on existing product names


64qam (approved):

client/ayon_core/tools/utils/widgets.py
```diff
@@ -104,6 +105,151 @@ def __init__(self, *args, **kwargs):
...
+def get_down_arrow_icon() -> QtGui.QIcon:
+    normal_pixmap = QtGui.QPixmap(
+        get_style_image_path("down_arrow")
+    )
+    on_pixmap = QtGui.QPixmap(
+        get_style_image_path("down_arrow_on")
+    )
+    disabled_pixmap = QtGui.QPixmap(
+        get_style_image_path("down_arrow_disabled")
+    )
```
BigRoy, line 117: I wonder if it'd make sense to make these image paths constants so we don't need to check the file's existence each time?

client/ayon_core/tools/utils/widgets.py
```diff
@@ -104,6 +105,151 @@ def __init__(self, *args, **kwargs):
...
+        tsz = self._text_input.sizeHint()
+        bsz = self._options_button.sizeHint()
+        hint.setHeight(max(tsz.height(), bsz.height()))
+        return hint
+
+    # Adds ability to change style of the widgets
+    # - because style change of the 'HintedLineEdit' may not propagate
+    #   correctly 'HintedLineEditInput' and 'HintedLineEditButton'
+    def set_text_widget_object_name(self, name: str):
+        self._text_input.setObjectName(name)
```
BigRoy, line 220: Can't we use Object Names for this in general? So that the inner object name is **always** "button" or "text" or whatever and when referencing them in the `qss` stylesheet we point to something like `HintedLineEdit button` or whatever? CSS allows such 'child object' referencing to dedicate styling for them? Or even without explicit object names we could even do something like `HintedLineEdit QPushButton:hover` etc. right?

client/ayon_core/tools/utils/widgets.py
```diff
@@ -104,6 +105,151 @@ def __init__(self, *args, **kwargs):
...
+        if not self._options:
+            return
+
+        menu = QtWidgets.QMenu(self)
+        for option in self._options:
+            if option in self.SEPARATORS:
+                menu.addSeparator()
+                continue
+            action = menu.addAction(option)
+            action.triggered.connect(self._on_option_action)
```
BigRoy, line 241: In this case - since you're responding to the action any way, you can also add the callback to the menu instead so we don't need to per action, and then in the callback respond to the action.

```python
menu.triggered.connect(self._on_option_action)

def on_option_action(self, action):
    self.setText(action.text())
```
Something like that?

client/ayon_core/tools/utils/widgets.py
iLLiCiTiT: You would have to re-define all styles again. Even if you want to change just one thing.

client/ayon_core/tools/utils/widgets.py
iLLiCiTiT: Didn't know could be done, will do...

client/ayon_core/tools/utils/widgets.py
iLLiCiTiT: Done.

PR #853, opened 2024-08-22
OIIO Transcode: Simplify Extract OIIO transcode settings

MustafaJafar:
## Changelog Description
This PR aims to ease and simplify the settings of `ExtractOIIOTranscode` plugin.

List of TODOs
- [x] Simplify the settings
- [x] Tweak the code to work with the new settings
- [x] Add Better Tool Tips
- [x] Add setting conversion

## Additional info
This PR was not requested by clients or community. 
The motivation came from documentation when this PR started initially as renaming the input fields.
but it evolved from continues discussions and testing to improve the user experience of the settings.

Therefore, we no longer need to explain why the display and view settings are shown if the transcode type is colorspace. 

This PR is not backward compatible as it changed the settings hierarchy a little bit.
But, conversion logic is implemented so we can copy old settings to the new settings without an issue.

For reference, I was testing BigRoy's PR [Collect required data for local and farm render jobs to have a 'colorspace' value specified #75](https://github.com/ynput/ayon-houdini/pull/75)

## Demo
Animation_100

## Testing notes:
1. Add your settings to the plugin. 
2. Verify it works as usual


MustafaJafar:
Here's a snippet from a run test on my side. 
```
DEBUG: Looking for matching profile for: hosts: "houdini" | product_types: "render" | product_names: "render_beauty" | task_names: "cfx" | task_types: "CFX"
DEBUG: Profile selected: {'product_types': ['render'], 'hosts': ['houdini'], 'task_types': [], 'task_names': [], 'product_names': [], 'delete_original': True, 'outputs': [{'name': 'render', 'extension': 'jpeg', 'transcoding_type': 'use_colorspace', 'use_colorspace': {'colorspace': 'Output - sRGB'}, 'use_display_view': {'display': '', 'view': ''}, 'oiiotool_args': {'additional_command_args': []}, 'tags': [], 'custom_tags': []}]}
```

image

iLLiCiTiT:
NOTE: If it's about changing labels, then please change just labels, otherwise we would have to implement conversion of settings (PITA).

MustafaJafar:
> NOTE: If it's about changing labels and adding conditional enum, then please change just labels and add conditional enums without changing attribute names, otherwise we would have to implement conversion of settings (PITA).

Can I add conditional enum without creating a new settings class ? 
Specifically, I mean `display` and `view` settings.

jakubjezek001:
> NOTE: If it's about changing labels and adding conditional enum, then please change just labels and add conditional enums without changing attribute names, otherwise we would have to implement conversion of settings (PITA).

I agree that it might be PITA but this could be implemented with conversion.py settings overrides ([example here ayon-core\server\settings\conversion.py](https://github.com/ynput/ayon-core/blob/develop/server/settings/conversion.py)).

jakubjezek001:
Consider making the colorspace section toggleable. For more details, check out the `ColorspaceConfigurationModel` in our codebase [](https://github.com/ynput/ayon-nuke/blob/develop/server/settings/common.py). This approach would better suit situations where we need to clearly indicate that adding a value to a field will override an upstream value. By using a toggle, we can make it clear to the user that they are about to override an existing **colorspaceData**.

BigRoy:
> This approach would better suit situations where we need to clearly indicate that adding a value to a field will override an upstream value. By using a toggle, we can make it clear to the user that they are about to override an existing colorspaceData.

I'm not sure what you mean. It's not overriding the colorspace of the image - it's literally transcoding it from the source colorspace to the destination colorspace? Yes, it's hence also marking it as the destination colorspace? (But is there really a case where it shouldn't if it's transcoded?)

I may be missing your point.

jakubjezek001:
> (But is there really a case where it shouldn't if it's transcoded?)

Yes, since you can use addinional oiiotool args, you could also use preset for resizing or other reposition methods.

BigRoy:
> > (But is there really a case where it shouldn't if it's transcoded?)
> 
> Yes, since you can use addinional oiiotool args, you could also use preset for resizing or other reposition methods.

Ah yes, we have that actually. 

But still - then the logic that specificying no colorspace value means "do not color convert" is totally sensible and when specifying a colorspace value will transcode it to that and mark that as the resulting file's colorspace.

jakubjezek001:
> But still - then the logic that specificying no colorspace value means "do not color convert" is totally sensible and when specifying a colorspace value will transcode it to that and mark that as the resulting file's colorspace.

If I was a user, I would rather prefer the explicit way, but this is just a suggestion. You can of course keep it implicit and add it into documentation.

BigRoy:
> If I was a user, I would rather prefer the explicit way, but this is just a suggestion. You can of course keep it implicit and add it into documentation.

Explicit as in - it's a separate option in the enum?

Like:
- no color conversion
- convert to target colorspace
    - colorspace: str (required target colorspace; must be a valid value)
- apply display/view from host
- apply display/view transform
    - display: str (required target display; must be a valid value)
    - view: str (required target view; must be a valid value)

server/settings/publish_plugins.py
```diff
@@ -256,8 +256,8 @@ class ExtractThumbnailModel(BaseSettingsModel):
 
 def _extract_oiio_transcoding_type():
     return [
-        {"value": "colorspace", "label": "Use Colorspace"},
-        {"value": "display", "label": "Use Display&View"}
+        {"value": "use_colorspace", "label": "Use Colorspace"},
+        {"value": "use_display_view", "label": "Use Display&View"}
```
iLLiCiTiT, line 260:
```suggestion
        {"value": "colorspace", "label": "Use Colorspace"},
        {"value": "display_view", "label": "Use Display&View"}
```

server/settings/publish_plugins.py
```diff
@@ -276,13 +287,20 @@ class ExtractOIIOTranscodeOutputModel(BaseSettingsModel):
...
     extension: str = SettingsField("", title="Extension")
     transcoding_type: str = SettingsField(
-        "colorspace",
+        "use_colorspace",
         title="Transcoding type",
-        enum_resolver=_extract_oiio_transcoding_type
+        enum_resolver=_extract_oiio_transcoding_type,
+        conditionalEnum=True
+    )
+    use_colorspace: UseColorspaceModel = SettingsField(
```
iLLiCiTiT, line 295: You don't need to wrap single attribute under model.

server/settings/publish_plugins.py
```diff
@@ -276,13 +287,20 @@ class ExtractOIIOTranscodeOutputModel(BaseSettingsModel):
...
+        title="Use Colorspace",
+        default_factory=UseColorspaceModel
     )
-    colorspace: str = SettingsField("", title="Colorspace")
-    display: str = SettingsField("", title="Display")
-    view: str = SettingsField("", title="View")
+    use_display_view: UseDisplayViewModel = SettingsField(
```
iLLiCiTiT, line 299:
```suggestion
    display_view: UseDisplayViewModel = SettingsField(
```

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -122,12 +122,15 @@ def process(self, instance):
...
+                    #       or should we compute the target colorspace from current view and display ?
+                    target_colorspace = (output_def["use_colorspace"]["colorspace"] or
                                          colorspace_data.get("colorspace"))
-                else:
-                    view = output_def["view"] or colorspace_data.get("view")
-                    display = (output_def["display"] or
+                elif transcoding_type == "use_display_view":
+                    view = output_def["use_display_view"]["view"] or colorspace_data.get("view")
+                    display = (output_def["use_display_view"]["display"] or
                                colorspace_data.get("display"))
```
iLLiCiTiT, line 134:
```suggestion
                elif transcoding_type == "display_view":
                    display_view = output_def["display_view"]
                    view = display_view["view"] or colorspace_data.get("view")
                    display = (
                        display_view["display"]
                        or colorspace_data.get("display")
                    )
```

BigRoy:
What's needed to finalize this @MustafaJafar ?

MustafaJafar:
> What's needed to finalize this @MustafaJafar ?

This PR was created as a suggestion therefore it was marked as  a draft. 
if you believe this suggestion is a good one, Then I'll happily to mark it as `ready for review` and then finish the last TODO.

Also, this PR works on my side.

BigRoy:
I think the settings UI is much clearer _with_ the changes proposed here so I'd definitely get it ready for merge if you can.

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -122,13 +122,19 @@ def process(self, instance):
                 transcoding_type = output_def["transcoding_type"]
 
                 target_colorspace = view = display = None
+                # NOTE: we use colorspace_data as the fallback values for the target colorspace.
```
iLLiCiTiT, line 125:
```suggestion
                # NOTE: we use colorspace_data as the fallback values for
                #     the target colorspace.
```

iLLiCiTiT:
There should be implemented conversion of overrides for new settings structure.

MustafaJafar:
> There should be implemented conversion of overrides for new settings structure. Let me know if you don't know how to do it.

I'd appreciate your help. I don't know indeed how to implement the conversion.

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -122,13 +122,19 @@ def process(self, instance):
...
                 if transcoding_type == "colorspace":
+                    # TODO: Should we fallback to the colorspace (which used as source above) ?
+                    #       or should we compute the target colorspace from current view and display ?
```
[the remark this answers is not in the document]
MustafaJafar: https://github.com/ynput/ayon-core/pull/853/commits/3c5603c9241560b578b28d4b27ab8eb3da486085

client/ayon_core/plugins/publish/extract_color_transcode.py
MustafaJafar: https://github.com/ynput/ayon-core/pull/853/commits/3c5603c9241560b578b28d4b27ab8eb3da486085

MustafaJafar:
@iLLiCiTiT I managed to add the settings conversion. 

https://github.com/user-attachments/assets/c7ccfb65-5df3-41d7-af58-4e34a0c511e5

server/settings/conversion.py
```diff
@@ -71,10 +71,36 @@ def _convert_validate_version_0_3_3(publish_overrides):
     validate_version["plugin_state_profiles"] = [profile]
 
 
+def _convert_oiio_transcode_0_4_5(publish_overrides):
+    """ExtractOIIOTranscode plugin changed in 0.4.5."""
+    if "ExtractOIIOTranscode" not in publish_overrides:
+        return
+
+    transcode_profiles = publish_overrides["ExtractOIIOTranscode"]["profiles"]
```
iLLiCiTiT, line 79:
```suggestion
    transcode_profiles = publish_overrides["ExtractOIIOTranscode"].get("profiles")
    if not transcode_profiles:
        return
```

server/settings/conversion.py
```diff
@@ -71,10 +71,36 @@ def _convert_validate_version_0_3_3(publish_overrides):
...
+                continue
+
+            output["display_view"] = {}
+            if output["display"]:
+                output["display_view"].update({"display": output["display"]})
+                output.pop("display")
+            if output["view"]:
+                output["display_view"].update({"view": output["view"]})
+                output.pop("view")
+
```
iLLiCiTiT, line 99:
```suggestion
        outputs = profile.get("outputs")
        if outputs is None:
            break

        for output in outputs :
            # Already new settings
            if "display" not in output and "view" not in output:
                break

            # Fix 'display' -> 'display_view' in 'transcoding_type'
            transcode_type = output.get("transcoding_type")
            if transcode_type == "display":
                output["transcoding_type"] = "display_view"
            
            # Convert 'display' and 'view' to new values
            output["display_view"] = {
                "display":  output.pop("display", ""),
                "view": output.pop("view", ""),
            }

```

iLLiCiTiT: That's why condition `if "display" not in output and "view" not in output:` is there.

jakubjezek001: Lets merge the suggestion so this could be approved @MustafaJafar

MustafaJafar: Sorry one more question. 
should we `continue` instead of `break` ?

```suggestion
        outputs = profile.get("outputs")
        if outputs is None:
            return

        for output in outputs :
            # Already new settings
            if "display" not in output and "view" not in output:
                continue

            # Fix 'display' -> 'display_view' in 'transcoding_type'
            transcode_type = output.get("transcoding_type")
            if transcode_type == "display":
                output["transcoding_type"] = "display_view"
            
            # Convert 'display' and 'view' to new values
            output["display_view"] = {
                "display": output.pop("display", ""),
                "view": output.pop("view", ""),
            }

```

iLLiCiTiT: > should we continue instead of break ?

If one output does not have it, then none of them have it.

server/settings/conversion.py
```diff
@@ -71,10 +71,42 @@ def _convert_validate_version_0_3_3(publish_overrides):
...
+        return
+
+    for profile in transcode_profiles:
+        outputs = profile.get("outputs")
+        if outputs is None:
+            return
+
+        for output in outputs :
+            # Already new settings
+            if "display" not in output and "view" not in output:
```
iLLiCiTiT, line 90: I probably made wrong assumption, this was better.
```suggestion
            if "display_view" in output:
```

MustafaJafar:
@iLLiCiTiT @jakubjezek001 I think this PR is ready.

iLLiCiTiT (approved):
LGTM

robin-ynput (approved):
I gave it a try with previous version and adjusted UI settings, and I can confirm it works perfectly. The new UI makes way more sense indeed, so thanks for that :muscle: .

As mentioned, the only caveat I found is that I had to re-create my `Extract OIIO Transcode`settings when I uploaded the new addons.

Before:
oiio_before


After:
after_otio_transcode


That's probably out of the scope of this review, but I think this really needs to be communicated properly on roll-out (new major ? non backward compatible flag ?).

MustafaJafar:
> That's probably out of the scope of this review, but I think this really needs to be communicated properly on roll-out (new major ? non backward compatible flag ?).

Thank you for pointing that out. 
I gave it another test on my side and currently it copies the exact list + doing settings conversion. 

in this video some tests (sorry no voice) where I copy from older version to the current version and they all have the same result.
However, it won't work when copying settings from the current to the older version. 

https://github.com/user-attachments/assets/5dbaaf2d-608b-4861-822d-89bc0437431b

MustafaJafar:
Let's merge this PR and if there any issue, we can fix them later.

PR #854, opened 2024-08-23
Use proper aac argument for ExtractReview

kalisp:
## Changelog Description
Original `acc` doesnt work for audio

## Additional info
Solves error `Unknown encoder 'acc'` in ExtractReview

## Testing notes:
1. publish something which will have review with attached audio file (from Hiero, maybe AE)
2. ExtractReview shouldnt fail


BigRoy (approved):
Didn't test it - but change makes sense.

64qam (commented):
haven't tested, code looks good:-)

64qam (approved):

PR #871, opened 2024-09-05
fix running builder on start if scene is the default empty scene

MustafaJafar:
## Changelog Description
This is a follow PR for https://github.com/ynput/ayon-houdini/pull/74
to achieve test results mentioned in https://github.com/ynput/ayon-houdini/pull/74#issuecomment-2329590726

## Additional Info
I developed this and tested in Houdini addon.. 
and since this fix is in Core, then this also should work in other Addons since `host.get_current_workfile` always return `None` if scene is the default empty file.

## Testing notes:
_Basically, whenever an "empty scene" would initialize in Houdini - then use the template_
1. Open last workfile: True
a. Launch with NO existing workfiles -> Workfile template build ✅ 
b. Launch with existing workfiles -> NO workfile template build, because it opens the existing workfile. 🔴 
2. Open last workfile: False
a. Launch with NO existing workfiles -> Workfile template build ✅ 
b. Launch with existing workfiles -> Workfile template build ✅ 
When doing "New scene" (e.g. CTRL+N), run the Workfile Template Build ✅ 

**Create First Version: False**
Never create and save workfile after template build if first version doesn't exist.


fabiaserra:
Also related to this I found a bug in our template use case if the placeholders don't explicitly define a representation name (which is what you'd want if you have varied representations named such as `bgeo.sc`, `bgeo.gz`...) on this function https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/pipeline/workfile/workfile_template_builder.py#L1523

So I changed it to this:
```
        # Not return early if there's no representation name given as
        # that should be optional
        representation_names = None
        representation_name = placeholder.data["representation"]
        if representation_name:
            representation_names = [representation_name]
```
As you can see in our fork at https://github.com/fabiaserra/ayon-core/blob/release/alkemyx/client/ayon_core/pipeline/workfile/workfile_template_builder.py#L1523

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -533,29 +533,20 @@ def build_template(
         if create_first_version:
```
fabiaserra, line 533: Like I suggested [here](https://github.com/ynput/ayon-houdini/pull/74#issuecomment-2331975304), the code is more readable to me like this:
```python
def build_template(
    self,
    template_path=None,
    level_limit=None,
    keep_placeholders=None,
    create_first_version=None,
    workfile_creation_enabled=False
):
    # Get default values if not provided
    if template_path is None or keep_placeholders is None or create_first_version is None:
        template_path = template_path or preset["path"]
        keep_placeholders = keep_placeholders if keep_placeholders is not None else preset["keep_placeholder"]
        create_first_version = create_first_version if create_first_version is not None else preset["create_first_version"]

    created_version_workfile = self.create_first_workfile_version() if create_first_version else False

    # Handle workfile creation if enabled and a new version was created
    if workfile_creation_enabled and created_version_workfile:
        self.import_template(template_path)
        self.populate_scene_placeholders(level_limit, keep_placeholders)
        self.save_workfile(created_version_workfile)
        return

    # If workfile creation is not enabled or no new version was created,
    # proceed with template import and population
    if not workfile_creation_enabled:
        self.import_template(template_path)
        self.populate_scene_placeholders(level_limit, keep_placeholders)
```

fabiaserra: can you expand why? is it because otherwise the template import/populate runs when scene is not empty?

BigRoy: > Please note that your logic misses `self.host.get_current_workfile() is None`.. from my testing, it's essential.

I don't think that's foolproof currently.

I think some integrations _unfortunately_ may return `None` or even `"untitled.hip"` (houdini?) depending on what the software returns as 'current filename' - even in an empty scene.

**Also** I'd argue that the logic to decide that this runs only on a new EMPTY scene should potentially be outside of this? Like, allow the host integration to define whether it opened something or not and then decide whether to run the template build to begin with?

BigRoy:
> Also related to this I found a bug in our template use case if the placeholders don't explicitly define a representation name (which is what you'd want if you have varied representations named such as `bgeo.sc`, `bgeo.gz`...) on this function https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/pipeline/workfile/workfile_template_builder.py#L1523

@fabiaserra thanks - I've separated that into PR of its own since it's separate change/functionality. It's here: https://github.com/ynput/ayon-core/pull/872

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: > I think some integrations _unfortunately_ may return `None` or even `"untitled.hip"` (houdini?) depending on what the software returns as 'current filename' - even in an empty scene.

I checked Maya, Nuke and Houdini they all return None. 

> **Also** I'd argue that the logic to decide that this runs only on a new EMPTY scene should potentially be outside of this? Like, allow the host integration to define whether it opened something or not and then decide whether to run the template build to begin with?

tbh, I don't know.

--- 

tbh, figuring out how this feature should work is not easy for me. 
I may/might be adding my personal opinion in this PR but by looking at the available settings and reading through your discussions made me come up with the following user story: 
1. if a workfile has a matched template profile, that profile will be picked.
2. the workfile will be built on any empty scene no matter there are existing files or not. 
3. if there are no existing files and the option `create first version` is enabled, then the work file will be saved.

and, the current state of this PR at this commit https://github.com/ynput/ayon-core/pull/871/commits/e4881e2fec7599863e311fd9124db89bc35fd2b8 exactly does what I've just described. but I still think the code can be furtherly improved.

MustafaJafar: Good catch, 
so having this profile 
image
leads to this error and _no files are created._
```python
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.805/houdini/python3.9libs\hdefereval.py", line 155, in _processDeferred
    result = code(*args, **kwargs)
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\lib.py", line 1404, in start_workfile_template_builder
    build_workfile_template(workfile_creation_enabled=True)
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\workfile_template_builder.py", line 245, in build_workfile_template
    builder.build_template(*args, **kwargs)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\workfile\workfile_template_builder.py", line 517, in build_template
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\workfile_template_builder.py", line 83, in get_template_preset
    raise TemplateLoadFailed((
ayon_core.pipeline.workfile.workfile_template_builder.TemplateLoadFailed: Template path is not set.
Path need to be set in Houdini\Template Workfile Build Settings\Profiles
```

fabiaserra: that error makes sense to me if we assume that `Create first version` it's associated to the Workfile template build framework and not outside of it (which is the case given how the settings are shown). For it to not error out it should be part of a generic `Workfile` settings where workfiles (even if empty) are created on launch

MustafaJafar: > I've asked now says "yes, make sense for this to work to save a first version even without a template" so if it's easier to update the code now to allow that.

going quickly through the code, template builder always expect a template to be set.
as there many places that will/may error if template path is empty or None.  
e.g. `open_template`

https://github.com/ynput/ayon-core/blob/167ec7a3ee6bf0b239b9f4115f4997ea251e2fa1/client/ayon_core/pipeline/workfile/workfile_template_builder.py#L590-L594

BigRoy: > therefore, I think decoupling building the template from saving the workfile needs another PR..

Perfect.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -512,50 +512,37 @@ def build_template(
...
+                create_first_version = preset["create_first_version"]
+
+        # Create, open and return the first workfile version.
+        # This only works if there are no existent files and
+        #  create_first_version is enabled.
+        created_version_workfile = create_first_version and self.create_first_workfile_version()
+
+        # # Abort the process if workfile_creation_enabled.
+        # if workfile_creation_enabled:
+        #     return
```
MustafaJafar, line 531: So, I've to say. 
from my test runs this `workfile_creation_enabled` was useless.
Also, after good sleep and rereading the docstring and the old code. 
I can say that this variable is meant to abort the process just after workfile creation if enabled. I've no idea about the use case it solves. 

also, keep in mind that creating the workfile is dependent only on `create_first_version` and `os.path.exists(last_workfile_path)`.

MustafaJafar (commented):
Hey, 
The PR still works as before.
I've added few comments to make it clearer.
Also, stopped using some unnecessary flag.

For reference, I tried the same cases mentioned in the PR description + the buttons in AYON menu
image

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: also, when using ai to proofread the docstring, it tells me:
```
`workfile_creation_enabled` (bool): When set to True, this option initiates the creation of the initial version. However, it will skip the process if the version already exists.
```
which makes me think does using `os.path.exists(last_workfile_path)` replaces this flag.

BigRoy:
@fabiaserra So - I've changed some things. Can you check whether you understand the arguments and the function now?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,52 @@ def build_template(
...
-        # if first version is created, import template
-        # and populate placeholders
-        if (
-            create_first_version
-            and workfile_creation_enabled
-            and created_version_workfile
-        ):
+        # Build the template if current workfile is a new unsaved file
+        # (that's detected by checking if it returns any current filepath)
+        if not self.host.get_current_workfile():
```
BigRoy, line 534: Actually - this check is incorrect. We _should_ be allowed to build the template even in an existing scene - import an additional template and build it. I guess I did make a fault in my thinking. The `build_template` is NOT only triggered on launch of new workfiles, but it is ALSO a separate function in some menus...

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,61 @@ def build_template(
...
-            create_first_version (bool): create first version of a workfile
-            workfile_creation_enabled (bool): If True, it might create
-                                              first version but ignore
-                                              process if version is created
+            create_first_version (bool): Create first version of a workfile.
+                 When set to True, this option initiates the saving of the
+                 workfile for an initial version. It will skip saving if
+                 a version already exists.
+            workfile_creation_enabled (bool): Whether we are creating a new
+                workfile. If False, we assume we explicitly want to build the
```
fabiaserra, line 514: This is very confusing to me. Why would `workfile_creation_enabled == False` mean that we want to build the template in the current scene? I suggest you first explain the condition when the variable is True and then explain the alternative

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,61 @@ def build_template(
...
-        # but a version is already created
-        if workfile_creation_enabled:
+        if not create_first_version:
+            # Do not save whatsoever
             return
 
-        self.import_template(template_path)
-        self.populate_scene_placeholders(
-            level_limit, keep_placeholders)
+        # If there is no existing workfile, save the first version
```
fabiaserra, line 559: where are you checking if there is no existing workfile? just reading the function `get_first_workfile_version_to_create` that's not obvious

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,61 @@ def build_template(
...
-
-        # check if first version is created
-        created_version_workfile = False
-        if create_first_version:
-            created_version_workfile = self.create_first_workfile_version()
+                create_first_version: bool = preset["create_first_version"]
 
-        # if first version is created, import template
-        # and populate placeholders
+        # Build the template if current workfile is a new unsaved file
```
fabiaserra, line 534: this general docstring should cover all the possible cases of the conditionals within. If you read this and after you read the docstring that follows with `# If not a workfile creation, an explicit load template was requested, so we always want to build the template` your head explodes

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,61 @@ def build_template(
...
+        # (that's detected by checking if it returns any current filepath)
         if (
-            create_first_version
-            and workfile_creation_enabled
-            and created_version_workfile
+                # If not a workfile creation, an explicit load template
+                # was requested, so we always want to build the template
```
fabiaserra, line 538: I still don't understand why that would be the case?

fabiaserra (changes requested):

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -608,7 +614,7 @@ def import_template(self, template_path):
 
         pass
 
-    def create_first_workfile_version(self):
+    def get_first_workfile_version_to_create(self):
```
fabiaserra, line 617: How about `get_first_workfile_path`?

MustafaJafar (commented):
Test cases in PR description are fulfilled. 
Also, explicit build and update workfile from template work.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,61 @@ def build_template(
...
-            value is None
-            for value in [
-                template_path,
-                keep_placeholders,
-                create_first_version,
-            ]
+
+        # Get default values if not provided
+        if (
+                template_path is None
```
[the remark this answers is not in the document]
MustafaJafar: I guess. the code passed all linting validations
image

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: tbh, I find the variable name confusing. 
Furtherly, BigRoy's comment inside the if statement describes how we are using it. 

https://github.com/ynput/ayon-core/blob/0faaadab51b0e7b74c483a32e406f90b49f45636/client/ayon_core/pipeline/workfile/workfile_template_builder.py#L537-L538

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: This happens inside `get_first_workfile_version_to_create`
the way it works: 

If my task doesn't have any workfiles then this environment variable `AYON_LAST_WORKFILE` will always refer to the first version of the workfile. 
Then, we just use `os.path.exists(last_workfile_path)` to check if the file exists. 

The only caveat with the logic above is that the mentioned env variable is not updated on context change.
This may result in an expected behavior.. but currently, we are aware of this potential issue but didn't debug it nor having any reproducible steps yet.

image

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: Could you make a suggestion please?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: Here's the execution flow: 
 
No. 1 Explicit Build: 
when triggering this button
image

It runs `build_workfile_template()` where `workfile_creation_enabled` argument will fall to `False`.
which tells the builder: hey, don't create any workfiles. just perform a regular build.

No. 2 `on_new()` callback
when starting a new scene, we runs `start_workfile_template_builder()` 
which runs the build function with the argument set to true,
`build_workfile_template(workfile_creation_enabled=True)`  which tells the builder: hey it's fine to create a workfile. after performing the regular build

client/ayon_core/pipeline/workfile/workfile_template_builder.py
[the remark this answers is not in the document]
fabiaserra: it looks like there's an extra indent level

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: I understand the code but I'm talking about the meaning behind the variable name and what it's doing. If you had to guess as an user, disabling the "workfile creation" (workfile_creation_enabled == False) it would mean something like the workfile is not going to be created no?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: yeah I know how the code works but I was talking more about how readable the code and its corresponding docstring look. It's not too bad but to me a function that is called `get_first_workfile_version_to_create` and returns a variable "workfile_path" is hard to associate it to a boolean. I think it's the name of the function and its logic is confusing from an API point of view. But perhaps we just need to expand the docstring and say something like `# the function returns False if a workfile already exists`

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: Yeah but I think we should wait to resolve this discussion here https://github.com/ynput/ayon-core/pull/871#discussion_r1747851707 first

But I would expect it to be something along these lines if there's two conditionals:
```
Build the template if we are explicitly requesting it or it's an empty scene
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: With those use cases in mind, an API interface like this one would make more sense to me:
```
   def build_template(
        self,
        template_path=None,
        level_limit=None,
        keep_placeholders=None,
        create_first_version=False,
        force_build=False,
    ):
```
where `force_build` is set to True only on the number 1 use case

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: I always try to avoid boolean variable names that are negations, they are harder to read

fabiaserra: apparently I'm not the only one https://www.serendipidata.com/posts/naming-guidelines-for-boolean-variables

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: it would also help with the readability to move this part into its own variable:
```
is_new_scene = self.host.get_current_workfile() is None
```
so at the end it reads like:
```python
if is_new_scene or force_build:
```
much easier to read than negations

client/ayon_core/pipeline/workfile/workfile_template_builder.py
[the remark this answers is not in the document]
BigRoy: I've had this battle with @iLLiCiTiT sometimes - I have a feeling he'd say the same here. My IDE is ok with both, but defaults to this :/

BigRoy:
@fabiaserra and @MustafaJafar can you check whether you like the code style changes I just made? 

@fabiaserra thanks for the notes - they were very helpful. I totally agree that the arguments are super weird on this function. I'd personally even argue that the _new file_ toggle should be _outside_ of the method that calls `build_template`. I'd expect a function to `build_template` to build regardless, and to have integrations either consider manually whether it's a new scene OR to have a dedicated `build_template_for_new_file` method that can have the initial create first version and save check. Anyway, that'd require refactoring into the other integrations and I've left that out for now.

fabiaserra:
> @fabiaserra thanks for the notes - they were very helpful. I totally agree that the arguments are super weird on this function. I'd personally even argue that the _new file_ toggle should be _outside_ of the method that calls `build_template`. I'd expect a function to `build_template` to build regardless, and to have integrations either consider manually whether it's a new scene OR to have a dedicated `build_template_for_new_file` method that can have the initial create first version and save check. Anyway, that'd require refactoring into the other integrations and I've left that out for now.

100% agree, I was going to suggest this but I didn't want to be too critic and ask for too many changes

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,64 @@ def build_template(
...
+        explicit_build_requested = not workfile_creation_enabled
+        if is_new_file or explicit_build_requested:
+            self.log.info(f"Building the workfile template: {template_path}")
             self.import_template(template_path)
             self.populate_scene_placeholders(
                 level_limit, keep_placeholders)
 
-            # save workfile after template is populated
-            self.save_workfile(created_version_workfile)
+        if not workfile_creation_enabled:
```
fabiaserra, line 548:
```suggestion
        if explicit_build_requested:
```
?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,64 @@ def build_template(
...
+            # Do not consider saving a first workfile version, if this
```
fabiaserra, line 549: and then I'd probably rephrase this to something more like this:
```
# If we are explicitly requesting to build the template
# we assume it's only doing that and not saving a new workfile
# so we return early
```
* And any reason why you are adding the comment within the scope of the conditional and not before?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,64 @@ def build_template(
...
+            # is not set to be a "workfile creation". Then we assume
+            # only the template should get build.
+            return
 
-        # ignore process if first workfile is enabled
-        # but a version is already created
-        if workfile_creation_enabled:
+        if not create_first_version:
+            # Do not save whatsoever
```
fabiaserra, line 555: this is not really that helpful to me, I'd probably add the docstring before saying something like:
`# Return early if create first version is disabled`
Or... I would change it to encapsulate the create version code inside the conditional:
```python
# Create the first workfile version if requested
if create_version_version:
        workfile_path = self.get_workfile_path()
        if not os.path.exists(workfile_path):
            self.log.info("Saving first workfile: %s", workfile_path)
            self.save_workfile(workfile_path)
        else:
            self.log.info(
                "A workfile already exists. Skipping save of workfile as "
                "initial version.")
```

fabiaserra (changes requested):
Looking much better already @BigRoy , thank you!

client/ayon_core/pipeline/workfile/workfile_template_builder.py
BigRoy: I had thought of that - but I felt the "skipping of saving" felt WAY weirder if it'd skip it just because an explicit build was requested so I figured that keeping it as it was is actually clearer in this if statement. What do you think? Happy to change it though.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
BigRoy: > And any reason why you are adding the comment within the scope of the conditional and not before?

Merely because it felt related to that condition. I'm fine with it either way, no idea whether there's any style guide for this!

client/ayon_core/pipeline/workfile/workfile_template_builder.py
fabiaserra: not sure I follow, `not workfile_creation_enabled == explicit_build_requested` no? To me it's much easier to read conditionals without negation clauses so I personally find it better but up to you

fabiaserra: yeah I see what you mean... but isn't the other conditional the same then?

BigRoy: It's the same value as `explicit_build_requested` yes, but it just made more sense to "skip saving" `if not workfile_creation_enabled`. Anyway, I'll just refactor your suggestion because there's also an argument to make for the `not` in that if statement making it harder to read.

fabiaserra (approved):
lgtm!

client/ayon_core/pipeline/workfile/workfile_template_builder.py
MustafaJafar: what about `disable_workfile_creation` ?

BigRoy: No matter how we rename the argument or the methods - it requires updating the dependent addons that call this function. Hence, I'd make that a separate PR. And, if we're going down that route I'd really do what I described [here](https://github.com/ynput/ayon-core/pull/871#issuecomment-2338251853). Building a template should not need to care at all whether it's a new file or not - it should just load template and build regardless. If we do need certain logic related to "build_and_save_on_new_file" or whatever then we should separate that out. It's really polluting this function (and its arguments) which should just be:

```python
builder.build_template()
```

Also, we have "full control" in a DCC on deciding when to 'build on new file' so I'd even argue we don't need to check then whether current file is saved - but just respond to the _new file_ callback and then have it initialize accordingly. Which means this is just as sensible:

```python
def on_new():
    builder.build_template()
    core.save_first_file()
```

Especially because we also mentioned that it may also make sense to allow saving that "first workfile" even if not template was used but we just want to initialize the workfile name for the artist potentially, etc.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -853,11 +844,10 @@ def get_template_preset(self):
...
         if not path:
-            raise TemplateLoadFailed((
-                "Template path is not set.\n"
-                "Path need to be set in {}\\Template Workfile Build "
-                "Settings\\Profiles"
-            ).format(host_name.title()))
+            self.log.info(
+                "Template path is not set."
+            )
+            return
```
BigRoy, line 850: @MustafaJafar I believe this path of entry would still error - because then the function returns `None` so usage of `get_template_preset` call result in `build_template` will still error. I'll need to still patch that up.

BigRoy: Hmm - not sure what the "best" fallback behavior would be. With the current logic it just seems weird if it'd still continue to save since I'd need to check whether an actual filepath was returned, etc. I'll need to think about this some more :')

iLLiCiTiT:
Can someone shortly explain what should this PR solve (and put it into description)? Reading all the conversations did not give me much reliable information.

Does this try to solve case when you run template builder, but used "skip last workfile" in launcher?

MustafaJafar:
> Does this try to solve case when you run template builder, but used "skip last workfile" in launcher

short answer is yes. 

--- 

This PR adds support to run the workfile template builder on startup and allow to override template path resolution per DCC which is helpful for supporting native DCC keys/variables.

This PR was made as a follow up for https://github.com/ynput/ayon-houdini/pull/74 the Houdini PR only allows calling the template builder on startup but we still needs to update the builder in the core to make it work with empty new scenes.

Maybe @BigRoy can explain it better than me.

BigRoy:
> This PR adds support to run the workfile template builder on startup and allow to override template path resolution per DCC which is helpful for supporting native DCC keys/variables.
> 
> This PR was made as a follow up for [ynput/ayon-houdini#74](https://github.com/ynput/ayon-houdini/pull/74) the Houdini PR only allows calling the template builder on startup but we still needs to update the builder in the core to make it work with empty new scenes.

As far as I know the current PR is only cosmetic. Multiple people: me, Mustafa and Fabia were working on code that touched this but understood nothing of what `build_template()` call was trying to do.

Code logic seemed odd with method names doing more than they described, return types were off and just the whole logic of what `workfile_creation_enabled` is was completely misleading. This doesn't **refactor** the problem that the function itself is doing too much. Building the template is for whatever reason also responsible for detecting whether the current file is an unsaved new file or not.

iLLiCiTiT:
Ok, please put that into PR description. Because current description has no value.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -506,55 +506,59 @@ def build_template(
...
+            template_path: str = template_path or preset["path"]
+            if keep_placeholders is None:
+                keep_placeholders: bool = preset["keep_placeholder"]
+            if create_first_version is None:
+                create_first_version: bool = preset["create_first_version"]
+
+        # Build the template if we are explicitly requesting it or if it's
+        # an unsaved "new file".
+        is_new_file = not self.host.get_current_workfile()
+        explicit_build_requested = not workfile_creation_enabled
```
iLLiCiTiT, line 541: Creating negated variable does help how?

iLLiCiTiT: Ok, if that is the case, we should change the argument name too.

From what I've looked for it is used only in nuke, so I suggest to add `explicit_build_requested` to args, keep `workfile_creation_enabled` there but default value is `None`. If `workfile_creation_enabled` won't be `None` we will warn about deprecation and use the value for `explicit_build_requested`.

MustafaJafar (commented):
Well, I can't approve my own PR. 
but, it works as expected on my side. 
Also, I like the informative log.
Also, I think it's redundant to log `__ last_workfile_path` . Also, it'll print file path with version 1 if there is not workfile yet.

Here are some test runs: 

1. Launching Houdini with **no** existing workfiles. 
```
>>> [  Found template at: '\\storage\studio\templates\my_houdini_template.hip'  ] 
>>> [  Building the workfile template: \\storage\studio\templates\my_houdini_template.hip  ] 
>>> [  __ last_workfile_path: \\storage\work\ayon_projects\Robo\Assets\Character\robococo\work\cfx\robo_robococo_cfx_v001.hip  ] 
>>> [  Saving first workfile: \\storage\work\ayon_projects\Robo\Assets\Character\robococo\work\cfx\robo_robococo_cfx_v001.hip  ] 
```
2. Launching Houdini with existing workfiles. 
```

>>> [  Found template at: '\\storage\studio\templates\my_houdini_template.hip'  ] 
>>> [  Building the workfile template: \\storage\studio\templates\my_houdini_template.hip  ] 
>>> [  __ last_workfile_path: \\storage\work\ayon_projects\Robo\Assets\Character\robococo\work\cfx\robo_robococo_cfx_v001.hip  ] 
>>> [  A workfile already exists. Skipping save of workfile as initial version.  ] 
```
3. Launch Houdini while having empty workfile template in template builder settings.
```
//storage/work/ayon_projects/Robo/Assets/Character/robococo/work/cfx
  File "C:\PROGRA~1/SIDEEF~1/HOUDIN~1.805/houdini/python3.9libs\hdefereval.py", line 155, in _processDeferred
    result = code(*args, **kwargs)
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\lib.py", line 1404, in start_workfile_template_builder
    build_workfile_template(workfile_creation_enabled=True)
  File "E:\Ynput\ayon-houdini\client\ayon_houdini\api\workfile_template_builder.py", line 135, in build_workfile_template
    builder.build_template(*args, **kwargs)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\workfile\workfile_template_builder.py", line 531, in build_template
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\workfile\workfile_template_builder.py", line 847, in get_template_preset
    raise TemplateLoadFailed((
ayon_core.pipeline.workfile.workfile_template_builder.TemplateLoadFailed: Template path is not set.
Path need to be set in Houdini\Template Workfile Build Settings\Profiles
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: @MustafaJafar can you address this comment?

BigRoy: Honestly - if we're refactoring. I'd change the system around that `build_template` always builds, regardless and that we need OTHER logic handling the case that defines whether or not we want to trigger it in a new file, etc. It's just weird functionality for a `build_template` function?

iLLiCiTiT: That is change of functionality, not better argument names and more separated functionality.

Not saying "no", but we should maybe find out why it was added at first place, and why?

iLLiCiTiT: My point is that the method expects `workfile_creation_enabled` which is used only to convert it to different variable -> we should change the arguments of the method, with backwards compatibility, if the argument is wrong.

BigRoy: And my point is - that if we are renaming the argument now; we're wasting time, because if we do - we should refactor the full `build_template` to just always build... and have the decision made on "whether it's a new file and if it should be saved or not" to be made outside of it". So that calling `build_template` does just that, build it. I'd even argue that because it also **loads** a template that it's a bit ambiguous with the name.

Or at least clean it up so that it's doing something more understandable. :D

iLLiCiTiT: I don't think we'll get to refactor of this code in next few months. Unless it is done by community, it won't change for a long time.

If you insist on not doing it here, then move this line to first line of function and add comment above it like
```
# More accurate variable name
# - logic related to workfile creation should be moved out in future
```

jakubjezek001 (approved):
Tested in Nuke and works as expected

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: You've reacted with emoji, but nothing changed?

MustafaJafar: https://github.com/ynput/ayon-core/pull/871/commits/7298ddb745e003ca8e317eae3baccab246a2ad37

Issue #878, opened 2024-09-10
Publisher: Allow context promise


iLLiCiTiT:
## Description
Enhancement is tightly connected to editorial publishing where is common that folder and task entities do not exist yet, because they will be created during publishing. In that case it is not possible to fill `folderPath` and `task` to correct "future" values.

## How it could be implemented
Add new key to instance data that would mark instance as it's context should not be validated, something like `contextPromise`. These instances could not change value of `folderPath` and `task` in UI -> It is responsibility of create plugin to change it.

This will require to modify current logic of validation, which is a little bit more complicated.

PR #880, opened 2024-09-10
Create: Allow context promise for editorial workflow

iLLiCiTiT:
## Changelog Description
Create logic does allow to fill folder path and task name that are not validated in Publisher UI.

## Additional info
This change required change of context validation on instances. From now `CreatedIntance` does not have information about having valid context or not, it must be validated against `CreateContext`.

Added option to add `"has_promised_context"` to instance data. When set to `True` it does mark the instance as it's context is not validated in Publisher and cannot be modified by artist. Must be prefilled by create plugin, this does allow editorial create plugins to create instances before their target folder and task do exist.

### Note
I'm not sure if `"has_promised_context"` should be stored to instance data as it does not have to be stored to scene. Maybe using transient data would be more sufficient?

## Testing notes:
1. Create package and upload it to server.
2. Choose one of editorial hosts.
3. Modify create plugin that defines instances to set `"has_promised_context"` to `True`.
4. That should allow to set `"folderPath"` and `"task"` even if they do not exist.

Resolves https://github.com/ynput/ayon-core/issues/878

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -6,7 +6,8 @@ import traceback
 import collections
 import inspect
 from contextlib import contextmanager
-from typing import Optional
+import typing
+from typing import Optional, Iterable, Dict
 
 import pyblish.logic
 import pyblish.api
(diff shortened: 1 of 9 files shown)


BigRoy:
This would be _very lovely_ to get in for this PR https://github.com/ynput/ayon-resolve/pull/15 so I just wanted to "link" them by mentioning it here.

robin-ynput:
Confirmed it works with Resolve, there is a small typo in the PR description tho.
It's not `has_context_promise` but `has_promised_context` that requires to be added in the instance data.

robin-ynput (approved):

iLLiCiTiT:
So the only question is, should it be based on instance data or transient data?

robin-ynput:
> So the only question is, should it be based on instance data or transient data?

Just flagging this to @jakubjezek001 as Im not sure Im the best one to answer this one.
I haven't used `transient_data` that much, what is it designed for ?

iLLiCiTiT:
> I haven't used transient_data that much, what is it designed for ?

All instance data are (should be) stored to scene, but transient data are not -> less data in scene.

jakubjezek001:
> So the only question is, should it be based on instance data or transient data?

That is a very good question. I believe that better would be to add it to transient data. The fact is that we might need to have promise only at initial publishing session and following republishing times context on instances will need to be validated. It seems to me that if we keep this at the instance (scene stored) data, we would be skipping it every time, right?

PR #889, opened 2024-09-16
Attribute definitions: Positive argument names

iLLiCiTiT:
## Changelog Description
Changed arguments in attribute definitions from `disabled` to `enabled` and `hidden` to `visible`.

## Additional info
This is pure cosmetics PR. It is more natural to set `enabled=False` than `disabled=True` and the same for `visible` vs. `hidden`.

All changes are be backwards compatible.

## Testing notes:
1. Validate changes made in code.
2. If there is existing attribute definition using `hidden` or `disabled` it should still work the same way, but print deprecated message.


client/ayon_core/lib/attribute_definitions.py
```diff
@@ -90,6 +91,30 @@ def __call__(cls, *args, **kwargs):
         return obj
 
 
+def _convert_reversed_attr(
+    main_value, depr_value, main_label, depr_label, default
+):
+    if main_value is not None and depr_value is not None:
+        if main_value == depr_value:
+            print(
+                f"God invalid '{main_label}' and '{depr_label}' arguments."
```
BigRoy, line 100:
```suggestion
                f"Got invalid '{main_label}' and '{depr_label}' arguments."
```

BigRoy (commented):
I have my doubts about this PR. I understand the tendency with positive arguments, etc. but in this case I really have my doubts whether the 'transition period' to change it around makes this better... it'll be worse for some time.

`hidden=True` to me is just as clear as `visible=False`, similarly with disabled/enabled. Do we really care?

iLLiCiTiT:
> but in this case I really have my doubts whether the 'transition period' to change it around makes this better...

It's used, there is not way how to do it without transition period, unless you want to break some addons.

BigRoy (approved):
As said before, I have my doubts about the need for the API change itself - but it does work! ✅ So good to go.

Issue #895, opened 2024-09-16
Collect OTIO subset resources -  get_media_range_with_retimes returns negative range values


robin-ynput:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

There seems to be something wrong when computing the retimed values from an image sequence.
Issue was originally flagged by @jakubjezek001 on https://github.com/ynput/ayon-resolve/pull/15

### Expected Behavior:

Currently retiming 
```
otio.schema.Clip(name='a01vfxd_sh010_renderCompMain_v023.[1005-1056].exr', media_reference=ImageSequenceReference(target_url_base='C:\\projects\\AY01_VFX_demo\\shots\\sq001\\sh010\\publish\\render\\renderCompMain\\v023\\', name_prefix='a01vfxd_sh010_renderCompMain_v023.', name_suffix='.exr', start_frame=1005, frame_step=1, rate=23.97602397602398, frame_zero_padding=4, missing_frame_policy=<MissingFramePolicy.error: 0>, available_range=otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=86530, rate=23.976), duration=otio.opentime.RationalTime(value=52, rate=23.976)), available_image_bounds=None, metadata={}), source_range=otio.opentime.TimeRange(start_time=otio.opentime.RationalTime(value=86530, rate=23.976), duration=otio.opentime.RationalTime(value=52, rate=23.976)), metadata={'Resolve_OTIO': {}}),

handleStart: 10
handleEnd: 10
```

returns this representation:
```
{'frameStart': 991, 'frameEnd': -85478, 'stagingDir': 'C:\\projects\\AY01_VFX_demo\\shots\\sq001\\sh010\\publish\\render\\renderCompMain\\v023\\', 'name': 'exr', 'ext': 'exr', 'files': []}

Note that the frameEnd being negative prevents the image file sequence from being properly retimed.
```

I think it should rather be:
```
{'frameStart': 995, 'frameEnd': 1066, 'stagingDir': 'C:\\projects\\AY01_VFX_demo\\shots\\sq001\\sh010\\publish\\render\\renderCompMain\\v023\\', 'name': 'exr', 'ext': 'exr', 'files': [file_sequence here]}
```

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Download serialized clip file: [out.json](https://github.com/user-attachments/files/17019630/out.json)
2. Run the following
```python
import clique

[...]

PR #897, opened 2024-09-17
Fix wrong retimed detection on image sequence clip.

robin-ynput:
## Changelog Description

This PR addresses the bug described on https://github.com/ynput/ayon-core/issues/895 regarding retiming OTIO clip from image sequence.
* Need to conform source_range to available_range rate
* Retiming logic was when input media does not start 0 (image sequence, metadata...)
* Ensure embedded TC are treated correctly
* Need to preserve accurate discreet frame number when dealing with img sequences

### Environment prep
1. copy paste items form both **tool.poetry.dependencies** and **ayon.runtimeDependencies** sections from `ayon-core\client\pyproject.toml` into root level pyproject.toml under **tool.poetry.dev-dependencies**
2. remove `ayon-core\poetry.lock`
3. run `.\tools\manage.ps1 create-env` (windows)
> Your .poetry and .venv should be installed in root of repository

### Testing
1. Run the unit tests
run `.\tools\manage.ps1 run pytest .\tests\client\ayon_core\pipeline\editorial\` (windows)

[AY-979](https://app.clickup.com/t/6658547/AY-979)


client/ayon_core/pipeline/editorial.py
```diff
@@ -224,38 +242,30 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
-        _handle_start = handle_end
-        _handle_end = handle_start
-        handle_start = _handle_start
-        handle_end = _handle_end
+        handle_start, handle_end = handle_end, handle_start
 
     source_in = source_range.start_time.value
 
-    media_in_trimmed = (
-        media_in + source_in + offset_in)
```
robin-ynput, line 248: If I understood correctly the logic here, I think adding `media_in` + `source_in` was wrong but still gave false-positive cause most media available range starts from 0.

jakubjezek001: Media sources often start with metadata for the starting timecode, so they're not always beginning from `0`.

client/ayon_core/pipeline/editorial.py
```diff
@@ -224,38 +242,30 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
     handle_start *= abs(time_scalar)
     handle_end *= abs(time_scalar)
 
-    # filip handles if reversed speed
+    # flip handles if reversed speed
     if time_scalar < 0:
-        _handle_start = handle_end
-        _handle_end = handle_start
-        handle_start = _handle_start
-        handle_end = _handle_end
```
jakubjezek001, line 243: well this code is kinda fun :D

client/ayon_core/pipeline/editorial.py
```diff
@@ -179,6 +179,24 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
+        media_in = media_ref.start_frame
+        media_out += media_ref.start_frame
+        source_range_start = otio.opentime.RationalTime(
+            media_in + source_range.start_time.value,
+        )
+        source_range = otio.opentime.TimeRange(
+            start_time=source_range_start,
+            duration=source_range.duration
+        )
+
```
jakubjezek001, line 199: We need to test this with actual case data, as some image sequences may start at frame 1001, while the metadata on the frame could show `00:20:36:12`. What would be the outcome in that scenario?

client/ayon_core/pipeline/editorial.py
```diff
@@ -224,38 +242,30 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
-        handle_end = _handle_end
+        handle_start, handle_end = handle_end, handle_start
 
     source_in = source_range.start_time.value
 
-    media_in_trimmed = (
-        media_in + source_in + offset_in)
-    media_out_trimmed = (
-        media_in + source_in + (
+    media_in_trimmed = (source_in + offset_in)
```
iLLiCiTiT, line 259:
```suggestion
    media_in_trimmed = source_in + offset_in
```

client/ayon_core/pipeline/editorial.py
```diff
@@ -179,6 +179,24 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
     media_in = available_range.start_time.value
     media_out = available_range.end_time_inclusive().value
 
+    # Ensure image sequence media_ref source and
+    # available range are absolute.
+    media_ref = otio_clip.media_reference
+    if (
+        hasattr(otio.schema, "ImageSequenceReference") and
+        isinstance(media_ref, otio.schema.ImageSequenceReference) and
+        media_in != media_ref.start_frame
```
iLLiCiTiT, line 188:
```suggestion
        hasattr(otio.schema, "ImageSequenceReference")
        and isinstance(media_ref, otio.schema.ImageSequenceReference)
        and media_in != media_ref.start_frame
```

jakubjezek001:
Before the introduction of Resolve's native OTIO, we mainly relied on two attributes:

*   The clip's source start frame as a relative range to the available media range.
*   The clip's media range as an absolute range, including the metadata timecode value.
  
With Resolve's interpretation of OTIO now in play, we consider:

*   The clip's source start frame refers to a trimmed absolute range that is available.
*   The media range remains absolute, including metadata.

We reference the media source to an absolute value with included metadata because of FFMPEG. Since we use FFMPEG for video trimming, it's crucial to have the timecode correctly defined.

First, figure out if the clip source range is relative or absolute to the media range. Next, find the relative source frame value and continue as before, making sure to clear up any confusing code that's mistakenly referring to me (old me 😄).

client/ayon_core/pipeline/editorial.py
robin-ynput: I'm gonna give it a try !

antirotor:
> ... especially since `ayon-core` doesn't have a test folder yet.

but it can - I added one for my own tests, see [feature/OP-1188_better-representation-model](https://github.com/ynput/ayon-core/tree/feature/OP-1188_better-representation-model) branch.

client/ayon_core/pipeline/editorial.py
```diff
@@ -179,6 +179,24 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
+    # Ensure image sequence media_ref source and
+    # available range are absolute.
+    media_ref = otio_clip.media_reference
+    if (
+        hasattr(otio.schema, "ImageSequenceReference")
+        and isinstance(media_ref, otio.schema.ImageSequenceReference)
+        and media_in != media_ref.start_frame
+    ):
+        media_in = media_ref.start_frame
+        media_out += media_ref.start_frame
```
iLLiCiTiT, line 191: If we'll keep it, should this first reduce by `media_in` (I'm fully out of context BTW, never worked with OTIO).
```suggestion
        media_out = (media_out - media_in) + media_ref.start_frame + 1
```

client/ayon_core/pipeline/editorial.py
robin-ynput: Added a unit tests and logic with image sequence with embedded TC.

client/ayon_core/pipeline/editorial.py
```diff
@@ -179,6 +179,24 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
+        source_range_start = otio.opentime.RationalTime(
+            media_in + source_range.start_time.value,
+        )
+        source_range = otio.opentime.TimeRange(
+            start_time=source_range_start,
+            duration=source_range.duration
+        )
+
```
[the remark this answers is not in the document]
robin-ynput: I believe we kind of need to apply different logic cause the computed retimed_values might be used in 2 ways:
* trimming for media inputs through ffmpeg which from my tests ignores embedded metadata and assumes input start a 0
* file copy for image sequence inputs, and we need to preserve in-out and handles as discrete frame numbers in order to remap the range to the available file list

client/ayon_core/pipeline/editorial.py
```diff
@@ -224,55 +250,75 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
-        handle_end = _handle_end
+        offset_in, offset_out = offset_out, offset_in
+        handle_start, handle_end = handle_end, handle_start
 
-    source_in = source_range.start_time.value
+    # compute retimed range
+    media_in_trimmed = conformed_source_range.start_time.value + offset_in
+    media_out_trimmed = media_in_trimmed + (
+            (conformed_source_range.duration.value * abs(
+                time_scalar) + offset_out) - 1)
```
iLLiCiTiT, line 266:
```suggestion
    media_out_trimmed = media_in_trimmed + (
        (
            conformed_source_range.duration.value
            * abs(time_scalar)
            + offset_out
        ) - 1
    )
```

tests/client/ayon_core/pipeline/editorial/test_media_range_with_retimes.py
```diff
@@ -0,0 +1,149 @@
...
+        "img_seq_with_handles.json",
+        expected_data
+    )
+
+
+def test_img_sequence_with_embedded_tc_and_handles():
+    """
+    Img sequence clip (embedded timecode 1h)
+    available files = 1000-1100
+    source_range =  91046.625-91,120.625 25fps
```
jakubjezek001, line 137: This value hint is unclear to me, could you elaborate a bit on what does it refer to?

jakubjezek001:
The method `get_media_range_with_retimes` is used in [collect\_timeline\_instances.py](https://github.com/ynput/ayon-flame/blob/develop/client/ayon_flame/plugins/publish/collect_timeline_instances.py) on line 300 and also in [extract\_product\_resources.py](https://github.com/ynput/ayon-flame/blob/develop/client/ayon_flame/plugins/publish/extract_product_resources.py) on line 413.

  
I originally used this method to calculate how much material needs to be prepared for on-the-fly Flame transcoding. While looking into [flame\_export.py](https://github.com/ynput/ayon-flame/blob/develop/client/ayon_flame/otio/flame_export.py), which handles ocio clip data, I noticed that from line 256, the available frame range on `ImageSequenceReference` is set by frame number rather than Timecode. This confirms your assumption was correct.

When I wrote the Flame integration two years ago, I faced the challenge of using the embedded TC on the image sequence. From the code in `flame_export.py` on line 305, I can see that I was using the embedded TC to calculate the true source in/out for the clip, and later for calculating retime speed (though this equation might be incorrect).

client/ayon_core/pipeline/editorial.py
```diff
@@ -224,55 +250,75 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
+    # mediaIn/mediaOut have to correspond
+    # to frame numbers from source sequence.
+    media_ref = otio_clip.media_reference
+    is_input_sequence = (
+        hasattr(otio.schema, "ImageSequenceReference") and
+        isinstance(media_ref, otio.schema.ImageSequenceReference)
+    )
+
+    if is_input_sequence:
+        # preserve discreet frame numbers
```
robin-ynput, line 281: typo discrete

tests/client/ayon_core/pipeline/editorial/test_media_range_with_retimes.py
robin-ynput: Totally, so embedded timecode for the img sequence is 1h meaning 90000 frames at 25fps.
Absolute soure range being `91046.625 - 91120.625` that gives us relative source range `1046.625 - 1120.625`.

So in exprimed as seconds the img sequence clip source range is:
- start_time : `1046 frames / 25 fps` = `41,84` seconds
- duration : `1120.625 - 1046.625` = `74 frames / 25 fps` = `2.94` seconds

Now as recompute duration as frames with an available range at 24.0 fps:
- start_time = `41,84` seconds x `24.0` fps = `1004.16` frames
- duration = `2.94` seconds x `24.0fps` = `70.56` frames

Finally remapping those as discrete frame numbers to the sequence:
Untitled-2024-09-19-1305

- first frame displayed is frame at duration `1004.16` = frame number `1005`
- last frame displayed is frame at duration `1004.16 + 70.56 = 1074.72` = frame number `1075`

Hopes that clarifies it.

tests/client/ayon_core/pipeline/editorial/test_media_range_with_retimes.py
```diff
@@ -0,0 +1,149 @@
...
+        'speed': 2.5,
+        'versionData': {
+            'retime': True,
+            'speed': 2.5,
+            'timewarps': [],
+            'handleStart': 0,
+            'handleEnd': 25
+        }
+    }
+#    import rpdb ; rpdb.Rpdb().set_trace()
```
robin-ynput, line 88: todo: remove this

tests/client/ayon_core/pipeline/editorial/test_media_range_with_retimes.py
jakubjezek001: I see, so it seems to be just a typo: `91046.625-91,120.625 25fps` should be actually `91046.625-91120.625 25fps`. The inconsistency of coma had confused me. Thank you.

jakubjezek001: > first frame displayed is frame at duration 1004.16 = frame number 1005

The calculation makes sense for sure the only point we need to validate is if the rounding up even below **.5** will work.

jakubjezek001:
I was doing some tests within Hiero editorial and noticed some differences in the resulting calculations. Here is the outcome before and after the fix. Since it's an old publisher, I could only copy and paste some resulting logs.

[otio_clip_publish_data_hiero.zip](https://github.com/user-attachments/files/17071832/otio_clip_publish_data_hiero.zip)

tests/client/ayon_core/pipeline/editorial/test_media_range_with_retimes.py
robin-ynput: Oh missed that, thanks for this ! It's adjusted.
For the rounding work below .5, do you want another tests for that ? 
Confirm this is properly handled by otio:
image

robin-ynput:
> I was doing some tests within Hiero editorial and noticed some differences in the resulting calculations. Here is the outcome before and after the fix. Since it's an old publisher, I could only copy and paste some resulting logs.
> 
> [otio_clip_publish_data_hiero.zip](https://github.com/user-attachments/files/17071832/otio_clip_publish_data_hiero.zip)

Thanks for the additional tests in Hiero, I've investigated that a bit more. 
It looks like the issues comes from AYON native OTIO exporter which introduce a mismatch between the available and source ranges.

When I use Hiero native OTIO exporter, our clips are retimed as expected.
I've made an issue to fix this in `ayon-hiero`: https://github.com/ynput/ayon-hiero/issues/13

client/ayon_core/pipeline/editorial.py
```diff
@@ -224,55 +250,79 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
-        media_in + source_in + (
-            ((source_range.duration.value - 1) * abs(
-                time_scalar)) + offset_out))
+    # If media source is an image sequence, returned
+    # mediaIn/mediaOut have to correspond
+    # to frame numbers from source sequence.
+    media_ref = otio_clip.media_reference
+    is_input_sequence = (
+        hasattr(otio.schema, "ImageSequenceReference") and
+        isinstance(media_ref, otio.schema.ImageSequenceReference)
```
iLLiCiTiT, line 281:
```suggestion
        hasattr(otio.schema, "ImageSequenceReference")
        and isinstance(media_ref, otio.schema.ImageSequenceReference)
```

iLLiCiTiT:
I'm not sure what is state of this PR?

robin-ynput:
> For the testing data, we also need to check the retime effects (**LinearTimeWarp**, **FreezeFrame**, **TimeEffect**). Since the method name relates to retiming, this will cover the full range of functions. Great job on practically developing the unittest we really need! I'm thinking about whether we should add the test to the appropriate folder. @kalisp and @antirotor could lend a hand here, especially since `ayon-core` doesn't have a test folder yet.



> I'm not sure what is state of this PR?

I'm waiting on approval on this one.
I have addressed the feedback and added the tests, the Hiero problem that was reported will be tackled as its own in `ayon-hiero` (https://github.com/ynput/ayon-hiero/issues/13).

tools/manage.sh
```diff
@@ -157,6 +157,7 @@ default_help() {
   echo -e "  ${BWhite}ruff-check${RST}      ${BCyan}Run Ruff check for the repository${RST}"
   echo -e "  ${BWhite}ruff-fix${RST}        ${BCyan}Run Ruff fix for the repository${RST}"
   echo -e "  ${BWhite}codespell${RST}       ${BCyan}Run codespell check for the repository${RST}"
+  echo -e "  ${BWhite}run${RST}             ${BCyan}Run the repository${RST}"
```
iLLiCiTiT, line 160: ???
What "Run the repository" means?

jakubjezek001 (approved):
Tested via pytests and also via Resolve and works perfect!

tools/manage.sh
```diff
@@ -175,6 +176,12 @@ run_codespell () {
   "$POETRY_HOME/bin/poetry" run codespell
 }
 
+run_command () {
+  echo -e "${BIGreen}>>>${RST} Running ..."
+  shift;  # will remove first arg ("run") from the "$@"
+  "$POETRY_HOME/bin/poetry" run "$@"
```
iLLiCiTiT, line 182: So, we're replacing `run` with `run`?

robin-ynput: Technically yes, I I thought it made more sense to keep it explicit.
* first `run` is purely arbitrary from `manage.sh` sub command choices and can be changed in the future
* second `run` here is dictated by `poetry`

tools/manage.sh
robin-ynput: I stole that from the Windows implementation in `manage.ps1`. Basically it allow you to run any command line in the context of the repository (`peotry run` underneath).
Maybe something like `Run a command in the repository environement` would make more sense ?

iLLiCiTiT: I would say it is connected to https://github.com/ynput/ayon-core/pull/897/files#r1776786031 comment. The `run` is actually `poetry run` -> `Run a poetry command in the repository environement`?

robin-ynput: This is adjusted, thanks for that.

iLLiCiTiT (commented):
Code looks ok. @jakubjezek001 merge if is ready please.

iLLiCiTiT (approved):

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -99,48 +100,68 @@ def process(self, instance):
...
+
+            # Create handle offsets.
+            handle_start = otio.opentime.RationalTime(
+                handle_start,
+                rate=self.actual_fps,
+            )
+            handle_end = otio.opentime.RationalTime(
+                handle_end,
+                rate=self.actual_fps,
+            )
```
jakubjezek001, line 164: This seems to be breaking the backward compatibility in my Hiero testing scene:

It looks like the **handle_start** and **handle_end** (originally intager) is overriden here into `opentime.RT` type.

```
  File "C:\Users\jakub\AppData\Local\Ynput\AYON\dependency_packages\ayon_2408091537_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\CODE\__YNPUT\ayon-core\client\ayon_core\plugins\publish\extract_otio_review.py", line 158, in process
TypeError: float() argument must be a string or a real number, not 'opentimelineio._opentime.RationalTime'
```

jakubjezek001 (changes requested):
Tests in Hiero didn't go well.

client/ayon_core/plugins/publish/extract_otio_review.py
robin-ynput: Thanks a lot for the double check, I completely forgot about retesting the multi-review-clips case. Totally my bad, apologies for this.

I've fixed the problem and added 2 additional unit tests to enforce those.
Also confirmed that it worked OK on Hiero (Linux) and Resolve (Windows).

jakubjezek001 (approved):
Fantastic work @robin-ynput ! Thanks a lot! 💫 💥

PR #906, opened 2024-09-23
Fix delivering UDIMs using {udim} in delivery template

### LINKED ISSUE #905: AY-6775_Deliver version skips UDIM tiles

Client:
When I tried to Deliver Version of png textures with udim, only _1001 got into the deliver folder. There should be 1001 - 1005. It shows Delivery finished successfully but it is only at 20%.  Here is the Directory template {root[work]}/{project[name]}/_io_test/{yy}{mm}{dd}/ File name template {project[code]}_{asset}_{subset}_{@version}<_{output}><.{@frame}><_{udim}>.{ext}

Roy:
I think this has to do with the _ being before the frame instead of . t goes through logic here:
from here: 

The 'frames' pattern of clique requires the dot in front. See pattern in the docs 
https://clique.readthedocs.io/en/stable/api_reference/#clique.PATTERNS

Note - this may be tricky because technically we should find the 'frame' and 'udim' from the filename (if for whatever reason someone would have both; I have never seen that in production) so the easiest fix would be:

1. Fix the `clique` call to detect the sequence using `_` before the number?
2. Here https://github.com/ynput/ayon-core/blob/43b4be851ede5043729446de171d30b0fd717640/client/ayon_core/plugins/load/delivery.py#L265-L266 we'll need to put the value into udim formatting instead of frame depending on how the original file was formatted, etc.

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1286628098745438208)

_This issue was automatically created from Clickup ticket [AY-6775](https://app.clickup.com/t/86c0d1bfu)_

BigRoy:
## Changelog Description

Support the `{udim}` key in delivery template to deliver UDIM sequences.

## Additional info

Fixes #905 

## Testing notes:

1. Publish a UDIM sequence from e.g. Substance Painter
2. Run "Deliver Versions" action to deliver the UDIMs

Use e.g. delivery template:
```json
{
  "name": "udims",
  "directory": "{root[work]}/{project[name]}/io_test/{yy}{mm}{dd}",
  "file": "{project[code]}{asset}{subset}{@version}<{output}><.{@frame}><_{udim}>.{ext}"
}
```
image

### DIFF
--- a/client/ayon_core/lib/path_tools.py
+++ b/client/ayon_core/lib/path_tools.py
@@ -81,7 +81,10 @@ def collect_frames(files):
         dict: {'/folder/product_v001.0001.png': '0001', ....}
     """
 
-    patterns = [clique.PATTERNS["frames"]]
+    # clique.PATTERNS["frames"] supports only `.1001.exr` not `_1001.exr` so
+    # we use a customized pattern.
+    pattern = "[_.](?P<index>(?P<padding>0*)\\d+)\\.\\D+\\d?$"
+    patterns = [pattern]
     collections, remainder = clique.assemble(
         files, minimum_items=1, patterns=patterns)
 
--- a/client/ayon_core/plugins/load/delivery.py
+++ b/client/ayon_core/plugins/load/delivery.py
@@ -231,6 +231,11 @@ class DeliveryOptionsDialog(QtWidgets.QDialog):
                 self.log
             ]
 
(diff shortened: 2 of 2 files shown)


moonyuet (commented):
Looks good for me in terms of delivering UDIMs. 
image

moonyuet (approved):
Tested with delivering renders(non-udims) also working smoothly (just to ensure everything works)
image

PR #908, opened 2024-09-23
Fix: Not removing absolute path in Maya DL submission

kalisp:
## Changelog Description
This occured only in Maya (or DCC with AOV), caused pasting whole absolute path to metadata.json which could cause `Given file names contain full paths` (if ValidateExpectedFiles was disabled). 'files' is expecting only file names, not full paths.

## Testing notes:
1. disable `ValidateExpectedFiles`
2. publish from Maya
3. check metadata.json, shouldn't contain absolute paths, but only file names in `files`

### DIFF
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -788,6 +788,11 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
                 colorspace = product.colorspace
                 break
 
+        if isinstance(files, (list, tuple)):
+            files = [os.path.basename(f) for f in files]
+        else:
+            files = os.path.basename(files)
+
         rep = {
             "name": ext,
             "ext": ext,


BigRoy:
This also influences other DCCs and should also need confirmation that it works from other hosts, like Houdini and potentially others that submit to the farm?

Technically it's by [all of these](https://github.com/ynput/ayon-deadline/blob/f747671128e8da1e3553f3b09cc696a77c2172fe/client/ayon_deadline/plugins/publish/global/submit_publish_job.py#L87-L97). So we may want to confirm at least a few:

- [ ] Blender
- [ ] Celaction
- [ ] Fusion
- [ ] Harmony
- [ ] Houdini
- [ ] Max
- [ ] Maya
- [ ] Nuke
- [ ] Unreal

antirotor (approved):

antirotor:
> This also influences other DCCs and should also need confirmation that it works from other hosts, like Houdini and potentially others that submit to the farm?

We should test it, but on the other hand it is returning back the code that was there before and was removed by mistake (mine ☝🏻 )

kalisp:
> This also influences other DCCs and should also need confirmation that it works from other hosts, like Houdini and potentially others that submit to the farm?
> 
> Technically it's by [all of these](https://github.com/ynput/ayon-deadline/blob/f747671128e8da1e3553f3b09cc696a77c2172fe/client/ayon_deadline/plugins/publish/global/submit_publish_job.py#L87-L97). So we may want to confirm at least a few:
> 
> * [ ]  Blender
> * [ ]  Celaction
> * [ ]  Fusion
> * [ ]  Harmony
> * [ ]  Houdini
> * [ ]  Max
> * [ ]  Maya
> * [x]  Nuke
> * [ ]  Unreal
> 
> We should also confirm the publish jobs work.

It would be affecting only DCCs which have AOVs, which are not these on the list, imho.

BigRoy (approved):
Didn't test but this is basically the logic that it should be, yes.

LiborBatek (approved):
I have made full test round with proper rendering via DL and all good...enclosing also JSON scrnshot that no full paths got in (when ```Validate expected files``` been ```OFF``` )

Screenshot 2024-09-23 155914

LiborBatek:
Also works in ```blender``` host . Again full testing with rendering and publish job done. LGTM

LiborBatek:
so now also ```3dsmax``` been tested and all good too

PR #913, opened 2024-09-24
Fix: download referenced representations when synchronizing workfile via SiteSync addon

kalisp:
## Changelog Description
When workfile representation is being downloaded with SiteSync addon, it should also download all referenced, eg loaded representations if they are not present on target site.

## Additional information
No need to test it in Maya, links should work in any DCC.

## Testing notes:
1. enable SiteSync addon, set studio-studio in Site Settings
1. publish workfile with something loaded via Loader
2. this should create `reference` link between workfile and loaded representation (could be checked by Node Addon??)
3. switch Site Settings to `local` - `studio` (to actually engage syncing) and set there your root folder for your local site ( restart Tray)
4. go to Loader, find published workfile from step 1 and `Download` it (bottom right corner)
5. it should eventually download both workfile and referenced representation

### DIFF
--- a/client/ayon_core/plugins/publish/integrate_inputlinks.py
+++ b/client/ayon_core/plugins/publish/integrate_inputlinks.py
@@ -9,7 +9,14 @@ from ayon_api import (
 
 
 class IntegrateInputLinksAYON(pyblish.api.ContextPlugin):
-    """Connecting version level dependency links"""
+    """Connecting version level dependency links
+
+    Handles links:
+        - generative - what gets produced from workfile
+        - reference - what was loaded into workfile
+
+    It expects workfile instance is being published.
+    """
 
     order = pyblish.api.IntegratorOrder + 0.2
     label = "Connect Dependency InputLinks AYON"
@@ -47,6 +54,11 @@ class IntegrateInputLinksAYON(pyblish.api.ContextPlugin):
         self.create_links_on_server(context, new_links_by_type)
 
     def split_instances(self, context):
+        """Separates published instances into workfile and other
+
+        Returns:
+            (tuple(pyblish.plugin.Instance), list(pyblish.plugin.Instance))
+        """
         workfile_instance = None
         other_instances = []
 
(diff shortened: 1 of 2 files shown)


client/ayon_core/plugins/publish/integrate_inputlinks.py
```diff
@@ -83,6 +95,15 @@ def add_link(self, new_links_by_type, link_type, input_id, output_id):
     def create_workfile_links(
         self, workfile_instance, other_instances, new_links_by_type
     ):
+        """Adds links (generative and reference) for workfile.
+
+        Args:
+            workfile_instance (pyblish.plugin.Instance): published workfile
+            other_instances (list[pyblish.plugin.Instance]): other published
+                instances
+            new_links_by_type (dict[str, list[str]): dictionary collecting new
```
BigRoy, line 104:
```suggestion
            new_links_by_type (dict[str, list[str]]): dictionary collecting new
```

client/ayon_core/plugins/publish/integrate_inputlinks.py
```diff
@@ -9,7 +9,14 @@
...
 
 class IntegrateInputLinksAYON(pyblish.api.ContextPlugin):
-    """Connecting version level dependency links"""
+    """Connecting version level dependency links
+
+    Handles links:
+        - generative - what gets produced from workfile
+        - reference - what was loaded into workfile
+
+    It expects workfile instance is being published.
```
BigRoy, line 18: So - shouldn't this basically have `families = ["workfile"]` applied? (It won't log the warning message then though)

kalisp: Bit irrelevant to this PR. Not sure why Kuba set it up as ContextPlugin. Maybe because it should handle different links in the future, not depending on workfile instance? (IDK if we have/plan to have such.)

BigRoy: Note: `ContextPlugin` can also filter to `families`.

iLLiCiTiT (approved):
Changes look ok, needs testing of site sync addon part (in loader).

PR #920, opened 2024-09-26
improved OIIO subimages handling

ReeceMulley:
## Changelog Description
Ayon's OIIOtool handling currently removes all subimages (all AOVs that aren't RGBA). This PR does the following: 

- takes out the argument that restricts to 1 subimage
- modifies the colour conversion arguments to only occur on the first subimage (rgba). 
 
The results of the OIIOtool will be unchanged, unless the input files have >1 subimages and the OIIOtool argument "-a" is added (should this be documented somewhere?). Then, extra channels will pass through the transcode, and can have reformats etc applied if desired.

## Additional info
To have subimages identified, inputs must be a proper multipart EXR. In Nuke, this can be done by setting "interleave" to "channels" in the write node (very easy to define in the Ayon Nuke addon). Don't forget to set your channels to "all"!

## Testing notes:
1. Configure an Extract OIIO Transcode profile for a multipart exr (see additional info for more)
2. Add "-a" into the profile's OIIOtool arguments
3. Check that your extra AOVs have survived the transcode process

### DIFF
--- a/client/ayon_core/lib/transcoding.py
+++ b/client/ayon_core/lib/transcoding.py
@@ -1152,9 +1152,7 @@ def convert_colorspace(
         input_arg, input_path,
         # Tell oiiotool which channels should be put to top stack
         #   (and output)
-        "--ch", channels_arg,
-        # Use first subimage
-        "--subimage", "0"
+        "--ch", channels_arg
     ])
 
     if all([target_colorspace, view, display]):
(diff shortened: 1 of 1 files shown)


robin-ynput (approved):
Thanks a lot for your fix.

I can confirm this works great. I tested with the following inputs:
* regular EXR
* multipart EXR
* JPEG

Colorspace is correctly applied to RGBA.

jakubjezek001 (approved):
looking good, Thanks!

iLLiCiTiT (approved):

PR #924, opened 2024-09-27
Raise Publisher UI to front when already opened

BigRoy:
## Changelog Description

Make Publisher UI raised to the front when clicking `AYON > Create…` or `AYON > Publish...` in host integrations if it was already opened.

## Additional info

Especially useful for integrations that do not have Qt UI to parent to and the publisher UI sometimes going behind the DCC itself, etc.

## Testing notes:

1. Publisher UI should raise to the front when currently behind another window (or behind the DCC window)
2. Publisher UI should still work across different integrations.

### DIFF
--- a/client/ayon_core/tools/utils/host_tools.py
+++ b/client/ayon_core/tools/utils/host_tools.py
@@ -252,6 +252,9 @@ class HostToolsHelper:
             if tab:
                 window.set_current_tab(tab)
             window.make_sure_is_visible()
+            window.raise_()
+            window.activateWindow()
+            window.showNormal()
 
     def get_tool_by_name(self, tool_name, parent=None, *args, **kwargs):
         """Show tool by it's name.


client/ayon_core/tools/utils/host_tools.py
```diff
@@ -252,6 +252,9 @@ def show_publisher_tool(self, parent=None, controller=None, tab=None):
             if tab:
                 window.set_current_tab(tab)
             window.make_sure_is_visible()
+            window.raise_()
```
iLLiCiTiT, line 255: So instead of adding it to `make_sure_is_visible`, you've added it here?

BigRoy: It doesn't raise it currently without this - it just activates it. Yes, that raises it for hosts where it's parented to the UI - but it does not when it's not (and it's e.g. behind Resolve, Fusion, Blender, Cinema4D, etc.). At least, that didn't work for me. Maybe @LiborBatek can confirm behavior in Blender? Or @jakubjezek001 @robin-ynput in Resolve or Fusion.

Also, I didn't put it [in `make_sure_is_visible`](https://github.com/ynput/ayon-core/blob/66353ec8c4f0d81d29c7b1ef7fd997e07bf754ec/client/ayon_core/tools/publisher/window.py#L439-L444) just because it seemed like some specific 'workaround' for something with the Publisher UI itself - since none of the other tools had the same. Also, also the other host tools had it in this `host_tools.py` file like this, and I figured that was maybe for a reason because we maybe didn't want this to ALWAYS be the case for running the app and NOT calling it via e.g. host tools - so I just matched the rest.

iLLiCiTiT: Put it to `make_sure_is_visible`.

BigRoy: Done with https://github.com/ynput/ayon-core/pull/924/commits/d667f21625bdd043a3642fad5f36e7f0b6357ec0

iLLiCiTiT (approved):

PR #926, opened 2024-09-30
Chore: Remove 'HOST_WORKFILE_EXTENSIONS'

iLLiCiTiT:
## Changelog Description
Constant `HOST_WORKFILE_EXTENSIONS` is deprecated and unused for a long time. It is time to remove it.

## Additional info
Workfile extensions are defined by host addons by implementing `get_workfile_extensions`.

## Testing notes:
1. Validate `HOST_WORKFILE_EXTENSIONS` is not really used.
2. Validate changes.

### DIFF
--- a/client/ayon_core/pipeline/__init__.py
+++ b/client/ayon_core/pipeline/__init__.py
@@ -3,7 +3,6 @@ from .constants import (
     AVALON_INSTANCE_ID,
     AYON_CONTAINER_ID,
     AYON_INSTANCE_ID,
-    HOST_WORKFILE_EXTENSIONS,
 )
 
 from .anatomy import Anatomy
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
I can't get to the links, it says 404.

> I still think the "todo" listed in the code [here](https://github.com/ynput/ayon-core/blob/98a0f72f0363574f9141f175907b667b53c6065a/client/ayon_core/pipeline/constants.py#L8) is what we should aim for instead of removing this now.

The TODO is already implemented by `get_workfile_extensions` is host addons. This constant is not used for more than a year.

BigRoy:
> The TODO is already implemented by `get_workfile_extensions` in host addons (`IHostAddon` interface). This constant is not used for more than a year.

Ah I see - and those addons are always accessible even outside of the dccs (like we could iterate over the host addons). Great!
In that case I'll refactor those bits and pieces.

BigRoy (approved):

PR #936, opened 2024-10-03
Fix NTSC framerates floating issue comparison.

robin-ynput:
## Changelog Description

Fixes issue reported by @jakubjezek001 on Discord when source_range and available_range rates are technically the same rate but not exactly the same value in the OTIO exported file.

### Testing
1. Run the unit tests
run `.\tools\manage.ps1 run pytest .\tests\client\ayon_core\pipeline\editorial\` (windows)
run `.\tools\manage.sh run pytest .\tests\client\ayon_core\pipeline\editorial\` (linux)

### DIFF
--- a/client/ayon_core/pipeline/editorial.py
+++ b/client/ayon_core/pipeline/editorial.py
@@ -292,13 +292,23 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
     # Note that 24fps is slower than 25fps hence extended duration
     # to preserve media range
 
-    # Compute new source range based on available rate
-    conformed_src_in = source_range.start_time.rescaled_to(available_range_rate)
-    conformed_src_duration = source_range.duration.rescaled_to(available_range_rate)
-    conformed_source_range = otio.opentime.TimeRange(
-        start_time=conformed_src_in,
-        duration=conformed_src_duration
-    )
+    # Compute new source range based on available rate.
+    # NSTC compatibility might introduce floating rates, when these are
+    # not exactly the same (23.976 vs 23.976024627685547)
+    # this will cause precision issue in computation.
+    # Round to 2 decimals for comparison.
+    rounded_av_rate = round(available_range_rate, 2)
+    rounded_src_rate = round(source_range.start_time.rate, 2)
+    if rounded_av_rate != rounded_src_rate:
+        conformed_src_in = source_range.start_time.rescaled_to(available_range_rate)
+        conformed_src_duration = source_range.duration.rescaled_to(available_range_rate)
+        conformed_source_range = otio.opentime.TimeRange(
+            start_time=conformed_src_in,
+            duration=conformed_src_duration
+        )
+
+    else:
+        conformed_source_range = source_range
 
     # modifiers
     time_scalar = 1.
--- a/client/ayon_core/pipeline/editorial.py
+++ b/client/ayon_core/pipeline/editorial.py
@@ -293,10 +293,13 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
     # to preserve media range
 
     # Compute new source range based on available rate.
+
+    # Backward-compatibility for Hiero OTIO exporter.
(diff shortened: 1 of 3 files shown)


client/ayon_core/pipeline/editorial.py
```diff
@@ -292,13 +292,23 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
 
-    # Compute new source range based on available rate
-    conformed_src_in = source_range.start_time.rescaled_to(available_range_rate)
-    conformed_src_duration = source_range.duration.rescaled_to(available_range_rate)
-    conformed_source_range = otio.opentime.TimeRange(
-        start_time=conformed_src_in,
-        duration=conformed_src_duration
-    )
+    # Compute new source range based on available rate.
+    # NSTC compatibility might introduce floating rates, when these are
```
BigRoy, line 296: NTSC I think?
```suggestion
    # NTSC compatibility might introduce floating rates, when these are
```

client/ayon_core/pipeline/editorial.py
```diff
@@ -292,13 +292,23 @@ def get_media_range_with_retimes(otio_clip, handle_start, handle_end):
...
+    # Round to 2 decimals for comparison.
+    rounded_av_rate = round(available_range_rate, 2)
+    rounded_src_rate = round(source_range.start_time.rate, 2)
+    if rounded_av_rate != rounded_src_rate:
+        conformed_src_in = source_range.start_time.rescaled_to(available_range_rate)
+        conformed_src_duration = source_range.duration.rescaled_to(available_range_rate)
+        conformed_source_range = otio.opentime.TimeRange(
+            start_time=conformed_src_in,
+            duration=conformed_src_duration
+        )
```
BigRoy, line 308: Out of curiosity - what problems would we have if we ALWAYS do this, instead of only when the rounded values differ?

client/ayon_core/pipeline/editorial.py
robin-ynput: Good spot, thanks for that !

client/ayon_core/pipeline/editorial.py
robin-ynput: That's a good question, and if you look at the diff that what's was happening before (always rescale to available_range rate). 

The issue this PR fixes is that sometimes we process OTIO input with consistent clips rate, but with different rounding precision e.g.
* source_rate = 23.976024627685547
* available_rate =  23.976

This is spotted as 2 different rates and forces the rescale operation which introduces a slight shift in the duration : `1s / 23.976 fps == 1.000001027s/23.976024627685547 fps`.
And this can create a drop when remapped to frame number (frame number are discrete while duration in seconds in continuous). 

This logic is fine and the result is mathematically correct. The problem we got here is that we needs to differentiate a valid frame rate scale (e.g. `25.0fps vs 24.0fps`) from a rounding mismatch (`24.0fps with NTSC fps` as `23.976024627685547 vs 23.976`)

Hope that clarify

jakubjezek001: It is assumed that this is mainly to maintain backward compatibility of the AYON Hiero OTIO exporter. It is also assumed that Hiero/Resolve native OTIO maintains consistent floating point values. It might be beneficial to comment on this as a backward compatibility fix.

jakubjezek001 (approved):
Tested in Hiero and works perfect! Thanks

client/ayon_core/pipeline/editorial.py
BigRoy: Agree - let's clarify the comment with that note - and then feel free to resolve this conversation. I'm always very wary of these types of "workarounds" because they seem to always just over time pop up new issues down the line of something we didn't anticipate either hehe.

robin-ynput: @BigRoy the actual proper fix for this one will likely be:
https://github.com/ynput/ayon-hiero/issues/13

Issue #937, opened 2024-10-03
Publish Validation Errors show traceback in artist facing reports


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Since #877 it seems that all validation errors now include the message twice, once with the full traceback:

![logstraceback3](https://github.com/user-attachments/assets/d407ca00-fbe8-4be0-9ff4-7cf938ec2a33)
![logstraceback2](https://github.com/user-attachments/assets/9b2a4d1b-b53e-408f-9324-7c77b569ae89)
![logstraceback1](https://github.com/user-attachments/assets/99400f94-eb7a-4a65-8776-6f958aae4105)
![logstraceback4](https://github.com/user-attachments/assets/dfbda2cc-1359-4dc0-bb50-55b355972fc3)

### Expected Behavior:

It should only include the nice messages for the artists

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Open publisher, get any publish invalidation report.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

PR #938, opened 2024-10-03
Publisher: Remove pyblish exception log from records

iLLiCiTiT:
## Changelog Description
Remove exception from log records.

## Additional info
Pyblish process method is logging exception which is added to records of our custom message handler.

## Testing notes:
1. Validation error does not add error log to logs with traceback.

Resolves https://github.com/ynput/ayon-core/issues/937

### DIFF
--- a/client/ayon_core/tools/publisher/models/publish.py
+++ b/client/ayon_core/tools/publisher/models/publish.py
@@ -32,17 +32,20 @@ PLUGIN_ORDER_OFFSET = 0.5
 class MessageHandler(logging.Handler):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
This does fix the issue:

image

image

But I'd consider it a hotfix really. We would really want to preferably 'catch' this differently if we can because it feels relatively hacky to just pick the last log entry and discard it like this. So maybe a `# TODO` alongside this in the code would be nice so we can keep it in mind for if we can think of better solutions.

iLLiCiTiT:
> But I'd consider it a hotfix really. We would really want to preferably 'catch' this differently if we can because it feels relatively hacky to just pick the last log entry and discard it like this.

Issue #961, opened 2024-10-22
Add descriptions to addon settings entries for better clarity


LiborBatek:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

We miss ```description``` entry for most of the addon's settings which can cause noise and worse UX.

### How would you imagine the implementation of the feature?

Just by adding description entry as a hint / hoover over tip will tremendously help user and his experience with the system and its settings.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

BigRoy:
@LiborBatek please point out which are still lacking the most. Otherwise, let's close this. Someone will need to chase which ones are lacking - so we can improve it. It's a bit much to maybe say that everything needs a description?

LiborBatek:
I would say mostly `ayon-core` related settings which arent having any sort of description in the header/on hoover over...

some examples of candidates for improvement:

1. `ayon+settings://core/disk_mapping`
2. `ayon+settings://core/version_start_category`
3. `ayon+settings://core/imageio/file_rules`
4. `ayon+settings://core/publish/CollectAnatomyInstanceData`
5. `ayon+settings://core/publish/ExtractUSDAssetContribution`
6. `ayon+settings://core/publish/ExtractUSDAssetContribution`
7. `ayon+settings://core/publish/ExtractReview`
8. `ayon+settings://core/publish/ExtractBurnin`
9. `ayon+settings://core/publish/IntegrateHeroVersion`
10. `ayon+settings://core/publish/CleanUp`
11. `ayon+settings://core/project_plugins`
12. `ayon+settings://core/filter_env_profiles`


I do agree not every single category needs description but the major ones (and usually not super clear ones yes)...it would be ok to go just with those above or some of them at least:)

PR #965, opened 2024-10-22
added descriptions to addons settings

LiborBatek:
## Changelog Description

Added descriptions to the settings categories in the addon for better communicating the purpose.

Resolves [https://github.com/ynput/ayon-core/issues/961](url)
## Testing notes:

1. build addon package and upload to the server / set it to bundle
2. inspect addons settings and newly added descriptions if correct/ clear

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -171,7 +171,8 @@ class CoreImageIOConfigProfilesModel(BaseSettingsModel):
 class CoreImageIOBaseModel(BaseSettingsModel):
     activate_global_color_management: bool = SettingsField(
         False,
-        title="Enable Color Management"
+        title="Enable Color Management",
+        description="When enabled, global color management been activated for all hosts available.",
     )
     ocio_config_profiles: list[CoreImageIOConfigProfilesModel] = SettingsField(
         default_factory=list, title="OCIO config profiles"
@@ -276,31 +277,38 @@ class CoreSettings(BaseSettingsModel):
     disk_mapping: DiskMappingModel = SettingsField(
         default_factory=DiskMappingModel,
         title="Disk mapping",
+        description="Map virtual disk(s) on artist's machine for AYON when running",
     )
     tools: GlobalToolsModel = SettingsField(
         default_factory=GlobalToolsModel,
-        title="Tools"
+        title="Tools",
+        description="Configure basic aspects for workfiles, products, and publishing",
     )
     version_start_category: VersionStartCategoryModel = SettingsField(
         default_factory=VersionStartCategoryModel,
-        title="Version start"
+        title="Version start",
+        description="Configure versioning for products using profiles",
     )
     imageio: CoreImageIOBaseModel = SettingsField(
         default_factory=CoreImageIOBaseModel,
-        title="Color Management (ImageIO)"
+        title="Color Management (ImageIO)",
+        description="Enable global color management and configure OCIO",
     )
     publish: PublishPuginsModel = SettingsField(
         default_factory=PublishPuginsModel,
-        title="Publish plugins"
+        title="Publish plugins",
+        description="Configure various aspects of publishing process and components",
     )
     project_plugins: MultiplatformPathListModel = SettingsField(
         default_factory=MultiplatformPathListModel,
         title="Additional Project Plugin Paths",
+        description="(DEPRECATED!) Configure file paths to arbitrary publish plugins",
     )
     project_folder_structure: str = SettingsField(
         "{}",
         widget="textarea",
         title="Project folder structure",
+        description="Lists non-Ayon folders to be created via web action",
         section="---"
     )
     project_environments: str = SettingsField(
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -287,7 +287,7 @@ class CoreSettings(BaseSettingsModel):
     version_start_category: VersionStartCategoryModel = SettingsField(
         default_factory=VersionStartCategoryModel,
         title="Version start",
-        description="Configure versioning for products using profiles",
+        description="Configure the first version to use for new products using profiles",
     )
     imageio: CoreImageIOBaseModel = SettingsField(
         default_factory=CoreImageIOBaseModel,
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -172,7 +172,7 @@ class CoreImageIOBaseModel(BaseSettingsModel):
     activate_global_color_management: bool = SettingsField(
         False,
         title="Enable Color Management",
-        description="When enabled, global color management been activated for all hosts available.",
+        description="When enabled, global color management using OpenColorIO will be enabled for all hosts available. With this enabled the `OCIO` environment variable will automatically be supplied to any OCIO supporting host.",
     )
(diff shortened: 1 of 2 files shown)


server/settings/main.py
```diff
@@ -171,7 +171,8 @@ class CoreImageIOConfigProfilesModel(BaseSettingsModel):
 class CoreImageIOBaseModel(BaseSettingsModel):
     activate_global_color_management: bool = SettingsField(
         False,
-        title="Enable Color Management"
+        title="Enable Color Management",
+        description="When enabled, global color management been activated for all hosts available.",
```
BigRoy, line 175: Personally I'd prefer it more if the description tells me more than the title. That it enables color management is already quite clear. The description should explain what that means as a result? What changes? What does if have effect on? When should I enable/disable it? For complex things it should likely even refer to the documentation for more details. 
```suggestion
        description="When enabled, global color management using OpenColorIO will be enabled for all hosts available. With this enabled the `OCIO` environment variable will automatically be supplied to any OCIO supporting host.",
```

E.g. that, but then adhering to line length.

server/settings/main.py
```diff
@@ -276,31 +277,38 @@ class CoreSettings(BaseSettingsModel):
     disk_mapping: DiskMappingModel = SettingsField(
         default_factory=DiskMappingModel,
         title="Disk mapping",
+        description="Map virtual disk(s) on artist's machine for AYON when running",
```
BigRoy, line 280: This doesn't map _virtual_ disks, right? It maps / mounts a folder to another path. Virtual disks to me sound like a very different thing.
```suggestion
        description="Map a (network) drive to artist's machine for AYON when running",
```

See:
- https://support.microsoft.com/en-us/windows/map-a-network-drive-in-windows-29ce55d1-34e3-a7e2-4801-131475f9557d
- https://askubuntu.com/questions/46183/how-to-map-a-network-drive

But maybe I'm misunderstanding the feature. Also here, it'd be great to have in the description when to use it.

server/settings/main.py
```diff
@@ -276,31 +277,38 @@ class CoreSettings(BaseSettingsModel):
...
         default_factory=GlobalToolsModel,
-        title="Tools"
+        title="Tools",
+        description="Configure basic aspects for workfiles, products, and publishing",
     )
     version_start_category: VersionStartCategoryModel = SettingsField(
         default_factory=VersionStartCategoryModel,
-        title="Version start"
+        title="Version start",
+        description="Configure versioning for products using profiles",
```
BigRoy, line 290:
```suggestion
        description="Configure the first version to use for new products using profiles",
```

server/settings/main.py
```diff
@@ -276,31 +277,38 @@ class CoreSettings(BaseSettingsModel):
...
         default_factory=CoreImageIOBaseModel,
-        title="Color Management (ImageIO)"
+        title="Color Management (ImageIO)",
+        description="Enable global color management and configure OCIO",
     )
     publish: PublishPuginsModel = SettingsField(
         default_factory=PublishPuginsModel,
-        title="Publish plugins"
+        title="Publish plugins",
+        description="Configure various aspects of publishing process and components",
```
BigRoy, line 300: Way too generic? Or? Does such a description really help?

server/settings/tools.py
```diff
@@ -162,7 +162,7 @@ class AYONMenuModel(BaseSettingsModel):
     version_up_current_workfile: bool = SettingsField(
         False,
         title="Version Up Workfile",
-        description="Add 'Version Up Workfile' to AYON menu"
+        description="Adds tool for 'incremental save' to the AYON menu"
```
BigRoy, line 165: It's not really a "tool" because it doesn't really open any dialog?

```suggestion
        description="Adds menu item for 'incremental save' to the AYON menu"
```

BigRoy (changes requested):
Some comments added. However, a lot of these to me feel like bloat instead of explaining more.

There is a section "Workfiles" and the description says "Configure workfiles and related tools and its behaviour". To me, that's not providing me more information - it's just giving me more to read whilst telling me the same thing... so there's something about Workfiles.

It could explain e.g. "Configure how workfiles are saved, whether workfiles or workfiles tool open on launch of a host, whether to create additional folders automatically in the work area." Making it more descriptive easily shows that it's just each entry BELOW it that should get a description instead.  Because each of those _is_ more descriptive than its title.

server/settings/main.py
LiborBatek: I do get your points and have to agree, but it should just fit into single line of text to fit in the UI (namely category headers) 

I can think of hoover over tips could be more descriptive for sure (using longer more detailed info).

Thanks for more indepth description! Im not expert on these (ayon-core been a trickier addon to tackle for me)

server/settings/main.py
LiborBatek: I had to refer to our user doc for this one (as I wasnt so sure) , so it must be wrong too...

server/settings/main.py
LiborBatek: Again, I get your point, but how you approach a title (and short line of text) when the category has so many items...it must be sort of generic. To me, it tells user what he can expect there - Publish plugins just tells : some list of addons present, but it actually tells nothing about the publishing itself can be tweaked too...works for me (but could be better formated probably as anything else imho)

LiborBatek:
@BigRoy  could you also help me with this one? ```Filter Env Profiles``` at the very bottom of the addons setting...

Could you formulate it as a short description so we can add it there? pls?

server/settings/main.py
LiborBatek: I can think of e.g. such description instead: ```Configure the way publishing performs and components it produces```

?

server/settings/main.py
LiborBatek: I can think of  ```By enabling, 'OCIO' environment will be used globally for all supported host apps```

server/settings/main.py
BigRoy: > Again, I get your point, but how you approach a title (and short line of text) when the category has so many items...it must be sort of generic. To me, it tells user what he can expect there - Publish plugins just tells : some list of addons present, but it actually tells nothing about the publishing itself can be tweaked too...works for me (but could be better formated probably as anything else imho)

Id'd argue if it's too generic in the description - it likely means we should focus on adding description of the stuff inside of it instead of just describing "there's publish plugins stuff here" on a "Publish Plugins" section :D

BigRoy:
> @BigRoy could you also help me with this one? `Filter Env Profiles` at the very bottom of the addons setting...
> 
> Could you formulate it as a short description so we can add it there? pls?

I'd love to - but unfortunately I have no idea what that setting does. Maybe @antirotor or @kalisp ?

LiborBatek:
> I'd love to - but unfortunately I have no idea what that setting does. Maybe @antirotor or @kalisp ?

...so you are human after all! :)

server/settings/main.py
LiborBatek: will change it accordingly

LiborBatek:
Seems there is no need for such a thing bloating the UI anyway....so closing, thx for your time investment,

PR #967, opened 2024-10-23
Create Context: Helper function to get entities

iLLiCiTiT:
## Changelog Description
Create context has helper functions to collect and cache entities.

## Additional info
The need came with https://github.com/ynput/ayon-deadline/pull/49 . That PR is fetching entities on each instance change which slows down the whole logic a lot. This way the entities are cached on CreateContext which avoids unnecessary re-fetching of the same data. The PR needed specifically task type, but I guess there will come soon more that will want to use attributes etc.

The methods are using folder path and task name for filtering as that is what instances have available in data.

The same methods are not used for context validation and to get current context entities.

## Testing notes:
1. All should be ok when publisher UI is opened in any host.
2. Validation of context should still happen without any issues.
3. Invalid context on existing instance should be properly displayed as invalid.
4. It is possible to use new helper methods to get corresponding entities.

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -1481,9 +1481,8 @@ class CreateContext:
         }
         remainder_paths = set()
         for folder_path in output:
-            # Skip invalid folder paths (e.g. if only folder name
-            #   is passed in)
-            if "/" not in folder_path:
+            # Skip invalid folder paths (folder name or empty path)
+            if not folder_path or "/" not in folder_path:
                 continue
 
             if folder_path not in self._folder_entities_by_path:
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -1584,24 +1584,24 @@ class CreateContext:
         folder_entities_by_path = self.get_folder_entities(
             missing_folder_paths
         )
-        folder_ids = set()
+        folder_path_by_id = {}
         for folder_path, folder_entity in folder_entities_by_path.items():
             if folder_entity is not None:
-                folder_ids.add(folder_entity["id"])
+                folder_path_by_id[folder_entity["id"]] = folder_path
 
-        if not folder_ids:
+        if not folder_path_by_id:
             return output
 
         task_entities_by_parent_id = collections.defaultdict(list)
         for task_entity in ayon_api.get_tasks(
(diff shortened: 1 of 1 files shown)


client/ayon_core/pipeline/create/context.py
```diff
@@ -1468,6 +1467,267 @@ def execute_autocreators(self):
...
+            folder_paths (Iterable[str]): Folder paths.
+
+        Returns:
+            Dict[str, Optional[Dict[str, Any]]]: Folder entities by path.
+
+        """
+        output = {
+            folder_path: None
+            for folder_path in folder_paths
+            if folder_path is not None
```
BigRoy, line 1483: According to the input type hint this should never be 'allowed' as value in the Iterable?

client/ayon_core/pipeline/create/context.py
```diff
@@ -1468,6 +1467,267 @@ def execute_autocreators(self):
...
+        }
+        remainder_paths = set()
+        for folder_path in output:
+            # Skip empty/invalid folder paths
+            if folder_path is None or "/" not in folder_path:
```
BigRoy, line 1488: folder path is never None with the filtering happenign above?

client/ayon_core/pipeline/create/context.py
```diff
@@ -1468,6 +1467,267 @@ def execute_autocreators(self):
...
+                continue
+
+            folder_entity = self._folder_entities_by_id.get(folder_id)
+            if folder_entity:
+                output[folder_path] = folder_entity
+            else:
+                remainder_paths.add(folder_path)
+
+        if not remainder_paths:
+            return output
```
BigRoy, line 1507: Looking at all of the other logic and queries done here - why do we need the folder_entity_by_id? Why can't just store `folder_entity_by_folder_path`?

That would basically make this logic:
```
output = {}
not_cached = set()
for folder_path in folder_paths:
    if folder_path in self._folder_entity_by_folder_path:
        output[folder_path] = self._folder_entity_by_folder_path[folder_path]
    else:
        not_cached.add(folder_path)
```
Right?

iLLiCiTiT: Very good point, simplified.

kalisp (approved):
Tried `get_task_entity`, worked. Tried all other new methods, they work (or wont fail).

Issue #804, opened 2024-07-29
Make `PublishAttributeValues` behave like `dict` when iterating


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

For whatever reason it seems that iterating a `PublishAttributeValues` instance is possible but does not behave like a standard `dict` which makes it confusing.

See:

I'm not sure why but this object when iterating over it yields e.g. 0 etc.
https://github.com/ynput/ayon-core/blob/05f6b0f2ce000910004c3ce9913e17ec528bc04b/client/ayon_core/pipeline/create/context.py#L670

I'd expect that to be non-iterable OR iterate like a dict over the keys by default? Error was faced here: https://github.com/ynput/ayon-maya/pull/40#issuecomment-2255594300
I'm not sure why python allows it to be iterable to begin with 😄 Maybe due to the implementation of `__contains__` but not `__iter__`?

### How would you imagine the implementation of the feature?

Implement `__iter__` to explicitly iterate the keys?
Or force it to be non-iterable to avoid confusion.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

PR #968, opened 2024-10-24
Create Attribute values: Implemented iter

iLLiCiTiT:
## Changelog Description
Added implementation to iterate over `AttributeValues`.

## Additional info
`AttributeValues` should behave as dict but iterating over it does not behave the same way.

## Testing notes:
1. If is possible to iterate over `AttributeValues` object to get keys.

Resolves https://github.com/ynput/ayon-core/issues/804


BigRoy (approved):
Looks good to me.

Issue #803, opened 2024-07-25
USD Contributions: Default "Add to department layer" to the task or task type


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Currently when creating USD publishes it requires a lot of custom tweaking the publish instance settings, like the "target product" and "initalize as" setting (separate issue #791?) but also definitely the often overlooked **Add to department layer** which defines in which layer your contribution would end up in.

@Lypsolon also stated:
> Add to department layer: This is a place where i fail a lot of times. Because it just doesn't enter my mind that I need to set this. Would it be possible for us to **set this from the Task we are on**? It happened to me a few times now that i contributed to the model layer from my texturing work and then it tends to brick the asset (because i now publish only textures into my model and the default prim is missing and so on.) 
> Maybe we can have a validator that goes out and checks this? 
> 
> Also we are missing a department layer for texturing. It would be great if our defaults for the layers match the task defaults (maybe they should even be created from the task defaults?)
> I would like to have a texture layer: The thought comes from education at pixl visn via Giorgio Lorenzetti. In the studio he is currently at they apparently have a scene where a texture artist sets up the base material setup and then a lookdev artist adds to it.

### How would you imagine the implementation of the feature?

It would be much nicer if this would default to your current task or task type - and based on that define the default value for your current context.

Because the layers MAY currently already mismatch with the layer names or you may work in 'unique' task names like `FX_background` I was thinking of maybe defining the default using profiles against current context task name and task type so that complex defaults can be defined in settings to match the studio's available layers in USD.

---

- Note: The task type may be more granular (e.g. Model type, Sculpt type may both need to go into Model department layer).

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

Somehow dynamically define the list of layers and their strengths - but the ordering will be hard to define automatically without explicitly defining layers and orders in settings. We may just need to find a nice mapping from task (or even CONTEXT) to the department layer

PR #973, opened 2024-10-26
USD contribution: Set up different default values based on profiles

BigRoy:
## Changelog Description

Follow up PR to: https://github.com/ynput/ayon-core/pull/994 so that USD products will use different attribute defaults based on profiles.

## Additional info

TODO
- [x] Implement profiles in settings
- [x] Allow to also filter profiles against e.g. task type

Fix https://github.com/ynput/ayon-core/issues/803
Fix https://github.com/ynput/ayon-core/issues/1153

**Warning**
This has the downside that any existing "Houdini Look" instances would have their 'defaults' reverted to the default values because the specialized Houdini class has been removed. Nonetheless, I think we should disregard that.
- The new defaults should match what users **should have been** setting previously anyway
- Plus this PR fixes https://github.com/ynput/ayon-core/issues/1153 which basically means it's currently broken without this PR anyway.

---

## Testing notes:

1. Use together with PR here: https://github.com/ynput/ayon-houdini/pull/144
3. In Houdini, start publishing USD model, groom and look products. Each product should have different defaults and target the right 'default layer'


client/ayon_core/pipeline/publish/publish_plugins.py
```diff
@@ -205,9 +205,9 @@ def instance_matches_plugin_families(
         if not cls.__instanceEnabled__:
             return False
 
-        for _ in pyblish.logic.plugins_by_families(
-            [cls], [instance.product_type]
-        ):
+        families = [instance.product_type]
```
iLLiCiTiT, line 208: To be honest I don't agree with this change. I still think we should not use `families` during create phase, but other data driven workflow. Either it would be `is_usd: True` on instance data or more complex structure, but it should be usd specific so it can be targeted and deterministic, with possibly more information, instead of arbitrary list of strings.

Example data on instance
```python
{
    "usdData": {
        "someKey": "someValue",
        "renderer": "supaR2D3",
    }
}
```
If the values would not change during create phase and would be read only, they could be in transient data so we don't overfill stored data to scene.

This is my personal opinion, maybe caused by me not having full view into the issue.

BigRoy: Honestly - most of these things I'm explaining here should also 1:1 define whether the plug-in should process the instance. Hence assigning families as tag is the way to go. But yes, there will be edge case where we may want to act on specific metadata and more complex values - but I'd say that's WAY harder to track in consistency, especially across multiple addon repositories instead of just checking "does it have this tag".

However, I just want to state that I'd like to PREFIX all these families, e.g. `traits/reviewable` or `traits/frame_range` or `node/AlembicROP` or whatever to differentiate clearly from the families that are intended as a product type. Also, because it allows us to transition smoother from the legacy system over time.

Also wanted to link this Community topic where I mentioned families versus product type as well: https://community.ynput.io/t/difference-between-family-and-product-type/1923

iLLiCiTiT: Be carefull about traits, as that will be new concept in representations, it might be confusin to mix those.

BTW does trait in this case make sense? Shouldn't be traits defined based on set up attributes instead of set up attributes based on traits?

BigRoy: > BTW does trait in this case make sense? Shouldn't be traits defined based on set up attributes instead of set up attributes based on traits?

The idea is indeed that traits and attributes would be intertwined - in some case a toggle may enable/disable a trait, but to me that's essentially the same case with these families. A creator attribute may add/remove such a 'family'. But yes, whether traits is the best name for it I'm not sure. I just used it because it clarified the subject since there's huge overlap in that area (a much better overlap than "product type" is)

iLLiCiTiT: Like I've said, traits are dangerous, if you want to use `families` then use families, and don't call families traits.

BTW if there are families used only during create phase then I have another argument why not use `families`. Wouldn't make sense to add `usdTraits` instead of `families` to the instance data? So it is clear it is related to USD?

BigRoy: The concept I'm trying to portray is that this is **not** USD specific - the concept expands beyond that and is completely embedded into ayon. E.g. being able to target an instance to be `farm` compatible, or something to have a `framerange` or alike so that we can have a generic frame range validation (or something of the sort). The forum post I linked explains that well.

I understand your point regarding "traits" and the potential confusion. All I really mean to say is that I really want to differentiate from what 'families' are actually about.

I'd really hate these scattered attributes appearing:
```
isUsd: True
isFbx: True
usdTraits: a, b
etc.
```

Using e.g. `file/usd` as a family may at least set a standard so that whenever we target something to a filetype, that we prefix at such. So that whenever you see `file/fbx` you know what it's doing and that it is a file type being referred to. Personally I'd even like `producttype/review` (although bit lengthy) but it's way clearer that "review" family is being targeted based on a product type as opposed to some other scattered data.

Anyway, I've requested a discussion/call this week to discuss this more in-depth. Would be good to have you there too! :)

iLLiCiTiT: Did you discuss the concept with anyone else?

MustafaJafar:
Hey, 
I gave this PR a test as I made a draft PR https://github.com/ynput/ayon-houdini/pull/148
This PR enhances many aspects:
- **Dev:** From pyblish plugins dev perspective, this PR helps understanding the intention of the creators. also, it helps making the system more modular as I can add as many attributes using different plugins. (I think there's a few cases that it won't work, will include it in different comments)
- **Artist:** I can find attributes related to my publish instance.

client/ayon_core/pipeline/publish/publish_plugins.py
BigRoy: I discussed it with Milan - but more as "my proposal". And I mentioned it to Dee and Ondrej last week in a meeting. But mostly from the perspective that we should discuss this in a brainstorm - which I believe current indicative planning is Thursday morning. So this is indeed "to be discussed"

iLLiCiTiT:
I have a question: Why are used profiles? You have quite deterministic way how it's defined, and it looks like it is not based on settings, you can just have if statements -> more specific and clear why it actually happens...

BigRoy:
> Nevermind I just saw in description it is meant to be based on settings. But does it make sense to have it in settings? How someone, non-developer, will be able to fill the values correctly?

Basically by "product type" (which is the profile) define the defaults for the attributes. Somewhat similar to what the Collect Job Info now does in `ayon-deadline` but with much less attributes ;)

MustafaJafar (commented):
I can hardly remember what this PR was about. (I can only remember that we used to have a big PR which we eventually split)
But, it seems to affect only `extract_usd_layer_contributions.py` publish plugin. 
So, I think we may need to rename this PR.

(anyways, this PR seems to work.)

BigRoy:
I think the PR title is correct- just the branch name on my repo is still confusing, sorry about that :)

MustafaJafar (approved):
Default values work as expected on my side. 

Here's a published asset using this PR and https://github.com/ynput/ayon-houdini/pull/144 
Animation_61



I'm publishing my animation as part of my usd asset. 
image

----
Note: It affects the default value for component builder as its product type is `usd`. 
image

MustafaJafar:
I also like these profiles. 
image

PR #978, opened 2024-10-30
Delivery: Use current roots during delivery

iLLiCiTiT:
## Changelog Description
Make sure that root is not re-used from published representation during delivery.

## Additional info
Because we do copy `"context"` it also copied `"root"` there, if root is filled in template data, anatomy does not auto-fill it which causes issues during delivery.

## Testing notes:
1. Publish some representation (can skip if you have any).
2. Create new root in the project.
3. Use the new root in delivery template.
4. Run delivery action in loader.
5. It should work fine and should not raise error because of missing root key.

### DIFF
--- a/client/ayon_core/pipeline/delivery.py
+++ b/client/ayon_core/pipeline/delivery.py
@@ -411,7 +411,7 @@ def get_representations_delivery_template_data(
         _merge_data(template_data, repre_entity["context"])
 
         # Remove roots from template data to auto-fill them with anatomy data
-        template_data.pop("root")
+        template_data.pop("root", None)
 
         output[repre_id] = template_data
     return output


client/ayon_core/pipeline/delivery.py
```diff
@@ -409,5 +409,9 @@ def get_representations_delivery_template_data(
             "version": version_entity["version"],
         })
         _merge_data(template_data, repre_entity["context"])
+
+        # Remove roots from temlate data to auto-fill them with anatomy data
+        template_data.pop("root")
```
BigRoy, line 414: Could it be - for whatever reason.... that `root` does _not_ exist in context? If so, this may need to be `template_data.pop("root", None)` because otherwise [it'd raise an `IndexError`](https://docs.python.org/3/tutorial/datastructures.html#more-on-lists)

client/ayon_core/pipeline/delivery.py
```diff
@@ -409,5 +409,9 @@ def get_representations_delivery_template_data(
             "version": version_entity["version"],
         })
         _merge_data(template_data, repre_entity["context"])
+
+        # Remove roots from temlate data to auto-fill them with anatomy data
```
BigRoy, line 413:
```suggestion
        # Remove roots from template data to auto-fill them with anatomy data
```

client/ayon_core/pipeline/delivery.py
iLLiCiTiT: It should be, and I actually had it, now I'm a little bit confused, because I saw it in code.
```suggestion
        template_data.pop("root", none)
```

BigRoy (commented):
Question @iLLiCiTiT - before this PR ... they may be usage of delivery templates of `{root}` in delivery templates. That would break now because one would require now to specify the children? Or was `root` already a dict in `context` too?

iLLiCiTiT:
> there may be usage of delivery templates of {root} in delivery templates.

I don't think it was, you would still have to use `{root[...]}` in template.

> Or was root already a dict in context too?

Yes, but only with one root, and one platform.

64qam (approved):

PR #979, opened 2024-10-31
Representation defined by traits

antirotor:
## Changelog Description
Introducing new way of defining representations. Developer friendly, flexible yet strictly typed, easy to extend using Traits as building blocks defining what representation IS.

## Additional info

New representation is holder for traits. Each trait define some properties and together with other traits describes the representation. Traits can be anything- their purpose is to help publisher, loader (and basically any other systems) to clearly identify data needed for processing. Like file system path, or frame per second, or resolution. Those properties are grouped together to logical block - frame range specific, resolution specific, etc.

This PR is introducing several new features:

**TraitBase** - this is a model for all Traits. It defines abstract properties like `id`, `name`, `description` that needs to be in all Traits.

> [!NOTE]
> Every Trait can re-implement `validate()` method. The one in `TraitBase` always returns `True`. `Representation` is passed to that method so every Trait can validate against all other Trait present in representation.

**Representation** - this is "container" of sorts to hold all Traits. It holds representation`name` and `id`. And lot of "helper" methods to work with Traits:
 
- methods to check if Trait exists in the Representation
- methods to add and remove Traits
- methods to get Traits
- method to return whole Representation serialized as Python dict
- method to reconstruct Representation from the dict

Most of them can run on bulk of Traits and you can access Traits by their class or by their Id. More on that below.

There is also mechanism for upgrading and versioning Traits.

### Traits

There are some pre-defined Traits that should come with **ayon-core** to provide "common language". They are all based on `TraitBase` and are grouped to logical chunks or namespaces - **content**, **lifecycle**, **meta**, **three_dimensional**, **time**, **two_dimensional**. Their complete list can be found in the code and is obviously subject to change based on the need.

### Practical examples

So how to work with traits and representations? To define traits (for example in *Collector* plugin or in *Extractor* plugin:
```python
from ayon_core.pipeline.traits import (
	FileLocation,
	Image,,
	MimeType,
    PixelBased,
	Representation
)

# define traits
file_location = FileLocation(
    file_path=Path("/path/to/file.jpg"),
)

pixel_based = PixelBased(
	display_window_width=1920,
    display_window_height=1080,
    pixel_aspect_ratio=1.0 
)

# create representation itself

representation = Representation(
    name="Image test",
	traits=[file_location, Image(), pixel_based])

# add additional traits
mime_type = MimeType(mime_type="image/jpeg")
representation.add_trait(mime_type)

# remove trait
representation.remove_trait(MimeType)

if representation.contains_trait(PixelBased):
	print(f"width: {representation.get_trait(PixelBased).display_window_width)}")

```

You can work with Traits using classes, but you can also utilize their ids. That is useful when working with representation that was serialized into "plain' dictionary:

```python

# some pseudo-function to get representation as dict
representation = Representation.from_dict(get_representation_dict(...))

# get trait by its id
# note: the use of version in the ID. You can test if trait is present in representation, or if trait of specific version is present.
if representation.contains_trait_by_id("ayon.content.FileLocation"):
	print(f"path: {representation.get_trait_by_id("ayon.content.FileLocation.v1")

# serialize back to dict
representation.traits_as_dict()
```

There is also feature of version upgrade on Traits. Whenever you want to de-serialize data that are using older version of trait, `upgrade()` method on newer trait definition is called to reconstruct new version (downgrading isn't possible). So, you can have serialized trait like so (type trait without properties):

```python
{
		ayon.2d.Image.v1: {}
}
```
But your current runtime has **Image** trait `ayon.2d.Image.v2` that is also adding property `foo`.

Whenever you run `representation.get_trait_by_id("ayon.2d.Image")` without version specifier, it can try to find out the latest Trait definition available and if it differs - `v1 != v2` it tries to call `upgrade()` method on the latest trait.

### Notes

Traits are no longer Pydantic models because Pydantic 2 is based on rust and we would need to support it in all possible host (distribute it somehow in AYON too). So until this issue is resolved, it is better to have them as pure Python dataclasses.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@
 
 [tool.poetry]
 name = "ayon-core"
-version = "1.1.6+dev"
+version = "1.1.7+dev"
 description = ""
 authors = ["Ynput Team <team@ynput.io>"]
 readme = "README.md"
--- a/client/ayon_core/hooks/pre_global_host_data.py
+++ b/client/ayon_core/hooks/pre_global_host_data.py
@@ -1,12 +1,15 @@
 from ayon_api import get_project, get_folder_by_path, get_task_by_name
 
+from ayon_core.pipeline import Anatomy
+from ayon_core.pipeline.anatomy import RootMissingEnv
+
 from ayon_applications import PreLaunchHook
+from ayon_applications.exceptions import ApplicationLaunchFailed
 from ayon_applications.utils import (
     EnvironmentPrepData,
     prepare_app_environments,
     prepare_context_environments
 )
-from ayon_core.pipeline import Anatomy
 
 
 class GlobalHostDataHook(PreLaunchHook):
@@ -67,9 +70,12 @@ class GlobalHostDataHook(PreLaunchHook):
         self.data["project_entity"] = project_entity
 
         # Anatomy
-        self.data["anatomy"] = Anatomy(
-            project_name, project_entity=project_entity
-        )
+        try:
+            self.data["anatomy"] = Anatomy(
+                project_name, project_entity=project_entity
+            )
+        except RootMissingEnv as exc:
+            raise ApplicationLaunchFailed(str(exc))
 
         folder_path = self.data.get("folder_path")
         if not folder_path:
--- a/client/ayon_core/pipeline/anatomy/__init__.py
+++ b/client/ayon_core/pipeline/anatomy/__init__.py
@@ -1,5 +1,6 @@
 from .exceptions import (
     ProjectNotSet,
+    RootMissingEnv,
     RootCombinationError,
     TemplateMissingKey,
     AnatomyTemplateUnsolved,
@@ -9,6 +10,7 @@ from .anatomy import Anatomy
 
 __all__ = (
     "ProjectNotSet",
+    "RootMissingEnv",
     "RootCombinationError",
     "TemplateMissingKey",
     "AnatomyTemplateUnsolved",
--- a/client/ayon_core/pipeline/anatomy/exceptions.py
+++ b/client/ayon_core/pipeline/anatomy/exceptions.py
@@ -5,6 +5,11 @@ class ProjectNotSet(Exception):
     """Exception raised when is created Anatomy without project name."""
 
 
+class RootMissingEnv(KeyError):
+    """Raised when root requires environment variables which is not filled."""
+    pass
+
+
 class RootCombinationError(Exception):
     """This exception is raised when templates has combined root types."""
 
--- a/client/ayon_core/pipeline/anatomy/roots.py
+++ b/client/ayon_core/pipeline/anatomy/roots.py
@@ -2,8 +2,10 @@ import os
 import platform
 import numbers
 
-from ayon_core.lib import Logger
+from ayon_core.lib import Logger, StringTemplate
 from ayon_core.lib.path_templates import FormatObject
+
+from .exceptions import RootMissingEnv
 
 
 class RootItem(FormatObject):
@@ -21,18 +23,36 @@ class RootItem(FormatObject):
             multi root setup otherwise None value is expected.
     """
     def __init__(self, parent, root_raw_data, name):
-        super(RootItem, self).__init__()
+        super().__init__()
         self._log = None
-        lowered_platform_keys = {}
-        for key, value in root_raw_data.items():
-            lowered_platform_keys[key.lower()] = value
+        lowered_platform_keys = {
+            key.lower(): value
+            for key, value in root_raw_data.items()
+        }
         self.raw_data = lowered_platform_keys
         self.cleaned_data = self._clean_roots(lowered_platform_keys)
         self.name = name
         self.parent = parent
 
         self.available_platforms = set(lowered_platform_keys.keys())
-        self.value = lowered_platform_keys.get(platform.system().lower())
+
+        current_platform = platform.system().lower()
+        # WARNING: Using environment variables in roots is not considered
+        #   as production safe. Some features may not work as expected, for
+        #   example USD resolver or site sync.
+        try:
+            self.value = lowered_platform_keys[current_platform].format_map(
+                os.environ
+            )
+        except KeyError:
+            result = StringTemplate(self.value).format(os.environ.copy())
+            is_are = "is" if len(result.missing_keys) == 1 else "are"
+            missing_keys = ", ".join(result.missing_keys)
+            raise RootMissingEnv(
+                f"Root \"{name}\" requires environment variable/s"
+                f" {missing_keys} which {is_are} not available."
+            )
+
         self.clean_value = self._clean_root(self.value)
 
     def __format__(self, *args, **kwargs):
@@ -105,10 +125,10 @@ class RootItem(FormatObject):
 
     def _clean_roots(self, raw_data):
         """Clean all values of raw root item values."""
-        cleaned = {}
-        for key, value in raw_data.items():
-            cleaned[key] = self._clean_root(value)
-        return cleaned
+        return {
+            key: self._clean_root(value)
+            for key, value in raw_data.items()
+        }
 
     def path_remapper(self, path, dst_platform=None, src_platform=None):
         """Remap path for specific platform.
--- a/client/ayon_core/pipeline/context_tools.py
+++ b/client/ayon_core/pipeline/context_tools.py
@@ -27,7 +27,8 @@ from .workfile import (
     get_workdir,
     get_custom_workfile_template_by_string_context,
     get_workfile_template_key_from_context,
-    get_last_workfile
+    get_last_workfile,
+    MissingWorkdirError,
 )
 from . import (
     register_loader_plugin_path,
@@ -251,7 +252,7 @@ def uninstall_host():
     pyblish.api.deregister_discovery_filter(filter_pyblish_plugins)
     deregister_loader_plugin_path(LOAD_PATH)
     deregister_inventory_action_path(INVENTORY_PATH)
-    log.info("Global plug-ins unregistred")
+    log.info("Global plug-ins unregistered")
 
     deregister_host()
 
@@ -617,7 +618,18 @@ def version_up_current_workfile():
     last_workfile_path = get_last_workfile(
         work_root, file_template, data, extensions, True
     )
-    new_workfile_path = version_up(last_workfile_path)
+    # `get_last_workfile` will return the first expected file version
+    # if no files exist yet. In that case, if they do not exist we will
+    # want to save v001
+    new_workfile_path = last_workfile_path
     if os.path.exists(new_workfile_path):
         new_workfile_path = version_up(new_workfile_path)
+
+    # Raise an error if the parent folder doesn't exist as `host.save_workfile`
+    # is not supposed/able to create missing folders.
+    parent_folder = os.path.dirname(new_workfile_path)
+    if not os.path.exists(parent_folder):
+        raise MissingWorkdirError(
+            f"Work area directory '{parent_folder}' does not exist.")
+
     host.save_workfile(new_workfile_path)
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -2303,10 +2303,16 @@ class CreateContext:
                 for plugin_name, plugin_value in item_changes.pop(
                     "publish_attributes"
                 ).items():
+                    if plugin_value is None:
+                        current_publish[plugin_name] = None
+                        continue
                     plugin_changes = current_publish.setdefault(
                         plugin_name, {}
                     )
-                    plugin_changes.update(plugin_value)
+                    if plugin_changes is None:
+                        current_publish[plugin_name] = plugin_value
+                    else:
+                        plugin_changes.update(plugin_value)
 
             item_values.update(item_changes)
 
--- a/client/ayon_core/pipeline/create/structures.py
+++ b/client/ayon_core/pipeline/create/structures.py
@@ -160,28 +160,25 @@ class AttributeValues:
         return self._attr_defs_by_key.get(key, default)
 
     def update(self, value):
-        changes = {}
-        for _key, _value in dict(value).items():
-            if _key in self._data and self._data.get(_key) == _value:
-                continue
-            self._data[_key] = _value
-            changes[_key] = _value
-
+        changes = self._update(value)
         if changes:
             self._parent.attribute_value_changed(self._key, changes)
 
     def pop(self, key, default=None):
-        has_key = key in self._data
-        value = self._data.pop(key, default)
-        # Remove attribute definition if is 'UnknownDef'
-        # - gives option to get rid of unknown values
-        attr_def = self._attr_defs_by_key.get(key)
-        if isinstance(attr_def, UnknownDef):
-            self._attr_defs_by_key.pop(key)
-            self._attr_defs.remove(attr_def)
-        elif has_key:
-            self._parent.attribute_value_changed(self._key, {key: None})
+        value, changes = self._pop(key, default)
+        if changes:
+            self._parent.attribute_value_changed(self._key, changes)
         return value
+
+    def set_value(self, value):
+        pop_keys = set(value.keys()) - set(self._data.keys())
+        changes = self._update(value)
+        for key in pop_keys:
+            _, key_changes = self._pop(key, None)
+            changes.update(key_changes)
+
+        if changes:
+            self._parent.attribute_value_changed(self._key, changes)
 
     def reset_values(self):
         self._data = {}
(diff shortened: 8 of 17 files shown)


client/ayon_core/plugins/__init__.py
BigRoy, line 1: Please remove these init files.

client/ayon_core/pipeline/traits/lifecycle.py
```diff
@@ -0,0 +1,37 @@
...
+class Persistent(TraitBase):
+    """Persistent trait model.
+
+    This model represents a persistent trait.
+
+    Attributes:
+        name (str): Trait name.
+        description (str): Trait description.
+        id (str): id should be namespaced trait name with version
+    """
```
BigRoy, line 33: Can we improve the docstring to describe the differences between these and how/why to use them?

client/ayon_core/pipeline/traits/time.py
```diff
@@ -0,0 +1,86 @@
...
+        forbidden (int): Gaps are forbidden.
+        missing (int): Gaps are interpreted as missing frames.
+        hold (int): Gaps are interpreted as hold frames (last existing frames).
+        black (int): Gaps are interpreted as black frames.
+    """
+    forbidden = auto()
+    missing = auto()
+    hold = auto()
+    black = auto()
+
```
BigRoy, line 25:
```suggestion


```

client/ayon_core/plugins/__init__.py
antirotor: Then unit test will stop working. I want to be able to import integrator class and test its methods. Is there better way? Also, is it bad thing? When I tried to run Sphinx auto-doc on OpenPype, I had to make modules from all code directories I wanted to include.

client/ayon_core/pipeline/traits/lifecycle.py
antirotor: Definitely, that is still in the progress once I am sure I have most needed. There are some Traits that won't be needed probably.

client/ayon_core/plugins/__init__.py
antirotor: But it shouldn't be here anyway, I've noticed that some of the integrator work was merged here. This PR should be about the traits.

BigRoy: Wasn't the need for `__init__.py` as a requirement dropped in Py 3.3+?

MustafaJafar:
Hey, would you like to me do some test runs for this PR?

antirotor:
> Hey, would you like to me do some test runs for this PR?
no tests yet (apart unit tests that were passing recently). It is more work in progress now - main development is now done on the integrator. This is here to discuss traits and if we have everything we need. I guess more will be clear one we try to use them in existing workflows.

MustafaJafar:
> I guess more will be clear one we try to use them in existing workflows.

This is most likely how I may test run it by trying to use it in draft PR. 
Personally, I find it cool to try new things.

client/ayon_core/pipeline/traits/utils.py
```diff
@@ -0,0 +1,38 @@
...
+    Note that this cannot guess frame rate, so it's set to 25.
+
+    Args:
+        paths (list[Path]): List of file paths.
+
+    Returns:
+        FrameRanged: FrameRanged trait.
+
+    """
+    col = assemble([path.as_posix() for path in paths])[0][0]
```
BigRoy, line 26: This will error on:

- File paths that do not have a sequence identifier
- File paths that only contain a single frame

Because both won'tbe detected as collections but as remainder

client/ayon_core/addon/interfaces.py
```diff
@@ -24,7 +35,11 @@ class AYONInterface(metaclass=_AYONInterfaceMeta):
     in the interface. By default, interface does not have any abstract parts.
     """
 
-    pass
+    log = None
+
+    def __init__(self):
+        """Initialize interface."""
+        self.log = logging.getLogger(self.__class__.__name__)
```
iLLiCiTiT, line 42: This is already defined on `AYONAddon`.

client/ayon_core/addon/interfaces.py
```diff
@@ -38,10 +53,25 @@ class IPluginPaths(AYONInterface):
     """
 
     @abstractmethod
-    def get_plugin_paths(self):
-        pass
+    def get_plugin_paths(self) -> dict[str, list[str]]:
```
iLLiCiTiT, line 56: We can't us `dict` and `list`.
```suggestion
    def get_plugin_paths(self) -> Dict[str, List[str]]:
```

client/ayon_core/addon/interfaces.py
```diff
@@ -117,77 +147,85 @@ def get_inventory_action_paths(self, host_name):
...
-    would do nothing.
-    """
+    Important:
+        The addon. still must be usable if is not used in tray even if it
+        would do nothing.
 
+    """
     tray_initialized = False
-    _tray_manager = None
+    manager: AddonsManager = None
```
iLLiCiTiT, line 164: Why this was added?

client/ayon_core/addon/interfaces.py
```diff
@@ -117,77 +147,85 @@ def get_inventory_action_paths(self, host_name):
...
-        """
+        prepared
 
-        pass
+        """
+        raise NotImplementedError
 
     @abstractmethod
-    def tray_menu(self, tray_menu):
+    def tray_menu(self, tray_menu: QtWidgets.QMenu) -> None:
```
iLLiCiTiT, line 181:
```suggestion
    def tray_menu(self, tray_menu: "QtWidgets.QMenu") -> None:
```

client/ayon_core/addon/interfaces.py
```diff
@@ -117,77 +147,85 @@ def get_inventory_action_paths(self, host_name):
...
-    def tray_start(self):
+    def tray_start(self) -> None:
         """Start procedure in tray tool."""
-
-        pass
+        raise NotImplementedError
 
     @abstractmethod
-    def tray_exit(self):
+    def tray_exit(self) -> None:
```
iLLiCiTiT, line 192: Is it necessary to typehint return value of `None`?

pyproject.toml
```diff
@@ -63,13 +69,29 @@ exclude = [
...
+
 # Assume Python 3.9
 target-version = "py39"
 
 [tool.ruff.lint]
+pydocstyle.convention = "google"
 # Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`)  codes by default.
-select = ["E", "F", "W"]
-ignore = []
+select = ["ALL"]
```
iLLiCiTiT, line 79: Change for different PR.

client/ayon_core/addon/interfaces.py
```diff
@@ -117,77 +147,85 @@ def get_inventory_action_paths(self, host_name):
...
-
-    def show_tray_message(self, title, message, icon=None, msecs=None):
+        self._tray_manager.tray_manager.execute_in_main_thread(callback)
+
+    def show_tray_message(
+            self,
+            title: str,
+            message: str,
+            icon: Optional[QtWidgets.QSystemTrayIcon]=None,
+            msecs: Optional[int]=None) -> None:
```
iLLiCiTiT, line 228:
```suggestion
            self,
        title: str,
        message: str,
        icon: Optional["QtWidgets.QSystemTrayIcon"] = None,
        msecs: Optional[int] = None
    ):
```

client/ayon_core/addon/interfaces.py
```diff
@@ -265,36 +304,40 @@ def tray_menu(self, tray_menu):
...
+    def tray_exit(self) -> None:
+        """Cleanup method which is executed on tray shutdown."""
         return
 
 
 class ITrayService(ITrayAddon):
+    """Tray service Interface."""
     # Module's property
-    menu_action = None
+    menu_action: QtWidgets.QAction = None
```
iLLiCiTiT, line 319: How did not the `QtWidgets` usages crash for you?

client/ayon_core/addon/interfaces.py
```diff
@@ -378,18 +424,31 @@ class IHostAddon(AYONInterface):
 
     @property
     @abstractmethod
-    def host_name(self):
+    def host_name(self) -> str:
         """Name of host which addon represents."""
+        raise NotImplementedError
```
iLLiCiTiT, line 429: ??? you're raising class....
```suggestion
        pass
```

client/ayon_core/pipeline/traits/content.py
```diff
@@ -0,0 +1,467 @@
...
+        id (str): id should be namespaced trait name with version
+        mime_type (str): Mime type like image/jpeg.
+
+    """
+
+    name: ClassVar[str] = "MimeType"
+    description: ClassVar[str] = "MimeType Trait Model"
+    id: ClassVar[str] = "ayon.content.MimeType.v1"
+    mime_type: str = Field(..., title="Mime Type")
+
```
iLLiCiTiT, line 45:
```suggestion


```

client/ayon_core/pipeline/traits/content.py
```diff
@@ -0,0 +1,467 @@
...
+    For more information, see RFC 2046 and RFC 4288 (and related RFCs).
+
+    Attributes:
+        name (str): Trait name.
+        description (str): Trait description.
+        id (str): id should be namespaced trait name with version
+        mime_type (str): Mime type like image/jpeg.
+
+    """
+
```
iLLiCiTiT, line 40:
```suggestion
```

client/ayon_core/pipeline/traits/content.py
```diff
@@ -0,0 +1,467 @@
...
+        location (str): Location.
+
+    """
+
+    name: ClassVar[str] = "LocatableContent"
+    description: ClassVar[str] = "LocatableContent Trait Model"
+    id: ClassVar[str] = "ayon.content.LocatableContent.v1"
+    location: str = Field(..., title="Location")
+    is_templated: Optional[bool] = Field(default=None, title="Is Templated")
+
```
iLLiCiTiT, line 68:
```suggestion


```

client/ayon_core/pipeline/traits/meta.py
```diff
@@ -0,0 +1,113 @@
...
+        This is not a persistent trait.
+
+    """
+    name: ClassVar[str] = "KeepOriginalName"
+    description: ClassVar[str] = "Keep Original Name Trait Model"
+    id: ClassVar[str] = "ayon.meta.KeepOriginalName.v1"
+    persistent: bool = Field(default=False, title="Persistent")
+
+
+class SourceApplication(TraitBase):
```
iLLiCiTiT, line 104: This trait is weird, why is there platform (I would expect that on some "base" trait)? Where we gonna get those information? Attributes `application` and `variant` is `ayon-applications` addon specific information, and `version` can be collected only by host integration.

client/ayon_core/pipeline/traits/representation.py
```diff
@@ -0,0 +1,710 @@
...
+
+        result = next(
+            (
+                self._data.get(trait_id)
+                for trait_id in self._data
+                if trait_id.startswith(trait_id)
+            ),
+            None,
+        )
+        if not result:
```
iLLiCiTiT, line 336:
```suggestion
        if result is None:
```

client/ayon_core/pipeline/traits/representation.py
```diff
@@ -0,0 +1,710 @@
...
+            msg = f"Trait with ID {trait_id} not found."
+            raise MissingTraitError(msg)
+        return result
+
+    def get_traits(self,
+                     traits: Optional[list[Type[T]]]=None
+                   ) -> dict[str, T]:
```
iLLiCiTiT, line 343:
```suggestion
    def get_traits(
        self, traits: Optional[list[Type[T]]]=None
    ) -> dict[str, T]:
```

client/ayon_core/pipeline/traits/representation.py
```diff
@@ -0,0 +1,710 @@
...
+            name: str,
+            representation_id: Optional[str]=None,
+            traits: Optional[list[T]]=None):
+        """Initialize the data.
+
+        Args:
+            name (str): Representation name. Must be unique within instance.
+            representation_id (str, optional): Representation ID.
+            traits (list[TraitBase], optional): List of traits.
+        """
```
iLLiCiTiT, line 410:
```suggestion

        """
```

client/ayon_core/pipeline/traits/representation.py
```diff
@@ -0,0 +1,710 @@
...
+
+    def __len__(self):
+        """Return the length of the data."""
+        return len(self._data)
+
+    def __init__(
+            self,
+            name: str,
+            representation_id: Optional[str]=None,
+            traits: Optional[list[T]]=None):
```
iLLiCiTiT, line 403:
```suggestion
            representation_id: Optional[str] = None,
            traits: Optional[list[T]] = None):
```

client/ayon_core/pipeline/traits/representation.py
```diff
@@ -0,0 +1,710 @@
...
+        """
+        self.name = name
+        self.representation_id = representation_id or uuid.uuid4().hex
+        self._data = {}
+        if traits:
+            for trait in traits:
+                self.add_trait(trait)
+
+    @staticmethod
+    def _get_version_from_id(trait_id: str) -> Union[int, None]:
```
iLLiCiTiT, line 419: `Union` has same issue as `list` and `dict` typehints, it was introduced later than what we do support.
```suggestion
    def _get_version_from_id(trait_id: str) -> "Union[int, None]":
```

client/ayon_core/pipeline/traits/three_dimensional.py
```diff
@@ -0,0 +1,81 @@
...
+    Type trait for shader data.
+
+    Sync with OpenAssetIO MediaCreation Traits.
+    """
+
+    id: ClassVar[str] = "ayon.3d.Shader.v1"
+    name: ClassVar[str] = "Shader"
+    description: ClassVar[str] = "Shader trait model."
+
+class Lighting(TraitBase):
```
iLLiCiTiT, line 61:
```suggestion

class Lighting(TraitBase):
```

client/ayon_core/pipeline/traits/three_dimensional.py
```diff
@@ -0,0 +1,81 @@
...
+    Type trait for lighting data.
+
+    Sync with OpenAssetIO MediaCreation Traits.
+    """
+
+    id: ClassVar[str] = "ayon.3d.Lighting.v1"
+    name: ClassVar[str] = "Lighting"
+    description: ClassVar[str] = "Lighting trait model."
+
+class IESProfile(TraitBase):
```
iLLiCiTiT, line 73:
```suggestion

class IESProfile(TraitBase):
```

client/ayon_core/pipeline/traits/representation.py
iLLiCiTiT: Or maybe I'm wrong? `Union` is available since Python 3.8 and using `list` and `dict` is available since Python 3.9 . I think we still expect 3.8 to work?

client/ayon_core/pipeline/traits/two_dimensional.py
```diff
@@ -0,0 +1,189 @@
...
+from typing import TYPE_CHECKING, ClassVar, Optional
+
+from pydantic import Field, field_validator
+
+from .trait import TraitBase
+
+if TYPE_CHECKING:
+    from .content import FileLocation, FileLocations
+
+class Image(TraitBase):
```
iLLiCiTiT, line 14:
```suggestion

class Image(TraitBase):
```

client/ayon_core/pipeline/traits/utils.py
iLLiCiTiT: I think that this function is potential danger, because it sounds like it can be used for any case, but it causing the same restriction we have now.

client/ayon_core/addon/interfaces.py
antirotor: you need to always annotate what it returns.

client/ayon_core/addon/interfaces.py
antirotor: we can, because of `__future__.annotation`, it works even on python 3.7

client/ayon_core/pipeline/traits/meta.py
antirotor: Think of it as metadata - it doesn't have any logic - only holds the information what was the application used to create this representation. This information needs to be filled by the host producing this representation.

client/ayon_core/addon/interfaces.py
antirotor: because it is used down the line like in `add_doubleclick_callback` and mypy doesn't like that it doesn't have type defined

client/ayon_core/addon/interfaces.py
antirotor: you don't need this because of the `__future__.annotations`

client/ayon_core/addon/interfaces.py
antirotor: again, `__future__.annotation` ftw

client/ayon_core/addon/interfaces.py
antirotor: This works because it is imported only in `TYPE_CHECKING`

client/ayon_core/pipeline/traits/representation.py
antirotor: this cannot hit valid boolean value, but yeah...

client/ayon_core/pipeline/traits/representation.py
antirotor: it works because of `__future__.annotations`

pyproject.toml
antirotor: I'll remove this from the PR just before the merge. I want it to be linted by these rules.

antirotor: I'll remove this from the PR just before the merge. I want it to be linted by these rules.

client/ayon_core/pipeline/traits/utils.py
antirotor: Main use of this function is on FrameRanged trait validation that should be sequence by default and if it fails then it's just right thing to do. But maybe it should be better annotated then.

client/ayon_core/pipeline/traits/meta.py
iLLiCiTiT: Well but it is trait that would combine information from 3 different sources... What if you're missing one of them?

client/ayon_core/addon/interfaces.py
iLLiCiTiT: Ok, `from __future__ import annotations` seems like magical replacer...

Issue #139, opened 2024-02-29
Enhancement: Display overriden vs. default values of attributes in publisher UI


antirotor:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

With attributes, there is no way to know if artist changed value on them from default.

### How would you imagine the implementation of the feature?

This could be easily shown just by changing font-face or color. Something like:

foo: [bar] (default)
*foo:* [goo] (overriden)

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

[cuID:[AY-2614](https://app.clickup.com/t/86bxpfm5u)]

BigRoy:
I personally like the way Houdini displays non-default values with bold value:

![image](https://github.com/ynput/ayon-core/assets/2439881/ad95be14-651a-45de-8a29-8a36e6cb6d00)

Note e.g. here the `Transform` tab being bold because there are children with non-default values.
Then the Translate Y and first dropdown of Transform Order here are non-default values, shown by bold values.

It's not super in your face that it gets annoying but it's obvious enough when you want to look at what is at a default value or not.

👍 

---

Might be a separate feature request, but bonus points for being able to right click and "Revert to defaults" or something along those lines.

PR #982, opened 2024-11-04
Publisher: Display overridden vs default values of attributes

iLLiCiTiT:
## Changelog Description
Visualy display if value of attributes on instance does not match default value.

## Additional info
It should also behave correctly on multiselection of instances even if they don't have same default values. But if at least one of selected instances does not use default value it is shown as overriden.

## Testing notes:
1. Open Publisher in host of your choice.
2. Create an instance (or use existing one).
3. Change any attribute value.
4. It should change color of it's label.

Resolves https://github.com/ynput/ayon-core/issues/139

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/product_attributes.py
+++ b/client/ayon_core/tools/publisher/widgets/product_attributes.py
@@ -1,8 +1,9 @@
-from typing import Dict
+import typing
+from typing import Dict, List, Any
 
 from qtpy import QtWidgets, QtCore
 
-from ayon_core.lib.attribute_definitions import UnknownDef
+from ayon_core.lib.attribute_definitions import AbstractAttrDef, UnknownDef
 from ayon_core.tools.utils import set_style_property
 from ayon_core.tools.attribute_defs import create_widget_for_attr_def
 from ayon_core.tools.publisher.abstract import AbstractPublisherFrontend
@@ -10,6 +11,9 @@ from ayon_core.tools.publisher.constants import (
     INPUTS_LAYOUT_HSPACING,
     INPUTS_LAYOUT_VSPACING,
 )
+
+if typing.TYPE_CHECKING:
+    from typing import Union
 
 
 def _set_label_overriden(label: QtWidgets.QLabel, overriden: bool):
@@ -23,6 +27,22 @@ def _set_label_overriden(label: QtWidgets.QLabel, overriden: bool):
 class _CreateAttrDefInfo:
     def __init__(self, attr_def, instance_ids, defaults, label_widget):
         self.attr_def = attr_def
+        self.instance_ids = instance_ids
+        self.defaults = defaults
+        self.label_widget = label_widget
+
+
+class _PublishAttrDefInfo:
+    def __init__(
+        self,
+        attr_def: AbstractAttrDef,
+        plugin_name: str,
+        instance_ids: List["Union[str, None]"],
+        defaults: List[Any],
+        label_widget: "Union[None, QtWidgets.QLabel]",
+    ):
+        self.attr_def = attr_def
+        self.plugin_name = plugin_name
         self.instance_ids = instance_ids
         self.defaults = defaults
         self.label_widget = label_widget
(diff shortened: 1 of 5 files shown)


client/ayon_core/tools/publisher/widgets/product_attributes.py
```diff
@@ -1,13 +1,52 @@
...
+        plugin_name: str,
+        instance_ids: List["Union[str, None]"],
+        defaults: List[Any],
+        label_widget: "Union[None, QtWidgets.QLabel]",
+    ):
+        self.attr_def = attr_def
+        self.plugin_name = plugin_name
+        self.instance_ids = instance_ids
+        self.defaults = defaults
+        self.label_widget = label_widget
```
BigRoy, line 48: Should these come with a tiny docstring?

client/ayon_core/tools/publisher/widgets/product_attributes.py
```diff
@@ -81,8 +119,7 @@ def _refresh_content(self):
             prev_content_widget.deleteLater()
 
         self._content_widget = None
-        self._attr_def_id_to_instances = {}
-        self._attr_def_id_to_attr_def = {}
+        self._attr_def_info_by_id = {}
```
BigRoy, line 122: Might be worth adding a type hint?

client/ayon_core/tools/publisher/widgets/product_attributes.py
```diff
@@ -121,8 +170,14 @@ def _refresh_content(self):
             col_num = 2 - expand_cols
 
             label = None
+            is_overriden = False
             if attr_def.is_value_def:
+                is_overriden = any(
+                    item["value"] != item["default"]
```
BigRoy, line 176: I wonder if it's worth having just a `is_default()` function on the info items so we can encapsulate the logic there?

BigRoy (approved):
I love this!

image

I do think this should **absolutely** come with a means to "reset to defaults" so that one could easily learn also that the color hinting means that it's "not at default value". But it does work very nice! Great work.

BigRoy:
@iLLiCiTiT should we somehow dim the color if it's 'disabled'?

image


And, the orange color is relatively 'negative'. Should we use the "green" color for saved studio override settings on web frontend instead? 
Or the blue color for unsaved overrides?

image

LiborBatek (commented):
Works nicely, the only occasion which throw some traceback  is when trying to switch settings for ```context``` and ```Validate outdated containers```



Screenshot 2024-11-05 094409

resulting in
```
// Traceback (most recent call last):
//   File "C:\Work\REPO\ayon-core\client\ayon_core\lib\events.py", line 332, in process_event
//     callback(event)
//   File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\create\create_maya_usd.py", line 30, in on_values_changed
//     if instance["creator_identifier"] != self.identifier:
// TypeError: 'NoneType' object is not subscriptable
```
Not failing but worth mentioning imho....

BigRoy:
> resulting in
> 
> ```
> // Traceback (most recent call last):
> //   File "C:\Work\REPO\ayon-core\client\ayon_core\lib\events.py", line 332, in process_event
> //     callback(event)
> //   File "c:\Work\REPO\ayon-maya\client\ayon_maya\plugins\create\create_maya_usd.py", line 30, in on_values_changed
> //     if instance["creator_identifier"] != self.identifier:
> // TypeError: 'NoneType' object is not subscriptable
> ```
> 
> Not failing but worth mentioning imho....

Thanks - that's actually a bug outside of this PR then. I'll fix it. 👍

client/ayon_core/tools/publisher/widgets/product_attributes.py
iLLiCiTiT: It has in `__init__`.

iLLiCiTiT:
> should we somehow dim the color if it's 'disabled'?

I've changed color to the "greenish" one, and with that I would keep it even if is disabled. But I don't have strong opinion about this, should be easy to change.

iLLiCiTiT:
> I would almost go as far as saying it should be part of this PR - but I'm technically fine with it as a follow up PR.

Separate issue, this was easy, reverting might be a little but more complicated.

BigRoy:
Just adding some screenshots with latest changes:

image

image

image

Looking really nice. I'd say, ready to merge.

iLLiCiTiT:
For comparison adding how it would look with the blue color 
image

BigRoy:
Yes - blue is much better!

image

Ready for me. 🚀

iLLiCiTiT:
I like blue the most.

PR #986, opened 2024-11-05
Publisher: Revert to default value

iLLiCiTiT:
## Changelog Description
Implemented context menu option on attribute definitions to reset attribute values to default states.

## Additional info
All attribute definition widgets should have implemented logic to be able to reset value of attributes to default value. In Publisher it does reset to default value per instance.

## Testing notes:
1. Open Publisher in host of your choice.
2. Create instance/s.
3. In Publish tab select instance/s and change their values.
4. When you right click them and choose `Revert to default` it should set value to default value defined by create/publish plugin. It should also work on multiselection.

### DIFF
--- a/client/ayon_core/style/style.css
+++ b/client/ayon_core/style/style.css
@@ -44,10 +44,6 @@ QLabel {
     background: transparent;
 }
 
-QLabel[overriden="1"] {
-    color: {color:font-overridden};
-}
-
 /* Inputs */
 QAbstractSpinBox, QLineEdit, QPlainTextEdit, QTextEdit {
     border: 1px solid {color:border};
@@ -1589,6 +1585,10 @@ CreateNextPageOverlay {
 }
 
 /* Attribute Definition widgets */
+AttributeDefinitionsLabel[overridden="1"] {
+    color: {color:font-overridden};
+}
+
 AttributeDefinitionsWidget QAbstractSpinBox, QLineEdit, QPlainTextEdit, QTextEdit {
     padding: 1px;
 }
--- a/client/ayon_core/tools/attribute_defs/__init__.py
+++ b/client/ayon_core/tools/attribute_defs/__init__.py
@@ -1,6 +1,7 @@
 from .widgets import (
     create_widget_for_attr_def,
     AttributeDefinitionsWidget,
+    AttributeDefinitionsLabel,
 )
 
 from .dialog import (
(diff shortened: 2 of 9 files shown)


BigRoy (approved):
Didn't check the code - but this does appear to work in Fusion 19!

image

Would still love to be able to see a shortcut label on the Action like CTRL+MMB and then be able to CTRL+MMB on the attribute to reset it that way - because it'd just be way quicker. But could be follow up PR.

Aside of that, this PR does have way more lines of code added than i anticipated for this feature? Any quick explanation on why it warrants this amount of changes? What was the biggest refactor needed for this?

iLLiCiTiT:
> Aside of that, this PR does have way more lines of code added than i anticipated for this feature? Any quick explanation on why it warrants this amount of changes? What was the biggest refactor needed for this?

What? I had to change widget of each attribute definition, in case of file input it required changes in 4 widgets, it's a lot less than I expected.

BigRoy:
I figured initially that it would've just been a call to the attribute def object like a `attr.revert_to_default()` type of call. But I suppose the Attribute Defs are so disconnected from the widgets and the data that it requires this additional 'layer' of code on top. 

It's also other things, like why do we need the wrapper widget on the file widget.

Also we seem to use the AttributesWidget that contains the attribute widgets themselves to define if something is overridden or not - but I had expected that to be the responsibility of the individual widgets knowing about their default and their current value changes - hence being aware about whether it's default or not.

But I suppose there are a myriad of reasons for the way it is and it's totally fine with me since I'm not seeing any performance issues or bugs due to it - it just got me by surprise. That's all.

Good to go. 🚀

iLLiCiTiT:
> I figured initially that it would've just been a call to the attribute def object like a attr.revert_to_default() type of call. But I suppose the Attribute Defs are so disconnected from the widgets and the data that it requires this additional 'layer' of code on top.

It's not that simple, there is not a 1:1 match on multiselection of instances.

> It's also other things, like why do we need the wrapper widget on the file widget.

You can't show context menu on disabled widget, which we do on multiselection. Instead of disabling whole widget it disables only the wrapper widget, so the rever to default still works.

> Also we seem to use the AttributesWidget that contains the attribute widgets themselves to define if something is overridden or not - but I had expected that to be the responsibility of the individual widgets knowing about their default and their current value changes - hence being aware about whether it's default or not.

Widget `AttributeDefinitionsWidget` is not used in Publisher, it is used e.g. in loader if `options` are defined with attr defs. And default implementation of "revert to default" is on those widgets as you described, but it works differently in Publisher. Also labels are not part of the attr widget, those are defined by parent that's why `AttributeDefinitionsWidget` also has to trigger revert to default if it is triggered on label related to the widget.

PR #988, opened 2024-11-06
ExtractOIIOTranscode: Log warning instead of raise error if media can't be transcoded due to unknown RGBA channels

BigRoy:
## Changelog Description

Allow ExtractOIIOTranscode to pass with a warning if it can't find the RGBA channels in source media instead of raising error

## Additional info

Over time we've had annoying reports where studios were unable to publish their renders because the publishing system (due to user configuration) tries to transcode e.g. a Z-depth AOV layer which may only have a named `Z` or `Y` channel and has no `R`, `G`, `B` or `A` channels and hence we don't really know what layer to process.

Instead of this we could also implement detecting more cases, so that if `RGBA` is not found in the file that maybe we allow to fall back to channels:
- XYZ
- Y
- Z (usually for z-depth?)

Here's metadata output for a file that fails without this PR:

```
<nchannels>1</nchannels>
<channelnames>
<channelname>Y</channelname>
</channelnames>
<alpha_channel>-1</alpha_channel>
<z_channel>-1</z_channel>
<deep>0</deep>
```
It's an Arnold Z-depth render AOV.

## Testing notes:

1. Unknown RGBA channels will now log a warning instead of disallow publishing altogether.


iLLiCiTiT:
I wonder how it ends up being marked for review if it cannot be used for review?

client/ayon_core/lib/transcoding.py
```diff
@@ -67,6 +67,11 @@
 }
 
 
+class UnknownRGBAChannelsError(ValueError):
```
iLLiCiTiT, line 70:
```suggestion
class MissingRGBAChannelsError(ValueError):
```

BigRoy:
Closing this PR in favor of https://github.com/ynput/ayon-core/pull/1271 which now also includes this PR.

PR #990, opened 2024-11-06
Publisher: Refresh of create/publish values under correct conditions

iLLiCiTiT:
## Changelog Description
Fixed conditions triggering refresh of values in UI.

## Additional info
I've noticed that the condition that should trigger refresh of values in UI has opposite condition to what it should be. Both create and publish attributes should be updated if happened changes in those keys.

## Testing notes:
As far as I know this feature is not used at this moment. It should be possible to fake it. For example in Create plugin, register listen to value changes and make sure instance has 2 checkboxes, when checkbox 1 changes value, also change value of checkbox 2 to the same value in the callback. That change done by create plugin should be propagated to UI.

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/product_attributes.py
+++ b/client/ayon_core/tools/publisher/widgets/product_attributes.py
@@ -232,7 +232,7 @@ class CreatorAttrsWidget(QtWidgets.QWidget):
         for instance_id, changes in event["instance_changes"].items():
             if (
                 instance_id in self._current_instance_ids
-                and "creator_attributes" not in changes
+                and "creator_attributes" in changes
             ):
                 self._refresh_content()
                 break
@@ -498,7 +498,7 @@ class PublishPluginAttrsWidget(QtWidgets.QWidget):
         for instance_id, changes in event["instance_changes"].items():
             if (
                 instance_id in self._current_instance_ids
-                and "publish_attributes" not in changes
+                and "publish_attributes" in changes
             ):
                 self._refresh_content()
                 break


BigRoy (approved):
I guess the changed code does make sense. :) didn't test.

iLLiCiTiT:
For future reference, this was my testing create plugin in traypublisher
```python
from ayon_core.lib.attribute_definitions import BoolDef

from ayon_traypublisher.api.plugin import TrayPublishCreator


class ExampleCreator(TrayPublishCreator):
    """Creates instance from file and retains its original name."""

    identifier = "io.ayon.example"
    label = "Example test"
    product_type = "example"
    description = "Example create plugin"
    extensions = [
        ".mov", ".mp4", ".mxf", ".m4v", ".mpg", ".exr",
        ".dpx", ".tif", ".png", ".jpg"
    ]

    def get_icon(self):
        return "fa.file"

        # Create new instance
        self._create_instance(product_name, instance_data, self.product_type)

    def register_callbacks(self):
        self.create_context.add_value_changed_callback(self._on_value_change)

    def _on_value_change(self, event):
        for item in event["changes"]:
            instance = item["instance"]
            if (
                instance is None
                or instance.creator_identifier != self.identifier
            ):
                continue
            changes = item["changes"].get("creator_attributes", {})

            if "bool_value_1" in changes:
                value = changes["bool_value_1"]
                instance.creator_attributes["bool_value_2"] = value

    def get_attr_defs_for_instance(self, instance):
        return [
            BoolDef(
                "bool_value_1",
                default=True,
                label="Bool 1",
            ),
            BoolDef(
                "bool_value_2",
                default=True,
                label="Bool 2",
            ),
        ]

    def get_pre_create_attr_defs(self):
        return []

    def get_product_name(self, *args, **kwargs):
        return f"An Instance"
```

BigRoy:
In that case - it looks like [this use case](https://github.com/ynput/ayon-maya/blob/8992e3b3920a0b69c5dedb349421e0b852106028/client/ayon_maya/plugins/create/create_maya_usd.py#L20-L46)? Are you sure it didn't work before?

iLLiCiTiT:
> Are you sure it didn't work before?

Not for me in traypublisher.

PR #998, opened 2024-11-07
Fix support for scriptsmenu running commands in PySide6

BigRoy:
## Changelog Description

Fix support of scripts menu in Qt6, e.g. Maya 2025.

## Additional info

This initially tried to fix it https://github.com/ynput/ayon-core/pull/669 however it did fix the "startup" error but it did make it fail to trigger the actual command from the menu entry due to the callback check when running a menu entry.

## Testing notes:

1. Launch Maya 2025
2. Run an action from the Custom Tools menu

This should also be backwards compatible in older Python versions, so also test e.g. Maya 2022, 2023 or 2024.

I personally tested on Windows in Maya 2022.3, Maya 2024.2 and 2025.3.

### DIFF
--- a/client/ayon_core/vendor/python/scriptsmenu/action.py
+++ b/client/ayon_core/vendor/python/scriptsmenu/action.py
@@ -1,6 +1,6 @@
 import os
 
-from qtpy import QtWidgets
+from qtpy import QtWidgets, QT6
 
 
 class Action(QtWidgets.QAction):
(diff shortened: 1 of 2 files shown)


client/ayon_core/vendor/python/scriptsmenu/action.py
```diff
@@ -112,20 +112,21 @@ def run_command(self):
...
         function will trigger all callbacks. When a callback function returns a
         non zero integer it will not execute the action's command
-
         """
 
         # get the current application and its linked keyboard modifiers
         app = QtWidgets.QApplication.instance()
         modifiers = app.keyboardModifiers()
+        if not QT6:
+            modifiers = int(modifiers)
```
iLLiCiTiT, line 124: We could use the advantage that modifiers in PySide6 are enum items (they have `name` and `value` attributes).
```suggestion
        if not hasattr(modifiers, "value"):
            modifiers = int(modifiers)
```

BigRoy: Feels a bit less explicit than it being Qt6 related - why do you feel that's nicer?

iLLiCiTiT: I didn't say nicer, but it does not require import of `QT6`.

iLLiCiTiT (approved):

Issue #886, opened 2024-09-12
Manager: Filter to outdated toggle does not work


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When **Filter to outdated** is enabled no filtering occurs.

![filter_to_outdated](https://github.com/user-attachments/assets/c936c0a3-1740-48d3-aeb9-63e98cd53590)

Reproduced in Maya 2024.2 and Fusion 18 - so I suppose issue is in core.
There is no logged error.

### Expected Behavior:

Only the "red versioned" entries should remain listed with the filter enabled.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Load multiple versions
2. Enable "Filter to outdated"

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

LiborBatek:
Quite important to fix it...yet over month without any progress @iLLiCiTiT

PR #999, opened 2024-11-07
Scene manager: Filter to outdated toggle does work

iLLiCiTiT:
## Changelog Description
Filter to outdated works again in scene inventory.

## Testing notes:
1. Open host of your choice.
2. Open loader.
3. Load some representations that are **not** from latest version and some that are from latest version.
4. In Scene manager check `Filter to outdated`.
5. It should keep only those that are outdated.

Resolves https://github.com/ynput/ayon-core/issues/886


BigRoy (approved):
Tested in Maya 2025 and Fusion 19. Works! 🚀

LiborBatek (approved):
Have been tested in ```maya 2024``` and all working nicely.

Issue #983, opened 2024-11-05
Load and manage products from a library project


BigRoy:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

The loader is currently able to load products from another (library) project. However, those have traditionally **never** been able to remain 'managed' and updated.

I'd like to propose a means to allow loading from another project and keep the full management working by persisting the `product_name` with the container data.

### How would you imagine the implementation of the feature?

Upon loading, we also imprint the loaded container with the `project_name` data alongside the representation id.

That way, anything that takes the `container` like `update_container` to set versions or alike is capable of also using the project name data to update correctly to a new version. Due to that we don't need to customize the function signature for `update_container`

There will be some more we'll need to do:

Required:

- Loaders will need to be able to imprint the `project_name` data.
- The `Loader.filepath_from_context` method should be tested with it's correctly resolving for another context than the current project.
- The Scene Inventory (manager) UI should be capable of also listing containers with this `project_name` metadata and also query the data/info regarding the other project.

Not critical for MVP:

- The "Switch Asset" UI may need to also support the `project_name` to allow switching these containers from another project too.
- The "Collect Scene Inputs" and [their input versions](https://github.com/ynput/ayon-core/blob/f54ff5622e8e5e00de3531a71006c224cd30c001/client/ayon_core/plugins/publish/collect_input_representations_to_versions.py#L6-L59) are collected solely using representation id for "links". To make that work correctly it _may_ need to also allow representation ids NOT part of the current project also with this `project_name` data attached to it to correctly identify it.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

moonyuet:
I studied a little bit with the code. The issue possibly contributes to the following function can't manage to find versions and the name of the project as we can always stick with searching the representation and version through the current project name. 
https://github.com/ynput/ayon-core/blob/4b5573d0a4ca1bc738ea55bfe451e5ee2a01e826/client/ayon_core/tools/sceneinventory/models/containers.py#L222-L243

https://github.com/ynput/ayon-core/blob/4b5573d0a4ca1bc738ea55bfe451e5ee2a01e826/client/ayon_core/tools/sceneinventory/models/containers.py#L297-L307

PR #1005, opened 2024-11-11
Load and manage products from a library project

moonyuet:
## Changelog Description
This PR is to support loading assets from the library project while the users can update and set versions with the containers from the loaded assets in the scene inventory tools. In the DCC hosts, you need to add the `project_name` as part of the container data, otherwise it will use the current project name instead(which results to the Entity N/A shown in the #983)
Resolve https://github.com/ynput/ayon-core/issues/983

## Additional info
Make sure using the self.filepath_from_context(context) for the updating function in the loaders.

## Testing notes:
1. Load Assets from the current project and library projects
2. Go to scene inventory to set version or update
3. Should be working.


client/ayon_core/tools/sceneinventory/control.py
```diff
@@ -4,7 +4,7 @@
 from ayon_core.host import HostBase
 from ayon_core.pipeline import (
     registered_host,
-    get_current_context,
+    get_current_context
```
iLLiCiTiT, line 7:
```suggestion
    get_current_context,
```

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -93,13 +94,15 @@ def __init__(
...
     ):
         self.representation_id = representation_id
         self.loader_name = loader_name
         self.object_name = object_name
         self.namespace = namespace
         self.item_id = item_id
+        self.project_name = project_name
 
     @classmethod
     def from_container_data(cls, container):
```
iLLiCiTiT, line 108: You should pass in the project name if container data is not enough.

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
     def get_representation_info_items(self, representation_ids):
         output = {}
         missing_repre_ids = set()
+        missing_repre_ids_by_project = {}
+        containers = self._controller.get_containers()
```
iLLiCiTiT, line 231:
```suggestion
        containers = self.get_containers()
```

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
...
         for repre_id in representation_ids:
             try:
                 uuid.UUID(repre_id)
             except ValueError:
                 output[repre_id] = RepresentationInfo.new_invalid()
                 continue
 
+            project_name = self._find_project_name(containers, repre_id)
+            if project_name is None:
+                project_name = self._controller.get_current_project_name()
```
iLLiCiTiT, line 242: Prepare this value outside of the loop.
```
current_project_name = self._controller.get_current_project_name()
for repre_id in representation_ids:
    ...
    project_name = self._find_project_name(containers, repre_id)
    if project_name is None:
        project_name = current_project_name

File: client/ayon_core/tools/sceneinventory/models/containers.py [hunk 5]
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
...
+
             repre_info = self._repre_info_by_id.get(repre_id)
             if repre_info is None:
                 missing_repre_ids.add(repre_id)
+                missing_repre_ids_by_project.update({project_name: repre_id})
```
iLLiCiTiT, line 247:
```suggestion
                missing_repre_ids_by_project.setdefault(
                    project_name, set()
                ).add(repre_id)
```

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
...
-                kwargs["version_id"] = version["id"]
-            if repre:
-                kwargs["representation_name"] = repre["name"]
-
-            repre_info = RepresentationInfo(**kwargs)
-            self._repre_info_by_id[repre_id] = repre_info
-            output[repre_id] = repre_info
+        for project_name, missing_ids in missing_repre_ids_by_project.items():
+            repre_hierarchy_by_id = get_representations_hierarchy(
+                project_name, {missing_ids}
```
iLLiCiTiT, line 256:
```suggestion
                project_name, missing_ids
```

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
...
                 output[repre_id] = RepresentationInfo.new_invalid()
                 continue
 
+            project_name = self._find_project_name(containers, repre_id)
+            if project_name is None:
+                project_name = self._controller.get_current_project_name()
+
             repre_info = self._repre_info_by_id.get(repre_id)
             if repre_info is None:
                 missing_repre_ids.add(repre_id)
```
iLLiCiTiT, line 246: Do we need `missing_repre_ids` for anything?

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
...
+                missing_repre_ids_by_project.update({project_name: repre_id})
             else:
                 output[repre_id] = repre_info
 
         if not missing_repre_ids:
```
iLLiCiTiT, line 251:
```suggestion
        if not missing_repre_ids_by_project:
```

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -293,20 +307,37 @@ def get_version_items(self, product_ids):
 
             def version_sorted(entity):
                 return entity["version"]
+            containers = self.get_containers()
+            for repre_id in representation_ids:
```
iLLiCiTiT, line 311: This change makes it uber slow when there is a lot of loaded containers in the scene.

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -222,61 +227,70 @@ def get_container_items_by_id(self, item_ids):
...
+                if repre:
+                    kwargs["representation_name"] = repre["name"]
+
+                repre_info = RepresentationInfo(**kwargs)
+                self._repre_info_by_id[repre_id] = repre_info
+                output[repre_id] = repre_info
         return output
 
-    def get_version_items(self, product_ids):
+    def get_version_items(self, product_ids, representation_ids):
```
iLLiCiTiT, line 292: Why do we need to pass in representation ids?

iLLiCiTiT:
I'm afraid you chose easy way, which does complicate things a lot.

One we decide we want to work with different projects on containers, we have to work with that in UI and api too. I'll use one example `get_version_items` which did expect product ids, but now also expects representation ids, those representation ids are used to find corresponding container, to find corresponding project. That is by design bad, that method should expect project name and product ids, when you call it, you should have prepared mapping to which projects which product ids belong.

client/ayon_core/tools/sceneinventory/models/containers.py
BigRoy: This could be much prettier when just using a `defaultdict(set)`?

client/ayon_core/tools/sceneinventory/models/containers.py
BigRoy: Yes - we should really group by project first, then batch the queries that way. I'd say this is required to fix within this PR

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -337,12 +368,18 @@ def version_sorted(entity):
...
             product_id: dict(self._version_items_by_product_id[product_id])
             for product_id in product_ids
         }
 
+    def _find_project_name(self, containers, representation_id):
+    # Function to find the project name by representation
+        for container in containers:
+            if container.get('representation') == representation_id:
+                return container.get('project_name', get_current_project_name())
+        return None
```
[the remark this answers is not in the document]
BigRoy: Ah yes, re-reading that now. That'd be the same.

We would either call separate per project, like `get_version_items(project, representation_ids)` or have a dataclass like:
```python
class RepresentationInfo:
    project_name: str
    representation_id: str
```

And we'd be passing `List[RepresentationInfo]` to the method instead.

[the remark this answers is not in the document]
BigRoy: In that case even better - if we can pass the full container 👍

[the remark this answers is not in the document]
iLLiCiTiT: Again, no, please don't complicate thigs. Existing metho `get_version_items` expects product ids, with this change it should expect project name and product ids. The UI logic is capable of collecting that data and calling the method accordingly, only change, add `project_name` to `get_version_items`.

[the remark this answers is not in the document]
BigRoy: So we're grouping the `product_ids` by project name elsewhere then? There may be many different projects across the different containers (and hence the product ids)

So we would need to know for EACH product id - to which project it belongs.. or make sure to group the product ids per project name outside of it and query it each time per project. I don't see how it would work otherwise?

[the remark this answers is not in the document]
iLLiCiTiT: This conversation is just confusing Kayla.

> So we're grouping the product_ids by project name elsewhere then? There may be many different projects across the different containers (and hence the product ids)

We're not, we should.

> There may be many different projects across the different containers (and hence the product ids)

Yes, one list of product ids per project == one call of `get_version_items`.

> So we would need to know for EACH product id - to which project it belongs.. or make sure to group the product ids per project name outside of it and query it each time per project. I don't see how it would work otherwise?

I told MANY times, all places using `get_version_items` have access to container items (and are based on them) contanier items have project name on them. So all the places can collect project > product ids mapping themselves and call the method accordingly to get version items, controller changes 2 things, adds project name to arguments and removed line `project_name = self._controller.get_current_project_name()`.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -150,7 +153,7 @@ def refresh(self, selected=None):
             if repre_info.is_valid
         }
         version_items_by_product_id = self._controller.get_version_items(
-            product_ids
+            product_ids, project_names
         )
```
BigRoy, line 157: I think this is the wrong approach because we'd need to know which product id belongs to which project. What @iLLiCiTiT mentioined before is to separate the query to `self._controller.get_version_items()` to have separate calls per project. So above, we'd get the product ids per project and then here we would do:

```python
version_items_by_product_id = self._controller.get_version_items(project_name, product_id)
```

As stated before - as far as I know on the backend nothing is keeping the database from having a product entity with the same `id` across different projects. Is that correct @martastain ? (It's unlikely by just creating products to happen, but when cloning e.g. projects I suppose it may very well share the same ids?) As such, we should still make sure to be explicit about which project each entry belongs to.

The same goes for `version_items_by_product_id` dict. That should technically also be "per project" because otherwise `product_id` may overlap accidentally.

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -221,70 +230,81 @@ def get_container_items_by_id(self, item_ids):
...
+                    kwargs["representation_name"] = repre["name"]
+
+                repre_info = RepresentationInfo(**kwargs)
+                self._repre_info_by_id[repre_id] = repre_info
+                self._product_id_by_project[project_name] = repre_info.product_id
+                output[repre_id] = repre_info
         return output
 
-    def get_version_items(self, product_ids):
+    def get_version_items(self, product_ids, project_names):
```
iLLiCiTiT, line 295: Please swap the arguments so project is first.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -150,7 +153,7 @@ def refresh(self, selected=None):
             if repre_info.is_valid
         }
         version_items_by_product_id = self._controller.get_version_items(
-            product_ids
+            product_ids, project_names
```
iLLiCiTiT, line 156: I thought you'll call this method for each project name once.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -208,13 +208,15 @@ def _build_item_menu_for_selection(self, menu, indexes, active_index):
         filtered_items = []
         product_ids = set()
         version_ids = set()
+        project_names = set()
```
iLLiCiTiT, line 211: You should store product ids by project name and call `get_version_items` per project.

BigRoy (commented):
Aside of their being plenty of code style notes I gave this a quick test run - and it seems to work fine.

Opening the scene inventory with containers from another project showed these entities just fine. However, visually there's NO clue whatsoever in the scene inventory that these are from another project.

I wonder whether it warrants somehow highlighting them. I was thinking of adding a `project_name` column to the Scene Inventory UI and potentially highlighting the name AYON blue using a delegate if the project name is not the current project.

But this is looking really nice.

I tested it with this PR: https://github.com/ynput/ayon-maya/pull/176

client/ayon_core/tools/sceneinventory/model.py
iLLiCiTiT: Yes

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -149,9 +152,20 @@ def refresh(self, selected=None):
             for repre_info in repre_info_by_id.values()
             if repre_info.is_valid
         }
-        version_items_by_product_id = self._controller.get_version_items(
-            product_ids
-        )
+
+        project_products = {project_name: set() for project_name in project_names}
```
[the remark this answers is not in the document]
moonyuet: We can consider using that.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -226,10 +230,12 @@ def _build_item_menu_for_selection(self, menu, indexes, active_index):
...
-
-        version_items_by_product_id = self._controller.get_version_items(
-            product_ids
-        )
+        version_items_by_product_id = {}
+        for project_name, product_ids in project_products.items():
+            version_items_by_product_id.update(
+                self._controller.get_version_items(
+                project_name, product_ids)
+            )
```
BigRoy, line 238: This resulting dict is technically invalid - a project id may be re-used across projects - it's not enforced to be unique.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -927,24 +947,37 @@ def update_all(self):
...
+        project_name_to_repre_ids = {}
+        for container_item in containers_items_by_id.values():
+            project_name = container_item.project_name
+            repre_id = container_item.representation_id
+            if project_name not in project_name_to_repre_ids:
+                project_name_to_repre_ids[project_name] = set()
+            project_name_to_repre_ids[project_name].add(repre_id)
+
+        # Get representation info items by ID
+        repre_info_by_id = self._controller.get_representation_info_items(repre_ids)
```
BigRoy, line 966: This likely would also need to run per project and receive a project name so we know which project to query in?

BigRoy (changes requested):

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -358,6 +373,7 @@ def _update_cache(self):
         container_items = []
         containers_by_id = {}
         container_items_by_id = {}
+        project_name_by_repre_id = {}
```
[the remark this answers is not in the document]
moonyuet: can you suggest any option to look up without repre id?

client/ayon_core/tools/sceneinventory/view.py
BigRoy: @iLLiCiTiT this would only ever make sense if `get_representation_info_items` takes `project_name` too, right?

Meaning all the logic here would just run per project.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -206,16 +206,20 @@ def _build_item_menu_for_selection(self, menu, indexes, active_index):
         # Exclude items that are "NOT FOUND" since setting versions, updating
         # and removal won't work for those items.
         filtered_items = []
-        product_ids = set()
+        project_products = {}
```
iLLiCiTiT, line 209:
```suggestion
        product_ids_by_project = collections.defaultdict(set)
```

client/ayon_core/tools/sceneinventory/view.py
iLLiCiTiT: You meant product id? If yes, I would not bother (at least not now).

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,6 +741,7 @@ def _show_version_dialog(self, item_ids, active_repre_id):
         container_items_by_id = self._controller.get_container_items_by_id(
             item_ids
         )
+        print(container_items_by_id, "container")
```
iLLiCiTiT, line 744: Please remove all the print statements.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -747,12 +754,25 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
+        for container_item in container_items_by_id.values():
+            repre_id = container_item.representation_id
+            project_name = container_item.project_name
+            repre_info = repre_info_by_id.get(repre_id)
+            if repre_info and repre_info.is_valid:
+                if project_name not in project_products:
+                    project_products[project_name] = set()
+                    product_id = repre_info.product_id
+                    project_products[project_name].add(product_id)
+        print("proj_product", project_products)
```
iLLiCiTiT, line 767:
```suggestion
        product_ids_by_project = collections.defaultdict(set)
        for container_item in container_items_by_id.values():
            repre_id = container_item.representation_id
            project_name = container_item.project_name
            repre_info = repre_info_by_id.get(repre_id)
            if not repre_info or not repre_info.is_valid:
                continue
            product_ids_by_project[project_name].add(
                repre_info.product_id
            )

```

client/ayon_core/tools/sceneinventory/view.py
iLLiCiTiT: I'm afraid so.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -747,12 +754,25 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
         active_version_id = active_repre_info.version_id
         active_product_id = active_repre_info.product_id
-        version_items_by_product_id = self._controller.get_version_items(
-            product_ids
-        )
+        version_items_by_product_id = {}
+        for project_name, product_ids in project_products.items():
+            version_items_by_product_id.update(
+                self._controller.get_version_items(
+                project_name, product_ids))
```
iLLiCiTiT, line 775: Much easier to read...
```suggestion
            version_items = self._controller.get_version_items(
                project_name, product_ids
            )
            version_items_by_product_id.update(version_items)
                
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -927,24 +947,37 @@ def update_all(self):
...
-            for repre_info in repre_info_by_id.values()
-            if repre_info.is_valid
+        # Extract project names and their corresponding representation IDs
+        project_name_to_repre_ids = {}
+        for container_item in containers_items_by_id.values():
+            project_name = container_item.project_name
+            repre_id = container_item.representation_id
+            if project_name not in project_name_to_repre_ids:
+                project_name_to_repre_ids[project_name] = set()
+            project_name_to_repre_ids[project_name].add(repre_id)
```
iLLiCiTiT, line 963:
```suggestion
        repre_ids_by_project = collections.defaultdict(set)
        for container_item in containers_items_by_id.values():
            project_name = container_item.project_name
            repre_id = container_item.representation_id
            repre_ids_by_project[project_name].add(repre_id)
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -927,24 +947,37 @@ def update_all(self):
...
+
+        # Get representation info items by ID
+        repre_info_by_id = self._controller.get_representation_info_items(repre_ids)
+
+        # Create a dictionary to map project names to sets of product IDs
+        project_products = {
+            project_name: set() for project_name in project_name_to_repre_ids.keys()
         }
```
iLLiCiTiT, line 971:
```suggestion
        product_ids_by_project = {
            project_name: set()
            for project_name in repre_ids_by_project.keys()
        }
```

client/ayon_core/tools/sceneinventory/models/containers.py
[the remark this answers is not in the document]
iLLiCiTiT: Require project name where you have only representation id.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -129,29 +129,54 @@ def refresh(self, selected=None):
 
         self._clear_items()
 
-        items_by_repre_id = {}
+        items_by_repre_id = collections.defaultdict(list)
```
BigRoy, line 132: This should be `items_by_project_and_repre_id` because `repre_id` may not be unique across projects. So two projects may use the same repre id, but would need different `items`.

BigRoy: This may be better off as a nested instead:

```python
item_by_repre_id_by_project_id = defaultdict(lambda: defaultdict(set))
for container_item in container_items:
        project_name = (
            container_item.project_name or
            self._controller.get_current_project_name()
        )
        repre_id = container_item.representation_id
        item_by_repre_id_by_project_id[project_name][repre_id].add(container_item)
``` 
Maybe like this ☝️

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -219,72 +226,79 @@ def get_container_items_by_id(self, item_ids):
             for item_id in item_ids
         }
 
-    def get_representation_info_items(self, representation_ids):
+    def get_representation_info_items(self, project_name, representation_ids):
         output = {}
-        missing_repre_ids = set()
+        missing_repre_ids_by_project = {}
```
BigRoy, line 231: Funnily enough - this now doesn't need to be "per project" anymore - because the full method operates within the project's context. (by having a `project_name` input)

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -219,72 +226,79 @@ def get_container_items_by_id(self, item_ids):
...
+        if project_name is None:
+            project_name = self._controller.get_current_project_name()
```
BigRoy, line 233: Calling this function should be called in a way that representation ids are pre-grouped by project name. So it should never be allowed to be `None`.  Right?
```suggestion
```

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -129,29 +129,54 @@ def refresh(self, selected=None):
...
         for container_item in container_items:
             # if (
             #     selected is not None
             #     and container_item.item_id not in selected
             # ):
             #     continue
+            project_name = (
+                container_item.project_name or
+                self._controller.get_current_project_name()
+            )
```
iLLiCiTiT, line 143: I though `project_name` is filled with current project on container item creation.
```suggestion
            project_name = container_item.project_name
```

client/ayon_core/tools/sceneinventory/models/containers.py
iLLiCiTiT: I meant to pass it the current project, not to use `None`.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -52,6 +53,7 @@ class InventoryModel(QtGui.QStandardItemModel):
         "Object name",
         "Active site",
         "Remote site",
+        "Project"
```
iLLiCiTiT, line 56:
```suggestion
        "Project",
```

client/ayon_core/tools/sceneinventory/models/containers.py
iLLiCiTiT: I meant to pass current project name as argument to `from_container_data` method.

```python
@classmethod
    def from_container_data(cls, container):
        return cls(
            representation_id=container["representation"],
            loader_name=container["loader"],
            namespace=container["namespace"],
            object_name=container["objectName"],
            item_id=uuid.uuid4().hex,
            project_name=container.get(
                "project_name", get_current_project_name()
            )
        )
```

client/ayon_core/tools/sceneinventory/model.py
moonyuet: I find out that using this code would cause the scene inventory loading the invalid name and version(Entity n/a) It can contribute to this below https://github.com/ynput/ayon-core/pull/1005#discussion_r1843904610

client/ayon_core/tools/sceneinventory/models/containers.py
iLLiCiTiT: > maybe we can do this in separate PR as enhancement.

We should implement the support correctly. It should not be that hard, it requires a lot of changes in code, but in practice it means we just have to consider project name as one hierarchy level above everything else (everywhere). If you struggle to achieve that, let me know.

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -358,10 +358,12 @@ def _update_cache(self):
         container_items = []
         containers_by_id = {}
         container_items_by_id = {}
+        project_cache = collections.defaultdict(dict)
```
iLLiCiTiT, line 361: Why this was introduces? You don't need to store container items by project.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,24 +745,41 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
+        for container_item in container_items_by_id.values():
+            repre_id = container_item.representation_id
+            project_name = container_item.project_name
+            repre_ids_by_project[project_name].add(repre_id)
+        repre_info_by_id = {}
+        for project_name, repre_ids in repre_ids_by_project.items():
+            repre_info = self._controller.get_representation_info_items(
+                project_name, repre_ids
+            )
+            repre_info_by_id.update(repre_info)
```
iLLiCiTiT, line 758:
```suggestion
            repre_info_by_id[project_name] = repre_info
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,24 +745,41 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
-        }
-        repre_info_by_id = self._controller.get_representation_info_items(
-            repre_ids
-        )
+        repre_ids_by_project = collections.defaultdict(set)
+        for container_item in container_items_by_id.values():
+            repre_id = container_item.representation_id
+            project_name = container_item.project_name
+            repre_ids_by_project[project_name].add(repre_id)
+        repre_info_by_id = {}
```
iLLiCiTiT, line 753:
```suggestion
        repre_info_by_project = {}
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,24 +745,41 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
 
         product_ids = {
             repre_info.product_id
             for repre_info in repre_info_by_id.values()
         }
+        product_ids_by_project = collections.defaultdict(set)
+        for container_item in container_items_by_id.values():
+            repre_id = container_item.representation_id
+            project_name = container_item.project_name
+            repre_info = repre_info_by_id.get(repre_id)
```
iLLiCiTiT, line 768:
```suggestion
            project_name = container_item.project_name
            repre_info_by_id = repre_info_by_project.get(project_name)
            repre_id = container_item.representation_id
            repre_info = repre_info_by_id.get(repre_id)
```

iLLiCiTiT: The same for version, that will be more complicated.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,42 +749,58 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
-        repre_info_by_id = self._controller.get_representation_info_items(
-            repre_ids
-        )
+        repre_ids_by_project = collections.defaultdict(set)
+        for container_item in container_items_by_id.values():
+            repre_id = container_item.representation_id
+            project_name = container_item.project_name
+            repre_ids_by_project[project_name].add(repre_id)
+
+        versions = set()
```
BigRoy, line 758: Anything using this `set` down the line wouldn't know which version id the project belongs to. This should be per project because each project may have overlapping version ids. Also, please name this variable better. What is this storing? I assume version ids? If so, name this `version_ids` - if not then I'm not sure what this will be holding.

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -126,32 +131,41 @@ def refresh(self, selected=None):
         """Refresh the model"""
         # for debugging or testing, injecting items from outside
         container_items = self._controller.get_container_items()
-
         self._clear_items()
-
-        items_by_repre_id = {}
+        repre_id = set()
```
BigRoy, line 135: Whatever will be using this ids further down wouldn't know about which project the `repre_id` belongs to - introducing potentially subtle bugs. It may be good to add a comment here now and create an issue where relevant OR fix it now. Since this potentially touches site sync I may be worth adding todo comment here now and make an issue on ayon-sitesync to orchestrate fixing that down the line.

Also, this holds representation ids, so please make it plural.

```suggestion
        repre_ids = set()
```

client/ayon_core/tools/sceneinventory/models/containers.py
```diff
@@ -361,7 +361,8 @@ def _update_cache(self):
         invalid_ids_mapping = {}
         for container in containers:
             try:
-                item = ContainerItem.from_container_data(container)
+                current_project_name = self._controller.get_current_project_name()
```
BigRoy, line 364: This is better stored outside of the for loop.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,42 +749,58 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
+            )
+            version_items_by_product_id_by_project[project_name] = version_items
+            active_repre_info = repre_info_by_id[active_repre_id]
+            active_version_id = active_repre_info.version_id
+            active_product_id = active_repre_info.product_id
+            version_items = list(
+                version_items_by_product_id_by_project[project_name][active_product_id].values()
+            )
+            all_versions = {version_item.version for version_item in version_items}
+            versions.update(all_versions)
```
BigRoy, line 788:
```suggestion
            versions.update(version_item.version for version_item in version_items)
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -192,30 +192,40 @@ def _build_item_menu_for_selection(self, menu, indexes, active_index):
...
+            project_name = container_item.project_name
+            repre_info_by_id = repre_info_by_project.get(project_name)
             repre_id = container_item.representation_id
+            all_valid_repre_ids = {
+                repre_id
+                for repre_id, repre_info in repre_info_by_id.items()
+                if repre_info.is_valid
+            }
+            valid_repre_ids.update(all_valid_repre_ids)
+
```
iLLiCiTiT, line 222:
```suggestion
        product_ids_by_project = collections.defaultdict(set)
        for container_item in container_items_by_id.values():
            project_name = container_item.project_name
            repre_id = container_item.representation_id
            repre_info_by_id = repre_info_by_project[project_name]

```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -192,30 +192,40 @@ def _build_item_menu_for_selection(self, menu, indexes, active_index):
...
             repre_info = repre_info_by_id.get(repre_id)
             if repre_info and repre_info.is_valid:
                 filtered_items.append(container_item)
                 version_ids.add(repre_info.version_id)
-                product_ids.add(repre_info.product_id)
-
+                product_id = repre_info.product_id
+                product_ids_by_project[project_name].add(product_id)
```
iLLiCiTiT, line 228:
```suggestion
                product_ids_by_project[project_name].add(
                    repre_info.product_id
                )
```

iLLiCiTiT:
It should be somewhat functional. Please re-test and report bugs.

QUESTION: What to do about switch dialog? Even I'm not able to make it work across projects, and it will be PITA to allow it for different than current project.

moonyuet:
> It should be somewhat functional. Please re-test and report bugs.
> 

Works okay in my side.
image

> QUESTION: What to do about switch dialog? Even I'm not able to make it work across projects, and it will be PITA to allow it for different than current project.

But we need to think about if there is any case of which same folder path and task name across different project....

BigRoy:
Set Version does not work in scene inventory in Houdini on product from another project. Error:
```python
  File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 392, i
n <lambda>
    lambda: self._show_version_dialog(item_ids, active_repre_id)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 821, i
n _show_version_dialog
    versions |= {
                ^
  File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, i
n <setcomp>
    version_item.version
AttributeError: 'dict' object has no attribute 'version'
```

moonyuet:
> Set Version does not work in scene inventory in Houdini on product from another project. Error:
> 
> ```python
>   File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 392, i
> n <lambda>
>     lambda: self._show_version_dialog(item_ids, active_repre_id)
>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 821, i
> n _show_version_dialog
>     versions |= {
>                 ^
>   File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, i
> n <setcomp>
>     version_item.version
> AttributeError: 'dict' object has no attribute 'version'
> ```

I am testing with other hosts too(Maya, Zbrush and Unreal), it is generally not working when setting version....

MustafaJafar:
> Set Version does not work in scene inventory in Houdini on product from another project. Error:
> 
> ```python
>   File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 392, i
> n <lambda>
>     lambda: self._show_version_dialog(item_ids, active_repre_id)
>             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 821, i
> n _show_version_dialog
>     versions |= {
>                 ^
>   File "F:\dev\ayon-core\client\ayon_core\tools\sceneinventory\view.py", line 822, i
> n <setcomp>
>     version_item.version
> AttributeError: 'dict' object has no attribute 'version'
> ```

I was able to reproduce the same error.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,44 +772,71 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
+            version_items_by_project.items()
+        ):
+            product_ids_by_version = collections.defaultdict(set)
+            product_ids_by_version_by_project[project_name] = (
+                product_ids_by_version
+            )
+            versions |= {
+                version_item.version
+                for version_item in version_items_by_product_id.values()
+            }
```
MustafaJafar, line 824: do we need to have this loop? 
I think we don't need it because we are going to loop through it in the next for loop.

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -805,7 +805,7 @@ def _show_version_dialog(self, item_ids, active_repre_id):
             version_items_by_project[project_name] = version_items_by_product_id
 
         active_version_id = active_repre_info.version_id
-        # active_product_id = active_repre_info.product_id
+        active_product_id = active_repre_info.product_id
```
MustafaJafar, line 808: This line is duplicated
```suggestion
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,44 +772,71 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
+            product_ids_by_version = collections.defaultdict(set)
+            product_ids_by_version_by_project[project_name] = (
+                product_ids_by_version
+            )
+            versions |= {
+                version_item.version
+                for version_item in version_items_by_product_id.values()
+            }
+            for version_item in version_items_by_product_id.values():
                 version = version_item.version
```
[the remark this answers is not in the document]
MustafaJafar: There's also something weird here, can you notice the version item id `98ba2c91a8e311efabcacff42558f871`.. it's duplicated!

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -735,45 +772,72 @@ def _show_version_dialog(self, item_ids, active_repre_id):
...
+        product_ids_by_version_by_project = {}
+        for project_name, version_items_by_product_id in (
+            version_items_by_project.items()
+        ):
+            product_ids_by_version = collections.defaultdict(set)
+            product_ids_by_version_by_project[project_name] = (
+                product_ids_by_version
+            )
+            for version_item_by_id in version_items_by_product_id.values():
+                for version_item in version_item_by_id.values():
```
MustafaJafar, line 824: The issue is fixed, but, I'm not sure if it fixes the problem or it's just a workaround.
out of curiosity I printed `version_items_by_product_id` 
and I found out that the version id is duplicated. is it expected? 

```python
{
    '98ba2c91a8e311efabcacff42558f871': {
        '98bc68c1a8e311efb681cff42558f871': <ayon_core.tools.sceneinventory.models.containers.VersionItem object at 0x000000013B2875D0>,
        '9df32065a8e311ef9b1ccff42558f871': <ayon_core.tools.sceneinventory.models.containers.VersionItem object at 0x0000000F82DC3D10>, 
        'c7310386a8e311efb146f02f74db125e': <ayon_core.tools.sceneinventory.models.containers.VersionItem object at 0x000000013EC05C90>}
}
```

MustafaJafar: yeah, it makes sense because the first one refers to the version id of the  latest publish and dictionary refers to every version in that product.

MustafaJafar (approved):
Tested along https://github.com/ynput/ayon-houdini/pull/185 and it works.

BigRoy (approved):
Works for me now (tested Houdini and Fusion).

Would be nice however if the "Project" was inherited upwards to the parent group, so you don't need to expand the entries to see the Project name.

image

iLLiCiTiT:
What about switch dialog? Allow to show it only if items from one project are selected?

BigRoy:
> What about switch dialog? Allow to show it only if items from one project are selected?

I guess that'd be a first step yes. But honestly, I'd just add a Project field in there. If not changed by user, it acts as "do not change project", etc. just like the other fields.

iLLiCiTiT:
>I guess that'd be a first step yes. But honestly, I'd just add a Project field in there. If not changed by user, it acts as "do not change project", etc. just like the other fields. However, I imagine that's also quite a refactor - as such, we may want to leave that for separate PR.

Made it stupid and show the dialog per project (so if containers from 3 projects are selected it will show the dialog 3 times).

moonyuet:
Can we merge this or there is something else needed to be fixed?

iLLiCiTiT:
It's approved, so I guess we can.

Issue #1007, opened 2024-11-11
Loader tool: Selection changes upon just hover over Version Column


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

Currently when hovering over ```Version``` column user selection will change without even clicking other product.

As below:

![Image](https://github.com/user-attachments/assets/2a23bf8a-6ed2-49d4-b6cb-6205ec8ddd5f)

### Expected Behavior:

The selection shouldnt change without user clicking any other product(s) but stay as is.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1) Open Loader and select some product via left click
2) Hover over ```Version``` column to other product
3) Selection from step 1) will change without users will

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell

```

BigRoy:
I can't reproduce this on my end. Is this tray loader, or within a certain host?

LiborBatek:
> I can't reproduce this on my end. Is this tray loader, or within a certain host?
> 
> It does do it for me if I keep middle mouse selected when hovering it though.

been experienced in ```blender``` and ```UE```....

iLLiCiTiT:
I did replicate it in tray loader, in maya and in nuke.

PR #1008, opened 2024-11-11
Loader UI: Don't change selection on version combobox hover

iLLiCiTiT:
## Changelog Description
Product/version selection does not change on hover over version combobox.

## Additional info
It looks like views do change selection based on focus and having available editor widget. It skips that logic if editor has `NoFocus` policy. Based on documentation they should have set focus policy ti `NoFocus` as we don't want to be able to "focus" into the widget.

## Testing notes:
1. Hover over version combobox does not change selection of products in loader tool.
2. It is possible to change version in loader tool ->this should be probably tested in Qt native DCC (e.g. maya) and some non-native DCC (e.g. blender).

Resolves https://github.com/ynput/ayon-core/issues/1007


LiborBatek (approved):
Have been tested in ```maya``` ```blender``` ```unreal``` and all working as expected...

PR #1010, opened 2024-11-12
Allow CSV ingest to create new shots.

robin-ynput:
## Changelog Description
Paragraphs contain detailed information on the changes made to the product or service, providing an in-depth description of the updates and enhancements. They can be used to explain the reasoning behind the changes, or to highlight the importance of the new features. Paragraphs can often include links to further information or support documentation.

## Additional info

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_core/plugins/publish/collect_hierarchy.py
+++ b/client/ayon_core/plugins/publish/collect_hierarchy.py
@@ -13,8 +13,8 @@ class CollectHierarchy(pyblish.api.ContextPlugin):
 
     label = "Collect Hierarchy"
     order = pyblish.api.CollectorOrder - 0.076
-    families = ["shot"]
-    hosts = ["resolve", "hiero", "flame"]
+    families = ["shot", "csv_ingest_shot"]
+    hosts = ["resolve", "hiero", "flame", "traypublisher"]
 
     def process(self, context):
         project_name = context.data["projectName"]
@@ -38,8 +38,9 @@ class CollectHierarchy(pyblish.api.ContextPlugin):
             ):
                 continue
 
-            # exclude if not masterLayer True
-            if not instance.data.get("heroTrack"):
+            # exclude if not CSV ingest shot and not masterLayer True
+            if ("csv_ingest_shot" not in families and
+                not instance.data.get("heroTrack")):
                 continue
 
             shot_data = {
@@ -49,7 +50,10 @@ class CollectHierarchy(pyblish.api.ContextPlugin):
                 "folder_type": "Shot",
                 "tasks": instance.data.get("tasks") or {},
                 "comments": instance.data.get("comments", []),
-                "attributes": {
+            }
+
+            if "csv_ingest_shot" not in families:
+                shot_data["attributes"] =  {
                     "handleStart": instance.data["handleStart"],
                     "handleEnd": instance.data["handleEnd"],
                     "frameStart": instance.data["frameStart"],
@@ -60,8 +64,8 @@ class CollectHierarchy(pyblish.api.ContextPlugin):
                     "resolutionWidth": instance.data["resolutionWidth"],
                     "resolutionHeight": instance.data["resolutionHeight"],
                     "pixelAspect": instance.data["pixelAspect"],
-                },
-            }
+                }          
+
             # Split by '/' for AYON where asset is a path
             name = instance.data["folderPath"].split("/")[-1]
             actual = {name: shot_data}
--- a/client/ayon_core/plugins/publish/extract_hierarchy_to_ayon.py
+++ b/client/ayon_core/plugins/publish/extract_hierarchy_to_ayon.py
@@ -22,7 +22,7 @@ class ExtractHierarchyToAYON(pyblish.api.ContextPlugin):
 
     order = pyblish.api.ExtractorOrder - 0.01
     label = "Extract Hierarchy To AYON"
-    families = ["clip", "shot"]
+    families = ["clip", "shot", "csv_ingest_shot"]
 
     def process(self, context):
         if not context.data.get("hierarchyContext"):


iLLiCiTiT:
We're adding a lot of csv ingest specific logic into core global plugin. There is `"csv_ingest_shot"` family, why it is not `"shot"` in traypublisher? Why we don't set attributes during csv ingest?

robin-ynput:
> We're adding a lot of csv ingest specific logic into core global plugin. There is `"csv_ingest_shot"` family, why it is not `"shot"` in traypublisher? Why we don't set attributes during csv ingest, could we handle that in a different way?

That's a very good point. I tried to explain it in the other PR but let me report it here as well. 
The `"shot"` family expects a lot of timeline/resolution data that comes mostly from OTIO and the editorial host but which do not not exist for CSV ingestion. So I though about those options:
a- tweak `ayon_core` to allow CSV to re-use a subset of the `"shot"` plugins for shot creation
b- duplicate some `"shot"` core logic and create new plugins for CSV ingestion in `ayon_traypublisher`
c- Somehow compute some OTIO and resolution data from ingested material in CSV.

Imo options `b` was making me duplicate a lot of code and `c` was way out of the scope of the PR so I went with `a`.. But very happt to change this if you guys disagree or think of something better.

jakubjezek001 (approved):
Looking good

iLLiCiTiT:
100% add comments why are we doing stuff when there is csv publish for future reference (nobody will remember why it is there). My point is that we might not make it based on `csv_ingest_shot` but on some more "generic" family (without `csv` in it). Or make the logic data based instead of family based.

And maybe think about why `heroTrack` is not filled in traypublisher, and why we don't fill attributes?

jakubjezek001:
> Imo options `B` was making me duplicate a lot of code and `C` was too way out of the scope of the initial request so I went with `A`.. But very happy to change this if you guys disagree or think of something better.

I like `A` too. Perhaps if it would be possible we could just simply enhance `ayon-traypublisher\client\ayon_traypublisher\plugins\publish\collect_shot_instances.py` plugin to allow both shot type instances. Since you can identify editorial type of shot instances by "creator_identifier" instance data key, it would make more sense perhaps stau within the scope of `shot` family but change the way the data are collected in mentioned plugin. 

```
creator_identifier = instance.data["creator_identifier"]
    if creator_identifier == "editorial_shot":
          # do editorial related otio clip evaluation
	else:
		# do csv ingest logic.
```

client/ayon_core/plugins/publish/collect_hierarchy.py
```diff
@@ -13,8 +13,8 @@ class CollectHierarchy(pyblish.api.ContextPlugin):
 
     label = "Collect Hierarchy"
     order = pyblish.api.CollectorOrder - 0.076
-    families = ["shot"]
-    hosts = ["resolve", "hiero", "flame"]
+    families = ["shot", "csv_ingest_shot"]
```
iLLiCiTiT, line 16: This is context plugin, it does not need families in filters.
```suggestion
```

client/ayon_core/plugins/publish/extract_hierarchy_to_ayon.py
```diff
@@ -22,7 +22,7 @@ class ExtractHierarchyToAYON(pyblish.api.ContextPlugin):
 
     order = pyblish.api.ExtractorOrder - 0.01
     label = "Extract Hierarchy To AYON"
-    families = ["clip", "shot"]
+    families = ["clip", "shot", "csv_ingest_shot"]
```
iLLiCiTiT, line 25: The same as in Collect Hierarchy, it is context plugin, does not need families filtering.
```suggestion
```

client/ayon_core/plugins/publish/collect_hierarchy.py
```diff
@@ -38,8 +38,9 @@ def process(self, context):
             ):
                 continue
 
-            # exclude if not masterLayer True
-            if not instance.data.get("heroTrack"):
+            # exclude if not CSV ingest shot and not masterLayer True
+            if ("csv_ingest_shot" not in families and
+                not instance.data.get("heroTrack")):
```
iLLiCiTiT, line 43:
```suggestion
            # Skip masterLayer check for traypubliser CSV ingest
            if (
                "csv_ingest_shot" not in families
                and not instance.data.get("heroTrack")
            ):
```

client/ayon_core/plugins/publish/collect_hierarchy.py
```diff
@@ -49,7 +50,10 @@ def process(self, context):
                 "folder_type": "Shot",
                 "tasks": instance.data.get("tasks") or {},
                 "comments": instance.data.get("comments", []),
-                "attributes": {
+            }
+
+            if "csv_ingest_shot" not in families:
```
iLLiCiTiT, line 55:
```suggestion
            # TODO Fill in reason why we don't set attributes for csv_ingest_shot
            if "csv_ingest_shot" not in families:
```

client/ayon_core/plugins/publish/collect_hierarchy.py
```diff
@@ -49,19 +48,32 @@ def process(self, context):
...
+            for shot_attr in SHOT_ATTRS:
+                if shot_attr not in instance.data:
+                    # Shot attribute might not be defined (e.g. CSV ingest)
+                    self.log.debug(
+                        "%s shot attribute is not defined for instance.",
+                        shot_attr
+                    )
+                    continue
+
+                shot_data["attributes"][shot_attr] = instance.data[shot_attr]
```
iLLiCiTiT, line 75: Maybe we should skip even if the key is filled with `None`?
```suggestion
                attr_value = instance.data.get(shot_attr)
                if attr_value is None:                
                    # Shot attribute might not be defined (e.g. CSV ingest)
                    self.log.debug(
                        "%s shot attribute is not defined for instance.",
                        shot_attr
                    )
                    continue

                shot_data["attributes"][shot_attr] = attr_value
```

jakubjezek001 (approved):
it seems that all works as it suppose to.

client/ayon_core/plugins/publish/collect_hierarchy.py
```diff
@@ -32,36 +31,49 @@ def process(self, context):
...
             if not instance.data.get("heroTrack"):
+                self.log.debug("Skipping not a shot from hero track")
                 continue
 
             shot_data = {
                 "entity_type": "folder",
-                # WARNING Default folder type is hardcoded
-                # suppose that all instances are Shots
-                "folder_type": "Shot",
+                # WARNING unless overwritten, default folder type is hardcoded to shot
```
jakubjezek001, line 46: yeah lets keep it as it is. Thanks https://github.com/ynput/ayon-traypublisher/pull/36#issuecomment-2488175999

krishnaavril:
Hey! Can I get the csv template/sample test file, please? for testing

PR #1021, opened 2024-11-22
Editorial: Don't require 'hierarchy' key on instance

iLLiCiTiT:
## Changelog Description
Use `folderPath` to calculate `hierarchy` instead of requiring `hierarchy` on instance data.

## Additional info
The hierarchy was required from OpenPype but with AYON we already have it available on `folderPath`.

## Testing notes:
1. Editorial instances don't have to fill `hierarchy` on instance to be integrated.


jakubjezek001 (approved):
LGTM

PR #1023, opened 2024-11-25
Chore: Set environment variables for farm

iLLiCiTiT:
## Changelog Description
Moved some of logic to define farm job environment variables from ayon-deadline to ayon-core.

## Additional info
There already is plugin https://github.com/ynput/ayon-deadline/blob/develop/client/ayon_deadline/plugins/publish/global/collect_deadline_job_env_vars.py that is collecting environment variables, but that is not extedable because the constant defined in deadline should be supplied from ayon core, and most of the variables defined in the plugin should be defined by ayon-core, ayon-ftrack, ayon-shotgrid or ayon-applications addon. This is first step to be able to actually do it in the other plugins. Right now they would have to import from ayon-deadline and have dependency to ayon-deadline addon.

This PR also will take effect only with changes in this PR https://github.com/ynput/ayon-deadline/pull/67

### Future steps
- Remove those env variables from ayon-deadline.
- Use constant `FARM_JOB_ENV_DATA_KEY` defined in ayon core instead of constant defined in ayon deadline.
- Move remaining env variables to corresponding repositories (ftrack > ftrack addon, shotgrid > shotgrid addon, applications > applications addon.

## Testing notes:
1. With this PR https://github.com/ynput/ayon-deadline/pull/67 the environment variables should be set by ayon-core.

### DIFF
--- a/client/ayon_core/plugins/publish/collect_farm_env_variables.py
+++ b/client/ayon_core/plugins/publish/collect_farm_env_variables.py
@@ -14,15 +14,15 @@ class CollectCoreJobEnvVars(pyblish.api.ContextPlugin):
     def process(self, context):
         env = context.data.setdefault(FARM_JOB_ENV_DATA_KEY, {})
         for key in [
-            # AYON
             "AYON_BUNDLE_NAME",
             "AYON_DEFAULT_SETTINGS_VARIANT",
             "AYON_PROJECT_NAME",
             "AYON_FOLDER_PATH",
             "AYON_TASK_NAME",
-            "AYON_WORKDIR",
             "AYON_LOG_NO_COLORS",
             "AYON_IN_TESTS",
+            # NOTE Not sure why workdir is needed?
+            "AYON_WORKDIR",
         ]:
             value = os.getenv(key)
             if value:
--- a/client/ayon_core/plugins/publish/collect_farm_env_variables.py
+++ b/client/ayon_core/plugins/publish/collect_farm_env_variables.py
@@ -23,8 +23,6 @@ class CollectCoreJobEnvVars(pyblish.api.ContextPlugin):
             "AYON_WORKDIR",
             "AYON_LOG_NO_COLORS",
             "AYON_IN_TESTS",
-            # backwards compatibility
-            "IS_TEST",
         ]:
             value = os.getenv(key)
             if value:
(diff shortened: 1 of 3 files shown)


client/ayon_core/plugins/publish/collect_farm_env_variables.py
```diff
@@ -0,0 +1,33 @@
...
+    def process(self, context):
+        env = context.data.setdefault(FARM_JOB_ENV_DATA_KEY, {})
+        for key in [
+            # AYON
+            "AYON_BUNDLE_NAME",
+            "AYON_DEFAULT_SETTINGS_VARIANT",
+            "AYON_PROJECT_NAME",
+            "AYON_FOLDER_PATH",
+            "AYON_TASK_NAME",
+            "AYON_WORKDIR",
```
BigRoy, line 23: I know it's technically not related to this PR - but _why_ do we need to pass the workdir? Isn't this workdir path also local to the current machine and not rootless? I have a feeling we should maybe, after this PR investigate this and potentially clean that up?

client/ayon_core/plugins/publish/collect_farm_env_variables.py
```diff
@@ -0,0 +1,33 @@
...
+            "AYON_LOG_NO_COLORS",
+            "AYON_IN_TESTS",
+            # backwards compatibility
+            "IS_TEST",
```
BigRoy, line 27: Looking at the codebase I can only find this `IS_TEST` used in GlobalJobPreLoad but only for the **openpype** backwards compatibility environment. As such, should we remove this already?

BigRoy (approved):
Looks good to me - with the exception of the comments as potential follow ups.

client/ayon_core/plugins/publish/collect_farm_env_variables.py
iLLiCiTiT: Validated, and you're right, removed.

client/ayon_core/plugins/publish/collect_farm_env_variables.py
BigRoy: It may be used here by the way: 
- https://github.com/ynput/ayon-core/blob/a08a1af331163492644fa8f5713328c95c51cc44/client/ayon_core/plugins/publish/collect_rendered_files.py#L176-L187

But I guess that should still be set by the "launch context" regardless?

iLLiCiTiT: I did try to find out where it is used, and I didn't find anything. If you can create issue it would be nice.

iLLiCiTiT: > It may be used here by the way:

Yes, but is it then used for anything?

PR #1034, opened 2024-12-03
Fix extract_otio_review

robin-ynput:
## Changelog Description
Minor fix, ensure `extract_otio_review` is skipped when no OTIO clip is available.

This fixes the following issue:
image


iLLiCiTiT (approved):
Make sense

Issue #788, opened 2024-07-19
Tools: Ignore containers with invalid data/schema with a warning instead of error


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

When containers are returned by a host with invalid data, e.g. invalid representation id, or that data completely lacking or in any way mismatching with the [container schema](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/pipeline/schema/container-3.0.json) then the tools should not fail to work, but log a warning or error and continue listing those entries that it _can_ resolve correctly.

![error](https://github.com/user-attachments/assets/c77b240b-bc7b-4e47-9b9d-9edbdccdf6c6)

Which results in empty loader:

![empty_loader](https://github.com/user-attachments/assets/9bb44ea7-22e9-4b1e-a1eb-f4d2364ad1aa)

### Expected Behavior:

Log a warning or error and continue listing those entries that it _can_ resolve correctly.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

Produce some invalid container results from a hosts `get_containers()` implementation.

The exact data can be confirmed/tested with:
```python
from ayon_core.pipeline import registered_host
import pprint

host = registered_host()
for container in host.get_containers():
    pprint.pprint(container)
```

And e.g. invalid data is: 
![invalid_data](https://github.com/user-attachments/assets/fe635011-23ed-43b9-a626-1504532c089d)

Because it should adhere to this schema: https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/pipeline/schema/container-3.0.json

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Additional context:

See [issue reported on Discord here](https://discord.com/channels/517362899170230292/563751989075378201/1263834351305756753)

PR #1037, opened 2024-12-05
Loader & Scene inventory: Ignore containers with invalid representation id

iLLiCiTiT:
## Changelog Description
Loader and scene inventory should not cause issues when container in scene contains invalid representation id.

## Additional info
Capture all possible errors when validating representation ids. In case of Loader tool, we really need only the representation ids and we don't care about anything else. In scene inventory already happens some validation of the data. The validation does not fully preserve issues, but that probably should not be job of the tool.

Partially resolves https://github.com/ynput/ayon-core/issues/788 . The issue is more aboud full validation of container data, which might be actually production breaking change at this moment as we don't enforce container schema now. And it might be more work than a simple capture of unhandled exceptions. We do miss a lot of load related api, which is epic on it's own.

## Testing notes:
1. Container with invalid representation id (for example being `None`) does not break tools.

### DIFF
--- a/client/ayon_core/tools/loader/control.py
+++ b/client/ayon_core/tools/loader/control.py
@@ -382,7 +382,7 @@ class LoaderController(BackendLoaderController, FrontendLoaderController):
                 try:
                     uuid.UUID(repre_id)
                     repre_ids.add(repre_id)
-                except ValueError:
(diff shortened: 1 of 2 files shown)


antirotor (approved):
looks like my local patch!

PR #1039, opened 2024-12-05
Refactor filtering by explicit frames

kalisp:
## Changelog Description
Simplified filtering of explicit frames set by artist in Publisher UI. This should be more explicit and safer.
Removed filtering on remainders, eg. single files as there is no real safe way how to get frame pattern there and it doesn't make sense for artist to filter on single file.

## Testing notes:
1. publish in your DCC of choice as usuall
2. try set `Frames` to explicitly filter only specific frames
image

### DIFF
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -385,8 +385,7 @@ def prepare_representations(
         if skeleton_data.get("slate"):
             frame_start -= 1
 
-        files = _get_real_files_to_rendered(collection, frames_to_render)
-
+        files = _get_real_files_to_render(collection, frames_to_render)
         # explicitly disable review by user
         preview = preview and not do_not_add_review
         rep = {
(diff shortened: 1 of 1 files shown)


client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -495,31 +491,47 @@ def _get_real_frames_to_render(frames):
...
+    ["foo_v01.0001.exr"] - only explicitly requested frame returned
+    """
+    found_frame_pattern_length = collection.padding
     normalized_frames_to_render = {
         str(frame_to_render).zfill(found_frame_pattern_length)
         for frame_to_render in frames_to_render
     }
 
+    head_name = os.path.basename(collection.head)
+
```
iLLiCiTiT, line 528:
```suggestion
    normalized_filenames = {
        f"{head_name}{frame}{collection.tail}"
        for frame in normalized_frames_to_render
    }
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -495,31 +491,47 @@ def _get_real_frames_to_render(frames):
...
+    file_names = [os.path.basename(f) for f in collection]
     return [
         file_name
-        for file_name in files
+        for file_name in file_names
         if any(
```
iLLiCiTiT, line 533:
```
        if file_name in normalized_filenames
```

64qam:
Return another error:
```Traceback (most recent call last):
  File "C:\Users\qam\AppData\Local\Ynput\AYON\dependency_packages\ayon_2411211808_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\qam\AppData\Local\Ynput\AYON\addons\deadline_0.5.0+dev\ayon_deadline\plugins\publish\nuke\submit_nuke_deadline.py", line 51, in process
    self._deadline_url = instance.data["deadline"]["url"]
KeyError: 'deadline'
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -495,33 +491,49 @@ def _get_real_frames_to_render(frames):
...
-    found_frame_pattern_length = len(extracted_frame)
+    expectedFiles = [
+        "foo_v01.0001.exr",
+        "foo_v01.0002.exr",
+    ]
+    frames_to_render = '0001'
+    >>
+    ["foo_v01.0001.exr"] - only explicitly requested frame returned
+    """
+    found_frame_pattern_length = collection.padding
```
BigRoy, line 521: Do note that `clique.assemble` creates a `clique.Collection` where `padding` is `0` if it has no frame starting with `0` by default. See the `assume_padded_when_ambiguous` argument to `clique.assemble` in recent `clique` versions [here](https://clique.readthedocs.io/en/2.0.0/api_reference/index.html#clique.assemble).

The question is however whether that could be problematic. If e.g. someone has explicit frame to render 999 but the assembled sequence was 1001-1010 then padding would be zero instead of four meaning we'd get:
- 999.exr
- 1000.exr
- 1001.exr 

Instead of:
- 0999.exr
- 1000.exr
- 1001.exr

Whether that's problematic, I'm not sure - just wanted to clarify that.

BigRoy: Also, wouldn't it just work to do:

```python

# Get overlapping frame ranges
included_frames = set(collection.indexes).intersection(frames_to_render)

# Create matching collection that refers to those frames
# and get their filenames.
collection = 
collection.indexes.clear()
collection.indexes.update(included_frames)
return list(collection)
```

Note that this changes the input `collection` - if you don't want that you may need to create a copy collection or just do:
```python
included_frames = set(collection.indexes).intersection(frames_to_render)
real_collection = clique.Collection(
    collection.head,
    collection.tail,
    collection.padding,
    indexes=included_frames
)
return list(real_collection)
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -495,34 +491,41 @@ def _get_real_frames_to_render(frames):
...
-            for frame in normalized_frames_to_render
-        )
+    Example:
+    --------
+
+    expectedFiles = [
+        "foo_v01.0001.exr",
+        "foo_v01.0002.exr",
     ]
+    frames_to_render = '0001'
```
BigRoy, line 517: I think this example value is wrong because according to the type hints it's not supposed to be a `str` but `list[int]`?

BigRoy (approved):
**Didn't test** - but code changes make sense.

PR #1044, opened 2024-12-10
Move actions from modules to tray

iLLiCiTiT:
## Changelog Description
Moved `Launcher`, `Loader` and `Console` directly to tray actions and removed `ayon_core.modules`. Also added `Publish report viewer` action to `Admin` section.

## Additional info
Tools should work as before. Changed label of `Loader` to `Browser`, we've discussed it multiple times and if I remember correctly we came to conclusion that it should be Browser. Added `Pubish report viewer` action which was requested multiple times.

Because the actions are not defined by "addons" their order is not exactly the same.

## Testing notes:
1. All the actions should work as before.


MilaKudr:
One thing to discuss is the order of the tools, many users remember that the launcher was first from the top, now it is second...

MilaKudr (approved):

LiborBatek:
@iLLiCiTiT  is there a way to keep ```Launcher``` on very top of the menu? I know seems not important but it really is...to maintain the muscle memory.

@m-u-r-p-h-y  thougts?

here is the original order:

image

iLLiCiTiT:
> @iLLiCiTiT is there a way to keep Launcher on very top of the menu? I know seems not important but it really is...to maintain the muscle memory.

I can't keep it in previous order, I can move it under addons actions of above addons actions, both can be confusing. If it is under addons actions then Publisher is at different place. If it is above then e.g. ftrack would be at different place.

64qam (approved):

m-u-r-p-h-y:
> I can't keep it in previous order, I can move it under addons actions of above addons actions, both can be confusing. If it is under addons actions then Publisher is at different place. If it is above then e.g. ftrack would be at different place.

do not understand the reasoning, but if not possible, then we can do nothing.

I would also rather keep the same order from the previous one as @LiborBatek suggests

m-u-r-p-h-y:
also @MustafaJafar should be aware of the changes as it will affect the documentation . . .

iLLiCiTiT:
> do not understand the reasoning, but if not possible, then we can do nothing.

`Publisher` and `ftrack` (and potentially other) actions are added by addons (dynamically). `Launcher` and `Browser` are now defined by tray, which happens at completelly different place not knowing about other actions. I would have to hardcode dependncy on other actions by label, which is possible, but not very systematic.

Note: The order before this was defined by title of the addon, so changing `Loader` to `Browser` would mean that `Browser` would be always at the top...

BigRoy:
> > do not understand the reasoning, but if not possible, then we can do nothing.
> 
> `Publisher` and `ftrack` (and potentially other) actions are added by addons (dynamically). `Launcher` and `Browser` are now defined by tray, which happens at completelly different place not knowing about other actions. I would have to hardcode dependncy on other actions (addons) by label, which is possible, but not very systematic.
> 
> Note: The order of addon actions was defined by title of the addon, so changing `Loader` to `Browser` would mean that `Browser` would be always at the top... So I can't use "previous" logic to sort it by title without changing order of `Browser` action.
> 
> Long story short, making this better means that we have to make something a little bit worse.

Could we mark the menu entries with custom data like `order` that could be used to sort by. So the sorting becomes by:
```
def sort(action):
    if action.order is not None:
        return action.order
    # sort by label
    action.text
```

I do think the Publisher is not so high priority over the Launcher that it should be at the top - and hence I feel the change is very confusing and we should look into ways to _do_ set a preferred order.

iLLiCiTiT:
> Could we mark the menu entries with custom data like order that could be used to sort by. So the sorting becomes by:

That would require some api for actions, which we don't have.

BigRoy:
> And should be launcher always at the top? Meaning that ftrack would move under it? (Can be achieved even now.)

I personally would say so. But definitely go based off of the opinion of others here please.

LiborBatek:
```Launcher``` as the very top/first in the menu? Yes, please!

iLLiCiTiT:
And `Browser` at the bottom is ok?

Issue #1051, opened 2024-12-12
Select frame for thumbnail extraction


Dreamink-Official:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Right now when using the extract_thumbnail plugin AYON is picking the middle frame by default.
It would be nice to have an option in the settings to select a specific frame which is used for that.

### How would you imagine the implementation of the feature?

 ![Image](https://github.com/user-attachments/assets/18775457-de53-44c5-8183-8a26060c3a90)

- Add a optional field to type in the frame which should be selected for the thumbnail
- When no frame is selected it should use the middle frame by default

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

### Additional context:

I have already implemented that locally but a question about the field in the AYON publisher: How are you handling frame ranges in AYON? For example in my DCC i am starting with frame 1001. But the extraction uses the video file so there the frame count starts by 0. 

In my current implementation for that I prefere to select the frame which is displayed in my dcc so frame 1005 for example.

jakubjezek001:
Hey @Dreamink-Official thank you for the request. I was wondering if it is necessary to know exact number of the frame for thumbnail or whether it could be just explicit frame agnostic approach which is already implemented now. 

Here in settings you are able to set float number referring to what part of total 100% range the frame should be taken. Yes it is approximating a little. 

`ayon+settings://core/publish/ExtractThumbnail/duration_split`

Dreamink-Official:
For me, (as a developer), the explicit frame-agnostic approach would be fine. But my artists specifically asked for a way to select the frame shown as the thumbnail, so I added this feature.

They told me:
- Sometimes the selected frame just doesn’t represent the shot well. Being able to pick a specific frame lets them show the most important moment.
- It gives them more control over how their work is presented, which matters a lot to them.
- It’s just a bit more flexible. Most of the time, the frame-agnostic approach works fine, but when you really need to highlight something specific, this makes it possible.

BigRoy:
> But my artists specifically asked for a way to select the frame shown as the thumbnail, so I added this feature.

Any chance you can contribute back (or can share) your changes?

Dreamink-Official:
Of course! I did created a PR back in time for that (https://github.com/ynput/ayon-core/pull/1052)

PR #1052, opened 2024-12-12
Select frame for thumbnail extraction

Dreamink-Official:
Changes: Instead of using the middle frame of the video, the user can now select a specific frame for thumbnail generation. If no valid value is provided, the middle frame will still be used as a fallback.

## Changelog Description
- Added get_attribute_defs() to introduce the frame selection field in the publisher.
- Created _get_thumbnail_frame() to retrieve the correct frame.
- Adjusted _create_frame_from_video() to select the specified frame from the video instead of using timestamps.

## Additional info
This PR is related to #1051.
As mentioned in the issue, I am not entirely sure which option aligns best with the default behavior of AYON:
A. Enter the real frame number to be used as the thumbnail (e.g., 1005) (my preferred method).
B. Enter the frame number of the extracted video (e.g., 5).

In the first commit, I used my preferred method [A].

## Testing notes:
1. Create a product type that also generates a thumbnail.
2. Enter your frame number.
3. Publish.

image

### DIFF
--- a/client/ayon_core/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_core/plugins/publish/extract_thumbnail.py
@@ -14,13 +14,17 @@ from ayon_core.lib import (
 
     path_to_subprocess_arg,
     run_subprocess,
+
+    TextDef,
 )
+from ayon_core.pipeline import AYONPyblishPluginMixin
+
 from ayon_core.lib.transcoding import convert_colorspace
 
 from ayon_core.lib.transcoding import VIDEO_EXTENSIONS
 
 
-class ExtractThumbnail(pyblish.api.InstancePlugin):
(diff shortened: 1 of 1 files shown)


jakubjezek001:
hey @technicaldirector thank you for you submission. Apologize for the delay. We will be dealing with the code shortly ;)

moonyuet (approved):
Published the selected frames as thumbnail as expected.
image

rdelillo:
Looking at the code changes, I think this overall makes perfect sense ! But I'd wait for the merge with latest code and retest it properly before giving an approval.

technicaldirector:
@moonyuet Thanks for the review! I’ll see if I can carve out some time in the next few days to look into it. Currently not working with Ayon since our film just wrapped up :)

PR #1054, opened 2024-12-15
Increment workfile even when no unsaved changes

BigRoy:
## Changelog Description

Always increment workfile when requested - instead of only when no unsaved changes

## Additional info

Fixes https://github.com/ynput/ayon-blender/issues/86
But also changes the same behavior for Maya and Blender.

## Testing notes:

1. Version up workfile menu entry should increment workfile even when no unsaved changes


moonyuet (approved):
Looks good for me.

LiborBatek (approved):
Been tested both with the menu entry and via shortcut keys and both were working nicely!

image

LGTM!

PR #1056, opened 2024-12-16
Editorial: Fix clip_media source for review track.

robin-ynput:
## Changelog Description

This PR fixes the issue when Hiero does not publish `review` product properly.
Ensure that the following use-cases works properly:
* review disabled
* `[Clip's media]`
* `Track: current track`
* `Track: another track`

## Testing notes:
1. Setup a multi-tracks timeline in Hiero
2. Publish a timeline with review source set as `clip_media` and another one as `track:XX`
3. Ensure review media works for each scenarios

### DIFF
--- a/client/ayon_core/plugins/publish/collect_otio_subset_resources.py
+++ b/client/ayon_core/plugins/publish/collect_otio_subset_resources.py
@@ -178,7 +178,8 @@ class CollectOtioSubsetResources(
             repre = self._create_representation(
                 frame_start, frame_end, collection=collection)
 
-            if "review" in instance.data["families"]:
+            if ("review" in instance.data["families"] and
+                not instance.data.get("otioReviewClips")):
                 review_repre = self._create_representation(
                 frame_start, frame_end, collection=collection,
                 delete=True, review=True)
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/publish/collect_otio_subset_resources.py
```diff
@@ -178,7 +178,8 @@ def process(self, instance):
             repre = self._create_representation(
                 frame_start, frame_end, collection=collection)
 
-            if "review" in instance.data["families"]:
+            if ("review" in instance.data["families"] and
+                not instance.data.get("otioReviewClips")):
```
iLLiCiTiT, line 182:
```suggestion
            if (
                not instance.data.get("otioReviewClips")
                and "review" in instance.data["families"]
            ):
```

iLLiCiTiT (approved):
Changes do make sense but @jakubjezek001 please confirm.

robin-ynput:
> Changes do make sense but @jakubjezek001 please confirm.

To detail a bit the initial bug, when input source for review was set to a track the logic was adding 2 representations:
* the OTIO based review from the track (correct)
* the `clip_media` source media (wrong)

This was causing an issue when integrating the review.

jakubjezek001 (approved):
Looking good!

PR #1060, opened 2024-12-16
Refactor OTIO frame range collection

jakubjezek001:
## Additional info
- Removed unused function import.
- Added detailed logging for data updates.
- Streamlined frame range calculations and handling.
- Introduced a new class for collecting source frame ranges.
- Improved readability by cleaning up code structure.

## Testing notes:
All should work as before
Run the tests: `.\tools\manage.ps1 run pytest .\tests\ -vv`

Blocking https://github.com/ynput/ayon-traypublisher/pull/58


client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -24,11 +24,65 @@ def process(self, instance):
...
             otio_range_to_frame_range,
             otio_range_with_handles
         )
 
+        if not instance.data.get("otioClip"):
+            self.log.debug("Skipping collect OTIO frame range.")
+            return
+
+        # get basic variables
+        otio_clip = instance.data["otioClip"]
```
iLLiCiTiT, line 36:
```suggestion
        otio_clip = instance.data.get("otioClip")
        if otio_clip is None:
            self.log.debug("Skipping collect OTIO frame range.")
            return

        # get basic variables

```

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -24,11 +24,65 @@ def process(self, instance):
         # Not all hosts can import these modules.
         import opentimelineio as otio
         from ayon_core.pipeline.editorial import (
-            get_media_range_with_retimes,
             otio_range_to_frame_range,
             otio_range_with_handles
         )
 
+        if not instance.data.get("otioClip"):
```
iLLiCiTiT, line 31: Can we get rid of `hosts` filtering with this change?

robin-ynput (commented):
I gave it a try through Hiero and it all works perfectly.

My only concern would be implementation-wise, currently it does:
```
Collect OTIO Frame Ranges [resolve, hiero, flame, traypublisher]:
* frameStart
* frameEnd
* clipIn
* clipOut
* clipInH
* clipOutH
```
then
```
Collect OTIO Frame Ranges (with media range) [hiero, flame]:
* frameStart (same value)
* frameEnd
* clipIn (same value)
* clipOut (same value)
* clipInH (same value)
* clipOutH (same value)
* sourceStart
* sourceEnd
* sourceStartH
* sourceEndH
```

I feel like there is quite some code duplication but more importantly to keep a single source of truth, I'd rather prefer to have :
```
Collect OTIO Frame Ranges [resolve, hiero, flame, traypublisher]:
* frameStart
* frameEnd
* clipIn
* clipOut
* clipInH
* clipOutH
```
then
```
Collect Retimed and Source OTIO Frame Ranges [hiero, flame]:
* frameEnd (retimed)
* sourceStart
* sourceEnd
* sourceStartH
* sourceEndH
```

What do you think ?

iLLiCiTiT:
I would like to avoid hosts filtering completelly if possible, but that would probably require more changes....

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -1,93 +1,204 @@
...
+        sourceStart (int): Source media start frame
+        sourceEnd (int): Source media end frame
+        sourceStartH (int): Source media start frame with handles
+        sourceEndH (int): Source media end frame with handles
+    """
+
+    label = "Collect Source OTIO Frame Ranges"
+    order = pyblish.api.CollectorOrder - 0.07
+    families = ["shot", "clip"]
+    hosts = ["hiero", "flame"]
```
iLLiCiTiT, line 120: Shouldn't we instead use different family from product types to filter instances to avoid `hosts` filtering?

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -1,93 +1,204 @@
...
         }
         instance.data.update(data)
-        self.log.debug(
-            "_ data: {}".format(pformat(data)))
-        self.log.debug(
-            "_ instance.data: {}".format(pformat(instance.data)))
+        self.log.debug(f"Added source ranges: {pformat(data)}")
+
+
+class CollectOtioRetimedRanges(pyblish.api.InstancePlugin):
```
iLLiCiTiT, line 159: There are plugins `CollectOtioFrameRanges`, `CollectOtioSourceRanges`, `CollectOtioRetimedRanges`, is there reason why it's 3 plugins instead of one? From what I see in the code, 1 plugin could be 1 method...

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
jakubjezek001: Not sure this is part of the scope of this PR. Also not up to me to decide how the workflow is designed.

jakubjezek001: I am assuming it would be better to create Issue with redesign suggestion.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
jakubjezek001: Only reason we are splitting it into different plugins is that we need to approach collection of some attributes in particular context, with host and families filtering.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -1,93 +1,204 @@
...
-"""
+
 from pprint import pformat
 
 import pyblish.api
 
+try:
+    import opentimelineio as otio
+except ImportError:
+    raise RuntimeError("OpenTimelineIO is not installed.")
```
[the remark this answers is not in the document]
iLLiCiTiT: Just remove the try > except block, it is unnecessary to capture error just to reraise different error hiding real error message.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -1,93 +1,204 @@
...
+        instance: The instance to validate
+        logger: Logger object to use for debug messages
+
+    Returns:
+        bool: True if valid, False otherwise
+    """
+    if not instance.data.get("otioClip"):
+        logger.debug("Skipping collect OTIO range - no clip found.")
+        return False
+    return True
```
iLLiCiTiT, line 38:
```suggestion
    return True

```

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
iLLiCiTiT: We're adding more and more plugins making it more and more hard to achieve that. Plugins in ayon-core should not have any filters in `hosts`.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
iLLiCiTiT: `CollectOtioSourceRanges` and `CollectOtioRetimedRanges` have exactly same filters. Hosts filtering overall is related to this comment https://github.com/ynput/ayon-core/pull/1060/files#r1927050138 .

jakubjezek001: implemented in https://github.com/ynput/ayon-core/pull/1060/commits/971c4aef43b5dad3f74b8f8ff321ef66c2927381

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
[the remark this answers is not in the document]
jakubjezek001: implemented in [`971c4ae` (#1060)](https://github.com/ynput/ayon-core/pull/1060/commits/971c4aef43b5dad3f74b8f8ff321ef66c2927381)

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -1,93 +1,158 @@
...
+        # Convert to frames
+        src_start, src_end = otio_range_to_frame_range(otio_src_range)
+        src_start_h, src_end_h = otio_range_to_frame_range(otio_src_range_handles)
+
+        data = {
             "sourceStart": src_starting_from + src_start,
             "sourceEnd": src_starting_from + src_end - 1,
             "sourceStartH": src_starting_from + src_start_h,
             "sourceEndH": src_starting_from + src_end_h - 1,
         }
```
robin-ynput, line 140: Correct me if I'm wrong but I believe the logic to compute the source range is actually wrong:
* source range is assumed as relative to the available range which is no more the cases with our OTIO exporters (except legacy Hiero).
* frame rate difference between source (e.g. 23.976fps) and available range (e.g. 50fps) of the clip is not handled by these maths

I had a check with a Hiero export and confirmed computed values were off.
This made me wondering where are those values actually used and what are the exepcted numbers (embedded timecode, relative to start_frame for sequence, ...) ?

If we are just looking for source range as frame numbers based on the available range I'd suggest this instead.

```suggestion
        # Backward-compatibility for Hiero OTIO exporter.
        # NTSC compatibility might introduce floating rates, when these are
        # not exactly the same (23.976 vs 23.976024627685547)
        # this will cause precision issue in computation.
        # Currently round to 2 decimals for comparison,
        # but this should always rescale after that.
        rounded_av_rate = round(otio_available_range.start_time.rate, 2)
        rounded_src_rate = round(otio_src_range.start_time.rate, 2)
        if rounded_av_rate != rounded_src_rate:
            conformed_src_in = otio_src_range.start_time.rescaled_to(
                otio_available_range.start_time.rate
            )
            conformed_src_duration = otio_src_range.duration.rescaled_to(
                otio_available_range.duration.rate
            )
            conformed_source_range = otio.opentime.TimeRange(
                start_time=conformed_src_in,
                duration=conformed_src_duration
            )

        else:
            conformed_source_range = source_range

        source_start = conformed_source_range.start_time
        source_end = source_start + conformed_source_range.duration

        handle_start = otio.opentime.RationalTime(
            instance.data.get("handleStart", 0),
            source_start.rate
        )
        handle_end = otio.opentime.RationalTime(
            instance.data.get("handleEnd", 0),
            source_start.rate
        )

        source_start_h = source_start - handle_start
        source_end_h = source_end + handle_end

        data = {
            "sourceStart": source_start.to_frames(),
            "sourceEnd": source_end.to_frames() - 1,
            "sourceStartH": source_start_h.to_frames(),
            "sourceEndH": source_end_h.to_frames() - 1,
        }
```

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -1,93 +1,158 @@
...
+        """Handle retimed clip frame ranges."""
+        workfile_source_duration = instance.data.get("shotDurationFromSource")
+        frame_start = instance.data["frameStart"]
+
+        # Handle retimed clip frame range
+        retimed_attributes = get_media_range_with_retimes(otio_clip, 0, 0)
+        self.log.debug(f"Retimed attributes: {retimed_attributes}")
+
+        media_in = int(retimed_attributes["mediaIn"])
+        media_out = int(retimed_attributes["mediaOut"])
```
robin-ynput, line 154: should we also update `sourceStart, sourceEnd, sourceStartH` and `sourceEndH` from the retimed values ?

robin-ynput (changes requested):

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
jakubjezek001: I am guessing that we have to.

robin-ynput:
@jakubjezek001 @iLLiCiTiT this one should be ready to review.
I have added some automated tests to validate the computations (sequence, video and retimed).

robin-ynput (approved):

PR #1064, opened 2024-12-18
Fix OTIO range computation 23.976 fps rounding with embedded timecode

robin-ynput:
## Changelog Description
help resolve https://github.com/ynput/ayon-flame/issues/28

Fixes another of those 23.976... fps rouding issue when computing the handles and available frame range.
Issue was initially reported on [Discord](https://discord.com/channels/517362899170230292/1248298172816822424/1318535309264289873).

## Additional info
The failing computation had been added as  unit test.

## Testing notes:
1. Run the automated tests
`.\tools\manage.ps1 run pytest .\tests\`


client/ayon_core/pipeline/editorial.py
```diff
@@ -196,11 +196,11 @@ def is_clip_from_media_sequence(otio_clip):
     return is_input_sequence or is_input_sequence_legacy
 
 
-def remap_range_on_file_sequence(otio_clip, in_out_range):
+def remap_range_on_file_sequence(otio_clip, otio_range):
```
iLLiCiTiT, line 199: NOTE: We should make this backwards compatible if we did pass in `tuple[int, int]` we should still support it.

jakubjezek001 (approved):
tests were workig as expected

client/ayon_core/pipeline/editorial.py
robin-ynput: I initially wanted to turn this function protected, then realized it is also used in another plugin of `ayon-core`.
Since it's not used by any other addons, I'll simply resolve this.

PR #1066, opened 2024-12-20
Store in instance data whether the staging dir set is a custom one

BigRoy:
## Changelog Description

Store in instance data whether the staging dir set is a custom one

## Additional info

Required to fix issue https://github.com/ynput/ayon-maya/issues/194#issuecomment-2556797055 using PR https://github.com/ynput/ayon-maya/pull/204

I did wonder a bit as well whether it may have been nicer to begin with to actually store the `StagingDir` class instance on the `instance.data` instead of the individual separate keys like we're doing now. But also felt that if we do want to - it may be better refactoring that separately.

## Testing notes:

1. Test with https://github.com/ynput/ayon-maya/pull/204
2. Submitting renders should work as intended with both custom staging dir set and when disabled.

### DIFF
--- a/client/ayon_core/pipeline/publish/lib.py
+++ b/client/ayon_core/pipeline/publish/lib.py
@@ -717,6 +717,7 @@ def get_instance_staging_dir(instance):
     instance.data.update({
         "stagingDir": staging_dir_path,
         "stagingDir_persistent": staging_dir_info.persistent,
+        "stagingDir_custom": staging_dir_info.custom
     })
 
     return staging_dir_path
--- a/client/ayon_core/pipeline/staging_dir.py
+++ b/client/ayon_core/pipeline/staging_dir.py
@@ -12,6 +12,7 @@ from .tempdir import get_temp_dir
 class StagingDir:
     directory: str
     persistent: bool
+    custom: bool  # Whether the staging dir is a custom staging dir
 
 
 def get_staging_dir_config(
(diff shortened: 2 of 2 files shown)


antirotor (approved):

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -717,6 +717,7 @@ def get_instance_staging_dir(instance):
     instance.data.update({
         "stagingDir": staging_dir_path,
         "stagingDir_persistent": staging_dir_info.persistent,
+        "stagingDir_custom": staging_dir_info.custom
```
iLLiCiTiT, line 720: Why `_custom`? That does not indicate it should hold boolean, the same for the attribut, it should be `is_custom`.

BigRoy: I did it personally because it matched the `persistent` one - and as described in the PR description I'd have personally preferred to store the `staging_dir_info` data class on the instance data instead.

Anyway, should we still refactor to `is_custom`? Both this core PR and the maya one aren't in a released version yet.

antirotor: I must say that if we change the key name, we should change both. I prefer consistency and having:

```
"stagingDir_persistent": staging_dir_info.persistent,
"stagindDirIsCustom": staging_dir_info.is_custom
```
looks off to me even if it is technically correct. `StagingDir` is a dataclass with hints available in standard IDEs so even if the naming doesn't suggest boolean, hints do so I don't see that big issue with it.

iLLiCiTiT: Don't know if we can change `"stagingDir_persistent"`, because it is already used in multiple other addons that are dependent on it, whereas "is custom" is new, not yet released.

I prefer to use expected naming, instead of keeping consistency with wrong one.

We can change attribute `persistent` to `is_persistent` in backwards compatible way, but not the key on instance.

PR #1069, opened 2025-01-07
Chore: Staging dir enhancements

iLLiCiTiT:
## Changelog Description
Added typehints. Function `get_staging_dir_info` returns `StagingDir` instead of string.
 Change bool attribute names to have `is_` prefix.

## Additional info
Also changed key `stagingDir_custom` to `stagingDir_is_custom`. I would like to use `stagingDirIsCustom` but there were opinions against that would not match existing `stagingDir_persistent` which cannot be changed because of backwards compatibility.

Attribute `is_persistent` has backwards compatibility to still support `persistent` even tho it is not used anywhere else. Key `stagingDir_persistent` cannot be changed at this moment, because it would break backwards compatibility. We could support both keys, but I have no idea how it is being used at other addons (e.g. if it is being changed?).

Function `get_staging_dir_info` could be maybe cleaned up a little, some of the arguments are not used anywhere.

## Testing notes:
1. Staging dir logic works as before.

### DIFF
--- a/client/ayon_core/pipeline/staging_dir.py
+++ b/client/ayon_core/pipeline/staging_dir.py
@@ -161,11 +161,15 @@ def get_staging_dir_info(
         )
 
     if force_tmp_dir:
-        return get_temp_dir(
-            project_name=project_entity["name"],
-            anatomy=anatomy,
-            prefix=prefix,
-            suffix=suffix,
+        return StagingDir(
+            get_temp_dir(
+                project_name=project_entity["name"],
+                anatomy=anatomy,
+                prefix=prefix,
+                suffix=suffix,
+            ),
+            is_persistent=False,
+            is_custom=False
         )
 
     # making few queries to database
--- a/client/ayon_core/pipeline/staging_dir.py
+++ b/client/ayon_core/pipeline/staging_dir.py
@@ -1,3 +1,6 @@
+import logging
+import warnings
+from typing import Optional, Dict, Any
 from dataclasses import dataclass
 
 from ayon_core.lib import Logger, filter_profiles
@@ -16,16 +19,16 @@ class StagingDir:
 
 
 def get_staging_dir_config(
-    project_name,
-    task_type,
-    task_name,
-    product_type,
-    product_name,
-    host_name,
-    project_settings=None,
-    anatomy=None,
-    log=None,
-):
+    project_name: str,
+    task_type: Optional[str, None],
+    task_name: Optional[str, None],
+    product_type: str,
+    product_name: str,
+    host_name: str,
+    project_settings: Optional[Dict[str, Any]] = None,
+    anatomy: Optional[Anatomy] = None,
+    log: Optional[logging.Logger] = None,
+) -> Optional[Dict[str, Any]]:
     """Get matching staging dir profile.
 
     Args:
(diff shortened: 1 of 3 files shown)


antirotor (approved):
I can't quickly test custom staging dir now, but the code looks good

BigRoy (approved):
Overall this seems fine - didn't do a test run but the changes are mostly cosmetic anyway.

I created this matching maya PR https://github.com/ynput/ayon-maya/pull/212

I still feel that it makes more sense to store the `StagingDir` instance in `instance.data` somewhere instead of each attribute separated. That way we just operate directly on our data class which I think is way more descriptive than these individual data keys (which also clutter the `instance.data` but I'm ok with it being the way it is.

For completeness sake I expected e.g. type `StagingDir` 
```python
staging_dir_info: StagingDir
instance.data["stagingDir_object"] = staging_dir_info
```

iLLiCiTiT:
> For completeness sake I expected e.g. type StagingDir

It should be `stagindDirInfo` (keep camel case or snake case, not both) -> which would replace all 3 current keys over time. But at this moment, that change the same issue as with `stagingDir_persistent`. We need to do research on how it is used, so we know we can support both for some time, before changing it.

If you have time to look, if `stagingDir_persistent` or `stagingDir` are being changed in some addon we could do it right now. If it's being only filled with pre-defined functions from ayon-core and then being only read.

client/ayon_core/pipeline/staging_dir.py
```diff
@@ -11,21 +14,41 @@
...
-    product_type,
-    product_name,
-    host_name,
-    project_settings=None,
-    anatomy=None,
-    log=None,
-):
+    project_name: str,
+    task_type: Optional[str, None],
+    task_name: Optional[str, None],
```
robin-ynput, line 44:
```suggestion
    task_type: Optional[str],
    task_name: Optional[str],
```

I got the following error without this change.
```
    from ayon_core.pipeline import install_ayon_plugins
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-core\client\ayon_core\pipeline\__init__.py", line 12, in <module>
    from .staging_dir import get_staging_dir_info
  File "C:\Users\robin\OneDrive\Bureau\dev_ayon\dev\ayon-core\client\ayon_core\pipeline\staging_dir.py", line 43, in <module>
    task_type: Optional[str, None],
  File "C:\Python\Python39\lib\typing.py", line 277, in inner
    return func(*args, **kwds)
  File "C:\Python\Python39\lib\typing.py", line 354, in __getitem__
    return self._getitem(self, parameters)
  File "C:\Python\Python39\lib\typing.py", line 478, in Optional
    arg = _type_check(parameters, f"{self} requires a single type.")
  File "C:\Python\Python39\lib\typing.py", line 166, in _type_check
    raise TypeError(f"{msg} Got {arg!r:.100}.")
TypeError: typing.Optional requires a single type. Got (<class 'str'>, None).
```
https://stackoverflow.com/questions/73893814/optional-typing-for-more-than-a-single-type

robin-ynput (changes requested):

robin-ynput:
I have run a Nuke publish using staging dir and custom temp dir with these changes and it went through correctly.

robin-ynput (approved):

Issue #899, opened 2024-09-10
AY-6316_Launcher: show only assigned tasks


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

Launcher should only show tasks assigned to the user not everything

### Additional context:
link to discussion on [Discord]()
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6316](https://app.clickup.com/t/86bzwk3g5)_

PR #1083, opened 2025-01-13
Show logged-in user with their avatar in tray menu

BigRoy:
## Changelog Description

Shows the logged in user in tray menu:

image

Instead of plain action labeled "Login"

## Additional info

Not optimized, up for discussion - and just an idea.
I wonder whether it'd be better to show it in the launcher UI instead.

Here's a mockup of that (very quick and dirty) as launcher action:

image

```python
from ayon_core.pipeline import LauncherAction

from qtpy import QtGui, QtCore, QtWidgets
import ayon_api
import webbrowser


def _crop_pixmap_to_circle(pixmap: QtGui.QPixmap) -> QtGui.QPixmap:
    size = min(pixmap.width(), pixmap.height())
    cropped_pixmap = QtGui.QPixmap(size, size)
    cropped_pixmap.fill(QtCore.Qt.transparent)

    painter = QtGui.QPainter(cropped_pixmap)
    painter.setRenderHint(QtGui.QPainter.Antialiasing)
    path = QtGui.QPainterPath()
    path.addEllipse(QtCore.QRectF(0, 0, size, size))
    painter.setClipPath(path)
    painter.drawPixmap(0, 0, pixmap)
    painter.end()

    return cropped_pixmap


class CurrentUser(LauncherAction):
    name = "currentuser"
    order = 999999

    @property
    def icon(self):
        # Show logged-in user's name and icon
        user = ayon_api.get_user()
        username = user["name"]
        response = ayon_api.get(f"users/{username}/avatar")
        image = response.content
        pixmap = QtGui.QPixmap()
        pixmap.loadFromData(image)
        pixmap = _crop_pixmap_to_circle(pixmap)

        import tempfile

        filepath = tempfile.NamedTemporaryFile(suffix=".png", delete=False)
        pixmap.save(filepath.name)

        return {
            "type": "path",
            "path": filepath.name
        }

    @property
    def label(self):
        user = ayon_api.get_user()
        return user.get("attrib", {}).get("fullName") or user["name"]

    def is_compatible(self, selection):
        return True

    def process(self, selection, **kwargs):
        url = ayon_api.get_server_api_connection().get_url()
        webbrowser.open(url + "/account/profile")
```

## Testing notes:

1. Start Tray Launcher
2. Logged-in user with icon should be visible in the menu

### DIFF
--- a/client/ayon_core/tools/tray/ui/tray.py
+++ b/client/ayon_core/tools/tray/ui/tray.py
@@ -48,6 +48,22 @@ from .dialogs import (
 )
 
 
+def _crop_pixmap_to_circle(pixmap: QtGui.QPixmap) -> QtGui.QPixmap:
+    size = min(pixmap.width(), pixmap.height())
+    cropped_pixmap = QtGui.QPixmap(size, size)
+    cropped_pixmap.fill(QtCore.Qt.transparent)
+
+    painter = QtGui.QPainter(cropped_pixmap)
+    painter.setRenderHint(QtGui.QPainter.Antialiasing)
+    path = QtGui.QPainterPath()
+    path.addEllipse(QtCore.QRectF(0, 0, size, size))
+    painter.setClipPath(path)
+    painter.drawPixmap(0, 0, pixmap)
+    painter.end()
+
+    return cropped_pixmap
+
+
 class TrayManager:
     """Cares about context of application.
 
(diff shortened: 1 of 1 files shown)


BigRoy:
Probably nicer if we could have a top menu bar like e.g this in the menu:

image

(That's from Dropbox app)

Or even better - maybe just show it at the top of the launcher UI top right? Especially if we also implement e.g. user-specific tasks filtering #899

PR #1086, opened 2025-01-15
Cosmetics + minor optimization for `CleanUp` plugin

BigRoy:
## Changelog Description

- Use `PublishError` instead of assertion
- Optimize the 'check for errors'
- Optimize the product type check + log debug message when skipped due to that
    - Do note that previously it'd also match if the exclude family was e.g. `clip` yet the product type of the instance was `shotclip` or `clippy` or `clip.audio` because it'd check if the string was IN the product type, it didn't check for exact match (which it does now)
- Move imports to top.

## Additional info

n/a

## Testing notes:
1. Cleanup should still work
2. "Exclude families" should still work as needed


client/ayon_core/plugins/publish/cleanup.py
```diff
@@ -71,10 +72,12 @@ def process(self, instance):
             self.log.debug("Cleaning renders new...")
             self.clean_renders(instance, skip_cleanup_filepaths)
 
-        if [ef for ef in self.exclude_families
-                if instance.data["productType"] in ef]:
+        product_type = instance.data["productType"]
```
iLLiCiTiT, line 75: Oh lord, this plugin...

This condition should probably happen before error checks. And we should change `exclude_families` to `exclude_product_types`.

I don't know why there was "str in str" check instead of "str == str"? I guess that is question for @jakubjezek001 who might be able to tell if there are product types that should be skipped too and contain `"clip`"?

jakubjezek001: I don't remember it anymore.. Its been reeeealy old code. But there could be case, or rather used to be, where **render** in nuke was **render.local**, **render.farm** and so on..

iLLiCiTiT: So it should be safe to replace with explicit product type check...?

jakubjezek001: > So it should be safe to replace with explicit product type check...?

I am just guessing but it might be a solution.

BigRoy: This is all that's keeping this from getting merged? Should I revert to behavior like before this PR?

BigRoy: Just pinging... what to do?

iLLiCiTiT: > This is all that's keeping this from getting merged? Should I revert to behavior like before this PR?

Probably yes. It is changing behavior that looks like is unpredictable at this moment.

client/ayon_core/plugins/publish/cleanup.py
```diff
@@ -71,10 +72,17 @@ def process(self, instance):
             self.log.debug("Cleaning renders new...")
             self.clean_renders(instance, skip_cleanup_filepaths)
 
-        if [ef for ef in self.exclude_families
-                if instance.data["productType"] in ef]:
+        # TODO: Figure out whether this could be refactored to just a
+        #  product_type in self.exclude_families check.
+        product_type = instance.data["productType"]
+        if any(
+            exclude_family in product_type
```
iLLiCiTiT, line 79:
```suggestion
            product_type in exclude_family
```

iLLiCiTiT (approved):

PR #1092, opened 2025-01-16
Reduce margins on Workfiles tool due to nested layouts

BigRoy:
## Changelog Description

Of all the AYON UIs, the workfiles tool was the only one that relatively had big margins around its full layout:

image

This PR reduces that to:
image

## Additional info

For comparison, some other tools:

Loader:
image

Publisher:

image

Scene Inventory:

image

## Testing notes:

1. Workfiles tools should look better

### DIFF
--- a/client/ayon_core/tools/workfiles/widgets/window.py
+++ b/client/ayon_core/tools/workfiles/widgets/window.py
@@ -113,6 +113,7 @@ class WorkfilesToolWindow(QtWidgets.QWidget):
 
         main_layout = QtWidgets.QHBoxLayout(self)
         main_layout.addWidget(pages_widget, 1)
+        main_layout.setContentsMargins(0, 0, 0, 0)
 
         overlay_messages_widget = MessageOverlayObject(self)
         overlay_invalid_host = InvalidHostOverlay(self)


client/ayon_core/tools/workfiles/widgets/window.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, controller=None, parent=None):
 
         main_layout = QtWidgets.QHBoxLayout(self)
         main_layout.addWidget(pages_widget, 1)
+        main_layout.setContentsMargins(0, 0, 0, 0)
```
iLLiCiTiT, line 116: Maybe we should set `pages_widget.layout().setContentsMargins(0, 0, 0, 0)` rather than on main layout.

BigRoy: I had my doubts about picking either of the two - I'm fine with swapping them, the result should be quite similar.
Do you want me to move it there?

iLLiCiTiT: If this `pages_widget.layout().setContentsMargins(0, 0, 0, 0)` work, I would use it, but maybe as
```python
pages_layout = pages_widget.layout()
pages_layout.setContentsMargins(0, 0, 0, 0)
```

BigRoy: So I gave this a go. Interestingly `pages_widget.setContentsMargins(0, 0, 0, 0)` also seemed to raise no errors... however, both your example and `pages_widget.setContentsMargins(0, 0, 0, 0)` had no visible effect on the margins of the dialog and hence did not give the intended result for me.

As such, that leaves the current options - or I might be doing something wrong.

iLLiCiTiT: I think I know why, but that's for future. Approving...

iLLiCiTiT (approved):

Issue #1091, opened 2025-01-16
Allow more flexibility in executable path resolving


Tilix4:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

As an admin, I'd like to be able to specify dynamic paths to find executables on users' machines. For examble, I'd like to be able to write something  like `~/subpath/to/my/app` or `$HOME/subpath/to/my/app` (this is unix system syntax, but env vars could be handy even for Windows systems).

### How would you imagine the implementation of the feature?

Add this simple line to https://github.com/ynput/ayon-core/blob/7d06cc00bd441e53e122ffa3486d820821aa3219/client/ayon_core/lib/vendor_bin_utils.py#L70 :
```py
executable = os.path.expandvars(os.path.expanduser(executable))
```

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

BigRoy:
Makes sense. For what it's worth, PRs are welcome.

m-u-r-p-h-y:
It is already possible

![Image](https://github.com/user-attachments/assets/84462aa2-d45a-47af-9995-a6c5970199d2)

alainxi:
In executable path, environment variables do **not** work if the env vars were declared in the "Arguments" environment (which would be very useful, as it is placed inside the DCC version : 4.0, 4.2, etc...)

Example :

![Image](https://github.com/user-attachments/assets/066b9ec7-d4d8-4dda-a592-c0ae7a1e1267)

![Image](https://github.com/user-attachments/assets/142a1077-4746-4e32-a259-3dbe1ae13e1a)

But the same env var declared in the "Arguments" environment, is correctly expanded when it is used in the main Blender env var field (above/outside the Variants part).

alainxi:
But the same env var declared in the "Arguments" environment, is correctly expanded when it is used in the main Blender env var field (above/outside the Variants part).

So, this works all right :

![Image](https://github.com/user-attachments/assets/d0492914-3a7e-494c-8106-52b1299608b4)

iLLiCiTiT:
In that case it is not issue for ayon-core but for ayon-applications addon.

alainxi:
Thanks.    I have declared an issue here :

https://github.com/ynput/ayon-applications/issues/51

Tilix4:
> It is possible to use environment variables, but it is not possible to use user's home directory (`~/`).

[...]

PR #1095, opened 2025-01-17
Feature: Executable path expand user and standard env var

Tilix4:
## Changelog Description
Executable path expand user `~` and standard environment variables syntax `$ENV_VAR`.

## Additional info
Fix of #1091 

## Testing notes:
1. In `Applications` addon, add a `~` or `$ENV_VAR` (don't forget to set it at some point) to an executable path, like `~/path/to/executable` or `$HOME/path/to/executable` (Unix syntax)

### DIFF
--- a/client/ayon_core/lib/vendor_bin_utils.py
+++ b/client/ayon_core/lib/vendor_bin_utils.py
@@ -67,6 +67,8 @@ def find_executable(executable):
         Union[str, None]: Full path to executable with extension which was
             found otherwise None.
     """
+    # Expand user and environment variables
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
> NB: I figured out this path resolving isn't processed when launching the application (clicking on the icon).

I think it is. But harmony integration wraps harmony process under ayon-launcher process running `launch_script.py` which will then launch the application. But the executable path is resolved the same way as others in ayon-applications addon inside `ayon_applications/defs.py` .

Issue #1098, opened 2025-01-22
ERROR while attempting to bind on address when launching ayon


Liametc:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

When launching ayon, there is an issue that arises from the webserver not being able to bind on an address. I have narrowed it down to client/ayon_core/tools/tray/webserver/server.py where it is resorting to "localhost" as a hostname. [This](https://github.com/lm-sys/FastChat/issues/1714#issuecomment-1594345815) suggestion to fix the error works as I have tried it on my local addons install. I replaced all "localhost" instances in the aforementioned file with "0.0.0.0".

### Expected Behavior:

That error should go way

### Version

1.0.10

### What platform you are running on?

Linux / Centos

### Steps To Reproduce:

Launch ayon, the error appears in the console

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
>>> Connected to AYON server http://*********:5000
*** AYON [1.1.1] -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
>>> Using AYON from [ /home/cache/liam.hoflay/AYON-1.1.1-linux-rocky9 ]
... AYON variant: [ production ]
... AYON bundle:  [ BLACKKITE-2025-01-21-01 ]
>>> loading environments ...
  - global AYON ...
  - for addons ...
Registered font families: Noto Sans, Noto Sans, Noto Sans, Noto Sans, Noto Sans Mono, Noto Sans Mono, Noto Sans Mono SemiBold, Noto Sans Mono, Noto Sans Mono Thin, Noto Sans Mono, Noto Sans Mono Medium, Noto Sans Mono, Noto Sans Mono Light, Noto Sans Mono, Noto Sans Mono ExtraBold, Noto Sans Mono, Noto Sans Mono ExtraLight, Noto Sans Mono, Noto Sans Mono Black, Noto Sans Mono
>>> [  Starting WebServer server  ] 
*** WRN: >>> { WebServerThread }: [  Web Server service has FAILED  ] 
==============================
[ Errno 99 ] ERROR while attempting to bind on address ('::1', 14419, 0, 0): cannot assign requested address
==============================
Traceback (most recent call last):
  File "/home/cache/liam.hoflay/AYON/addons/core_1.0.10/ayon_core/tools/tray/webserver/server.py", line 212, in run
    self.loop.run_until_complete(self.start_server())
[...]

PR #1099, opened 2025-01-22
Tray: Server uses 127.0.0.1 instead of localhost

Liametc:
## Changelog Description
Fix tray server issues if DNS returned IPv6 instead of IPv4 IP address.

## Additional information
Looks like python socker module does not support binding to IPv6 IP addresses. Using explicit IPv4 127.0.01, consideres as host loophole IP address should resolve the issues.
Fix for #1098. Swapping `"localhost"` for `"127.0.0.1"` in client/ayon_core/tools/tray/webserver/server.py

## Testing notes:
1. Tray can be launched even if machine is connected to IPv6 network.

### DIFF
--- a/package.py
+++ b/package.py
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "1.0.15-dev"
+version = "1.0.14-bk.1"
 
 client_dir = "ayon_core"
 
--- a/package.py
+++ b/package.py
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "1.0.14-bk.1"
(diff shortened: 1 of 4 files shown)


package.py
```diff
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "1.0.15-dev"
+version = "1.0.14-bk.1"
```
iLLiCiTiT, line 3: Please remove version changes.

Liametc: whoops they weren't meant to get in there

iLLiCiTiT (approved):

PR #1103, opened 2025-01-24
Create: Auto-fill project entity in get product name if is not passed in

iLLiCiTiT:
## Changelog Description
Auto fill project entity from current context if is not passed into the create plugin method.

## Additional info
There is low chance we will call this method from different than current project, but I've added check for the project name (just safety).

## Testing notes:
1. Product name calculation might be a little bit faster for creators which are calling the method on their own (e.g. auto-creators).


BigRoy (approved):
Makes sense. Didn't test but should be fine.

PR #1104, opened 2025-01-24
Publisher: Report page can stretch logs

iLLiCiTiT:
## Changelog Description
Report page can be stretched so long logs should be accessible.

## Additional info
Set minimum size of content widget to be able to stretch it's size.

## Testing notes:
1. Long logs on Report page can be readable.

### DIFF
--- a/client/ayon_core/style/style.css
+++ b/client/ayon_core/style/style.css
@@ -1171,6 +1171,8 @@ ValidationArtistMessage QLabel {
 
 #PublishLogMessage {
     font-family: "Noto Sans Mono";
+    border: none;
+    padding: 0;
 }
 
 #PublishInstanceLogsLabel {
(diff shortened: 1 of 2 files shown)


alainxi:
I tested this PR, but long logs are still not readable.
And it added unexpected vertical space between the first lines.

compare

alainxi:
Instead, in the same function, I tried to activate horizontal scroll bar : 

scroll_area.setHorizontalScrollBarPolicy(QtCore.Qt.ScrollBarAsNeeded)

While it allows to see very long lines, it kills word wrap for the rest of the page.
So, moderately long lines cannot be seen at first glance anymore (because word wrap is dead), you now need to use the scroll bar for a lot of lines, which is far from ideal.

alainxi:
I found a solution to still use line returns : I replace` \n` by `<br/>`

BigRoy (approved):
Works for me.

Now it shows the example validator I did like this:
image

alainxi:
Also works for me :  I tried `\n` and `\t`
Thanks !

Issue #1106, opened 2025-01-25
ExtractReview does not process representations with uppercase extension


JoseCovaOS:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

The `ExtractReview` plugin is not creating a reviewable representation of files that have its extension in uppercase. 

You can see the following message in the `ExtractReview` logs:  `INFO: Representation has unsupported extension ".MOV"`.
Also, in ftrack, you can see there is no reviewable component for the published version.

### Expected Behavior:

The `ExtractReview` plugin should be extension-agnostic, meaning it should accept both `.MOV` and `.mov` as long as the extension is supported. 
If configured, a reviewable component should appear for the assetversion in ftrack. 

### Version

1.0.14

### What platform you are running on?

Linux / Centos

### Steps To Reproduce:

1. Using the traypublisher, try to publish a render whose representation has `.MOV` extension. 
2. Check ftrack, an assetversion should appear but with no ftrack review component. 

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell

```

PR #1107, opened 2025-01-25
Support uppercase extensions in ExtractReview plugin

JoseCovaOS:
## Changelog Description
`ExtractReview` plugin supports representations with an uppercase extension.

## Additional info
This would make `ExtractReview` plugin process files that may not be in uppercase as in the case described in the [issue](https://github.com/ynput/ayon-core/issues/1106). For instance, a representation with its extension like `.Mov` would be processed as well. 

## Testing notes:
**We are assuming there is a ExtractReview profile configured for this**
1. Launch traypublisher and try to publish a render whose representation has `.MOV` extension.
2. Check ftrack, assetversion should have a reviewable component.


client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -200,7 +200,7 @@ def _get_outputs_per_representations(self, instance, profile_outputs):
             if input_ext.startswith("."):
                 input_ext = input_ext[1:]
 
-            if input_ext not in self.supported_exts:
+            if input_ext.lower() not in self.supported_exts:
```
iLLiCiTiT, line 203: Should be moved to line 199.

JoseCovaOS: Pushed :)

BigRoy:
Just want to remark that doing this would still generate a representation in the end that's named e.g. `MOV` instead of `mov` which may or may not influence other areas too. Whether that's preferred or not I'll leave for discussion but be aware that that's the case.

iLLiCiTiT:
> Just want to remark that doing this would still generate a representation in the end that's named e.g. MOV instead of mov which may or may not influence other areas too. Whether that's preferred or not I'll leave for discussion but be aware that that's the case.

I kinda agree with Roy here. Where are the files comming from? If it is traypublisher, maybe we should change how traypublisher creates them.

JoseCovaOS:
@iLLiCiTiT files came from the traypublisher

iLLiCiTiT:
In that case I would move the fix to traypublisher, as this would fix only reviewables, we probably should fix all other cases too...

antirotor (approved):
In my opinion this is fine (see the comment above) but I'd still like some arguments before merging @iLLiCiTiT @BigRoy  :D

iLLiCiTiT:
I really think this should be fixed in traypublisher. Extract review only fixes reviewable, but the "real" representation would be wrong.

PR #1108, opened 2025-01-27
Enum attribute definition: Allow placeholder text

iLLiCiTiT:
## Changelog Description
Added option to define placeholder in EnumDef.

## Testing notes:
1. Multiselection enum def does show placeholder if nothing is selected.


BigRoy (commented):
So this does work - I'm just not entirely convinced it looks the way I expected it.
It looks like this (see value **Export all shapes**):

image


However, I would have expected the text to appear dimmed similar to this field:

image

As such, I'm not entirely sure visually it gives the UX hint I had hoped. If it were maybe grey like other placeholder text maybe it'd work however.

BigRoy:
Looking nice now: 
image

For whatever the reason teh font size still renders smaller than other fields. (It seems to have a large padding to the placeholder text?)

image

Above is just an `EnumDef` tested in Fusion.
Below is the `EnumDef` tested in Silhouette (which has had styling issues before). Both appear similar.
image

It's cosmetics - but any chance we can get that font size right?

BigRoy:
Oh, another note. The color differs from placeholder text on e.g. `TextDef` too:

image

iLLiCiTiT:
> It's cosmetics - but any chance we can get that font size right?

The font size is smaller because if something is selected it paints colored frame around it.

> Oh, another note. The color differs from placeholder text on e.g. TextDef too:

The different color is related to the smaller size. It changes alpha + applied antialiasing = different color.

BigRoy:
> The font size is smaller because if something is selected it paints colored frame around it.

That may be - but it looks fine with selected entries:
image
image
image

iLLiCiTiT:
> That may be - but it looks fine with selected entries:

You mean the size or color? Color is without alpha so it looks nicer. Text size is same as placeholder text, but has the frame around (which visually makes difference).

BigRoy:
- With an entry selected in multiselection enumdef the font size seems ok.
- The placeholder text of "TextDef" is much more vivid than placeholder text of the "EnumDef" which feels off as can be seen [here](https://github.com/ynput/ayon-core/pull/1108#issuecomment-2637154075).

BigRoy:
The font size looks better now - nice!
However, placeholder text color still mismatches from other placeholder texts:

image

Also, the placeholder text draws over the button for the dropdown:

image

iLLiCiTiT:
I did not use attr definitions to compare but custom widget (where I used `PlaceholderLineEdit`) so it was the same on my side -> we don't use `PlaceholderLineEdit` in attribute definitions, changing.

BigRoy (approved):
Looks the same now! :)

image

PR #1110, opened 2025-01-29
Create: Allow to pass in transient data on instance creation

iLLiCiTiT:
## Changelog Description
It is possible to set transient data on instance creation.

## Additional info
Create plugins cannot fill attribute definitions on instance initialization, if their logic is based on transient data.

## Testing notes:
1. Validate code changes.

### DIFF
--- a/client/ayon_core/pipeline/create/structures.py
+++ b/client/ayon_core/pipeline/create/structures.py
@@ -1,6 +1,7 @@
 import copy
 import collections
 from uuid import uuid4
+import typing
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Flawless - seems to work perfectly.

I do think for the `transient_data` I'm not _entirely sure_ whether the key is required to be a `str` because it'd support any key a regular `dict` would support? Although I don't think the type hint is problematic in practice.

PR #1111, opened 2025-01-29
AY-7063 Fix Version follow up from Workfile set by hosts.

robin-ynput:
## Changelog Description

resolve: https://github.com/ynput/ayon-hiero/issues/19

Initial requests was to be able to "disable" the workfile-version-follow-up feature per host.
The setting to enable this feature is `ayon+settings://core/publish/CollectSceneVersion/hosts`.

I realized there already was a settings to  discriminate the workfile version collection per host: `ayon+settings://core/publish/CollectSceneVersion/hosts`

Changes:
* [X] Ensure that `ayon+settings://core/publish/CollectSceneVersion/hosts` is used in `collect_scene_version`
* [X] Report this in the `{version}` template computation for staging directory

## Testing notes:
1. Enable the version follow-up from `ayon+settings://core/publish/CollectSceneVersion/hosts`
2a. Ensure it follows workfile version if the host is part of `ayon+settings://core/publish/CollectSceneVersion/hosts`
2b. Ensure it does NOT follow workfile version otherwise

Tested in Hiero and Nuke (with/without staging directory).

### DIFF
--- a/client/ayon_core/pipeline/create/creator_plugins.py
+++ b/client/ayon_core/pipeline/create/creator_plugins.py
@@ -858,13 +858,24 @@ class Creator(BaseCreator):
             ["CollectAnatomyInstanceData"]
             ["follow_workfile_version"]
         )
+        follow_version_hosts = (
+            publish_settings
+            ["CollectSceneVersion"]
+            ["hosts"]
+        )
+
+        current_host = create_ctx.host.name
+        follow_workfile_version = (
+            follow_workfile_version and 
+            current_host in follow_version_hosts
+        )
 
         # Gather version number provided from the instance.
+        current_workfile = create_ctx.get_current_workfile_path()
         version = instance.get("version")
 
         # If follow workfile, gather version from workfile path.
-        if version is None and follow_workfile_version:
-            current_workfile = self.create_context.get_current_workfile_path()
+        if version is None and follow_workfile_version and current_workfile:
             workfile_version = get_version_from_path(current_workfile)
             version = int(workfile_version)
 
--- a/client/ayon_core/plugins/publish/collect_scene_version.py
+++ b/client/ayon_core/plugins/publish/collect_scene_version.py
@@ -46,6 +46,13 @@ class CollectSceneVersion(pyblish.api.ContextPlugin):
             self.log.debug("Skipping for headless publishing")
             return
 
+        if context.data["hostName"] not in self.hosts:
+            self.log.debug(
+                f"Host {context.data['hostName']} is"
+                " not setup for following version."
+            )
+            return
+
         if not context.data.get('currentFile'):
             self.log.error("Cannot get current workfile path. "
                            "Make sure your scene is saved.")


client/ayon_core/plugins/publish/collect_scene_version.py
```diff
@@ -46,6 +46,13 @@ def process(self, context):
             self.log.debug("Skipping for headless publishing")
             return
 
+        if context.data["hostName"] not in self.hosts:
+            self.log.debug(
+                f"Host {context.data['hostName']} is"
+                " not setup for collecting scene version."
+            )
+            return
+
```
BigRoy, line 55: This should be irrelevant - the pyblish plug-in _should not_ run if the currently registered host you're running with is not in `hosts` attribute of the plug-in.
```suggestion
```

robin-ynput: Good call so I've investigated a bit more on why it was still running even if current host was not part of the `hosts` list. 

This is a `Pyblish` order thing, it first check against defined hosts for the plugins THEN we have our function which apply project settings and potentially restrict/change this host. So we need an additional host filtering after the project settings are applied.

I have pushed the changes which fixes that in `lib.py`. When I run tests with Hiero and Nuke, `Collect Scene Version` seems to be the only plugin that expect its `hosts` value to be defined dynamically via settings.

I guess this is @iLLiCiTiT's choice to see if we handle this plugin specifically or add another host filtering after project settings are applied.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -460,8 +460,19 @@ class method 'def apply_settings(cls, project_settings)'
...
+        # Pyblish already operated a filter based on host.
+        # But applying settings might have changed "hosts"
+        # value in plugin so re-filter.
+        plugin_hosts = getattr(plugin, "hosts", None)
+        if (
+            plugin_hosts
+            and "*" not in plugin_hosts
+            and host_name not in plugin_hosts
+        ):
+            plugins.remove(plugin)
```
BigRoy, line 472: I initially though that:

> It should be totally allowed for a plug-in to be registered even if it does not match the current host. It would just not run - this removal code should not be needed.
> 
> The publishing iterator only runs instances that match the current host (or any hosts if `plugin.hosts = ["*"]`. This filtering as far as I know does not happen at registration time, but on runtime of processing the plugins - if not, then I think that's actually a bug in the iterator?

But, it seems that these plugins are [filtered in `discover`](https://github.com/pyblish/pyblish-base/blob/master/pyblish/plugin.py#L1376) due to logic [here](https://github.com/pyblish/pyblish-base/blob/master/pyblish/plugin.py#L1445-L1447).

However, this then shows a much greater issue at hand - because that means the plug-ins may never have been listed if we were to **add** new hosts. Unless we're using some custom logic to find plug-ins, @iLLiCiTiT ?

client/ayon_core/plugins/publish/collect_scene_version.py
BigRoy: Ignore me, see: https://github.com/ynput/ayon-core/pull/1111/files#r1933860944

client/ayon_core/pipeline/publish/lib.py
iLLiCiTiT: Why is this needed?

iLLiCiTiT: Aaah, changing `hosts` in settings? I hope those plugins which have that in setting do have set `hotst =["*"]` in the code, otherwise they even may not get to this point.

And the logic is not correct, you should use `host_is_compatible` from `pyblish.plugin`.

client/ayon_core/plugins/publish/collect_scene_version.py
iLLiCiTiT: Filtering then in filter function is better solution, so commit this suggestion.

client/ayon_core/pipeline/publish/lib.py
BigRoy: > Aaah, changing `hosts` in settings? I hope those plugins which have that in setting do have set `hosts = ["*"]` in the code, otherwise they even may not get to this point.

That's a valid point - we'll need to make sure that's the case and we're good to go.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -460,8 +460,19 @@ class method 'def apply_settings(cls, project_settings)'
...
 
+        # Pyblish already operated a filter based on host.
+        # But applying settings might have changed "hosts"
+        # value in plugin so re-filter.
+        plugin_hosts = getattr(plugin, "hosts", None)
+        if (
+            plugin_hosts
+            and "*" not in plugin_hosts
+            and host_name not in plugin_hosts
+        ):
```
iLLiCiTiT, line 471: Requires import `from pyblish.plugin import host_is_compatible`
```suggestion
        if not host_is_compatible(plugin):
```

robin-ynput: Yep, thanks a lot both for your feedback, I'll adjust this PR accordingly 🙏

client/ayon_core/plugins/publish/collect_scene_version.py
robin-ynput: This is done, thanks !

client/ayon_core/pipeline/publish/lib.py
robin-ynput: Adjusted, thanks for the feedback.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -460,8 +460,14 @@ class method 'def apply_settings(cls, project_settings)'
...
 
+        # Pyblish already operated a filter based on host.
+        # But applying settings might have changed "hosts"
+        # value in plugin so re-filter.
+        if not pyblish.plugin.host_is_compatible(plugin):
+            plugins.remove(plugin)
+
         # Remove disabled plugins
-        if getattr(plugin, "enabled", True) is False:
+        elif getattr(plugin, "enabled", True) is False:
```
BigRoy, line 470: This is a bit stupid - but can we reverse these two? I'm sure the second check is likely to be much faster and maybe even more likely to generate a removal.

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -858,13 +858,24 @@ def get_staging_dir(self, instance) -> Optional[StagingDir]:
...
         # Gather version number provided from the instance.
+        current_workfile = create_ctx.get_current_workfile_path()
         version = instance.get("version")
 
         # If follow workfile, gather version from workfile path.
-        if version is None and follow_workfile_version:
-            current_workfile = self.create_context.get_current_workfile_path()
+        if version is None and follow_workfile_version and current_workfile:
             workfile_version = get_version_from_path(current_workfile)
             version = int(workfile_version)
```
BigRoy, line 880: Not problematic to this PR - but I'm quite sure that if someone were to just save their scene "bla" manually, and then trigger this code that this may error - right? Because `get_version_from_path` would be unable to find a version in the filename?

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -858,13 +858,24 @@ def get_staging_dir(self, instance) -> Optional[StagingDir]:
             ["CollectAnatomyInstanceData"]
             ["follow_workfile_version"]
         )
+        follow_version_hosts = (
+            publish_settings
+            ["CollectSceneVersion"]
+            ["hosts"]
+        )
```
BigRoy, line 865: We might want to generate an issue after this PR - because it's quite confusing that **both** of these settings are not in the same location yet seemingly affect a similar thing? (or are the same thing?)

BigRoy (approved):
Didn't test - but aside of some minor cosmetic notes looks good to me!

client/ayon_core/pipeline/create/creator_plugins.py
iLLiCiTiT: It would crash indeed, and before this PR it would also crash if `get_current_workfile_path` would return `None`. I think we should change it to:
```python
if version is None and follow_workfile_version and current_workfile:
    workfile_version = get_version_from_path(current_workfile)
    if workfile_version is None:
        version = int(workfile_version)

if version is None:
```

Usage of `get_version_from_path` is also questionable, but that conversation for different times.

robin-ynput: I've changed it and tested it as per your suggestion. Confirmed it works fine for unsaved and "random-path" workfiles.
Thanks for the suggestions.

client/ayon_core/pipeline/publish/lib.py
robin-ynput: Good call, this is done.

iLLiCiTiT (approved):

client/ayon_core/pipeline/create/creator_plugins.py
robin-ynput: here it is: https://github.com/ynput/ayon-core/issues/1117

PR #1112, opened 2025-01-29
Fix [clip_media] reviewable.

robin-ynput:
## Changelog Description

While working on retimes, I realized I've over-engineered the logic to handle `[clip_media]` reviewable.
When selected the reviewable should be `otio_clip` original media source range + handles.

This change simplify the approach to make it leverage `collect_otio_review`, `extract_otio_review` and `extract_review`.

Changes:
* [X] Rework `[clip_media]` logic to make it go through `collect_otio_review`
* [X] Ensure single frame reviewables are properly handled by `extract_review` plugins

## Testing notes:
1. From Flame or Hiero publish shot clip with the `[clip_media]` reviewable enabled
2. Ensure reviewable is properly produce from clip source range (no retime applied) + handles


jakubjezek001 (approved):
works like a charm!

PR #1120, opened 2025-02-04
Speed Up Publishing Times

r42-chun:
1. Use concurrent.futures to copy files 
2. Use speedcopy.copyfile for windows when copying Hero files as shutil.copy is extremely slow.

Results.
On a test publish job on the farm, have cut down publishing times from 16 mins 51 seconds to 6 mins 50 seconds.
image
image


BigRoy (commented):
Very nice! :) I've added some comments which are mostly cosmetics. I wouldn't have expected that concurrency on the file transfers would've made such a big difference.

The amount of threads being started to do the transfer - I wonder whether we should still customize that - because it may be quite many (e.g. 32 threads), see: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

But if this doesn't give issues in production on "regular publishing" on a workstation and doesn't block the user's machine significantly I'd say we can keep the defaults just fine.

krishnaavril:
Commenting to show the intrest and following this up

This a major problem actually, our large FX caches are taking hours to copy, Finally some relief!

client/ayon_core/lib/file_transaction.py
```diff
@@ -109,41 +109,46 @@ def add(self, src, dst, mode=MODE_COPY):
...
+            for dst, (src, _) in self._transfers.items():
+                backup_futures.append(executor.submit(self.backup_file, dst, src))
+            concurrent.futures.wait(backup_futures)
+
+            transfer_futures = []
+            for dst, (src, opts) in self._transfers.items():
+                transfer_futures.append(executor.submit(self.transfer_file, dst, src, opts))
+            concurrent.futures.wait(transfer_futures)
+
+    def backup_file(self, dst, src):
```
iLLiCiTiT, line 123: Both `backup_file` and `transfer_file` should start with `_`. They should not be used by anything else than `FileTransaction` object.

client/ayon_core/lib/file_transaction.py
```diff
@@ -109,41 +109,46 @@ def add(self, src, dst, mode=MODE_COPY):
...
-            self.log.debug("Checking file ... {} -> {}".format(src, dst))
-            path_same = self._same_paths(src, dst)
-            if path_same or not os.path.exists(dst):
-                continue
-
-            # Backup original file
-            # todo: add timestamp or uuid to ensure unique
-            backup = dst + ".bak"
-            self._backup_to_original[backup] = dst
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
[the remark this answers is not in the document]
r42-chun: Ahhh yes, that is true. There isn't any handling of errors.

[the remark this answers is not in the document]
r42-chun: I have added some error handling.

client/ayon_core/lib/file_transaction.py
r42-chun: I have added the _ at the start of backup_file and transfer_file

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -2,6 +2,14 @@
 import copy
 import errno
 import shutil
+import sys
+# this is needed until speedcopy for linux is fixed
+if sys.platform == "win32":
+    from speedcopy import copyfile
+else:
+    from shutil import copyfile
```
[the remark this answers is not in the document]
antirotor: Speedcopy is in dependencies of AYON (https://github.com/ynput/ayon-core/blob/c7899fb3be31d215207c5e0cccf01a43e73408ff/client/pyproject.toml#L10) and it has its own fallback if it cannot use server-side copy. What is the issue with spedcopy on linux?

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -415,11 +423,11 @@ def integrate_instance(
...
-            for src_path, dst_path in src_to_dst_file_paths:
-                self.copy_file(src_path, dst_path)
-
-            for src_path, dst_path in other_file_paths_mapping:
-                self.copy_file(src_path, dst_path)
+            with concurrent.futures.ThreadPoolExecutor() as executor:
+                file_futures = []
+                for src_path, dst_path in src_to_dst_file_paths + other_file_paths_mapping:
+                    file_futures.append(executor.submit(self.copy_file, src_path, dst_path))
+                concurrent.futures.wait(file_futures)
```
BigRoy, line 430: Since the other code needed it to - I assume this would also need error handling? I'm wondering if it warrants making a small wrapper around `ThreadPoolExecutor` that allows to easily call the list of futures - yet error as soon as one errors along the way.

Preferably we'd not want the rest of the pool to continue execution either from that point onwards.

client/ayon_core/lib/file_transaction.py
```diff
@@ -109,41 +115,52 @@ def add(self, src, dst, mode=MODE_COPY):
...
-                create_hard_link(src, dst)
+        if opts["mode"] == self.MODE_COPY:
+            self.log.debug(f"Copying file ... {src} -> {dst}")
+            copyfile(src, dst)
+        elif opts["mode"] == self.MODE_HARDLINK:
+            self.log.debug(f"Hardlinking file ... {src} -> {dst}")
+            create_hard_link(src, dst)
 
-            self._transferred.append(dst)
+        self._transferred.append(dst)
```
BigRoy, line 160: Technically this runs in a thread now - and hence it _should not_ do this on an object outside of the thread. However, in Python implementations this should still be safe due to the GIL and how the interpreter locks, etc.

Nonetheless, in a potential non-GIL or free threaded Python 3 world this would be unsafe and instead we should be instead relying on the `future.result()` instead.

BigRoy (approved):
I've cleaned up the code to match code style guide + added error handling in both cases (or at least, have it raise and log the errors and stop as soon as possible on an error instead of continuing to finish the rest of the threaded queue.

@iLLiCiTiT @antirotor thoughts?

I have done some simple tests from Maya, Fusion and Silhouette - however I did not do any speed comparisons.

As far as I know, usually 'threading' file transfers should not make things much faster because the slowness would be IO-bound or network-speed bound regardless. So this making a huge performance impact might need some testing in a few production scenarios.

Also note that the `max_workers=8` seems relatively arbitrarily chosen, and depending on actual speed test runs and comparisons a more optimal figure could be found. Without doing actual speed comparisons we may just as well remove the `max_workers` and have it use the defaults of `ThreadPoolExecutor` instead (which may be much higher depending on the machine's CPU count in the running process however.)

---

For completeness sake, if it now DOES error then e.g. an output log turns into:
```
ERROR: Error occurred in worker
  File "C:\Program Files\BorisFX\Silhouette 2024.5\resources\python\lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "F:\dev\ayon-core\client\ayon_core\lib\file_transaction.py", line 155, in _transfer_file
    copyfile(src, dst)
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\speedcopy\__init__.py", line 332, in copyfile
    ret = COPYFILE('\\\\?\\' + source_file,
  File "_ctypes/callproc.c", line 997, in GetResult
PermissionError: [WinError -2147024891] Access is denied

  File "F:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 158, in process
  File "F:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 290, in register
    file_transactions.process()
  File "F:\dev\ayon-core\client\ayon_core\lib\file_transaction.py", line 129, in process
    as_completed_stop_and_raise_on_error(
  File "F:\dev\ayon-core\client\ayon_core\lib\threadpool.py", line 49, in as_completed_stop_and_raise_on_error
    raise exceptions[-1]
  File "C:\Program Files\BorisFX\Silhouette 2024.5\resources\python\lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "F:\dev\ayon-core\client\ayon_core\lib\file_transaction.py", line 155, in _transfer_file
    copyfile(src, dst)
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\speedcopy\__init__.py", line 332, in copyfile
    ret = COPYFILE('\\\\?\\' + source_file,
  File "_ctypes/callproc.c", line 997, in GetResult
PermissionError: [WinError -2147024891] Access is denied

  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "F:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 170, in process
    raise exc
  File "F:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 158, in process
  File "F:\dev\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 290, in register
    file_transactions.process()
  File "F:\dev\ayon-core\client\ayon_core\lib\file_transaction.py", line 129, in process
    as_completed_stop_and_raise_on_error(
  File "F:\dev\ayon-core\client\ayon_core\lib\threadpool.py", line 49, in as_completed_stop_and_raise_on_error
    raise exceptions[-1]
  File "C:\Program Files\BorisFX\Silhouette 2024.5\resources\python\lib\concurrent\futures\thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "F:\dev\ayon-core\client\ayon_core\lib\file_transaction.py", line 155, in _transfer_file
    copyfile(src, dst)
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\speedcopy\__init__.py", line 332, in copyfile
    ret = COPYFILE('\\\\?\\' + source_file,
  File "_ctypes/callproc.c", line 997, in GetResult
PermissionError: [WinError -2147024891] Access is denied
```
Note that this in particular is from the speedcopy on trying to copy a folder (which never worked anyway) but was the easiest error reproducable I could find. As can be seen - it does make for quite a verbose output.

r42-chun:
Thanks so much Roy for helping me with the error handling! Much appreciated

> Also note that the `max_workers=8` seems relatively arbitrarily chosen,

Yes, it was chosen randomly. Wonder if we can make it a parameter or just remove it entirely?

client/ayon_core/lib/threadpool.py
```diff
@@ -0,0 +1,49 @@
+import logging
+import concurrent.futures
+from concurrent.futures import ThreadPoolExecutor, Future
+from typing import List, Optional
+
+log = logging.getLogger(__name__)
+
+
+def as_completed_stop_and_raise_on_error(
```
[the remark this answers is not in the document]
BigRoy: I'm re-using it in two places currently - that's why I added it into a separate file. But you mean to add it to `file_transaction.py` but not necessarily the `FileTransaction` class itself. If so, I can definitely do that! :)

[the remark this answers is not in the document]
BigRoy: I've now moved it to the other file (but not a method because it's also used outside of the class). See https://github.com/ynput/ayon-core/pull/1120/commits/9c67bf1990d26fdcde5e3566c058aadfdda6bf8e

@iLLiCiTiT better?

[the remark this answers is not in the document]
BigRoy: Also, I assume the function may get more use if we were to move more things to thread pools, e.g. transcoding, etc.

krishnaavril:
This is working fine for us,
Can this be merged into the develop branch!
Anything still missing in this?

BigRoy:
@krishnaavril Did you happen to do a test-run with varying the maximum concurrent threads?

client/ayon_core/lib/threadpool.py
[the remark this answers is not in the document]
BigRoy: Marking this resolved @iLLiCiTiT

client/ayon_core/lib/file_transaction.py
BigRoy: Thoughts @iLLiCiTiT ?

client/ayon_core/plugins/publish/integrate_hero_version.py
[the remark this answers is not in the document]
BigRoy: I personally don't know unfortunately.

BigRoy:
Performance results of ["max workers" was shared on Discord here](https://discord.com/channels/517362899170230292/1349016217331171369)

image

Showing that fiddling with "max workers" didn't seem to make much of a difference. Of course, this may also be an issue with how these results are measured - but just wanted to link it.

BigRoy:
We've been running this in production as is for some time - it showed no issues and performance seemed fine or slightly better. Not as big of an improvement as it has seemed to be for others - but still a fine fix.

What's keeping this from merging now? @iLLiCiTiT @antirotor @philippe-ynput

antirotor (approved):
I think it looks solid, let's merge

client/ayon_core/lib/file_transaction.py
```diff
@@ -212,3 +226,46 @@ def _same_paths(self, src, dst):
             return os.stat(src) == os.stat(dst)
 
         return src == dst
+
+
+def as_completed_stop_and_raise_on_error(
```
iLLiCiTiT, line 231: The function name is wierd, it does not tell for what it should be called, maybe something like `wait_for_futures`, without the "raise on error".

client/ayon_core/lib/file_transaction.py
```diff
@@ -212,3 +226,46 @@ def _same_paths(self, src, dst):
...
+        exception = future.exception()
+        if exception:
+            # As soon as an error occurs, stop executing more futures.
+            # Running workers however, will still complete so we also want
+            # to log those errors if any occurred on them.
+            executor.shutdown(wait=True, cancel_futures=True)
+            break
+    else:
+        # Futures are completed, no exceptions occurred
+        return
```
iLLiCiTiT, line 255: I know this is technically ok, but would be more clear to return if there are not `exceptions` collected instead?

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -415,11 +424,14 @@ def integrate_instance(
...
-                self.copy_file(src_path, dst_path)
-
-            for src_path, dst_path in other_file_paths_mapping:
-                self.copy_file(src_path, dst_path)
+            with ThreadPoolExecutor(max_workers=8) as executor:
+                futures = [
+                    executor.submit(self.copy_file, src_path, dst_path)
+                    for src_path, dst_path
+                    in itertools.chain(src_to_dst_file_paths,
+                                       other_file_paths_mapping)
```
iLLiCiTiT, line 432:
```suggestion
                    for src_path, dst_path in itertools.chain(
                        src_to_dst_file_paths, other_file_paths_mapping
                    )
```

client/ayon_core/lib/file_transaction.py
antirotor: I think we will have to deal with this once GIL is dropped on many places anyway. So far I don't think AYON would work in non CPython based interpreter anyway.

munkybutt:
This is awesome! Something I have been considering for a long while but not got a chance to try out so thanks for this PR.
From what I understand about python and the GIL, the main reason for the initial speed up of transfers when moving to concurrency is that it decouples the one thread running at 100% from the storage which is only at 50% read/write and waiting for the GIL to come back to continue processing.
It removes the single thread as a bottle neck and the operation becomes close to completely io bound. Which also explains why throwing more threads at it doesn't show much of a speed up.
(I got obsessed with the GIL a few years ago)

PR #1121, opened 2025-02-04
String template: Fix index validation for list values

iLLiCiTiT:
## Changelog Description
List values are correctly resovled as have missing index/key.

## Additional info
The validation was not readable much and it was looking at length of input data instead of lenght of current value.

## Testing notes:
1. Use `<{folder[parents][100]}>` in anatomy template.
2. Run some process to try fill the template.
3. It should not crash because of `IndexError`.


BigRoy (approved):
Didn't test but code change makes sense

PR #1129, opened 2025-02-06
Add custom product name profiles for substance designer 

moonyuet:
## Changelog Description
Add custom product name profiles for substance designer
Fix https://github.com/ynput/ayon-substance-designer/issues/9

## Additional info
Test with https://github.com/ynput/ayon-substance-designer/pull/11

## Testing notes:
1. Launch Substance Designer
2. Create Texture
3. Play around with the new settings.

### DIFF
--- a/server/settings/tools.py
+++ b/server/settings/tools.py
@@ -484,6 +484,17 @@ DEFAULT_TOOLS_VALUES = {
                 "task_types": [],
                 "tasks": [],
                 "template": "{folder[name]}_{variant}"
+            },
+            {
+                "product_types": [
+                    "textureSet"
+                ],
+                "hosts": [
+                    "substancedesigner"
+                ],
+                "task_types": [],
+                "tasks": [],
+                "template": "T_{folder[name]}{variant}"
             }
         ],
         "filter_creator_profiles": []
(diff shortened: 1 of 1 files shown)


server/settings/tools.py
```diff
@@ -484,6 +484,17 @@ class GlobalToolsModel(BaseSettingsModel):
...
+            {
+                "product_types": [
+                    "textureSet"
+                ],
+                "hosts": [
+                    "substancedesigner"
+                ],
+                "task_types": [],
+                "tasks": [],
+                "template": "T_{folder[name]}{variant}"
```
BigRoy, line 497: This feels overly specific to unreal - and I have my doubts about whether this is a sensible default?

server/settings/tools.py
```diff
@@ -548,10 +559,13 @@ class GlobalToolsModel(BaseSettingsModel):
             },
             {
                 "product_types": [
-                    "simpleUnrealTexture"
+                    "simpleUnrealTexture",
+                    "image",
+                    "textures"
```
BigRoy, line 564: This change would now also force publishes from `standalonepublisher` that are just `image` and `textures` publishes to be renamed to the new template - that sounds incorrect behavior and overly specifically targeted to ingesting for Unreal.

server/settings/tools.py
moonyuet: We can remove this here, and if the users want to add the template, we can suggest them to add the parameter accordingly to this.

LiborBatek (approved):
LGTM
...merging I guess depends more on the devs pov atm

PR #1134, opened 2025-02-11
Chore: Use platformdirs instead of appdirs

iLLiCiTiT:
## Changelog Description
Replace usage of `appdirs` with `platformdirs`.

## Additional info
Moduly `appdirs` is deprecated and `platformdirs` is it's successor.

## Testing notes:
1. Hopefully nothing really changes.

### DIFF
--- a/client/ayon_core/lib/local_settings.py
+++ b/client/ayon_core/lib/local_settings.py
@@ -9,7 +9,7 @@ from datetime import datetime
 from abc import ABC, abstractmethod
 from functools import lru_cache
 
-import appdirs
+import platformdirs
 import ayon_api
 
 _PLACEHOLDER = object()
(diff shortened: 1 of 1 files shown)


BigRoy:
Should this update some dependencies somewhere?

iLLiCiTiT:
> Should this update some dependencies somewhere?

It is already part of AYON launcher dependencies, so should not be necessary (the same way as `appdirs` are).

kalisp (approved):
Looks fine.

Btw there are still couple of mentions of `appdirs` in `ayon-launcher`. They will be purged later, I am guessing?
(They are still installed into `.venv`, but checked that `platformdirs` are there too, so it shouldnt be problem.)

iLLiCiTiT:
> Btw there are still couple of mentions of appdirs in ayon-launcher.

But `appdirs` module should not be used anywhere.

> (They are still installed into .venv, but checked that platformdirs are there too, so it shouldnt be problem.)

We can't just remove it. We have to first stop using, and then remove it.

PR #1135, opened 2025-02-11
Chore: Remove acre dependency

iLLiCiTiT:
## Changelog Description
Moved functions from `acre` to ayon-core lib functions.

## Additional info
We're very dependent on `acre` module, but we're using fork of the module which is installed to ayon-launcher using git url, which makes it's maintanance very difficult, and does not allow us to define the dependency into addons that are using it (ayon-core and ayon-applications).

Made some small tweaks in the code.

Plan to remove the dependency was there for a long time, but we never got to it.

## Testing notes:
1. Global environments from settings and from addons are set and calculated as expected.

### DIFF
--- a/client/ayon_core/lib/env_tools.py
+++ b/client/ayon_core/lib/env_tools.py
@@ -13,11 +13,6 @@ if typing.TYPE_CHECKING:
     PlatformName = Literal["windows", "linux", "darwin"]
     EnvValue = Union[str, list[str], dict[str, str], dict[str, list[str]]]
 
-Results = collections.namedtuple(
-    "Results",
-    ["sorted", "cyclic"]
-)
-
 
 class CycleError(ValueError):
     """Raised when a cycle is detected in dynamic env variables compute."""
@@ -124,7 +119,8 @@ def parse_env_variables_structure(
         dict: The flattened environment for a platform.
 
     """
-    platform_name = platform_name or platform.system().lower()
+    if platform_name is None:
+        platform_name = platform.system().lower()
 
     result = {}
     for variable, value in env.items():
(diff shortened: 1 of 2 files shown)


client/ayon_core/lib/env_tools.py
```diff
@@ -27,7 +55,11 @@ def env_value_to_bool(env_key=None, value=None, default=False):
...
 
-def get_paths_from_environ(env_key=None, env_value=None, return_first=False):
+def get_paths_from_environ(
+    env_key: Optional[str] = None,
+    env_value: Optional[str] = None,
+    return_first: bool = False,
+) -> Optional[Union[str, list[str]]]:
     """Return existing paths from specific environment variable.
 
     Args:
```
BigRoy, line 65: Type hints in the docstring do not match the type hints of the arguments? (The `str` are not optional in the docstring?)

BigRoy (commented):
I'm not sure this is an improvement - I think we're better off keeping the `acre` as isolated part by itself to 100% ensure it remains an isolated piece of software instead of over time sneakily getting 'hotfixes' in the logic here that introduces dependency to other aspects inside AYON.

Acre is very simple **because** it runs in such an isolated manner.

It also feels weird that this logic is in something called `env_tools` but maybe that's just me.

Also, if we DO end up merging it into AYON - we should **absolutely** also include test coverage to be transferred along as well.

And last but not least, you're now technically copying source code into AYON core from a source that is GNU Lesser General Public License v3.0 licensed. The source code of `acre` is 'mine' and other contributors to of course there isn't necessarily the direct need for conflict - but it is something to keep in mind maybe?

iLLiCiTiT:
> I'm not sure this is an improvement - I think we're better off keeping the acre as isolated part by itself to 100% ensure it remains an isolated piece of software instead of over time sneakily getting 'hotfixes' in the logic here that introduces dependency to other aspects inside AYON.
>
> And last but not least, you're now technically copying source code into AYON core from a source that is GNU Lesser General Public License v3.0 licensed. The source code of acre is 'mine' and other contributors to of course there isn't necessarily the direct need for conflict - but it is something to keep in mind maybe?

True, but we already are using forked and modified version of acre to match needs in AYON (OpenPype originally), so we either merge them to your acre, and release them to pypi, or we have to do something else, but we can't keept in at current state.

BigRoy:
> True, but we already are using forked and modified version of acre to match needs in AYON (OpenPype originally), so we either merge them to your acre, and release them to pypi,

I'd say - fine with me - or if in some way possible, transfer the pypi ownership to the `ynput` repo?

iLLiCiTiT:
> I'd say - fine with me - or if in some way possible, transfer the pypi ownership to the ynput repo?

I'm ok with keeping it in your repository, but we have to merge the changes there. BTW we did make some changes in past that I don't think are really necessary at this moment and we can revert them, especially those that doesn't make any sense...

BigRoy (approved):
Works for me - I however did not do a run of the tests. Did you confirm those ran fine? (if so, how do I run the `core` tests?)

Other than that, the PR branch does work fine for me.

iLLiCiTiT:
> Works for me - I however did not do a run of the tests. Did you confirm those ran fine? (if so, how do I run the core tests?)

Updated `pyproject.toml` and scripts. Now it should be working using `./tools/manage.ps1 create-env` and `./tools/manage.ps run-tests`.

PR #1136, opened 2025-02-11
UI: Nice checkbox uses odd sizes

iLLiCiTiT:
## Changelog Description
NiceCheckbox does use odd sizes for frame to paint to avoid cropped paint.

## Additional information
It looks like it dependends where is the checkbox painted and how much space it gets, which might be based on UI scaling too. 

## Testing notes:
Requires to be able to achieve the issue on machine which might be tricky. It works ok on my 100% scaling monitor and changes on 150% scaling monitor on my work laptop, but also doesn't work on different machine with 100% scaling, so hard to tell how to replicate.
1. The checkbox is not cropped.

### DIFF
--- a/client/ayon_core/tools/utils/nice_checkbox.py
+++ b/client/ayon_core/tools/utils/nice_checkbox.py
@@ -383,7 +383,7 @@ class NiceCheckbox(QtWidgets.QFrame):
         else:
             radius = floor(checkbox_rect.width() * 0.5)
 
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Without this PR it also looks good for me and I can't reproduce the odd behavior. However, with this PR it **also** looks good for me in Maya, Fusion and Silhouette. So approving that it 'works' but not sure if it fixes the odd styling on some machines.

Issue #1062, opened 2024-12-17
Publish Error Report double line breaks


MustafaJafar:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

Currently, error messages with one line break `\n` work in log as expected but not in the report on the right side of the  window.
However, it works as expected when using bullet points `-`

Here's the output of using only one `\n` and no `-` in `detail`
![image](https://github.com/user-attachments/assets/93e12c12-2d03-456c-b743-e702d88c4615)
The output of combining `\n` and `-` in `detail`
![image](https://github.com/user-attachments/assets/caf91783-136e-4f0e-938d-c834adafcba7)
The output of using two `\n`
![image](https://github.com/user-attachments/assets/81c7b30f-cb26-49c1-b81c-72956381d235)

### Expected Behavior:

.

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

check my screenshots.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell

```

MustafaJafar:
Tagging @iLLiCiTiT for visibility.

PR #1137, opened 2025-02-11
Publisher: Publish description and detail respects new line characters

iLLiCiTiT:
## Changelog Description
New line characters in publish error and details are showed correctly now.

## Additional info
The issue is that we do allow HTML tags, so we use `setHtml` method, which does not respect new line characters. On the other side, multiple plugins are raising errors with new line characters that should be ignored, so this might actually break more than fix.

## Testing notes:
1. When `PublishError` raises error with `\n` it is shown in UI.

Resolves https://github.com/ynput/ayon-core/issues/1062

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/report_page.py
+++ b/client/ayon_core/tools/publisher/widgets/report_page.py
@@ -1693,6 +1693,9 @@ class PublishFailWidget(QtWidgets.QWidget):
             description = error_item.description or description
             detail = error_item.detail or detail
 
+        description = description.replace("\n", "<br/>")
+        detail = detail.replace("\n", "<br/>")
+
         self._error_details_widget.set_detail(detail)
 
         if commonmark:


BigRoy:
Just below in the code I'm also seeing a `setMarkdown` call (if not "commonmark" whatever that means).
So technically the enforced `\n` to `<br>` would be invalid because it'd break code blocks or alike, and likely other cases?

iLLiCiTiT:
You're right, and I don't have strong opinion about it, but I did spend at about an hour finding out what might go wrong. The idea is/was that we will use markdown to nicely visualize the errors to user.

On the other side, the reality is that most of the errors are not markdown, especially those that are not `PublishXmlValidationError`. I'm tempted to say, that we'll add `is_markdown` to `PublishError` which will be by default `False`, and by default `True` for `PublishXmlValidationError`, because I don't have proper solution without knowing if I should do the conversion or not.

BigRoy:
I think the best question asked now is - which current reports are not shown in the way we would expect them to be shown. And if they are shown invalidly - would it just make sense to include `<br>` explicitly in that particular case in that particular report?

Also, I'd say moving to markdown by default isn't a bad idea for me per se either - and maybe rewriting any that aren't actually markdown now to turn them into markdown instead of using html?

iLLiCiTiT:
I'm fine with having markdown as true by default, also to keep compatibility. But we need to be able to have that information on the exception so we can handle it accordingly. So if there is not markdown then e.g. printing to console could revert it back to "standard" text? Considering that text will be printed in deadline output too... I'm maybe overthinking it?

BigRoy:
I think considering the deadline logs, also there markdown would be more readable than HTML. So defaulting to markdown I think is still the best way forward.

About 'sanitizing' that for any console output in the future - I'd say, keep that separate from this PR here. Feels a bit out of scope and hasn't been an issue much yet. I'd rather have us solve being able to also get JSON reports for those headless publishes - so that if one really needed to, they could load up the JSON report from that session into the dedicated Publish JSON report viewer.

MustafaJafar:
Ehmmm 🤔 

https://github.com/ynput/ayon-houdini/blob/fa4dea35600c6e393603c0662b356d5c833374ce/client/ayon_houdini/plugins/publish/extract_render.py#L92-L100

| Before | After |
|--|--|
| image | image |

MustafaJafar:
I've gone through the discussion. personally, I'd love to stick to one consistent way to provide error messages. 
If the code currently is using markdown eventually (i.e. convert the input strings to markdown), then I'd be happy to tweak any input strings and make them markdown at the first place and skip any conversations.. so that what I see in code is what I get.

BigRoy:
@iLLiCiTiT what's the next step for this PR?

iLLiCiTiT:
> @iLLiCiTiT what's the next step for this PR?

Current PR state does not fix anything, it actually makes it worse. I think that we use the publisher error in 2 ways: 1. show it in UI, 2. print it in terminal. I think we need to know for which of those cases is the message meant so we can modify it for UI accordingly, and I don't have capacity to resolve this now as it is not "crutial" to be fixed ASAP. Right now it meants that you have to use doubled `\n` to make actuall new line character (if I get it correctly). In that regard closing PR for now. Re-open, or create new if willing to continue.

PR #1138, opened 2025-02-11
Attribute definitions: Multiselection EnumDef allows empty items

iLLiCiTiT:
## Changelog Description
EnumDef does not require to have items.

## Additional info
It is valid option to have empty items for multiselection enumerator.

## Testing notes:
1. It is possible to create EnumDef with empty items (e.g. in pre-create attributes).
2. It shows correctly in UI.

```python
EnumDef(
    "some_key",
    [],
    multiselection=True,
)
```
Resolves https://github.com/ynput/ayon-core/issues/810

### DIFF
--- a/client/ayon_core/lib/attribute_definitions.py
+++ b/client/ayon_core/lib/attribute_definitions.py
@@ -562,17 +562,18 @@ class EnumDef(AbstractAttrDef):
         multiselection: Optional[bool] = False,
         **kwargs
     ):
-        if not items:
-            raise ValueError((
-                "Empty 'items' value. {} must have"
+        if multiselection is None:
+            multiselection = False
+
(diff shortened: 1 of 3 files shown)


BigRoy (approved):
This does work indeed - it does feel a bit odd however. When clicking the box, nothing happens.. making it feel like it's broken even though there are just no item. 

I wonder if we should instead be doing something else to hint that there are no options to pick from?

iLLiCiTiT:
> This does work indeed - it does feel a bit odd however. When clicking the box, nothing happens.. making it feel like it's broken even though there are just no item.

That does make sense. Looking into it.

BigRoy (approved):
This is now very nice!
 
image

PR #1141, opened 2025-02-11
Host: Version up current workfile handles empty workdir

### LINKED ISSUE #1139: `version_up_current_workfile()` doesn't return the first workfile name if there are no workfiles

### Current Behavior:

`version_up_current_workfile()` doesn't return the first workfile name if there are no workfiles.
It actually returns the second version as it version up two times.

https://github.com/ynput/ayon-core/blob/48d94b69fdbb047eee915e1f93b414e42b550fce/client/ayon_core/pipeline/context_tools.py#L620-L622

### Expected Behavior:

.

### Version

1.0.14

Windows

### Steps To Reproduce:

1. Open an empty scene in any task with the correct AYON context and no previous workfiles.
2. Run
```
from ayon_core.pipeline.context_tools import version_up_current_workfile
version_up_current_workfile()
```

### Relevant log output:

```shell

```

### Additional context:

_No response_

### LINKED ISSUE #1140: `version_up(filepath)` errors if the workfile dir doesn't exist.

### Current Behavior:

`version_up(filepath)` errors if the workfile dir doesn't exist as there are no `dirname` to list its subdirs.

https://github.com/ynput/ayon-core/blob/48d94b69fdbb047eee915e1f93b414e42b550fce/client/ayon_core/lib/path_tools.py#L139

### Expected Behavior:

.

### Version

1.0.14

Windows

### Steps To Reproduce:

1. Open an empty scene in any task with the correct AYON context and no previous workfiles.
2. Run
```
from ayon_core.pipeline.context_tools import version_up_current_workfile
version_up_current_workfile()
```

### Relevant log output:

```shell

```

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
Version up current workfile does handle cases if workdir for current context does not exist yet or does not contain any workfiles yet.

## Testing notes:
1. Open an empty scene in any task with the correct AYON context and no previous workfiles.
2. Run
```
from ayon_core.pipeline.context_tools import version_up_current_workfile
version_up_current_workfile()
```
3. It should work as expected. 
4. if the task folder doesn't exist yet, this function raises an error.

Resolves #1139


client/ayon_core/pipeline/context_tools.py
```diff
@@ -617,7 +617,7 @@ def version_up_current_workfile():
     last_workfile_path = get_last_workfile(
         work_root, file_template, data, extensions, True
     )
-    new_workfile_path = version_up(last_workfile_path)
+    new_workfile_path = last_workfile_path
```
BigRoy, line 620:
```suggestion
        # `get_last_workfile` will return the first expected file version
        # if no files exist yet. In that case, if they do not exist we will
        # want to save v001
    new_workfile_path = last_workfile_path
```

client/ayon_core/lib/path_tools.py
```diff
@@ -136,8 +136,8 @@ def version_up(filepath):
         index += len(new_label)
         clash_basename = clash_basename[:index]
 
-    for file in os.listdir(dirname):
-        if file.endswith(ext) and file.startswith(clash_basename):
+    for f in os.scandir(dirname):
+        if f.name.endswith(ext) and f.name.startswith(clash_basename):
```
BigRoy, line 140: I personally think it raising an error if the folder does not exist is fine behavior.

MustafaJafar:
@iLLiCiTiT 
After a discussion with @BigRoy I realized I was using `version_up_current_workfile()` in a way that is not intended to. 
As `version_up_current_workfile()` should be used when there's at least one workfile to up version.

Regarding my usage: I used it to get/create the next workfile name and location for a task. the current state of this function in develop branch has these issues:
- if the workfile doesn't exist (i.e. it will be the first workfile), it's then up versioned and returning/saving `v002` instead of `v001`.
- the work directory doesn't exist where the `version_up` fails as it uses `os.listdir` which raises an error. 
- after replacing `os.listdir`,  the `host.save_workfile(new_workfile_path)` will mostly fail because the work directory doesn't exist.

Let me know what do you think?
.. I thought about getting the work directory and creating it before calling `version_up_current_workfile()` but this also means that the work directory will be queried twice making my code slower/doing unnecessary things.

BigRoy:
> As `version_up_current_workfile()` should be used when there's at least one workfile to up version.

This isn't _entirely true_. It's currently also used if there's no existing workfile yet **BUT** the work area already exists (afaik)

So:

> if the workfile doesn't exist (i.e. it will be the first workfile), it's then up versioned and returning/saving v002 instead of v001.

BUG: Should be fixed!

> the work directory doesn't exist where the version_up fails as it uses os.listdir which raises an error.

Maybe not the best error - but I'd say the function itself isn't necessarily responsible for initializing the work area (including all its 'extra work area folders, etc.'

client/ayon_core/pipeline/context_tools.py
MustafaJafar: https://github.com/ynput/ayon-core/pull/1141/commits/65d7404d428229e8573169040cb5fec4b5f4e522

MustafaJafar:
@BigRoy `version_up_current_workfile()` now:
- won't create any missing folders
- increments existing last work file or creates the first work file if it doesn't exist already.
- avoid the error in `version_up`
- raise a meaningful error if the parent folder of the work file doesn't exist.

client/ayon_core/pipeline/context_tools.py
```diff
@@ -617,7 +617,18 @@ def version_up_current_workfile():
...
+    # want to save v001
+    new_workfile_path = last_workfile_path
     if os.path.exists(new_workfile_path):
         new_workfile_path = version_up(new_workfile_path)
+
+    # Raise an error if the parent folder doesn't exist as `host.save_workfile`
+    # is not supposed/able to create missing folders.
+    parent_folder = os.path.dirname(new_workfile_path)
+    if not os.path.exists(parent_folder):
+        raise AssertionError(
```
BigRoy, line 631: I have a feeling this should be a different error than an AssertionError. @iLLiCiTiT thoughts?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -617,7 +617,18 @@ def version_up_current_workfile():
...
+            f"Folder {parent_folder} does not exist yet.")
```
BigRoy, line 632: Is this better if more specific?
```suggestion
            f"Work area directory '{parent_folder}' does not exist yet.")
```

BigRoy (approved):
Aside of cosmetic notes (and the type error) I think this looks. Approving it - but do check what @iLLiCiTiT has to say about the error type.

iLLiCiTiT:
I kinda got lost at what this PR should be handling. What is the issue the PR is solving, or feature it is adding? The description (from what I understand) describes what happens now, but why do we need to make it happen?

client/ayon_core/pipeline/context_tools.py
```diff
@@ -617,7 +617,18 @@ def version_up_current_workfile():
...
+
```
iLLiCiTiT, line 633: If I get it correctly, this is what we need (requires to implement `MissingWorkfile`).
```suggestion
    if not os.path.exists(last_workfile_path):
        raise MissingWorkfile(
            f"No workfile found for current context.")

    new_workfile_path = version_up(last_workfile_path)
```

BigRoy: This will generate many issues I'd say - because we have quite a few places in the code that expect this to return the first valid version number, if none exists yet. 

E.g. all the "Version Up Workfile" file menu entries are allowed to save the initial version (or the new version) if the current scene name is unsaved and there's no workfile yet.

MustafaJafar:
> What is the issue the PR is solving, or feature it is adding? 

Calling `version_up_current_workfile()` when the work dir is empty, creates a workfile with version `v002` instead of `v001`. 
This happens because is called `version_up(...)` twice in this situation.

Another issue that `version_up(filepath)` is expecting an existent work directory and the called function `version_up_current_workfile` is not supposed to initialize work directory therefore we should raise an error that work directory doesn't exist.

client/ayon_core/pipeline/context_tools.py
MustafaJafar: I agree, however, I don't what'd be better. And, I didn't find `FolderNoteFoundError` in compliance to `FileNotFoundError`.

client/ayon_core/pipeline/context_tools.py
iLLiCiTiT: Ok, so based on https://github.com/ynput/ayon-core/pull/1141#issuecomment-2678368973 , this suggestion is actually better?

1. Check if last workfile is existing file, if not, then there is nothing to update?
2. Continue as before...

iLLiCiTiT: Or if the workdir existence is the issue, then please first check the parent folder, before running version up.
```python
    if not os.path.exists(os.path.dirname(last_workfile_path)):
        raise MissingWorkdir(
            "Workdir for current context does not exist.")

    new_workfile_path = last_workfile_path
    if os.path.exists(new_workfile_path):
        new_workfile_path = version_up(new_workfile_path)
```
Again, technically the same, but more readable and understandable.

iLLiCiTiT: And if the issue is missing workdir, shouldn't we just create it instead of raising error?

BigRoy: We could, but creating the folder isn't sufficient. Because technically we should then "initialize the work area" including creating the "Work folders" as per settings, etc. to ensure the studio's work area is actually initialized as intended?

MustafaJafar: What would be the consequences of just creating the folder and saving the workfile?
will that cause skipping initializing the work directory later?

BigRoy: > What would be the consequences of just creating the folder and saving the workfile?
> will that cause skipping initializing the work directory later?

It may mean that we can't rely anymore on the work directory being correctly initialized on that particular run at least, even if it were created later. As such, we somehow are introducing a way where the scene file _may_ be generated and saved there without the work directory being initialized in the way a studio might have configured it should. Resulting in confusion.

So either, we disallow this and error here. Or we enforce this to initialize fully the work directory.

client/ayon_core/pipeline/context_tools.py
```diff
@@ -617,7 +618,18 @@ def version_up_current_workfile():
...
+    new_workfile_path = last_workfile_path
     if os.path.exists(new_workfile_path):
         new_workfile_path = version_up(new_workfile_path)
+
+    # Raise an error if the parent folder doesn't exist as `host.save_workfile`
+    # is not supposed/able to create missing folders.
+    parent_folder = os.path.dirname(new_workfile_path)
+    if not os.path.exists(parent_folder):
+        raise MissingWorkdirError(
+            f"Work area directory '{parent_folder}' does not exist yet.")
```
BigRoy, line 633:
```suggestion
            f"Work area directory '{parent_folder}' does not exist.")
```

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -2,6 +2,10 @@
 from ayon_core.settings import get_project_settings
 
 
+class MissingWorkdirError(Exception):
+    pass
```
BigRoy, line 6: I want to avoid the confusion that this may be something else than the directory not existing on disk, e.g. someone mistakenly assuming this may mean something of 'workdir' is missing in the database or whatsoever.

```suggestion
    """Raised when accessing a work directory not found on disk."""
    pass
```

BigRoy (approved):
Didn't test but looks good to me. Mostly cosmetic notes still though.

client/ayon_core/pipeline/context_tools.py
MustafaJafar: Let just disallow it. _whispering, It's currently disallowed._

iLLiCiTiT (approved):
I guess it is good to go.

MustafaJafar:
I gave it another  and it works.
Note for the future: This is helpful when implementing automations where context is changed and file upversioned by code. 
this can be done via something like

```python
import ayon_api
from ayon_core.pipeline.context_tools import (
    change_current_context,
    get_current_project_name,
    version_up_current_workfile
)

folder_path = "/things/simple_publish" # use an uninitialized folder path i.e. a folder path that wasn't opened via workfiles beforehand.
task_name = "modeling" 

task_entity = ayon_api.get_task_by_folder_path(project_name, folder_path, task_name)

task_entity = ayon_api.get_task_by_folder_path(project_name, folder_path, task_name)

changed = change_current_context(
    folder_entity,
    task_entity
)

version_up_current_workfile()
```

Issue #1142, opened 2025-02-12
Attach reviewables to other instances using EnumDef on 'attachable' integrator


BigRoy:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

The Maya integration has a feature where you can attach a review or render to another product, like e.g. attaching a playblast to a model publish so it's accompanied with a turntable.

However, this logic is unfortunately:
- Maya-specific
- Doesn't scale well to other integrations (each need custom logic)
- Supports only attaching to one other product instead of multiples

### How would you imagine the implementation of the feature?

I'd like to propose creating a dedicated "post-extractor / pre-integrator" plug-in which exposes a multiselection `EnumDef` to pick which instances to attach to.

![Image](https://github.com/user-attachments/assets/43bdb0ae-e324-4bbd-8574-25049fc88848)

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

### Describe alternatives you've considered:

This overlaps a lot with https://github.com/ynput/ayon-core/issues/915 but solving it a different way.

BigRoy:
Not entirely the same thing, but just wanted to link this somewhat relevant [Discord thread](https://discord.com/channels/517362899170230292/563751989075378201/1298252644137304066) about attaching dedicated (existing) reviewable files to a publish in Houdini.

Not 1:1 the same thing.

PR #1143, opened 2025-02-12
Implement Integrator to allow attaching reviewables to other products easily

BigRoy:
## Changelog Description

Allow attaching reviewables to other products directly from within the publisher UI.

image

## Additional info

Fix: https://github.com/ynput/ayon-core/issues/1142

### TODO

- [x] There is one small issue as when the Review been attached to some publish instance the actual review instance itself does not have anything to integrate so there is a warning of that... -> avoid warning
- [x] in ```maya``` and attaching the ```review``` to the ```workfile``` besides ```model``` ...I didnt get any reviewable for ```workfile``` also checked publish folder and there is just the maya work file nothing else.
- x] avoid allowing to attach reviewable to another reviewable?
- [x] ~~potentially allow to attach **but also** still publish separate instance?~~ -> should be separate PR
- [x] ~~Instance attributes should preferably be much tighter "targeted" to define when the attach feature exists (e.g. not all hosts maybe should support it? or should they?) as @iLLiCiTiT mentioned~~ -> for separate issue/PR
- [x] Should need testing with "farm" instance that do not publish locally, but on the farm - which likely would need tweaks in `ayon-deadline` currently. -> for now this is disallowed
- [x] There is a bug where instances (registered later than the review instances by the creators) may be lost as "attach to" selection on the review instances, because initially on reset the instance according to the attribute does not exist. See: https://github.com/ynput/ayon-core/pull/1143#issuecomment-2701521796 -> PR is up: https://github.com/ynput/ayon-core/pull/1192
- [ ] There is still a bug where multiselection `EnumDef` sometimes lose their 'selected values' (unrelated to above mentioned bug that's now fixed).

## Testing notes:

1. Set up dedicated review instance, yet attach it to another instance:

image

3. Publish.
4. The reviewables should now be on the 'attached instances' and the review instance should not be published separately.

image


iLLiCiTiT:
I don't understand what it solves, and where?

BigRoy:
> I don't understand what it solves, and where? Because this looks like issue of Maya, where we have instance with review product type, which probably should not exists, instead each instance should have review camera enumerator, or? Because what is purpose of the review instance with this change?

The review itself is heavily configurable and hence, having a singular place (like the review instance) is totally valid in my opinion. Similarly, it's also totally valid to generate just a reviewable without also requiring to publishing a model, or have it attached. Like being able to just generate a video of the scene (a playblast in maya, opengl render in houdini, etc.)

> This also affects all other hosts that have render product types which isn't really expected behavior.

Correct - the biggest downside of this feature is filtering to finding the right instance of what is actually generating a reviewable. However, _if_ those render instances, generate media that is a render or reviewable then it is actually _correctly_ listed. This feature could e.g. even work in tray publisher if there you'd publish a rendered sequence + an alembic, then attach the render to the alembic.

But I think it's likely best to discuss in a call @mkolar @philippe-ynput @iLLiCiTiT to see whether this is even worth a further pursuit at this point.

It's basically an equivalent of what we're doing in Maya but then supports attaching a review to multiple products and the logic would work across any host that generates the reviewable representations during Extraction.

iLLiCiTiT:
> It's basically an equivalent of what we're doing in Maya but then supports attaching a review to multiple products and the logic would work across any host that generates the reviewable representations during Extraction.

Even tho that is true, it is also something you might want to prevent of doing it. Imagine artist adding review of instane A to instance B in Nuke, with this it is possible. Don't get me wrong I think it does resolve the issues, but I think it should be much more targeted to host and the host products workflow, instead of being able to set review of any render/review instance to any render/review instance. But I might also be wrong and agree that the call would help.

mkolar:
> Even tho that is true, it is also something you might want to prevent of doing it. Imagine artist adding review of instance A to instance B in Nuke, with this it is possible, but not something you want to allow in production.

This is a good general solution to problem we have at the moment I'd say. It would certainly be nice to do it per host with specific solutions to each, it does require some thinking from the artists potentially, but ultimately unblock a lot of things

LiborBatek:
also worth mentioning that when in ```maya``` and attaching the ```review``` to the ```workfile``` besides ```model``` ...I didnt get any reviewable for ```workfile``` also checked publish folder and there is just the maya work file nothing else.

Screenshot 2025-02-14 104144


Worked well for the model product and also when not attaching any publish instances to the review > resulting into disctinct product ```Review``` which is correct.

LiborBatek:
As it is Approved, still something left.

@BigRoy  just want to mention that the ```Review``` when attached to the ```workfile``` publish product nothing been produced resulting into no reviewables integrated...

is it sort of way the things should be or malfunction? Im not sure if this should work too or not.

BigRoy:
Thanks @LiborBatek I have a checklist in the PR description that I will go through first before merging. It should include that one too.

MustafaJafar:
> Allow attaching reviewables to other products directly from within the publisher UI.
> 
> image

Does doing this collapse instances together in the publisher UI?.. 
They are going to be one product eventually.

BigRoy:
> Does doing this collapse instances together in the publisher UI?..
> They are going to be one product eventually.

No. The publisher UI is not aware of them 'belonging together' - also, the reviewable may be attached to more than one product too. I'm not convinced merging them is trivial in a meaningful way from a UX perspective.

philippe-ynput:
Looks like we are done here for now. Can we please merge ASAP ? Thanks 😊

mkolar:
@jakubjezek001 @BigRoy can we get this in soon please? I'd very much appreciate the functionality for some of the demo projects we're working on. Who can unblock this?

BigRoy:
> Who can unblock this?

@iLLiCiTiT most likely due to bug explained [here](https://github.com/ynput/ayon-core/pull/1143#issuecomment-2701521796).
In essence:

> It seems `get_attr_defs_for_instance` gets called before all instances are available to the CreateContext in publisher UI (it basically gets called A LOT OF TIMES)... but in those first calls the other instances do not exist yet (or may not exist yet, depending on the order the instances are created by the create context). As such "listing other instances" as items in the attribute definition fails loses the selected values because there's a brief moment where the other item did not appear in the list anymore and hence... it drops the value.
> Instead it'd make more sense to first collect/create all instances, and then call get_attr_defs_for_instance once all the instances are all populated? 🤷‍♂️

vivimage:
Will it work in tray publisher?

BigRoy:
> Will it work in tray publisher?

The attribute would display on anything that is of product type `render` or `review`, from there it would work in tray publisher. However, beware that many of the tray publisher instances also already have their own _reviewable presentations_ and mixing an "attached" review with reviewable representations is very likely to result in some failure.

But say you just throw in a model alembic to publish, and have a separate review instance to publish - then you should be able to attach the review instance to the model alembic (without needing 'reviewable representations' on the model instance.) That's where it should be able to then work too.

client/ayon_core/plugins/publish/integrate_attach_reviewable.py
```diff
@@ -0,0 +1,136 @@
...
+        for repre in representations:
+            self.log.debug(
+                "Marking representation as deleted because it was "
+                f"attached to other instances instead: {repre}"
+            )
+            repre.setdefault("tags", []).append("delete")
+
+        # Stop integrator from trying to integrate this instance
+        if attach_to:
+            instance.data["integrate"] = False
```
iLLiCiTiT, line 103: Is this good idea for `render` instance?

BigRoy: Yes - because what it integrates - then occurs on the _other_ instances. Basically representations get moved there instead.

iLLiCiTiT: There is a `render` product type that is used as "preview render"?

BigRoy: A turntable render in a lookdev scene for example, may be considered just a reviewable for the published look. So in that sense, I suppose so? You wouldn't do a viewport preview for that - you'd use a render.

iLLiCiTiT: And isn't that confusing? Because it is from my point of view, I have an instance that is suddenly not integrated because some other instance has it set as review source, the instance itself can't affect it at all, and has no visual information about it. And it might makes sense in Maya, but it doesn't make sense in e.g. Nuke or Fusion (and other 2D dccs), or?

mkolar: > And isn't that confusing? Because it is from my point of view, I have an instance that is suddenly not integrated because some other instance has it set as review source, the instance itself can't affect it at all, and has no visual information about it.

It is exactly the intention here. Of course it can be abused, but I'd argue that's not for us to police.

BigRoy: Thanks, resolving this conversation.

Issue #1132, opened 2025-02-10
Load menu : add sub-menus to group loaders


alainxi:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

When we publish images (for example from Substance Painter), we use Ayon's OIIO to automatically provide several resized versions of the images.
The Ayon Loader allows to choose which of the output images we want to load.
This is very useful, but it displays it as a flat list, which can quickly become too long and difficult to read.

### How would you imagine the implementation of the feature?

Please provide sub-menu (see my screenshot and mockup gui), to group by load types, to greatly enhance the use of that gui.

On my screenshot we only output 2 OIIO images, and it's already cumbersome to use with the current layout.
If we wanted to output 5 images, it would be impossible to use the current layout.

![Image](https://github.com/user-attachments/assets/d849ad39-7f65-46c0-973e-d7ebb1c5d941)

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

### Additional context:

I think the flat list entries are set here :
https://github.com/ynput/ayon-core/blob/e58a253511c4d3e9e443870e4c1c985e457bbc9a/client/ayon_core/tools/loader/models/actions.py#L284-L286

Issue #1058, opened 2024-12-16
Load API and Loader actions


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues. I thought we had one, but I couldn't find it.

### Please describe the feature you have in mind and explain what the current shortcomings are?

We're struggling with current load api a lot when it comes to how it's shown, how it is processed and what it can do. We do have `LoaderPlugin` and `ProductLoaderPlugin` that are used to be shown in loader tool and to be called. They usually have predefined way how they are filtered and they have no way how to affect how actions in loader tool are shown.

### Issues
- We use loader plugins as actions and not as loaders, so they don't have to have implement all some of "load" methods that should be mandatory.
- They can't define how their actions are shown in UI -> mostly related to those which are used as actions.
- Loaders can operate only on single item at a time (when loading, switching, removing).
- Loaders have to store metadata about loaded containers in a way so host do understand it.
- There is limitted api for loaded containers -> switch.

### Loader action
Add new plugin type that would work as loader tool action. It would get whole selection information and return possible action items, with submenus icons etc. This should replace existing loader plugins that are actions in fact.

This might need more conversation as we might want to have the actions more generic, not just loader tool specific?

### Load context and plugin
Similar to `CreateContext` we might need `LoadContext` that would wrap load plugins and new `LoadPlugin` as replacement of `LoaderPlugin`. `LoadPlugin` will initialized before any logic happens, it should be able to operate on multiple items at once, even might require to operate on multiple items. Is responsible for storing and collecting containers (like create plugins are) -> we can create a dataclass model for it to make it validated. Can tell if switch can happen or not (e.g. when changing representation).

There might be more requirements I'm not aware of, but from all convesations in past few years these are the most crutial pin points we have. Loader actions are quite clear to me, they can (should) be implemented before change in load api. Please share enhancement ideas and point to bad design choices.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

PR #1146, opened 2025-02-12
Loader: Add submenu to group loaders in load context menu: 

BigRoy:
## Changelog Description

Quick and dirty prototype to test whether grouping of loader actions is better UX or not

An env var `AYON_LOADER_GROUP_MODE` can be used to switch between `never`, `multiples`, `always`. The default is now "always". If an invalid value is set for the env var not from the aforementioned list it will silently use `always`

image

## Additional info

This is merely a draft prototype so one can test whether it actually improves the UX.

There are some issues:
- [ ] Submenu actions don't seem to get highlighting on mouse over
- [ ] The context menu on the individual representations bottom right in the loader UI are also always grouped, even though there should never appear 'duplicates' in that list.
- [ ] The "label" for the menu items is currently very hackily 'patched' from the action item by splitting of the last `" ("` which definitely isn't what we should do.

This may completely be superseded by way more thorough API overhaul: https://github.com/ynput/ayon-core/issues/1058

_This is not production ready ☝️ ; it's merely intended for cosmetically testing the change_

Intends to fix https://github.com/ynput/ayon-core/issues/1132

## Testing notes:

1. Play around and feedback on it cosmetically.

### DIFF
--- a/client/ayon_core/tools/loader/ui/actions_utils.py
+++ b/client/ayon_core/tools/loader/ui/actions_utils.py
@@ -1,6 +1,9 @@
+import os
+import itertools
 import uuid
+from typing import List
 
-from qtpy import QtWidgets, QtGui
+from qtpy import QtWidgets, QtGui, QtCore
 import qtawesome
 
 from ayon_core.lib.attribute_definitions import AbstractAttrDef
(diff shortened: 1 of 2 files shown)


alainxi:
Works for me, until I try to click on the option icons (when `get_options` is implemented).
The icon appears in the menu, but it has no effect (the loader behaves as if I clicked on the menu, not on the icon), the option popup does not appear.

options

BigRoy:
Closing this in favor of the Load/Actions API that will change things up anyway.

PR #1148, opened 2025-02-13
Chore: Environment variables in roots

iLLiCiTiT:
## Changelog Description
Allow to use environment variables in roots.

## Additional info
This is a draft of possibility to use environment variables in root values. Current implementation does simply uses `os.environ` to format root value.

Fix https://github.com/ynput/ayon-core/issues/1175

Resolves [AY-7489](https://app.clickup.com/t/86c23n71z)

### Caveats with this feature
- Roots don't have to be used only in `ayon-core` environment, so we would have to handle all other places.
- We can't map roots for other platform.
- (NOT VERIFIED) All machines where this would be used would have to have filled the environment variable, even if user has filled site root to different value.
- Not sure how site sync is working with roots, it would have to be validated what happens if root has env key and then is synchronized to different platform with root without env keys in value.
- USD resolver does not handle environment variable keys.
- ...Some more I don't know.


client/ayon_core/pipeline/anatomy/roots.py
```diff
@@ -23,16 +23,19 @@ class RootItem(FormatObject):
...
         self.raw_data = lowered_platform_keys
         self.cleaned_data = self._clean_roots(lowered_platform_keys)
         self.name = name
         self.parent = parent
 
         self.available_platforms = set(lowered_platform_keys.keys())
-        self.value = lowered_platform_keys.get(platform.system().lower())
+        self.value = lowered_platform_keys[platform.system().lower()].format(
+            **os.environ
+        )
```
BigRoy, line 38: I think `format_map` (added in Py3.2) should offer better performance.
```suggestion
        self.value = lowered_platform_keys[platform.system().lower()].format_map(
            os.environ
        )
```

By the way, also note the example in the Python docs: https://docs.python.org/3/library/stdtypes.html#str.format_map

Interesting approach with using a custom class with missing there.

philippe-ynput: Let's give sidefx a moment to test it for us... 👍

BigRoy (approved):
Can we merge this? - so far it has been fine on the client site.

@philippe-ynput

philippe-ynput:
@BigRoy I want to talk @iLLiCiTiT first: he has good points.

philippe-ynput:
@BigRoy There are 2 aspects I don't like here:
- Every artist has to set its env vars, otherwise they will get a cryptic error message if it is missing ?
- We already have a feature (roots) to do this, that is actually supported in all use cases.

BigRoy:
> * define a "template root" setting on the project side, i.e. "{SIDEFX_GOOGLE_DRIVE}Shared drive\ayon".
> 
> * The client side picks up the `SIDEFX_GOOGLE_DRIVE` on the artist machine and registers it with the server as this artist's root.

I think whichever way you turn it - you end up getting to a point where you're resolving some "dynamic" value to a static value based on the artist's current environment (which is what they're looking for). I personally don't see much difference with taking the local system's environment variable value and "moving it" somewhere else - but what you're describing would then make sense for resolving user paths server side - e.g. for a potential USD asset resolver?

So the idea is, on launch of the launcher, we allow a certain env var to register a 'user site override' for a certain path entry? Or when would this happen?

Interesting approach, but would easily grow the dev scope of this whereas the client is already using this happily. Meaning this may just linger here as just a client customization in the current state?

philippe-ynput:
I am not against using env vars to customize artist roots but I am against creating a quick and dirty solution that opens the door to more problems.

If we can find a way to grab a pre-determined user env var to set their roots (and garantee everything keeps working as usual), we should do it.

@iLLiCiTiT what do you think ? How much more work would that be ?

iLLiCiTiT:
> I personally don't see much difference with taking the local system's environment variable value and "moving it" somewhere else - but what you're describing would then make sense for resolving user paths server side - e.g. for a potential USD asset resolver?

USD resolver would have to fill the environment variables too, which it does not do now. As far as I know the path resolving does happen on server. It does fill roots based on site id - site id has platform and name, you can change the root of site id on server, so if it has overriden root, the server does know about it.

> So the idea is, on launch of the launcher, we allow a certain env var to register a 'user site override' for a certain path entry? Or when would this happen?

It's not env variable, it is defined on AYON server and Anatomy can resolve it, and site sync can resolve it.

> Interesting approach, but would easily grow the dev scope of this whereas the client is already using this happily. Meaning this may just linger here as just a client customization in the current state?

The issue for me is, that if we do allow it, it looks like it is safe to use, which is not true.

Why I don't like it? Because I know some possible issues, and I don't know consequences. If it would be for my studio where I can fix ad-hoc issues when they arise I would be fine with it, but we don't have access to client machines to do ad-hoc fixes in workfiles. The issues can be revealed at the end of the project and the studio would be screwed.

philippe-ynput:
@BigRoy We discussed this with @mkolar and @iLLiCiTiT and agreed on the following:
- Accept the current change.
- Add extra checks and meaningful error message if the env var is not defined.
- Document this in the code's docstrings, but don't advertise it: at this point we don't know how safe it really is.

iLLiCiTiT:
@BigRoy could you give it another try after changes made recently?

BigRoy (approved):
I set my project anatomy root to `{MYVAR}`, like so:

image

Then ran my local dev env with `set MYVAR=C:/projects` and launch AYON. It could launch apps fine in the correct folder.

Then I ran it without the var... and it would raise an error on launching an application like this:

image

So this seems to work as intended.

Issue #1156, opened 2025-02-18
Loader tool: Tasks filtering


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Add possibility of tasks filtering in loader UI.

### How would you imagine the implementation of the feature?

Fast solution is to use place of current product types filters for tasks widget, which means that product types has to be moved to the header next to product name and statuses filters.

Task widget has to support multiselection of folders, so we probably need to group them by name, and allow individual selection if necessary.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

### Additional context:

This issue is sub-issue of https://github.com/ynput/ayon-core/issues/1131 which also requests tags filtering. That is not available at this moment and we have to prepare server part first (comming next server release).

PR #1157, opened 2025-02-18
Loader tool: Tasks filtering

iLLiCiTiT:
## Changelog Description
Added tasks filtering to loader tool.

## Additional info
Product types filtering was moved next to other products filters, origin place was used for tasks widget. Tasks widget does group tasks by name if multiple folders are selected. Also shows folder label for multiselection of tasks. The selected tasks does filtering at version level, so if task has any version it will show it in UI.

Deselectable tree view was enhanced a little, when there is selected only one item, which is clicked on, it is deselected so it is not necessary to click "elsewhere" to clear selection.

## Testing notes:
1. Open loader tool in tray or in host.
2. Try out new filtering widgets both product types and tasks.

Resolves https://github.com/ynput/ayon-core/issues/1156


BigRoy (changes requested):
I wonder if putting the "Select all/deselect all/swap" option at the top of the list makes more sense than at the bottom.

Similarly is done on AYON web frontend:
image

Now those options are hidden quite far:
image
image

---

Is there any way we can influence the default 'length' the combobox list expands down? With many product types it's really nicer if it's a bit longer.

---

Visually I think it'd be nicer if the "Product type" icon is in the list next to the product types too (even if it's always the blue folder icon for now)...

Although technically we should be able to mimic product types from AYON frontend, even though that's currently still an hardcoded list [here](https://github.com/ynput/ayon-frontend/blob/b254c4aff450c75e2ef6465508dff8dbc846aa2b/src/features/project.js#L3-L22). Still - better than nothing?

---

Does this mean that the settings for `ayon+settings://core/tools/loader/product_type_filter_profiles` have now come irrelevant and should be removed? Or could it technically also set 'defaults' for the "Product types filter" field here?

---

The tasks icons are incorrect, here I'm comparing to tasks in launcher:

image


---

Some screenshots of this PR:

image
image

BigRoy:
Also, I had the UI crash once and have seen these errors a few times pop up - especially when clicking through the folders in the loader and/or when multiselecting folders:
```python
  File "F:\dev\ayon-core\client\ayon_core\tools\loader\ui\tasks_widget.py", line 131, in _on_refresh_thread
    self._fill_data_from_thread(thread)
  File "F:\dev\ayon-core\client\ayon_core\tools\loader\ui\tasks_widget.py", line 199, in _fill_data_from_thread
    parent.removeRow(item.row())
RuntimeError: Internal C++ object (PySide2.QtGui.QStandardItem) already deleted.
```

BigRoy:
And when multiselecting the tasks list doesn't seem entirely correct?

image

It seems to exclude certain things from groups?

image

BigRoy:
### Task list widget columns are 'editable'

The second and third column (when multiselecting folders) are both editable when double clicked. That should be locked instead:
image

---

### Taskless products

I can also have publishes that are taskless (e.g. from tray publisher?)
image

Should the "No task" be a valid 'filter' in the list too to list just those products?

64qam (approved):

iLLiCiTiT:
> Also, I had the UI crash once and have seen these errors a few times pop up - especially when clicking through the folders in the loader and/or when multiselecting folders:

Ynteresting, I'll try to reproduce.

> And when multiselecting the tasks list doesn't seem entirely correct?

Looks like related to the error.

> Should the "No task" be a valid 'filter' in the list too to list just those products?

No strong opinion about this. I think it doesn't make much sense, if you want to filter by task, you want the task, not "without task", or?

iLLiCiTiT:
> It's more that it may not be clear that there may be products that are not related to any of the listed tasks, that's mostly it.

I understand, but adding item for no tasks might be confusin too, or? What would be label and icon? `< Without task >` , or the same as we have for marking folders without task, but it would be enabled instead?

iLLiCiTiT:
> I wonder if putting the "Select all/deselect all/swap" option at the top of the list makes more sense than at the bottom.

Create an issue for this. I don't think it belongs to this PR and it would block it.

BigRoy:
Nice work - some solid improvements since I can't get it to bug out anymore.

Currently leaves:

- [ ] Is there any way we can influence the default 'length' the combobox list expands down? With many product types it's really nicer if it's a bit longer.
- [ ] Without task is nice, but...
    - [ ] Can we move it to the bottom of the list?
    - [ ] Can we hide it if there are no products currently without a task?
    - [ ] I'd personally prefer just the label without the `<>` and even the shorter "No task" 
- [ ] Task icons still incorrect? as described [here](https://github.com/ynput/ayon-core/pull/1157#pullrequestreview-2625214216)
- [ ] What to do with: `ayon+settings://core/tools/loader/product_type_filter_profiles`? It now does nothing, and hence has become irrelevant and should be removed? Or could it technically also set 'defaults' for the "Product types filter" field here?

This will need to become a separate issue after this PR:

- [ ] I wonder if putting the "Select all/deselect all/swap" option at the top of the list makes more sense than at the bottom

iLLiCiTiT:
> Nice work - some solid improvements since I can't get it to bug out anymore.

Please, filter that only to "Must have", the task does have limits on time spent.

BigRoy (approved):
Works for me!

PR #1162, opened 2025-02-21
Host tools: Validate host against correct interface for publisher tool

iLLiCiTiT:
## Changelog Description
Publisher tool validates host agains `IPublishHost` interface instead of `ILoadHost`.

## Additional info
I guess the bug is there since beginning on Publisher tool, but all hosts that do use host tools to show publisher tool, also have implemented load interface. I found out the wrong validating with changes in traypublisher.

## Testing notes:
1. All hosts can open publisher.

### DIFF
--- a/client/ayon_core/tools/utils/host_tools.py
+++ b/client/ayon_core/tools/utils/host_tools.py
@@ -7,7 +7,7 @@ import os
 
 import pyblish.api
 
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Tried Maya, Nuke, Adobe products, Traypublisher

PR #1163, opened 2025-02-24
Combobox: Move Select All, Deselect All and Swap menu entries to top

BigRoy:
## Changelog Description

Combobox: Move Select All, Deselect All and Swap menu entries to top

## Additional info

image

I vaguely have a feeling that this is breaking some behavior somewhere though - because if you refresh the loader UI many times... and then pop-open the dropdown menu then suddenly it force auto scrolls to the bottom of the list. Which may hint this has unintended side effects.

## Testing notes:

1. Menu entries should now appear at top.
2. Menu should still open consistently without issues
3. Refreshing the menu should still work as intended.


iLLiCiTiT:
The loader replaced the combobox filters with filters bar.

image

Not sure if the multiselection combobox is used anywhere now?

PR #1164, opened 2025-02-24
Removed frame recalculation for explicit frames

kalisp:
## Changelog Description
Instances with explicit frames to render (set via `Frames` box in Publisher UI) are marked with `explicit_frames`. Frames for represetations of such instances shouldn't be recalculated.

## Additional info
`frameStart` is not enough for gaps in frame ranges.
This behaviour resulted in issue that even if I set frames `1003-1005` (from `1001-1010` range), they were eventually renamed to `1001-1003` (as `1001` is still `frameStart` and it should stay the same(?))

For future posterity - first idea would be to delete recalculation altogether, but it might still be used and depending on :
~~~Maybe temporary commit as this naive approach won't be possible. Not sure if/what use case was this necessary and if it is still applicable. 
I am throwing this here as a kickoff for discussion. I couldnt find original PR for more details. (As this is there from dawn of time, it seems.)

Another option would be to mark, that we want to render only explicit frames on instance (most likely in `CollectJobInfo` or `AbstractSubmitDeadline` and add condition  to check it to `integrate`. 
This would encompass changes to `ayon-deadline` AND `ayon-core`. 
~~~

More details https://app.clickup.com/t/6658547/AY-7215

## Testing notes:
Requires https://github.com/ynput/ayon-deadline/pull/139
1. try to publish to DL explicit frames (via `Frames` in Publisher UI - 1003,1004,1005)
2. check that they were not renamed to `1003,1004,1005` in `publish` 

Relate to https://github.com/ynput/ayon-deadline/issues/78


iLLiCiTiT:
> Existing code recalculated/renamed rendered files according to values in AYON DB (frameStart). This limits explicit render of selected only frames.

It recalculates the files based on frameStart stored on representation data... Not according to AYON DB, or?

kalisp:
> So, this means that representation does have stored `1001` under `frameStart` on representation.

Does this mean that representation `frameStart` and `frameEnd` should be updated instead? So different `representations` (`versions`) will have different frame ranges? 
I am fine with that, won't this break something unexpectedly? (even in this case there are gaps expected, as clients would like to render every X frame etc.)

iLLiCiTiT:
> Does this mean that representation frameStart and frameEnd should be updated instead? So different representations (versions) will have different frame ranges?
> I am fine with that, won't this break something unexpectedly? (even in this case there are gaps expected, as clients would like to render every X frame etc.)

I didn't go through the code, but how is this related to version? Version has 1-n representations. Representation A has frame range  1001-1020, and Repreesntation B has frame range 1000-1020 (because it has slate).

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -247,7 +247,8 @@ def create_skeleton_instance(
         "useSequenceForReview": data.get("useSequenceForReview", True),
         # map inputVersions `ObjectId` -> `str` so json supports it
         "inputVersions": list(map(str, data.get("inputVersions", []))),
-        "colorspace": data.get("colorspace")
+        "colorspace": data.get("colorspace"),
+        "explicit_frames": data.get("explicit_frames")
```
iLLiCiTiT, line 251:
```suggestion
        "explicitRenderFrames": data.get("explicitRenderFrames")
```

kalisp: Ok.

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -719,8 +740,15 @@ def get_product_name_and_group_from_template(
...
-def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
-                              skip_integration_repre_list, do_not_add_review):
+def _create_instances_for_aov(
+    instance,
+    skeleton,
+    aov_filter,
+    additional_data,
+    skip_integration_repre_list,
+    do_not_add_review,
+    frames_to_render=None
```
iLLiCiTiT, line 750:
```suggestion
    frames_to_render
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -754,10 +783,30 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
+                collections[0], frames_to_render)
+        else:
+            frame_start = int(skeleton.get("frameStartHandle"))
+            frame_end = int(skeleton.get("frameEndHandle"))
+            frames_to_render = list(range(frame_start, frame_end + 1))
 
-        expected_filepath = collected_files
         if isinstance(collected_files, (list, tuple)):
-            expected_filepath = collected_files[0]
+            expected_filepath = os.path.join(staging_dir, collected_files[0])
```
iLLiCiTiT, line 806: How did it before worked as filepaths, and not it is filename? Isn't that because `_collect_expected_files_for_aov` returns filepaths, but `_get_real_files_to_render` does return filenames? If yes, then we should change it to filepaths...

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -475,17 +476,31 @@ def prepare_representations(
...
     """
+    pattern = r"(?:step|by|every|x|:)(\d+)$"
+
     frames_to_render = []
+    step = 1
     for frame in frames.split(","):
         if "-" in frame:
-            splitted = frame.split("-")
+            frame_start, frame_end = frame.split("-")
+            match = re.findall(pattern, frame_end)
```
iLLiCiTiT, line 496:
```suggestion
            match = step_pattern.findall(frame_end)
```

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -392,7 +393,7 @@ def prepare_representations(
         rep = {
             "name": ext,
             "ext": ext,
-            "files": files,
+            "files": [os.path.basename(file_path) for file_path in files],
```
iLLiCiTiT, line 396: I really think this function should expect that `files` is list of filenames.

iLLiCiTiT:
Made small changes. Renamed `get_real_frames_to_render` to `convert_frames_str_to_list` (tells what it actually does, instead for what it is used for).

iLLiCiTiT (approved):

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -506,7 +536,7 @@ def _get_real_files_to_render(collection, frames_to_render):
         collection (clique.Collection): absolute paths
         frames_to_render (list[int]): of int 1001
     Returns:
-        (list[str])
+        (list[str]): absolut paths of files to be rendered
```
BigRoy, line 539:
```suggestion
        (list[str]): absolute paths of files to be rendered
```

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -280,6 +280,10 @@ def _translate_to_sequence(self, files_to_convert):
 
             collection = collections[0]
             frames = list(collection.indexes)
+            real_range = list(range(frames[0], frames[-1] + 1))
+            if set(frames) != set(real_range):  # check for gaps
+                return files_to_convert
```
BigRoy, line 285: Isn't this just:

```suggestion
            if collection.holes():
                return files_to_convert
```

iLLiCiTiT: @kalisp ?

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -754,10 +799,23 @@ def _create_instances_for_aov(instance, skeleton, aov_filter, additional_data,
...
+        first_filepath = collected_files
+        if isinstance(first_filepath, (list, tuple)):
+            first_filepath = first_filepath[0]
+        staging_dir = os.path.dirname(first_filepath)
+
+        if (
+            frames_to_render is not None
+            and isinstance(collected_files, (list, tuple))  # not single file
+        ):
+            frames_to_render = convert_frames_str_to_list(frames_to_render)
```
BigRoy, line 811: This is wrong and bugged.
Client reported a bug, they got an error: `AttributeError: 'list' object has no attribute 'split'``

In this case it seems we're passing a `list` to `convert_frames_str_to_list` which does not handle `list` inputs

PR #1184, opened 2025-03-11
Remove usage of "intent"

BigRoy:
## Changelog Description

The new publisher does not use "intent" in any meaningful way, nor exposes it.

## Additional info

This removes "Validate Intent" and other "intent" related logic.

Whether this is a good idea or not, I'll leave up for discussion - hence this being a **draft PR.**
The other way would be to define what "intent" should be nowadays and how we'd implement that into the publisher from here on out.

## Testing notes:

1. Publishing should still work.
2. Extracting burnins should still work.


BigRoy:
May be good to decide what to do here @mkolar @philippe-ynput to decide what to do with this lingering PR.

antirotor:
I think we should reinstate intent instead. I think it would be as easy as adding one plugin with attribute and if I remember correctly, it was actually requested by one of our clients.

Issue #1186, opened 2025-03-12
Create context: Collect attribute definitions when all instances are collected


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Attribute definitions are collected when instance is created, which might be an issue during collection phase if instance A is dependent on instance B but that is not yet collected.

### How would you imagine the implementation of the feature?

Try to find out if we can skip the automated attributes preparation during collection phase.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

BigRoy:
Related to the issue mentioned [here](https://github.com/ynput/ayon-core/pull/1143#issuecomment-2701521796).

PR #1192, opened 2025-03-13
Create context: Do not trigger publish attributes collection after each add bulk

iLLiCiTiT:
## Changelog Description
Collect publish attributes when all add bulk context managers end.

## Additional info
Each add bulk context manager triggers publish attribute collection for all added instances in that single context, but that should happen at the bulk fully ends.

Positive side effect is that is makes the collection a lot faster.

## Testing notes:
This might be tricky to test as it might reveal possible issues during collection phase. In ideal case scenario we should test all possible instances in all hosts, especially those that have additional logic in collect instances. But I don't think this change does affect them in any way. Creators should not be dependent on publish attributes, so technically nothing should break.
- Collection of instances should not break anything.

Resolves https://github.com/ynput/ayon-core/issues/1186

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -839,7 +839,7 @@ class CreateContext:
                     publish_attributes.update(output)
 
         for plugin in self.plugins_with_defs:
-            attr_defs = plugin.get_attr_defs_for_context (self)
+            attr_defs = plugin.get_attr_defs_for_context(self)
             if not attr_defs:
                 continue
             self._publish_attributes.set_publish_plugin_attr_defs(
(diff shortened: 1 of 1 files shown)


BigRoy:
Just want to say that in Maya - this is _really great!_ It's a massive speed-up on "resetting" the publisher UI. 🚀 
However, I did find another bug which I _think_ is unrelated but am not entirely sure.

Whenever I have a multiselection `EnumDef` the selected items are lost on reset if they had multiple selected.
E.g.

image

After save, and reset:
image

In this PR https://github.com/ynput/ayon-core/pull/1143

However, it does work with `ayon-deadline` machine list `EnumDef`:

This one does persist: 
image

Which means it's something specific to that PR. :/ Any ideas?

BigRoy (approved):
Works perfect on my (very simple) test runs in Tray Publisher, Maya and Fusion.

PR #1193, opened 2025-03-13
Add simple Create Project Structure web action

### LINKED ISSUE #914: AY-5076_Ayon: Folder creation

we need the ability to create folders for production, 

similar we have for track

if implemented as webaction should/could this be a separate addon?

_No response_

_No response_

### Additional context:
()

_This issue was automatically created from Clickup ticket [AY-5076](https://app.clickup.com/t/86bycu87z)_

MustafaJafar:
## Changelog Description
resolve #914
A simple Create Project Structure web action that creates the folder structure as defined in project settings: `ayon+settings://core/project_folder_structure`

image

image

## Testing notes:
1. Update (create and upload) core addon
2. Test CLI action via, it should create folders as per settings.
```shell
ayon_console.exe  createprojectstructure --project <project_name>
```
3. Run the web action action, it's found on any folder in the project on the web UI, it should create folders as per settings.


BigRoy:
Nice! Probably good one for @LiborBatek or @m-u-r-p-h-y to test too?

server/__init__.py
```diff
@@ -26,3 +34,54 @@ async def convert_settings_overrides(
...
+        output.append(
+            SimpleActionManifest(
+                identifier=f"{IDENTIFIER_PREFIX}.createprojectstructure",
+                label="Create Project Folder Structure",
+                icon={
+                    "type": "material-symbols",
+                    "name": "create_new_folder",
+                },
+                order=100,
+                entity_type="folder",
```
BigRoy, line 56: This is an interesting caveat - we can't run "actions" currently in the web frontend on the **project level**. There is no UX for where actions may be displayed on project level. So there is the caveat here that you need to select an AYON folder entity to create the project structure... even though the folders on disk it actually creates have nothing to do with the selected AYON folder entity.

Slightly confusing - but I don't think we have anything better on the "web actions" frontend currently.

BigRoy (approved):
Didn't test, but code looks good. I did leave a comment about this being available on "folders" but I don't think we have better frontend UX for this yet.

client/ayon_core/cli.py
```diff
@@ -257,6 +257,28 @@ def _set_global_environments() -> None:
         os.environ["QT_AUTO_SCREEN_SCALE_FACTOR"] = "1"
 
 
+@main_cli.command()
+@click.option(
+    "--project",
+    type=str,
+    help="Define project (project must be set).",
```
iLLiCiTiT, line 264:
```suggestion
    help="Project name.",
```

iLLiCiTiT: Please move the function above the private functions.

server/__init__.py
```diff
@@ -26,3 +34,54 @@ async def convert_settings_overrides(
...
+
+        """
+
+        project_name = executor.context.project_name
+
+        if executor.identifier == \
+              f"{IDENTIFIER_PREFIX}.createprojectstructure":
+            return await executor.get_launcher_action_response(
+                args=[
+                    "createprojectstructure",
```
iLLiCiTiT, line 82:
```suggestion
                    "create-project-structure",
```

server/__init__.py
```diff
@@ -9,6 +14,9 @@
 )
 
 
+IDENTIFIER_PREFIX = "core.launch"
```
iLLiCiTiT, line 17:
```suggestion
ACTION_ID_PREFIX = "core"
```

server/__init__.py
```diff
@@ -26,3 +34,54 @@ async def convert_settings_overrides(
...
+        executor: "ActionExecutor",
+    ) -> "ExecuteResponseModel":
+        """Execute actions.
+
+        Note:
+            Executes CLI actions defined in the
+            addon's client code or other addons.
+
+        """
+
```
iLLiCiTiT, line 75:
```suggestion
        """Execute webactions."""
```

server/__init__.py
```diff
@@ -26,3 +34,54 @@ async def convert_settings_overrides(
...
+                entity_subtypes=None,
+                allow_multiselection=False,
+            )
+        )
+
+        return output
+
+    async def execute_action(
+        self,
+        executor: "ActionExecutor",
```
iLLiCiTiT, line 66:
```suggestion
        executor: ActionExecutor,
```

server/__init__.py
```diff
@@ -26,3 +34,54 @@ async def convert_settings_overrides(
...
+                icon={
+                    "type": "material-symbols",
+                    "name": "create_new_folder",
+                },
+                order=100,
+                entity_type="folder",
+                entity_subtypes=None,
+                allow_multiselection=False,
+            )
+        )
```
iLLiCiTiT, line 60:
```suggestion
        if project_name:
            output.append(
                SimpleActionManifest(
                    identifier=f"{IDENTIFIER_PREFIX}.createprojectstructure",
                    label="Create Project Folder Structure",
                    icon={
                        "type": "material-symbols",
                        "name": "create_new_folder",
                    },
                    order=100,
                    entity_type="folder",
                    entity_subtypes=None,
                    allow_multiselection=False,
                )
            )
```

client/ayon_core/cli.py
```diff
@@ -22,7 +22,7 @@
     compute_env_variables_structure,
     merge_env_variables,
 )
-
+from ayon_core.pipeline import project_folders
```
iLLiCiTiT, line 25: Do not import pipeline code at the top of this file please.

server/__init__.py
```diff
@@ -26,3 +34,54 @@ async def convert_settings_overrides(
...
+        if executor.identifier == \
+              f"{IDENTIFIER_PREFIX}.createprojectstructure":
+            return await executor.get_launcher_action_response(
+                args=[
+                    "createprojectstructure",
+                    "--project", project_name,
+                ]
+            )
+
+        raise ValueError(f"Unknown action: {executor.identifier}")
```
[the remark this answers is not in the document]
MustafaJafar: Is adding log done in using 
```python
from ayon_server.logging import logger
```
Is it fine to add a dependency ? The new core addon won't be much compatible with older server version. 

For reference, here's a similar situation https://github.com/ynput/ayon-nuke/pull/95

server/__init__.py
MustafaJafar: project name should always be present in this function. plus, we don't use it when appending the action.
so, even if it's `None` or `""`, it shouldn't cause an issue, or 🤔 ?

server/__init__.py
```diff
@@ -26,3 +34,48 @@ async def convert_settings_overrides(
...
+        project_name = executor.context.project_name
+
+        if executor.identifier == \
+              f"{IDENTIFIER_PREFIX}.create_project_structure":
+            return await executor.get_launcher_action_response(
+                args=[
+                    "create-project-structure",
+                    "--project", project_name,
+                ]
+            )
```
MustafaJafar, line 79: I believe there'll always be a `project_name` here. 
But, in case there is not, what should I return 
```suggestion
        if executor.identifier == \
              f"{IDENTIFIER_PREFIX}.create_project_structure":
            
            if not project_name:
                # Do something
                return

            return await executor.get_launcher_action_response(
                args=[
                    "create-project-structure",
                    "--project", project_name,
                ]
            )
```

server/__init__.py
MustafaJafar: Here's also a similar one https://github.com/ynput/ayon-core/pull/1193#discussion_r2003016054

BigRoy: @iLLiCiTiT could you follow up on this?

iLLiCiTiT: > project name should always be present in this function. plus, we don't use it when appending the action.
so, even if it's None or "", it shouldn't cause an issue, or 🤔 ?

This is "discovery" of the action, we should validate the project, just because current web UI does not allow it, doesn't mean we won't (soon) show it elsewhere where it will be possible. Without project, it doesn't make sense to run the action as it requires project to run.

> Here's also a similar one https://github.com/ynput/ayon-core/pull/1193#discussion_r2003016054

That's when the action is triggered, if you trigger the action without a project, it is a bug (I would not modify the code there).

server/__init__.py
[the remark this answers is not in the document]
iLLiCiTiT: Then use `nxtools.logging`.

client/ayon_core/cli.py
iLLiCiTiT: Why is the `"(project must be set)"` kept there?

m-u-r-p-h-y:
Having this available on every folder seems odd. It creates a false expectation that the folder you are running the action on will be created, not just project ones.

I think there is a better place for this on right mouse click on a project. 
image

would it be difficult to do?

BigRoy:
> Having this available on every folder seems odd. It creates a false expectation that the folder you are running the action on will be created, not just project ones.
> 
> I think there is a better place for this on right mouse click on a project. image
> 
> would it be difficult to do?

I don't believe web actions currently allow that, right @martastain @Innders ?

iLLiCiTiT:
> I don't believe web actions currently allow that, right @martastain @Innders ?

Not in Web UI. And if it would be added, it would be in next release of server, which would post-pone this PR again.

Innders:
We are definitely going to need to run actions in the context of just a project and maybe even with no context... Just run the action on a studio.

There was this idea of having a page where you could see all of your actions in one place. Maybe this is the place you can run any action.

If the action requires some context, you must provide it before it can run.

This could probably be an addon that's created.

server/__init__.py
iLLiCiTiT: Please do changes from suggestion.

server/__init__.py
MustafaJafar: I did it anyways https://github.com/ynput/ayon-core/pull/1193/commits/0ec82bd34d83542949ab27ae209c012082714916

server/__init__.py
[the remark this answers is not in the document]
MustafaJafar: I hope I did it correctly. 
https://github.com/ynput/ayon-core/pull/1193/commits/7a7f4b44a04af44d9d3fc8debaaefa52f297c81c

MustafaJafar:
> > Having this available on every folder seems odd. It creates a false expectation that the folder you are running the action on will be created, not just project ones.
> > I think there is a better place for this on right mouse click on a project. image
> > would it be difficult to do?
> 
> I don't believe web actions currently allow that, right @martastain @Innders ?

Should we do it in a follow up PR?

iLLiCiTiT:
> @iLLiCiTiT since I've added `nxtools`, poetry now requires to update the lock file. 
> I'm not able to do becasue of an issue in my dev environment.
> 
> ```
> pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock` to fix the lock file.
> !!! Poetry environment installation failed
> ```

You didn't have to add it :) . You can just remove it, or lock poetry (as suggested), or remove lock file (probably is not needed).

MustafaJafar:
> You didn't have to add it :) . You can just remove it, or lock poetry (as suggested), or remove lock file (probably is not needed) and add the filename to gitignore.

I fixed my issue and updated the poetry lock for consistency. 
If it isn't necessary, could we clean it up it a follow-up PR? Honestly, I don't want to touch something I don't know the effects of.

MustafaJafar:
The poetry file is gone now :D 
@iLLiCiTiT what should we do with this PR?

iLLiCiTiT (approved):
Made small changes in the code to use new executor methods, and simplified description of the settings -> it mentioned ftrack logic (not job of core addon to explain that) and removed explanations that are not relevant for user.

BigRoy (commented):
Where should this action appear?

It does not appear here?
image

And I'm not sure where else I'd find project level actions?

iLLiCiTiT:
> And I'm not sure where else I'd find project level actions?

It is ongoing PR. Both AYON launcher and Web UI.

MustafaJafar:
> So if I were to test this - what do I need? Currently in core, it doesn't show in either for me. :)

I think you need to upload the package.

MustafaJafar:
Making another test run. 
Tagging @iLLiCiTiT 

for server ≥1.9.1
1. I can find the action in the overview page and it works nicely.
image
2. I can't find project action in the launcher UI.

for server < 1.9.1
- I can't find the actions list in the overview page. didn't see any errors.
image

iLLiCiTiT (approved):

Issue #1199, opened 2025-03-17
Pyblish plugin discovery crashes plugins with type annotations and dataclasses


antirotor:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

Discovery function in AYON (and also pyblish-base) doesn't work with plugins with dataclass in the package and type annotations. That's because dataclass is looking into `sys.modules` when processing annotations and we are adding the imported modules (plugins) to `sys.modules` only after they are executed, not before.

**Affected code:**
https://github.com/ynput/ayon-core/blob/f54096a56a06f2804140da19d050fc3293bfde16/client/ayon_core/pipeline/publish/lib.py#L254-L261

**Similar issue:**
https://github.com/mkdocs/mkdocs/issues/3141

### Expected Behavior:

The solution would be to add plugin to `sys.modules` before executing it.

### Version

1.0.14

### What platform you are running on?

Windows

### Steps To Reproduce:

Add this plugin:

```python
"""Test dataclass in publish plugin."""
from __future__ import annotations

from dataclasses import dataclass
from typing import ClassVar

import pyblish.api


@dataclass
class TestPublishData:
    """Test dataclass in publish plugin."""
    id: str
    name: str
    description: str
    version: str


class TestPublishPlugin(pyblish.api.ContextPlugin):
    """Test publish plugin."""
    label = "Test Publish Plugin"
    families: ClassVar[list[str]] = ["*"]

    def process(self, context: pyblish.api.Context) -> None:
        """Process the plugin."""
        _ = TestPublishData(
            id="test",
            name="Test",
            description="Test dataclass in publish plugin.",
            version="1.0.0"
        )
```

Open publisher and you'll find it under *crashed plugins* with error like `'NoneType' object has no attribute '__dict__'`

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
'NoneType' object has no attribute '__dict__'
```

PR #1200, opened 2025-03-17
Fix pyblish plugin discovery when using dataclasses

antirotor:
## Changelog Description
When plugin is using type annotations and dataclasses, it crashed during discovery. This PR is fixing that (and possibly other type hint related issues). Closing #1199 

## Additional info
This PR can possibly affect how plugins are discovered so careful testing is needed.

## Testing notes:
Create plugin which is using dataclass and type annotations. Something like this:

```python
"""Test dataclass in publish plugin."""

from dataclasses import dataclass
from typing import ClassVar

import pyblish.api


@dataclass
class TestPublishData:
    """Test dataclass in publish plugin."""
    id: str
    name: str
    description: str
    version: str


class TestPublishPlugin(pyblish.api.ContextPlugin):
    """Test publish plugin."""
    label = "Test Publish Plugin"
    families: ClassVar[list[str]] = ["*"]

    def process(self, context: pyblish.api.Context) -> None:
        """Process the plugin."""
        _ = TestPublishData(
            id="test",
            name="Test",
            description="Test dataclass in publish plugin.",
            version="1.0.0"
        )
```

Make sure pyblish can see this.

Run publisher by your way of choice. Check *Crashed plugins* - it shouldn't be listed there, it should run as expected.

Also make sure other plugins are discovered as they used to be - in general, in *Crashed plugins* shouldn't be any new items except the usual suspects.


antirotor:
closing as this is based on wrong branch, will reopen with correct one

PR #1201, opened 2025-03-17
Fix publish plugin discovery

antirotor:
## Changelog Description
When plugin is using type annotations and dataclasses, it crashed during discovery. This PR is fixing that (and possibly other type hint related issues). Closing #1199 

## Additional info
This PR can possibly affect how plugins are discovered so careful testing is needed.

## Testing notes:
Create plugin which is using dataclass and type annotations. Something like this:

```python
"""Test dataclass in publish plugin."""

from dataclasses import dataclass
from typing import ClassVar

import pyblish.api


@dataclass
class TestPublishData:
    """Test dataclass in publish plugin."""
    id: str
    name: str
    description: str
    version: str


class TestPublishPlugin(pyblish.api.ContextPlugin):
    """Test publish plugin."""
    label = "Test Publish Plugin"
    families: ClassVar[list[str]] = ["*"]

    def process(self, context: pyblish.api.Context) -> None:
        """Process the plugin."""
        _ = TestPublishData(
            id="test",
            name="Test",
            description="Test dataclass in publish plugin.",
            version="1.0.0"
        )
```

Make sure pyblish can see this.

Run publisher by your way of choice. Check *Crashed plugins* - it shouldn't be listed there, it should run as expected.

Also make sure other plugins are discovered as they used to be - in general, in *Crashed plugins* shouldn't be any new items except the usual suspects.

>[!WARNING]
> There are places using plain pyblish discovery function and there the problem will resurface again (farm publishing, web publisher, ...?). We need to 1) push the fix to pyblish, 2) move these things to ayon-core

### DIFF
--- a/client/ayon_core/lib/python_module_tools.py
+++ b/client/ayon_core/lib/python_module_tools.py
@@ -12,8 +12,8 @@ log = logging.getLogger(__name__)
 
 def import_filepath(
         filepath: str,
-        module_name: Optional[str]=None,
-        sys_module_name: Optional[str]=None) -> types.ModuleType:
+        module_name: Optional[str] = None,
+        sys_module_name: Optional[str] = None) -> types.ModuleType:
     """Import python file as python module.
 
     Args:
--- a/client/ayon_core/lib/python_module_tools.py
+++ b/client/ayon_core/lib/python_module_tools.py
@@ -28,6 +28,7 @@ def import_filepath(filepath, module_name=None):
     module_loader = importlib.machinery.SourceFileLoader(
         module_name, filepath
     )
+    sys.modules[module_name] = module
     module_loader.exec_module(module)
     return module
 
@@ -193,7 +194,7 @@ def is_func_signature_supported(func, *args, **kwargs):
     Notes:
         This does NOT check if the function would work with passed arguments
             only if they can be passed in. If function have *args, **kwargs
-            in paramaters, this will always return 'True'.
+            in parameters, this will always return 'True'.
 
     Example:
         >>> def my_function(my_number):
--- a/client/ayon_core/lib/python_module_tools.py
+++ b/client/ayon_core/lib/python_module_tools.py
@@ -1,6 +1,8 @@
+"""Tools for working with python modules and classes."""
 import os
 import sys
 import types
+from typing import Optional
 import importlib
 import inspect
 import logging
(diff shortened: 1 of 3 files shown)


client/ayon_core/pipeline/publish/lib.py
```diff
@@ -252,17 +289,13 @@ def publish_plugins_discover(paths=None):
                 continue
 
             try:
-                module = import_filepath(abspath, mod_name)
+                module = _import_module(mod_name, Path(abspath))
```
BigRoy, line 292: Why not update the implementation of `import_filepath` instead? I assume a similar fix may be well suited for other cases where we're importing things this way too?

E.g. is the same issue also present in our other plug-in systems currently, like our `Creators`, etc? because those also use `import_filepath` in a similar manner through the `modules_from_path` function. Or?

antirotor: done. I've also made it more safe for backwards compatibility.

client/ayon_core/lib/python_module_tools.py
```diff
@@ -1,20 +1,31 @@
...
 import logging
 
 log = logging.getLogger(__name__)
 
 
-def import_filepath(filepath, module_name=None):
+def import_filepath(
+        filepath: str,
+        module_name: Optional[str]=None,
+        sys_module_name: Optional[str]=None) -> types.ModuleType:
```
iLLiCiTiT, line 16:
```suggestion
        module_name: Optional[str] = None,
        sys_module_name: Optional[str] = None) -> types.ModuleType:
```

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -361,7 +360,7 @@ def get_plugin_settings(plugin, project_settings, log, category=None):
     # Settings category determined from path
     # - usually path is './<category>/plugins/publish/<plugin file>'
     # - category can be host name of addon name ('maya', 'deadline', ...)
-    filepath = os.path.normpath(inspect.getsourcefile(plugin))
+    filepath = os.path.normpath(inspect.getfile(plugin.__class__))
```
iLLiCiTiT, line 363:
```suggestion
    filepath = os.path.normpath(inspect.getfile(plugin))
```

iLLiCiTiT (approved):

BigRoy (approved):
Does seem to fix the issue

munkybutt:
Hey - as a side effect of this any publish classes inside files of the same name can cause conflicts when applying settings from the server.
I just came across this where we have the following two publish plugins:
`ayon_core/plugins/publish/collect_scene_version.py` - class: `CollectSceneVersion`
`ayon_perforce/plugins/publish/collect_scene_version.py` - class: `CollectSceneVersionPerforce`
and the publisher got confused about where the class came from, saying it came from the second file and did not apply the server settings that would have disabled the first collector.
Changing the name of the second file to:
`ayon_perforce/plugins/publish/collect_scene_version_perforce.py` fixed this issue, but something to be aware of. 
The fix to allow dataclasses in the publish file is great though - I was blocked by this recently.

iLLiCiTiT:
> Hey - as a side effect of this any publish classes inside files of the same name can cause conflicts when applying settings from the server.

That's a very good point, I don't know how to handle duplicate file names, but the settings can be resolved by adding `settings_category = "core"` to ayon-core's plugins.

PR #1202, opened 2025-03-18
fixed After Effects unable to load template

miravassor:
"path" -> "template_path"

## Changelog Description
After Effects unable to load a template in the launcher and not opening because of non existing key "path"


BigRoy:
> All other host addons are using `path`, so it should be changed in AfterEffects addon settings model instead.

Created https://github.com/ynput/ayon-aftereffects/pull/25

PR #1204, opened 2025-03-19
Chore: Function to get addons resources directory

iLLiCiTiT:
## Changelog Description
Added new dedicated function `get_addons_resources_dir` where addons resources should be stored, if addon has any.

## Additional info
Addons resources might be ad-hoc files, the best example would be ffmpeg or oiio from 3rd part addon.

## Testing notes:
1. Validate the function name, logic and docstring.


kalisp (approved):
LGTM

PR #1208, opened 2025-03-23
Burnin size can auto adjust by image height

jrsndl:
## Changelog Description
This PR adds setting for Extract Burnin that allows auto adjusting the burnin size relative to image size height.

When creating full res and scaled down reviewables, it is hard to pick good burnin size. For example, good burnin size for UHD is unpleasantly big for 1080p.

## Testing notes:
1. Configure two extract review outputs with different output heights
2. Enable Size relative to height option in Extract Burnin settings
3. Check if Extracted reviewables burnin size matches (resize reviewables to the same size: fit height)

### DIFF
--- a/client/ayon_core/scripts/otio_burnin.py
+++ b/client/ayon_core/scripts/otio_burnin.py
@@ -34,6 +34,8 @@ CURRENT_FRAME_KEY = "{current_frame}"
 CURRENT_FRAME_SPLITTER = "_-_CURRENT_FRAME_-_"
 TIMECODE_KEY = "{timecode}"
 SOURCE_TIMECODE_KEY = "{source_timecode}"
+TEXT_SIZE_RELATIVE = False
+RELATIVE_HEIGHT = 1080
 
 
 def _get_ffprobe_data(source):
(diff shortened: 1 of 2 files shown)


jrsndl:
for issue #1209

iLLiCiTiT:
I don't understand what it means to do? The title says auto adjust, but settings has "unit height", what is that for?

BigRoy:
> I don't understand what it means to do? The title says auto adjust, but settings has "unit height", what is that for?

I suppose it's about resizing font sizes relatively?
And I'd say "unit size" is the size at which you entered the current font sizes?

So that if unit size is say 1000 pixels, and you're rendering now at 2000 pixels then the font size would be 2x what you've entered in settings.
Similarly if you're rendering 500 pixels, the font size would become 0.5x what you have in settings.

Like:
```
size_ratio = current_size / unit_size
font-size *= size_ratio
```

jrsndl:
@iLLiCiTiT @BigRoy 
> I suppose it's about resizing font sizes relatively? And I'd say "unit size" is the size at which you entered the current font sizes?

Exactly, it makes the text size (and other pixel values) to be the same relative size regardless of the image size.

There is often a need to make preview file that is typically >UHD, and editorial reviewable, typically HD.  With "one size for all" approach, text is often too big or too small. SOme projhects also require different resolutions for different shots, relative burnin size makes the burnin look more consistent.

Sorry for not being clear.

PR #1211, opened 2025-03-24
Utils widgets: Ceil height of 'ExpandingTextEdit'

iLLiCiTiT:
## Changelog Description
Ceil height of `ExpandingTextEdit` as it returns float that might be e.g. `0.1` which might cause issues in `QSize`.

## Additional info
Type `float` is "invalid" type for `QSize` in some PySide releases, and I think it might fix issue with fast changes of the height in publisher report.

## Testing notes:
There is not much to test.

### DIFF
--- a/client/ayon_core/tools/utils/widgets.py
+++ b/client/ayon_core/tools/utils/widgets.py
@@ -1,4 +1,5 @@
 import logging
+import math
 from typing import Optional, List, Set, Any
 
 from qtpy import QtWidgets, QtCore, QtGui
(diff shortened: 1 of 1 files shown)


BigRoy (commented):
Didn't test - but changes make sense.

However, I think we're better off fixing the return value for `heightForWidth()` method instead.
According to the docs it should return `int`, see [here](https://doc.qt.io/qt-6/qwidget.html#heightForWidth). So we're just returning the wrong type ourselves?

iLLiCiTiT:
> Didn't test - but changes make sense.
> 
> However, I think we're better off fixing the return value for `heightForWidth()` method instead. According to the docs it should return `int`, see [here](https://doc.qt.io/qt-6/qwidget.html#heightForWidth). So we're just returning the wrong type ourselves?

True, changed.

PR #1214, opened 2025-03-27
Allow disabling removal of rendered files on farm renders

BigRoy:
## Changelog Description

Allow disabling removal of rendered files on farm renders

## Additional info

image

With `ayon+settings://core/publish/CollectRenderedFiles/remove_files` disabled the `metadata.json` and work area renders will remain on disk after a publish even if the folder is not a persistent staging dir

## Testing notes:

1. Submit to farm with this disabled.
2. Publish job should not delete the renders or metadata json file.


kalisp:
I am just wondering if we really need to have another toggle in settings and not depend on staging dir persistence only? (If it is not working, its a bug.)
@philippe-ynput

BigRoy:
The issue with **custom staging dir persistence** is that you are **required** to configure a custom directory - whereas most don't want to define yet another template or directory definition but just tell the system to wherever you're rendering, please do not remove the renders... even by default.

I'd have preferred to be able to not have an additional toggle.

If we could just mark the render files as "persistent" without needing to additionally set up custom render staging dir paths, etc.

philippe-ynput:
@kalisp I agree with @BigRoy on that one. And the default should be reversed too.

kalisp:
But this approach would keep all rendered files from all DCCs just for sake of Nuke, wouldn't it?

BigRoy:
> But this approach would keep all rendered files from all DCCs just for sake of Nuke, wouldn't it?

It would do it for all DCCs, yes - as is intended. The request has come up for different hosts quite a few times:


Community public:
- https://community.ynput.io/t/how-to-disable-cleanup-feature/1726/6?u=bigroy
- https://community.ynput.io/t/ability-to-comp-frames-that-are-being-rendered-on-deadline/2308/7?u=bigroy
- https://community.ynput.io/t/nuke-deadline-cleanup/2417/6?u=bigroy
- https://discord.com/channels/517362899170230292/563751989075378201/1353943789735116852

Private clients:
- https://discord.com/channels/517362899170230292/920042152929677363/1354782287165919364
- https://discord.com/channels/517362899170230292/1343860429729828907/1343861705293303839

Covering Nuke, Maya and Houdini.

MustafaJafar (approved):
I've tested it on my side and it works. 

It doesn't take intermediate files into consideration and I believe this is correct as outlined in this comment https://github.com/ynput/ayon-houdini/blob/5be04ed5d30bbb53cad4405b62b180f4eb72ea9f/client/ayon_houdini/plugins/publish/collect_files_for_cleaning_up.py#L23-L30

server/settings/publish_plugins.py
```diff
@@ -1428,6 +1446,9 @@ class PublishPuginsModel(BaseSettingsModel):
     "AttachReviewables": {
         "enabled": True,
     },
+    "CollectRenderedFiles": {
+        "remove_files": True
```
BigRoy, line 1450: If I understood correctly from @philippe-ynput we actually, moving forward, want to disable this removing behavior by default. If so, this suggestion should be applied.

```suggestion
        "remove_files": False
```

client/ayon_core/plugins/publish/collect_rendered_files.py
```diff
@@ -31,6 +31,9 @@ class CollectRenderedFiles(pyblish.api.ContextPlugin):
     # Keep "filesequence" for backwards compatibility of older jobs
     targets = ["filesequence", "farm"]
     label = "Collect rendered frames"
+    settings_category = "core"
+
+    remove_files = True
```
BigRoy, line 36: See [here](https://github.com/ynput/ayon-core/pull/1214/files#r2047019732)
```suggestion
    remove_files = False
```

philippe-ynput (approved):

server/settings/publish_plugins.py
```diff
@@ -929,6 +929,20 @@ class IntegrateHeroVersionModel(BaseSettingsModel):
                     "hero versions.")
 
 
+class CollectRenderedFilesModel(BaseSettingsModel):
+    remove_files: bool = SettingsField(
+        True,
```
BigRoy, line 934:
```suggestion
        False,
```

PR #1216, opened 2025-03-31
Used new enum for template names from Anatomy

kalisp:
## Changelog Description
There was a new `anatomy_template_items_enum` added to server backend which provides dropdown instead of manually providing text value which could cause issues of wrong format/typos.

`anatomy_template_items_enum` could be used with selected template type as `publish`|`render` or without it. In case of without preselection template name is concatenated in format `tempateType_templateName` (eg. `publish_default`) as each template type contains at least same `default` value.

## Additional info
This PR depends on https://github.com/ynput/ayon-backend/pull/558 and should be merged ONLY after new release of server. (And bumping ayon server dependency in `package.py`)

Defaults contain `tycache` values which are not part of `Anatomy`. These defaults should be probably removed.

## Testing notes:
1. use `ayon-backend` PR for your server
2. install package for ayon-core
3. check `ayon+settings://core/tools/publish/template_name_profiles`


MilaKudr (approved):
its there
Screenshot 2025-03-31 152834

iLLiCiTiT:
This PR should bump required server version.

server/settings/tools.py
```diff
@@ -306,7 +311,11 @@ class CustomStagingDirProfileModel(BaseSettingsModel):
     custom_staging_dir_persistent: bool = SettingsField(
         False, title="Custom Staging Folder Persistent"
     )
-    template_name: str = SettingsField("", title="Template Name")
+    template_name: str = SettingsField(
+        "",
+        title="Template name",
+        enum_resolver=anatomy_template_items_enum
```
iLLiCiTiT, line 317: missing category

server/settings/tools.py
```diff
@@ -314,7 +314,7 @@ class CustomStagingDirProfileModel(BaseSettingsModel):
     template_name: str = SettingsField(
         "",
         title="Template name",
-        enum_resolver=anatomy_template_items_enum
+        enum_resolver=anatomy_template_items_enum(category="publish")
```
iLLiCiTiT, line 317:
```suggestion
        enum_resolver=anatomy_template_items_enum(category="staging")
```

iLLiCiTiT:
Ok, so this PR is created at a time the enumerators are even not in any release. I think we should wait for some time with merging this.

jakubjezek001 (approved):
Agree that this should be merged after ayon-backend#558, but tests were working as expected so it is approval from me ;)

kalisp:
This could get merged theoretically, correct? I bumped server dependency to >=1.7.6 (first version of server with enum)

PR #1217, opened 2025-03-31
Transcoding: Use single `oiiotool` call for sequences, instead of frames one by one

BigRoy:
## Changelog Description

Use single `oiiotool` command and some other optimizations.

- Prepare attributes to remove list just once.
- Process sequences as a single `oiiotool` call

## Additional info

This leaves the question whether this is the best approach - or whether pooling subprocesses of different `oiiotool` processes is a better approach to get the best performance.

Also, it seems a somewhat similar "deprecated" implementation existed in the codebase - it's unclear why excatly that was dropped.

Thoughts?

## Testing notes:

1. Publishing `.exr` with review should work.

### DIFF
--- a/client/ayon_core/lib/transcoding.py
+++ b/client/ayon_core/lib/transcoding.py
@@ -10,6 +10,8 @@ from typing import Optional
 
 import xml.etree.ElementTree
 
+import clique
+
 from .execute import run_subprocess
 from .vendor_bin_utils import (
     get_ffmpeg_tool_args,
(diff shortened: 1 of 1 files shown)


client/ayon_core/lib/transcoding.py
```diff
@@ -526,135 +528,36 @@ def should_convert_for_ffmpeg(src_filepath):
...
-        "--subimage", "0"
-    ])
-
-    # Add frame definitions to arguments
-    if is_sequence:
-        oiio_cmd.extend([
-            "--frames", "{}-{}".format(input_frame_start, input_frame_end)
-        ])
-
+def _get_attributes_to_erase(input_info: dict, logger: logging.Logger = None) -> list[str]:
```
iLLiCiTiT, line 531:
```suggestion
def _get_attributes_to_erase(input_info: dict, logger: logging.Logger) -> list[str]:
```

client/ayon_core/lib/transcoding.py
```diff
@@ -526,135 +528,36 @@ def should_convert_for_ffmpeg(src_filepath):
...
 
-    logger.debug("Conversion command: {}".format(" ".join(oiio_cmd)))
-    run_subprocess(oiio_cmd, logger=logger)
+    if logger is not None:
+        for attr_name, reason in zip(erase_attr_names, reasons):
+            # Set attribute to empty string
+            logger.info(
+                f"Removed attribute \"{attr_name}\" from metadata"
+                f" because {reason}."
+            )
```
iLLiCiTiT, line 559:
```suggestion
    for attr_name, reason in zip(erase_attr_names, reasons):
        # Set attribute to empty string
        logger.info(
            f"Removed attribute \"{attr_name}\" from metadata"
            f" because {reason}."
        )
```

client/ayon_core/lib/transcoding.py
```diff
@@ -707,7 +610,19 @@ def convert_input_paths_for_ffmpeg(
...
+    )
+
+    input_collections, input_remainders = clique.assemble(
+        input_paths,
+        patterns=[clique.PATTERNS["frames"]],
+        assume_padded_when_ambiguous=True,
+    )
+    process_inputs = list(input_collections)
+    process_inputs.extend(input_remainders)
+    for _input in process_inputs:
```
iLLiCiTiT, line 625:
```suggestion
    input_items = list(input_collections)
    input_items.extend(input_remainders)
    for input_item in input_items:
```

client/ayon_core/lib/transcoding.py
iLLiCiTiT, line 1: Missing import `from __future__ import annotations`

client/ayon_core/lib/transcoding.py
```diff
@@ -526,135 +528,36 @@ def should_convert_for_ffmpeg(src_filepath):
...
-
-        if erase_attribute:
-            # Set attribute to empty string
-            logger.info((
-                "Removed attribute \"{}\" from metadata because {}."
-            ).format(attr_name, erase_reason))
-            oiio_cmd.extend(["--eraseattrib", attr_name])
+            reason = f"has too long value ({len(attr_value)} chars)."
+            erase_attr_names.append(attr_name)
+            reasons.append(reason)
```
iLLiCiTiT, line 544:
```suggestion
            reasons.append(reason)
            continue
```

client/ayon_core/lib/transcoding.py
```diff
@@ -526,135 +528,36 @@ def should_convert_for_ffmpeg(src_filepath):
...
-    # Add frame definitions to arguments
-    if is_sequence:
-        oiio_cmd.extend([
-            "--frames", "{}-{}".format(input_frame_start, input_frame_end)
-        ])
-
+def _get_attributes_to_erase(input_info: dict, logger: logging.Logger = None) -> list[str]:
+    """FFMPEG does not support some attributes in metadata."""
+    erase_attr_names = []
+    reasons = []
```
iLLiCiTiT, line 534: These 2 could be maping attr name -> reason.

client/ayon_core/lib/transcoding.py
```diff
@@ -718,47 +633,29 @@ def convert_input_paths_for_ffmpeg(
...
+        if isinstance(_input, clique.Collection):
+            oiio_cmd.extend([
+                "--framepadding", _input.padding,
+                "--frames", _input.format("{ranges}"),
+            ])
+            _input: str = _input.format("{head}#{tail}")
+
         oiio_cmd.extend([
-            input_arg, input_path,
+            input_arg, _input,
```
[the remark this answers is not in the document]
BigRoy: remainder is `list[str]`

BigRoy:
If could update beyond 2.5.10 for OIIO, then we could also use the `--parallel-frames` argument here so that it's processing multiple frames in parallel within the one process.

BigRoy:
> If could update beyond 2.5.10 for OIIO, then we could also use the --parallel-frames argument here so that it's processing multiple frames in parallel within the one process.

As I've been testing with @antirotor with recent OIIO 3.x builds the performance in recent releases seems much better - and is even greatly improved beyond that with `--parallel-frames` implemented in OIIO with [this commit](https://github.com/AcademySoftwareFoundation/OpenImageIO/commit/f40f9800c83e2c596c127777bea1e468564fbb10) in [this PR](https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/3849) which seems to have been in there since OIIO 2.5.2.0.

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -279,15 +295,13 @@ def _translate_to_sequence(self, files_to_convert):
...
-            file_name = "{}{}{}".format(collection.head, frame_str,
-                                        collection.tail)
 
-            files_to_convert = [file_name]
+            # Support sequences with holes by supplying dedicated `--frames`
+            # Create `frames` string like "1001-1002,1004,1010-1012
+            frames: str = collection.format("{ranges}").replace(" ", "")
+            # Create `filename` string like "file.%04d.exr"
+            filename = collection.format("{head}{padding}{tail}")
+            return [(filename, frames)]
```
BigRoy, line 304: @iLLiCiTiT any quick ideas on how to improve this function to have a more consistent return type. :D My brain is tired today it seems.

BigRoy: Changed it around - not sure if I'm entirely happy cosmetically - but it should work fine.

kalisp:
`Warning: This requires` does this mean it will fail with current version of oiio, so this will need to get merged ONLY AFTER new ayon-third-party is created (and dependency bumped up in this PR)?

BigRoy:
> Warning: This requires does this mean it will fail with current version of oiio, so this will need to get merged ONLY AFTER new ayon-third-party is created (and dependency bumped up in this PR)?

Correct - the old OIIO will complain about the `--parallel-frames` argument being unknown.

jakubjezek001:
related pr #1277

BigRoy:
@iLLiCiTiT should this PR make a 'soft' requirement dependency to `ayon-third-parties 1.3.0`?

client/ayon_core/lib/transcoding.py
```diff
@@ -526,135 +529,36 @@ def should_convert_for_ffmpeg(src_filepath):
...
+            break
 
-    logger.debug("Conversion command: {}".format(" ".join(oiio_cmd)))
-    run_subprocess(oiio_cmd, logger=logger)
+    if logger is not None:
+        for attr_name, reason in erase_attrs.items():
+            logger.info(
+                f"Removed attribute \"{attr_name}\" from metadata"
+                f" because {reason}."
+            )
```
iLLiCiTiT, line 560: Why the check if is expected to be passed in.
```suggestion
    for attr_name, reason in erase_attrs.items():
        logger.info(
            f"Removed attribute \"{attr_name}\" from metadata"
            f" because {reason}."
        )
```

BigRoy: Actually - I think we should fix the type hint instead. It's used in `convert_input_paths_for_ffmpeg` which actually defaults `logger` to `None`

iLLiCiTiT:
> @iLLiCiTiT should this PR make a 'soft' requirement dependency to ayon-third-parties 1.3.0?

If we can check for the OIIO version and use the "enhancement arguments" only if the arguments are supported by the version, then that would be prefered. If no, then soft requirement in package.py it is.

BigRoy:
> If we can check for the OIIO version and use the "enhancement arguments" only if the arguments are supported by the version, then that would be prefered. If no, then soft requirement in package.py it is.

I personally feel that adding such a version check is just more hassle and will just add more slowness to the process due to the overhead. Basically running it more often is exactly the opposite of what we want to achieve, hehe. The OIIO version we were using was old. Like, really really old. Good riddance, really.

Maybe the better approach is to add the soft dependency and maybe wait a little longer before merging (as then third party package would've been released for a while).

iLLiCiTiT:
> I personally feel that adding such a version check is just more hassle and will just add more slowness to the process due to the overhead. Basically running it more often is exactly the opposite of what we want to achieve, hehe. The OIIO version we were using was old. Like, really really old. Good riddance, really.

It is not that slow, compared to real oiiotool conversion. And the version can be cached, so it's called only once.

> Maybe the better approach is to add the soft dependency and maybe wait a little longer before merging (as then third party package would've been released for a while).

The new versions were released, but not yet tested on all linux distributions we need to support. So adding soft dependency could mean that we might need freeze some studios on older ayon-core because of that.

BigRoy:
> It is not that slow, compared to real oiiotool conversion. And the version can be cached, so it's called only once.

It's anywhere between 0.15 and 0.02 seconds for me (oiiotool served over network share) with the 0.13 being a cold run, and all other subsequent runs between 0.04-0.02 seconds from server cache over 10 Gbps connections. 

The highest time I got was `0.2409 seconds` - but that was cold cache + cold run (regex needing to compile perhaps?) - however, that's most likely the case if e.g. a machine just randomly starts to run this every now and then, like in production on a publish or on farm.

antirotor:
I don't really mind soft requirement. We did the same with ocio recently. It is already used in productions and so far no issues reported.

jakubjezek001:
Is there any chance we could just merge this one plese, it is starting to be quite urgent on some clients. @philippe-ynput @antirotor @iLLiCiTiT

client/ayon_core/lib/transcoding.py
```diff
@@ -1017,6 +1073,24 @@ def convert_colorspace(
...
+
+    if frame_padding:
+        oiio_cmd.extend([
+            "--framepadding", str(frame_padding),
+        ])
+
+    if parallel_frames:
+        oiio_cmd.extend([
+            "--parallel-frames"
+        ])
```
iLLiCiTiT:
```suggestion
        oiio_cmd.append("--parallel-frames")
```

BigRoy:
@jakubjezek001 this works fine for me with `ayon-third-parties 1.4.0` - can you give it a whirl?

BigRoy:
> If we can check for the OIIO version and use the "enhancement arguments" only if the arguments are supported by the version, then that would be prefered. If no, then soft requirement in package.py it is.

Ended up adding soft requirement with https://github.com/ynput/ayon-core/pull/1217/commits/55eb4cccbe267af4f8560604e73a7214b13a276b - those releases have been out a while, and updating OIIO is getting to a point where it'd really be recommended.

image

Issue #928, opened 2024-10-01
AY-2218_Plugin hooks - Loader and Scene Inventory


ynbot:
## Plugin hooks 🪝 

Extending or changing functionality in current plugin system is difficult as extending classes of loaders or actions create dependency  between any custom code and the plugin functionality. If the loader changes, all custom modifications can break down the line. 

Hook system provides entry points for arbitrary code to run upon plugin execution without binding it to the code of the plugin itself. For example:

Loader that is loading model from OBJ file in Maya. That loader will invoke some Maya logic to load the actual format and it will containerize the data in Maya so it can be version controled. But what if studio wants to do something else with loaded data - what if it wants to alway apply some transformation, or automatically assign some default shader, fill some metadata somewhere?

Currently, the only way would be to create new loader inheriting the original one (but with that you would need to solve import issues), you would probably need to fork maya addon repo. Or you could change the logic directly, or you could duplicate the code into your loader in your studio addon and somehow disable the original one.

All this is supercumbersome and it is solved by hooks.

## Captain Hook 🚢 

Hooks are scripts that can be executed within some context on certain time. Registering hooks and executing them is sort of plugin system. In case of Loader and Actions execution point in time is probably "before load" and "after load" and something needs to be passed to the hook to make it useful (probably runtime content of arguments passed to `load()`)

You should be able to register your hooks to execution point in time and specific plugin. So back to the example above: lets say the OBJ loader class is `LoadObj`. I have hook called "`assign_shader_to_obj`" that is implemented in some hook class with some entry point method. I register my hook by calling AYON API that passes the hook object itself, the loader class name and "the execution time type" and when this loader is run, it will execute my hook.

_This issue was automatically created from Clickup ticket [AY-2218](https://app.clickup.com/t/862ke0qjw)_

PR #1222, opened 2025-04-03
Pre and post loader hooks

kalisp:
## Changelog Description
This PR introduces new concept of `Pre/Post Loader Hooks` which could be used by to enhance or modify loading processes without need of modifying official loaders.

The idea is for custom addons to be able to implement their own `PrePostLoaderHookPlugin` which would be triggered before/after Loader action.

Method to be implemented in `PrePostLoaderHookPlugin`:
- `pre_load`
- `post_load`
- `pre_update`
- `post_update`
- `pre_remove`
- `post_remove`

The function signature (arguments) to each of the pre/post methods matches the regular `LoadPlugin` except that `post_*` methods take an additional `result` argument which receives the output of the `LoadPlugin` method call, like simplified:
```python
pre_load()
result = load()
post_load(x, y, z, result=result)
```

## Additional info
Each addon must register its folder with pre/post hooks like this:
```        
    register_loader_hook_plugin_path(loader_hook_dir)
```

I am testing it with this dummy custom addon which shows hooks for `"CopyFile", "FileLoader", "LoadImage"` Loaders.
~~[my_custom_addon-1.0.0.zip](https://github.com/user-attachments/files/20777382/my_custom_addon-1.0.0.zip)~~ (outdated example files)

### Register example hook to just print all the hook methods

```python
from ayon_core.pipeline.load.plugins import LoaderHookPlugin
from ayon_core.pipeline.load import register_loader_hook_plugin


class LoadHookTest(LoaderHookPlugin):
    @classmethod
    def is_compatible(cls, Loader) -> bool:
        return True

    def pre_load(self, context, name=None, namespace=None, data=None, plugin=None):
        self.log.info("pre_loadaa")

    def post_load(self, context, name=None, namespace=None, data=None, plugin=None, result=None):
        self.log.info("post_load")

    def pre_update(self, container, context, plugin=None):
        self.log.info("pre_update")

    def post_update(self, container, context, plugin=None, result=None):
        self.log.info("post_update")

    def pre_remove(self, container, plugin=None):
        self.log.info("pre_remove")

    def post_remove(self, container, plugin=None, result=None):
        self.log.info("post_remove")
		
		
register_loader_hook_plugin(LoadHookTest)
```

Or when targeting specific loaders and e.g. writing to a file on disk on each hook call:
```python
from ayon_core.pipeline.load import get_loader_identifier
from ayon_core.pipeline.load.plugins import LoaderHookPlugin


class PrePostCopyFile(LoaderHookPlugin):
    """Example of hook which would be run before and after CopyFile.load()"""
    @classmethod
    def is_compatible(cls, Loader) -> bool:
        identifier = get_loader_identifier(Loader)
        return identifier in {"CopyFile", "FileLoader", "LoadImage"}

    def pre_load(self, context, name=None, namespace=None, data=None, plugin=None):
        self._write("pre_load")

    def post_load(self, context, name=None, namespace=None, data=None, plugin=None, result=None):
        self._write("post_load")

    def pre_update(self, container, context, plugin=None):
        self._write("pre_update")

    def post_update(self, container, context, plugin=None, result=None):
        self._write("post_update")

    def pre_remove(self, container, plugin=None):
        self._write("pre_remove")

    def post_remove(self, container, plugin=None, result=None):
        self._write("post_remove")

    def _write(self, value):
        with open("c:/projects/pre_post_load.txt", "a") as fp:
            fp.write(f"{value}\n")
```

## Testing notes:
1. update `copy_file_hook.py` in `my_custom_addon` however you feel fit for your testing
2. follow this step


client/ayon_core/tools/loader/models/actions.py
```diff
@@ -737,12 +781,26 @@ def _load_representations_by_loader(self, loader, repre_contexts, options):
             if version < 0:
                 version = "Hero"
             try:
+                for hook_plugin in hooks.get("pre", []):
+                    hook_plugin.process(
+                        loader,
+                        repre_context,
+                        options=options
```
BigRoy, line 788: The hook logic now seems to run in the UI - whereas it should most likely _actually_ trigger in the load API itself, right?

So that if you were to programmatically load or (or update?) that it'd also trigger the revelant hooks (or callbacks or what you may call it.)

BigRoy:
I wonder if new hook plug-ins here make more sense than just an `EventSystem` 😕 Seems somewhat overkill here.

BigRoy:
> > I wonder if new hook plug-ins here make more sense than just an `EventSystem` 😕 Seems somewhat overkill here.
> 
> Had the same question, but they should be blocking, I dont think you can achieve that with events.

The Event system in AYON _is_ blocking by default, right @iLLiCiTiT ?

client/ayon_core/tools/loader/models/actions.py
kalisp: Moved it out to utils.

iLLiCiTiT:
> The Event system in AYON is blocking by default, right @iLLiCiTiT ?

Yes it is. But it matters what should happen during the hooks, and what is the reason/use-case to add the hooks?

Because adding hooks (2 classes) to the current crap load api means that we will have to add more backwards compatibility with more classes to be duplicated when the api changes, which will complicate it.

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -1,5 +1,6 @@
 import os
```
iLLiCiTiT, line 1: from __future__ import annotations

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -264,6 +265,43 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        pass
+
+    def update(self, container, context):
+        pass
+
+    def switch(self, container, context):
+        pass
+
+
+class PostLoadHookPlugin:
```
iLLiCiTiT, line 287:
```suggestion
class PostLoaderHookPlugin:
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -264,6 +266,43 @@ class ProductLoaderPlugin(LoaderPlugin):
     """
 
 
+class PreLoaderHookPlugin:
```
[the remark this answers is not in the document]
BigRoy: Also, how should the hooks behave if the loading failed? Should post hooks still trigger, or should those not?
Also, what should happen if a pre hook errors, should it continue loading, or should it stop?

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -264,6 +266,43 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+
+    def switch(self, container, context):
+        pass
+
+
+class PostLoaderHookPlugin:
+    """Plugin that should be run after any Loaders in 'loaders'
+
+    Should be used as non-invasive method to enrich core loading process.
+    Any external studio might want to modify loaded data before or afte
```
BigRoy, line 292:
```suggestion
    Any external studio might want to modify loaded data before or after
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -264,6 +266,43 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        pass
+
+    def update(self, container, context):
+        pass
+
+    def switch(self, container, context):
+        pass
+
+
+class PostLoaderHookPlugin:
```
BigRoy, line 288: Should Post Loader hook plug-in get access to the loaded container? Or how would one actually detect _what_ was loaded?

iLLiCiTiT:
I wanted to ask very similar questions to @BigRoy 's but I wanted to wait for this first https://github.com/ynput/ayon-core/pull/1222#issuecomment-2782443206 . We really need to know what is the use-case. It's all related to current crap api. Loader plugins are also "actions" (delivery etc.), so I guess the pre/post hook needs to know "some" context, but because current api does not actually tell you anything about "if", "where" or "what" actually happened it's just a random class.

kalisp:
Summoning @antirotor to describe more details about https://github.com/ynput/ayon-core/issues/928

client/ayon_core/pipeline/load/plugins.py
kalisp: Added container argument

client/ayon_core/pipeline/load/plugins.py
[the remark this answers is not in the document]
kalisp: Thats probably question of who is making pre/post hook. If they want to loading to fail/be skipped, they need to raise an exception.

iLLiCiTiT:
> There are also often customization about hierarchy where loaded stuff should end up - if you want to customize this - to add loaded model into very specific structure, again - the whole loader logic is the same, you just need to move loaded items somewhere.

So how this will gonna help if you don't know what happened in `load`?

iLLiCiTiT:
Also important note is that this PR is adding 2 new classes and at 10 new functions just to register and discover them, changing api functions for load. All of that will make backwards compatibility even harder when we befin with new load api.

client/ayon_core/tools/loader/models/actions.py
```diff
@@ -307,22 +316,29 @@ def _get_loaders(self, project_name):
                 we want to show loaders for?
 
         Returns:
-            tuple[list[ProductLoaderPlugin], list[LoaderPlugin]]: Discovered
-                loader plugins.
+            tuple(
+                list[ProductLoaderPlugin],
+                list[LoaderPlugin],
+            ): Discovered loader plugins.
```
iLLiCiTiT, line 322: Why this changed?

antirotor:
We should move with this forward. So, what about one class for hooks (`PrePostLoadHook`) with `pre_load` and `post_load` methods - hook it on loader class name and monkey-patch loader `.load()` so it executes those hook methods before and after? Make it as simple as possible.

>So how will these changes help if you don't know what happened in load?

You can go around by listing the scene in pre_load and post_load and you'll get what loader pulled in and then you can manipulate the result.

kalisp:
Vastly simplified logic.

Currently monkey patched method is added to Loader classes who are compatible to any registered `PrePostLoaderHookPlugin` during `discover_loader_plugins`.

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +252,50 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        self,
+        context: dict,
+        name: str | None = None,
+        namespace: str | None = None,
+        options: dict | None = None,
+        container: dict | None = None,  # (ayon:container-3.0)
+    ):
+        pass
+
+    def switch(self, container, context):
```
iLLiCiTiT, line 285: There probably should be `pre_switch`, `post_switch`, `pre_update`, `post_update`.

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -268,11 +304,43 @@ def discover_loader_plugins(project_name=None):
                 "Failed to apply settings to loader {}".format(
                     plugin.__name__
                 ),
-                exc_info=True
+                exc_info=True,
             )
+        for hookCls in hooks:
+            if hookCls.is_compatible(plugin):
+                hook_loader_load(plugin, hookCls())
```
iLLiCiTiT, line 311:
```suggestion
        for hook_cls in hooks:
            if hook_cls.is_compatible(plugin):
                hook_loader_load(plugin, hook_cls())
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +252,50 @@ class ProductLoaderPlugin(LoaderPlugin):
...
     from ayon_core.lib import Logger
     from ayon_core.pipeline import get_current_project_name
 
     log = Logger.get_logger("LoaderDiscover")
-    plugins = discover(LoaderPlugin)
     if not project_name:
         project_name = get_current_project_name()
     project_settings = get_project_settings(project_name)
+    plugins = discover(LoaderPlugin)
+    hooks = discover(PrePostLoaderHookPlugin)
```
iLLiCiTiT, line 298: We might just create objects here and not?
```suggestion
    hooks = [cls() for cls in discover(PrePostLoaderHookPlugin)]
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +252,50 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+class PrePostLoaderHookPlugin:
+    """Plugin that should be run before or post specific Loader in 'loaders'
+
+    Should be used as non-invasive method to enrich core loading process.
+    Any studio might want to modify loaded data before or after
+    they are loaded without need to override existing core plugins.
+    """
+    @classmethod
+    def is_compatible(cls, Loader: LoaderPlugin) -> bool:
+        pass
```
BigRoy, line 264: I'd argue that this should default to `True` so that by default it applies to all.

BigRoy: Or it should be an abstract method because currently it does not return bool by default.

client/ayon_core/pipeline/load/plugins.py
iLLiCiTiT: filtered_hooks = [
            hook_cls
            for hook_cls in hooks
            if hook_cls.is_compatible(plugin)
        ]
        if filtered_hooks:
            setattr(plugin, "_load_hooks", filtered_hooks)
            ...add wrapper for 'load', 'switch' and 'update'...

BigRoy (commented):
The pre and post logic seems to work. Here's the example I set up in Maya:

```python
from ayon_core.pipeline.load import register_loader_hook_plugin
from ayon_core.pipeline.load.plugins import PrePostLoaderHookPlugin
from ayon_maya.api.lib import get_highest_in_hierarchy
from maya import cmds


def _add_to_group(nodes: list[str], group_name: str) -> bool:
    """Add nodes to a group.

    Return whether the group is newly created.
    """
    exists = cmds.objExists(group_name)
    if not exists:
        cmds.createNode("transform", name=group_name)
    cmds.parent(nodes, group_name)
    return not exists


def _add_to_display_layer(nodes: list[str], layer_name: str):
    """Add nodes to a display layer."""

    # Only dag objects can be in a display layer
    nodes = cmds.ls(nodes, dag=True, long=True)
    if not nodes:
        return

    if not cmds.objExists(layer_name):
        cmds.createDisplayLayer(name=layer_name, empty=True)
        cmds.setAttr(f"{layer_name}.displayType", 2)  # Set to reference layer
    cmds.editDisplayLayerMembers(layer_name, nodes, noRecurse=True)

class CategorizeLoadedContent(PrePostLoaderHookPlugin):
    """Automatically add loaded content into a group and display layer based on source asset type"""
    def __init__(self):
        super().__init__()
        self.before_nodes: set[str] = {}

    @classmethod
    def is_compatible(cls, Loader) -> bool:
        return True

    def pre_load(
        self,
        context: dict,
        name: str | None = None,
        namespace: str | None = None,
        options: dict | None = None,
    ):
        self.before_nodes = set(cmds.ls(type="transform"))

    def post_load(
        self,
        context: dict,
        name: str | None = None,
        namespace: str | None = None,
        options: dict | None = None,
        container: dict | None = None,  # (ayon:container-3.0)
    ):
        nodes = cmds.ls(type="transform")

        new = set(nodes) - self.before_nodes
        if not new:
            return
        
        roots = get_highest_in_hierarchy(new)
        print(roots)
        
        # Some very simple logic to define what asset
        # type a folder belongs to (just as example)
        folder_name = context["folder"]["name"]
        type_name = "UTIL"
        if folder_name.startswith("char_"):
            type_name = "CHAR"
        elif folder_name.startswith("prop_"):
            type_name = "PROP"
        
        _add_to_group(roots, type_name)
        _add_to_display_layer(roots, f"{type_name}_LR")


register_loader_hook_plugin(CategorizeLoadedContent)
```

But there's no way to intercept switch or update.

client/ayon_core/tools/loader/models/actions.py
```diff
@@ -737,6 +743,7 @@ def _load_representations_by_loader(self, loader, repre_contexts, options):
             if version < 0:
                 version = "Hero"
             try:
+
```
iLLiCiTiT, line 746:
```suggestion
```

client/ayon_core/pipeline/load/plugins.py
BigRoy: For what it's worth - as soon as I started implementing as a test it would have made my life easier if this were just `return True` by default - since I can build and test the implementation with less. (I missed adding the `is_compatible` initially).

But I suppose abstract method would similarly provide me with hints in that area to implement it. 👍

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +252,50 @@ class ProductLoaderPlugin(LoaderPlugin):
...
 
 
+class PrePostLoaderHookPlugin:
+    """Plugin that should be run before or post specific Loader in 'loaders'
+
+    Should be used as non-invasive method to enrich core loading process.
+    Any studio might want to modify loaded data before or after
+    they are loaded without need to override existing core plugins.
+    """
+    @classmethod
```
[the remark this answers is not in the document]
antirotor: I agree,  order would be great here

client/ayon_core/pipeline/load/plugins.py
antirotor: question is then if it should chain these in a case where `switch()` calls directly `update()` (which is the case probably most of the time). It probably should.

antirotor (commented):
I am not sure about the `switch()`, `update()` and possibly even `remove()` parts. Maybe they can be added later when demanded?

kalisp:
Implemented `update`, `remove` methods.

Solved all other comments.

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +253,82 @@ class ProductLoaderPlugin(LoaderPlugin):
     """
 
 
+class PrePostLoaderHookPlugin:
```
iLLiCiTiT, line 256:
```suggestion
class LoaderHookPlugin:
```

BigRoy (changes requested):
> Implemented `update`, `remove` methods.

These didn't work for me because you were passing a `container` kwarg argument even though the `post_update` and `post_remove` methods already received a `container` argument - resulting in an error being raised. I've refactored it to `result` and updated the examples in the PR description. Which now with the logs work for me @kalisp - please do another test run yourself and double check my commits whether they are to your liking.

### more complicated stack trace if `LoaderPlugin` method raises an error and hooks are registered

I did want to stress a side effect of this monkeypatching - and that is that IF an error occurs in a Loader Plug-in the stack trace becomes more complicated due to the `wrapped_method` becoming a part of it. 

### `self` on the `LoaderHookPlugin` refers to the `LoaderPlugin` when called from the wrapper 

~~There's also the caveat that when referring to `self` on the Loader Hook Plug-in, it will actually be referring to the `LoaderPlugin`. As such, you can't make e.g. a private method on the `LoaderHookPlugin` which you call from `pre_load`~~

I changed this with https://github.com/ynput/ayon-core/pull/1222/commits/a593516f29394b28f03cb14e8d2d1818ebbc9ca0 because it just felt so extremely odd that it wouldn't otherwise - feel free to revert if needed. Before that commit, belwo example failed - now it works if you add the `plugin=None` argument to `pre_load`.

```python
class Test(LoaderHookPlugin):
    """Automatically add loaded content into a group and display layer based on source asset type"""

    def __init__(self):
        super().__init__()
        self.before_nodes: set[str] = {}

    @classmethod
    def is_compatible(cls, Loader) -> bool:
        return True

    def pre_load(
            self,
            context: dict,
            name: str | None = None,
            namespace: str | None = None,
            options: dict | None = None,
    ):
        print(self._my_hello_world())  # this fails
		print(self.before_nodes)       # and this too
        
    def _my_hello_world(self):
        print("HELLO WORLD")
```

### `register_loader_hook_plugin` does not re-register the class

Also, when using `register_loader_hook_plugin` to explicitly register a class... it seems that after registering once and calling it again, it never updates the class - so if you're developing and changing the code and re-registering then it's not actually going to run the changed code. This makes the development cycle quite a burden.

**Note:** This issue does not seem to occur with `register_loader_hook_plugin_path`. Those discovered plug-ins appear to reload just fine.

### abstract methods, really abstract?

The fact that each of the methods is abstract means you MUST (from a typing perspective?) implement them - _even_ if for some pre/post hooks you may not need to do anything on e.g. `update` (see e.g. example code below). Do we care?

---

All in all, it now 'works' if you adhere to very strict and unexpected rules on how you implement things.

Here's an example Maya Pre Post Hook that now loads content into groups and display layers based on some arbitrary "type" and also uses a post remove hook to delete those groups if they turned empty on deleting loaded content.

```python
    get_representation_context,
    get_current_project_name
)
from ayon_core.pipeline.load.plugins import LoaderHookPlugin
from ayon_maya.api.lib import get_highest_in_hierarchy
from maya import cmds


def _add_to_group(nodes: list[str], group_name: str) -> bool:
    """Add nodes to a group.

    Return whether the group is newly created.
    """
    exists = cmds.objExists(group_name)
    if not exists:
        cmds.createNode("transform", name=group_name)
    cmds.parent(nodes, group_name)
    return not exists


def _add_to_display_layer(nodes: list[str], layer_name: str):
    """Add nodes to a display layer."""

    # Only dag objects can be in a display layer
    nodes = cmds.ls(nodes, dag=True, long=True)
    if not nodes:
        return

    if not cmds.objExists(layer_name):
        cmds.createDisplayLayer(name=layer_name, empty=True)
        cmds.setAttr(f"{layer_name}.displayType", 2)  # Set to reference layer
    cmds.editDisplayLayerMembers(layer_name, nodes, noRecurse=True)


class CategorizeLoadedContent(LoaderHookPlugin):
    """Automatically add loaded content into a group and display layer based on source asset type"""
    @classmethod
    def is_compatible(cls, Loader) -> bool:
        return True

    def pre_load(
            self,
            context: dict,
            name: str | None = None,
            namespace: str | None = None,
            options: dict | None = None,
    ):
        self.before_nodes = set(cmds.ls(type="transform"))

    def post_load(
            self,
            context: dict,
            name: str | None = None,
            namespace: str | None = None,
            options: dict | None = None,
            container: dict | None = None,  # (ayon:container-3.0)
            result = None
    ):
        nodes = cmds.ls(type="transform")

        new = set(nodes) - self.before_nodes
        if not new:
            return

        roots = get_highest_in_hierarchy(new)

        type_name = get_asset_type_name(context)

        _add_to_group(roots, type_name)
        _add_to_display_layer(roots, f"{type_name}_LR")

    def pre_update(self, container, context):
        pass

    def post_update(self, container, context, result=None):
        pass

    def pre_remove(self, container):
        pass

    def post_remove(self, container, result=None):
        project_name: str = container.get(
            "project_name",
            get_current_project_name()
        )
        context = get_representation_context(
            project_name=project_name,
            representation=container["representation"]
        )
        type_name = get_asset_type_name(context)

        # If the container is removed, we can remove the group and layer
        # if they are now empty.
        if cmds.objExists(type_name) and not cmds.listRelatives(type_name,
                                                                children=True):
            cmds.delete(type_name)

        # If the display layer is empty, remove it
        layer_name: str = f"{type_name}_LR"
        if cmds.objExists(layer_name) and not cmds.editDisplayLayerMembers(
                layer_name, query=True, noRecurse=True):
            cmds.delete(layer_name)


def get_asset_type_name(context: dict) -> str:

    # Some very simple logic to define what asset
    # type a folder belongs to (just as example)
    folder_name = context["folder"]["name"]
    type_name = "UTIL"
    if folder_name.startswith("char_"):
        type_name = "CHAR"
    elif folder_name.startswith("prop_"):
        type_name = "PROP"
    return type_name
```

BigRoy (approved):
This worked for me with the recent changes I made - but would need testing and confirming of code changes by someone else please. :)

kalisp:
> > Implemented `update`, `remove` methods.
> 
Yeah, you are right, it was broken, I misread log and got confused its working on my dummy test. I re tested (and this time double checked myself).

iLLiCiTiT (approved):

antirotor (changes requested):
Ok, I might be just down beause I had to recently remove all uses of `TypedDict` from my code because we still support Python 3.7. I believe `str | None` is working only for python 3.10+ so I guess this will kill any older Maya than 2024 (Maya 2023 has 3.9)? Sorry. I would like to use newer syntax so much...

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+
+    @abstractmethod
+    def post_load(
+        self,
+        context: dict,
+        name: str | None = None,
+        namespace: str | None = None,
+        options: dict | None = None,
+        plugin: LoaderPlugin | None = None,
+        result: Any = None,
```
iLLiCiTiT, line 293:
```suggestion
        name: Optional[str],
        namespace: Optional[str],
        options: Optional[dict],
        plugin: Optional[LoaderPlugin],
        result: Any,
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+    ):
+        pass
+
+    @abstractmethod
+    def pre_update(
+        self,
+        container: dict,  # (ayon:container-3.0)
+        context: dict,
+        plugin: LoaderPlugin | None = None,
```
iLLiCiTiT, line 302:
```suggestion
        container: dict,  # (ayon:container-3.0)
        context: dict,
        plugin: Optional[LoaderPlugin],
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+    ):
+        pass
+
+    @abstractmethod
+    def post_update(
+        self,
+        container: dict,  # (ayon:container-3.0)
+        context: dict,
+        plugin: LoaderPlugin | None = None,
```
iLLiCiTiT, line 311:
```suggestion
        plugin: Optional[LoaderPlugin],
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+    ):
+        pass
+
+    @abstractmethod
+    def post_remove(
+        self,
+        container: dict,  # (ayon:container-3.0)
+        plugin: LoaderPlugin | None = None,
+        result: Any = None,
```
iLLiCiTiT, line 329:
```suggestion
        plugin: Optional[LoaderPlugin],
        result: Any,
```

client/ayon_core/pipeline/load/plugins.py
iLLiCiTiT, line 311:
```suggestion
        plugin: LoaderPlugin,
        container: dict,  # (ayon:container-3.0)
        context: dict,
```

client/ayon_core/pipeline/load/plugins.py
iLLiCiTiT, line 329:
```suggestion
        plugin: LoaderPlugin,
        result: Any,
        container: dict,  # (ayon:container-3.0)
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        plugin: Optional[LoaderPlugin] = None,
+    ):
+        pass
+
+    @abstractmethod
+    def post_remove(
+        self,
+        container: dict,  # (ayon:container-3.0)
+        plugin: Optional[LoaderPlugin] = None,
+        result: Any = None,
```
iLLiCiTiT, line 329:
```suggestion
        plugin: LoaderPlugin,
        result: Any,
        container: dict,  # (ayon:container-3.0)
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+    ):
+        pass
+
+    @abstractmethod
+    def post_update(
+        self,
+        container: dict,  # (ayon:container-3.0)
+        context: dict,
+        plugin: Optional[LoaderPlugin] = None,
+        result: Any = None,
```
iLLiCiTiT, line 312:
```suggestion
        plugin: LoaderPlugin,
        result: Any,
        container: dict,  # (ayon:container-3.0)
        context: dict,
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+    ):
+        pass
+
+    @abstractmethod
+    def pre_update(
+        self,
+        container: dict,  # (ayon:container-3.0)
+        context: dict,
+        plugin: Optional[LoaderPlugin] = None,
```
iLLiCiTiT, line 302:
```suggestion
        plugin: LoaderPlugin,
        container: dict,  # (ayon:container-3.0)
        context: dict,
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+
+    @abstractmethod
+    def post_load(
+        self,
+        context: dict,
+        name: Optional[str] = None,
+        namespace: Optional[str] = None,
+        options: Optional[dict] = None,
+        plugin: Optional[LoaderPlugin] = None,
+        result: Any = None,
```
iLLiCiTiT, line 293:
```suggestion
        plugin: LoaderPlugin,
        result: Any,
        context: dict,
        name: Optional[str],
        namespace: Optional[str],
        options: Optional[dict],
```

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        pass
+
+    @abstractmethod
+    def pre_load(
+        self,
+        context: dict,
+        name: Optional[str] = None,
+        namespace: Optional[str] = None,
+        options: Optional[dict] = None,
+        plugin: Optional[LoaderPlugin] = None,
```
iLLiCiTiT, line 281:
```suggestion
        plugin: LoaderPlugin,
        context: dict,
        name: Optional[str],
        namespace: Optional[str],
        options: Optional[dict],
```

kalisp:
Reordered arguments in `pre_`, `post_` methods.

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        pass
+
+    @abstractmethod
+    def pre_load(
+        self,
+        plugin: LoaderPlugin,
+        context: dict,
+        name: Optional[str] = None,
+        namespace: Optional[str] = None,
+        options: Optional[dict] = None,
```
iLLiCiTiT, line 281:
```suggestion
        name: Optional[str],
        namespace: Optional[str],
        options: Optional[dict],
```

iLLiCiTiT (approved):

client/ayon_core/pipeline/load/plugins.py
```diff
@@ -251,15 +254,94 @@ class ProductLoaderPlugin(LoaderPlugin):
...
+        self,
+        plugin: LoaderPlugin,
+        context: dict,
+        name: Optional[str],
+        namespace: Optional[str],
+        options: Optional[dict],
+    ):
+        pass
+
+    @abstractmethod
```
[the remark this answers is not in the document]
iLLiCiTiT: But if I want to implement only `pre` methods then I also have to implement all `post` methods, but keep them empty.

iLLiCiTiT:
It is mergable, from my point of view the abstract methods are unnecessary, but I won't be the one who will use it.

PR #1224, opened 2025-04-04
Enhancing publisher card widget with folder and task name

jm22dogs:
## Changelog Description
This PR adds a sub-label to the publisher card widget, containing the folder and the task name.
This is a very minimal implementation.

The image below shows the end result. The first card has the folder and the task name and the second one is an example for when a task is missing:

image

## Additional info
Having many cards with similar names can be very confusing, this PR facilitates more information in the widget to minimize this problem.

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -211,7 +211,7 @@ class ContextCardWidget(CardWidget):
         icon_widget = PublishPixmapLabel(None, self)
         icon_widget.setObjectName("ProductTypeIconLabel")
 
-        label_widget = QtWidgets.QLabel(CONTEXT_LABEL, self)
+        label_widget = QtWidgets.QLabel(f"<span>{CONTEXT_LABEL}</span>", self)
 
         icon_layout = QtWidgets.QHBoxLayout()
         icon_layout.setContentsMargins(5, 5, 5, 5)


client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -492,10 +492,22 @@ def _validate_context(self, context_info):
...
         self._context_warning.setVisible(not valid)
 
+    @staticmethod
+    def get_card_widget_sub_label(folder_name, task_name=None):
+        sublabel = "<br><span style=\"font-size: 8pt;\">"
+        sublabel += "<b>{}</b>".format(folder_name)
+        if task_name:
+            sublabel +=  " - <i>{}</i>".format(task_name)
+        sublabel += "</span>"
+        return sublabel
```
iLLiCiTiT, line 502:
```suggestion
    def get_card_widget_sub_label(folder_path, task_name):
        
        parts = []
        if folder_path:
            folder_name = folder_path.split("/")[-1]
            parts.append(f"<b>{folder_name}</b>")
            if task_name:
                parts.append(folder_name)
        if not parts:
            return None
        sublabel = " - ".join(parts)
        return f"<span style=\"font-size: 8pt;\">{sublabel}</span>"
```

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -492,10 +492,22 @@ def _validate_context(self, context_info):
...
+
     def _update_product_name(self):
         variant = self.instance.variant
         product_name = self.instance.product_name
         label = self.instance.label
+        folder_name = self.instance.get_folder_path().split("/")[-1]
```
iLLiCiTiT, line 508:
```suggestion
        folder_path = self.instance.get_folder_path()
```

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -513,6 +529,9 @@ def _update_product_name(self):
             for part in found_parts:
                 replacement = "<b>{}</b>".format(part)
                 label = label.replace(part, replacement)
+                sublabel = self._get_card_widget_sub_label(folder_path, task_name)
+                if sublabel:
+                    label += f"<br/>{sublabel}"
```
iLLiCiTiT, line 534:
```suggestion
        sublabel = self._get_card_widget_sub_label(folder_path, task_name)
        if sublabel:
            label += f"<br/>{sublabel}"
```

iLLiCiTiT:
>  is that optional or not atm?

It is not, and I'm not sure how would that be. Where you would turn it on/off?

LiborBatek:
> > is that optional or not atm?
> 
> It is not, and I'm not sure how would that be. Where you would turn it on/off?

I mean in `ayon-core` as settings...not in the publisher UI...as I can imagine when having a lot of publish instances it makes the widgets quite dense/ slightly cluttered despite being it useful info... 

would be good to have it as optional thingy or personal preference of the studio

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -383,28 +385,53 @@ def _validate_context(self, context_info):
...
 
+    @staticmethod
+    def _get_card_widget_sub_label(
+        folder_path: Optional[str],
+        task_name: Optional[str],
+    ) -> str:
+        sublabel = ""
+        if folder_path:
+            folder_name = folder_path.rsplit("/", 1)[-1]
+            sublabel = f"- <b>{folder_name}</b>"
```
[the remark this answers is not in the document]
iLLiCiTiT: I added it to "indent" it, so it does not start at same position as product name.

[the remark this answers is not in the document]
BigRoy: I see, for me it's not needed but if you want forced indentation a tab is better than the `- ` prefix I'd say.

jm22dogs:
thank you for taking the time to put some work into this @iLLiCiTiT , I think that works nicely. I do agree with @BigRoy that may be confusing to have some widgets with the extra info and some without. My suggestion would be to do something like this:

image

but I am not sure how much time consuming adding a separate group for different context's products would be.
As for the indentation and the lack of a dash that @BigRoy mentions, I don't mind really, but if I had to choose I would go with the original  presented look as well.

MustafaJafar (approved):
This is so cool!
image

MustafaJafar:
> I wouldn't take it that far - there are certain contexts, like Tray Publisher where you're essentially always publishing to a different context. Also, it only makes it even more cluttered.

yeah, I agree having all of the instances as done currently feels more natural.

iLLiCiTiT:
> I guess the PR in the current state (with the revert) is good to go?

Yes for me.

PR #1231, opened 2025-04-09
Delete poetry.lock

antirotor:
## Changelog Description
Poetry lock shouldn't be tracked file to avoid conflics, it should be always recreated from dependencies defined in `pyproject.toml`. Note that this is affecting only developers, not the dependency packages, etc.


iLLiCiTiT (approved):

iLLiCiTiT:
We should add it to gitignore too.

antirotor:
> We should add it to gitignore too.

It is being added here - https://github.com/ynput/ayon-core/pull/1147 - but I suppose no harm done when adding it in separate PR

iLLiCiTiT:
Should be added to this PR (which is removing the file).

antirotor:
> Should be added to this PR (which is removing the file).

done, I thought it was already merged

PR #1233, opened 2025-04-09
Editorial: Updates OTIO review extract to improve quality

jakubjezek001:
## Changelog Description

Improves review extraction by:
- Switches output extension to '.png' for better image quality.
- Adds compression level to ffmpeg command.
- Adds scaling to the video filter.
- Forces re-encoding for lossy formats.

## Testing notes:
1. open any of your editorial dcc testing timeline with reference layer flagged for review target
2. publish clip and see all function as before


client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -520,9 +526,12 @@ def _render_segment(self, sequence=None,
             input_extension
             and self.output_ext == input_extension
         ):
-            command.extend([
-                "-c", "copy"
-            ])
+            if input_extension.lower() in [
```
iLLiCiTiT, line 529: Does this condition actually make sense? This happens only if input extension is same as `output_ext`. And `output_ext` is always `".png"`, so why do we need this condition?

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -520,9 +526,12 @@ def _render_segment(self, sequence=None,
...
+                    '.png', '.tif', '.tiff', '.dpx', '.exr']:
+                command.extend(["-c", "copy"])
+            else:
+                # For lossy formats, force re-encode
+                command.extend(["-pix_fmt", "rgba"])
```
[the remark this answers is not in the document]
jakubjezek001: just to make sure this re-encodes in case of loosy formats

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -474,6 +474,7 @@ def _render_segment(self, sequence=None,
 
             command.extend([
                 "-start_number", str(in_frame_start),
+                "-compression_level", "5",
```
rdelillo, line 477:
```suggestion
```

Put this way (before `-i`) `-compression_level 5` applies to the input which might be unsupported for some formats.
E.g. when encoding EXR files:
```
Codec AVOption compression_level () specified for input file #0 (C:/projects/AY01_VFX_demo/resources/plates/MER_sq001_sh020_P02\\MER_sq001_sh020_P02.%04d.exr) is not a decoding option.
```

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -510,6 +511,11 @@ def _render_segment(self, sequence=None,
                 "-tune", "stillimage"
             ])
 
+        if video or sequence:
+            command.extend([
+                "-vf", f"scale={self.to_width}:{self.to_height}:flags=lanczos"
```
rdelillo, line 516:
```suggestion
                "-vf", f"scale={self.to_width}:{self.to_height}:flags=lanczos",
                "-compression_level", "5",
```

rdelillo (approved):
LGTM

I did republish with review from Hiero and the new result looks way better.
image

I think a lot of people will be happy with this one 🚀

iLLiCiTiT:
@jakubjezek001 suggestions from @rdelillo look like they make sense. We want to define compression of output instead of input.

PR #1236, opened 2025-04-10
Chore: Update ffmpeg commands

iLLiCiTiT:
## Changelog Description
Update ffmpeg commands to work with newer ffmpeg releases.

## Additional info
Stop using deprecated commands (that were deprecated many years ago).

## Testing notes:
1. It should be possible to create thumbnails from video files.


64qam (approved):
It could work

kalisp:
Is there some dependency that should be added `package.py`? (When new `ffmpeg` started to show up?)

m-u-r-p-h-y:
I know it is out of the scope of this PR, just a question if those switches should not be exposed and editable the same way we have them in extract review for h264 generation for example.

iLLiCiTiT:
> I know it is out of the scope of this PR, just a question if those switches should not be exposed and editable the same way we have them in extract review for h264 generation for example.

Doesn't make sense, there is not much you could modify.

PR #1239, opened 2025-04-11
Tools: Fix thumbnails output value type

iLLiCiTiT:
## Changelog Description
Always return the same output when receiving thumbnails.

## Testing notes:
1. Loader does not crash if nothing is selected, or if all selected entities don't have thumbnails.


MustafaJafar (approved):
This works! 
I'm no longer seeing the error.

PR #1251, opened 2025-04-17
Try extract thumbnail from available representations on instance

moonyuet:
## Changelog Description
This PR is to make sure thumbnail would be exported when there is valid representation with image content files.
Resolve https://github.com/ynput/ayon-fusion/issues/16

## Additional info
Representations with `"review"` tag has privilege to be used before any other representation.

## Testing notes:
Test along with https://github.com/ynput/ayon-fusion/pull/43
1. Build this addon with the relevant fusion addon
2. Launch Fusion
3. Publish

### DIFF
--- a/client/ayon_core/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_core/plugins/publish/extract_thumbnail.py
@@ -372,15 +372,10 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
         Returns:
             bool: whether the representation has the valid image content
         """
-        files = repre.get("files")
-        if not files:
-            return False
-
         # Get first file's extension
-        if isinstance(files, (list, tuple)):
-            first_file = files[0]
-        else:
-            first_file = files
+        first_file = repre["files"]
+        if isinstance(first_file, (list, tuple)):
+            first_file = first_file[0]
 
         ext = os.path.splitext(first_file)[1].lower()
 
--- a/client/ayon_core/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_core/plugins/publish/extract_thumbnail.py
@@ -17,7 +17,7 @@ from ayon_core.lib import (
 )
 from ayon_core.lib.transcoding import convert_colorspace
 
-from ayon_core.lib.transcoding import VIDEO_EXTENSIONS
+from ayon_core.lib.transcoding import VIDEO_EXTENSIONS, IMAGE_EXTENSIONS
 
 
 class ExtractThumbnail(pyblish.api.InstancePlugin):
@@ -349,7 +349,8 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
                 continue
 
             if "review" not in tags:
-                continue
+                if not self._is_valid_media_repre(repre):
+                    continue
 
             if not repre.get("files"):
                 self.log.debug((
@@ -359,6 +360,21 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
 
             filtered_repres.append(repre)
         return filtered_repres
+
+    def _is_valid_media_repre(self, repre):
+        """Check if representation contains valid media files"""
+        files = repre.get("files")
+        if not files:
+            return False
+
+        # Get first file's extension
+        if isinstance(files, (list, tuple)):
+            first_file = files[0]
+        else:
+            first_file = files
+
+        ext = os.path.splitext(first_file)[1].lower()
+        return ext in IMAGE_EXTENSIONS
 
     def _create_thumbnail_oiio(
         self,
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -303,6 +303,8 @@ def _is_review_instance(self, instance):
         #   other way then checking for "review" key on instance data?
         if instance.data.get("review", True):
             return True
+        if instance.data.get("need_thumbnail", False):
+            return True
```
BigRoy, line 307: I wonder if we now should refactor the method name to `_is_thumbnail_instance` too to avoid confusion?

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -348,7 +350,7 @@ def _get_filtered_repres(self, instance):
                 # to be published locally
                 continue
 
-            if "review" not in tags:
+            if "review" not in tags or "need_thumbnail" not in tags:
                 continue
```
BigRoy, line 354: This would mean the representation **must** now have both otherwise it'd get skipped. is that correct?

moonyuet:
I have updated the PR regarding to the comments in https://github.com/ynput/ayon-fusion/pull/43

iLLiCiTiT:
The changes made broke the logic. Variable `instance_repres` is not used only for finding representations to use but also to find existing thumbnail representations. That's why `_get_explicit_repres_for_thumbnail` and `_get_filtered_repres` do happen after those validations, which is the place where the representation filtering should happen. I don't think it should be related to product type, it should just find representation with image/video files that can be used instead of using first one, which is case of model can be `.fbx` which is invalid representation for conversion.

Also if I remember correctly the idea was that it behaves the same way as before. First use review representation, and if there is no valid review representation to create thumbnail from, then look for other representations with valid image content.

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -349,7 +349,8 @@ def _get_filtered_repres(self, instance):
...
 
             if "review" not in tags:
```
iLLiCiTiT, line 351: I would add one modification. Collect representations with `"review"` tag into different list and withou the tag into other list and concatenate them at the end -> so they are in "prefered" order.

```
review_repres = []
other_repres = []
for repre in src_repres:
    ...

    has_review_tag = "review" in tags
    if not has_review_tag and not self._is_valid_images_repre(repre):
        continue
    ...

    if has_review_tag:
        review_repres.append(repre)
    else:
        other_repres.append(repre)

return review_repres + other_repres
```

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -360,6 +361,31 @@ def _get_filtered_repres(self, instance):
...
+            return False
+
+        # Get first file's extension
+        if isinstance(files, (list, tuple)):
+            first_file = files[0]
+        else:
+            first_file = files
+
+        ext = os.path.splitext(first_file)[1].lower()
+        return ext in IMAGE_EXTENSIONS
```
iLLiCiTiT, line 387: Requires import of `VIDEO_EXTENSIONS`
```suggestion
        return ext in IMAGE_EXTENSIONS or ext in VIDEO_EXTENSIONS
```

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -348,17 +349,45 @@ def _get_filtered_repres(self, instance):
...
-            filtered_repres.append(repre)
-        return filtered_repres
+            has_review_tag = "review" in tags
+            if not has_review_tag and not self._is_valid_images_repre(repre):
+                continue
+
+            if has_review_tag:
+                review_repres.append(repre)
+            else:
+                other_repres.append(repre)
```
BigRoy, line 365: Why not just do this?
```suggestion
            if has_review_tag:
                review_repres.append(repre)
            elif self._is_valid_images_repre(repre):
                other_repres.append(repre)
```

BigRoy:
We may want to update the PR title and description, since it does not require any `need_thumbnail` anymore.

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -348,17 +349,42 @@ def _get_filtered_repres(self, instance):
...
+        Args:
+            repre (dict): representation
+
+        Returns:
+            bool: whether the representation has the valid image content
+        """
+        files = repre.get("files")
+        if not files:
+            return False
+
```
iLLiCiTiT, line 378:
```suggestion
        files = repre["files"]

```

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -348,17 +349,42 @@ def _get_filtered_repres(self, instance):
...
+        # Get first file's extension
+        if isinstance(files, (list, tuple)):
+            first_file = files[0]
+        else:
+            first_file = files
```
iLLiCiTiT, line 383:
```suggestion
        # Get first file's extension
        first_file = repre.get("files")
        if isinstance(first_file, (list, tuple)):
            first_file = first_file[0]
```

PR #1255, opened 2025-04-23
Adds explicit resolution override to publisher

### LINKED ISSUE #1243: AY-7197_Override resolution on editorial publish

When doing editorial publish, there is no way to override resolution of the output and set it to custom one. The one taken from the project is used. There should be an option to set:

use resolution from source
use resolution set on project
use custom resolution specified

This can be done currently only in Flame. But it should be supported ideally in all editorial publishing, but at least in Hiero/Nuke Studio

_No response_

_No response_

### Additional context:
link to discussion on [Discord](None)

_This issue was automatically created from Clickup ticket [AY-7197](https://app.clickup.com/t/86c14eq8q)_

jakubjezek001:
## Changelog Description
Adds a plugin that allows users to explicitly override the resolution settings (width, height, pixel aspect) of instances during the publishing process.
This provides a way to ensure consistency and accuracy in resolution values across different tasks and product types.
The plugin is configurable through the AYON settings, allowing administrators to define the available resolution options and the product types for which the override is enabled.

## Testing notes:
1. Generate addon package and install it to server (change package.py version if necessary)
2. Add it to your bundle or dev bundle
3. New publishing addon was added with default settings `ayon+settings://core/publish/CollectExplicitResolution`
image
4. add multiple the same presets and try to save - this should result in none unique error blocking you to save
5. fix it and then save so you are having multiple presets
6. Open Traypublisher at some project and apply Simple Editorial workflow from following data https://drive.google.com/open?id=1Jn4JFfWSxedrrnQO4HHdrHM2c9Yo9xs9&usp=drive_fs
7. drop EDL and mp4 to required sections and hit create (project should be 25fps )
8. double click at any of the shot product instances and see following attribute
image
9. notice that all presets are offered in options of the enumerator
10. Go back to settings an either remove the `shot` product or disable the addon category completely, save and refresh Traypublisher. Notice that the attribute is not offered. 

resolves [flame#46](https://github.com/ynput/ayon-flame/issues/46)
resolves #1243


server/settings/publish_plugins.py
```diff
@@ -157,6 +159,77 @@ def validate_unique_outputs(cls, value):
...
+    'height' and 'pixel_aspect' properties.
+    """
+    options = []
+    for obj in objects:
+        item_test_text = f"{obj.width}x{obj.height}x{obj.pixel_aspect}"
+        if item_test_text not in options:
+            options.append(item_test_text)
+        else:
+            raise BadRequestException(
+                f"Duplicate option '{item_test_text}'")
```
iLLiCiTiT, line 207:
```suggestion
    options = set()
    for obj in objects:
        item_test_text = f"{obj.width}x{obj.height}x{obj.pixel_aspect}"
        if item_test_text in options:
            raise BadRequestException(
                f"Duplicate option '{item_test_text}'")

        options.add(item_test_text)
```

server/settings/publish_plugins.py
```diff
@@ -157,6 +159,77 @@ def validate_unique_outputs(cls, value):
...
+        ge=0.0,
+        le=100000.0,
+        description=(
+            "Pixel Aspect resolution decimal number value"),
+        placeholder="Pixel aspect"
+    )
+
+
+def ensure_unique_resolution_option(
+        objects: Iterable[Any], field_name: str | None = None) -> None:  # noqa: C901
```
iLLiCiTiT, line 194: I think it will be always `list`.
```suggestion
        objects: list[Any], field_name: str | None = None) -> None:  # noqa: C901
```

iLLiCiTiT: Why is expecting `field_name` if is not used?

client/ayon_core/plugins/publish/collect_explicit_resolution.py
```diff
@@ -0,0 +1,104 @@
...
+        if item_values:
+            # if the item is in the cache, get the values from it
+            return {
+                "resolutionWidth": item_values["width"],
+                "resolutionHeight": item_values["height"],
+                "pixelAspect": item_values["pixel_aspect"],
+            }
+        else:
+            raise PublishError(
+                f"Invalid resolution value: {resolution_value}")
```
iLLiCiTiT, line 67:
```suggestion

        raise PublishError(
            f"Invalid resolution value: {resolution_value}")
```

server/settings/publish_plugins.py
```diff
@@ -1162,6 +1239,19 @@ class PublishPuginsModel(BaseSettingsModel):
...
         ]
     },
+    "CollectExplicitResolution": {
+        "enabled": True,
+        "product_types": [
+            "shot"
+        ],
+        "options": [
+            {
+                "width": 2048,
```
iLLiCiTiT, line 1249: Is this a valid default value?

client/ayon_core/plugins/publish/collect_explicit_resolution.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class CollectExplicitResolution(
+    pyblish.api.InstancePlugin,
+    publish.AYONPyblishPluginMixin,
+):
+    """Collect explicit user defined resolution attributes for instances"""
+
+    label = "Choose Explicit Resolution"
+    order = pyblish.api.CollectorOrder + 0.49
```
rdelillo, line 14: I could not make this work for me for Hiero/Nuke Studio.
Even if I select the proper option, the resulting shot was not created with the "forced" resolution.

From my understanding:
* shot folders are created by `extract_hierarchy_to_ayon` based on the `hierarchyContext` entry
* unless overwritten by a plugin `hierarchyContext` is created by `collect_hierarchy` plugin
* `collect_explicit_resolution` currently runs AFTER `collect_hierarchy`

For this stuff to be picked up, I had to change the order to be
```python
    order = pyblish.api.CollectorOrder - 0.091
```

rdelillo (changes requested):
Unfortunately, the "Editorial Simple" example did not work for me, neither publishing through Hiero.

No resolution attributes was propagated through.
I *believe* this is because of the `order` issue I commented in `CollectExplicitResolution`.

Also believe some changes are missing in `ayon-traypublisher` for `Editorial Simple`. 
See my fixes here: https://github.com/ynput/ayon-traypublisher/pull/86

client/ayon_core/plugins/publish/collect_explicit_resolution.py
```diff
@@ -0,0 +1,104 @@
...
+        # Add all cached resolution items to the dropdown options
+        for item_text in resolution_items:
+            items.append((item_text, item_text))
+
+        return [
+            EnumDef(
+                "explicit_resolution",
+                items,
+                default="Don't override",
+                label="Override Resolution",
```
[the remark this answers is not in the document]
jakubjezek001: yes you are right, but rather `Force Short resolution`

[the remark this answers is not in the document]
jakubjezek001: Actually you are correct, it should be product

client/ayon_core/plugins/publish/collect_explicit_resolution.py
jakubjezek001: implemented

rdelillo (approved):

server/settings/publish_plugins.py
```diff
@@ -1171,6 +1248,13 @@ class PublishPuginsModel(BaseSettingsModel):
             },
         ]
     },
+    "CollectExplicitResolution": {
+        "enabled": True,
+        "product_types": [
+            "shot"
+        ],
```
BigRoy, line 1255: Note that having this empty by default but enabled... just means that we're running more logic (`get_attr_defs_for_instance`) by default, for ALL instances... regardless of whether `product_types` is defined.

Should this maybe have been disabled by default?

PR #1258, opened 2025-04-25
Set running from console even for triggering from code

kalisp:
## Changelog Description
`is_using_ayon_console` is used to denote if AYON is running with support of console. Adobe products are using this check to produce console. This update resolves it to true even if AYON is started directly from code, eg. from IDE (via `python.exe`).

## Additional info
I personally run AYON from IDE the most and I always need to tweak this to show console in Adobe products. Its nothing major, just PITA. I always felt that it would make a sense that running from code would mean "is_using_ayon_console".
If this update would make sense, it would be much easier for me.

### DIFF
--- a/client/ayon_core/lib/ayon_info.py
+++ b/client/ayon_core/lib/ayon_info.py
@@ -77,7 +77,7 @@ def is_using_ayon_console():
     executable_filename = os.path.basename(executable_path)
     return (
         "ayon_console" in executable_filename
-        or "python.exe" in executable_filename
+        or "python" in executable_filename
     )
 
 
--- a/client/ayon_core/lib/ayon_info.py
+++ b/client/ayon_core/lib/ayon_info.py
@@ -75,7 +75,10 @@ def is_using_ayon_console():
         return True
     executable_path = os.environ["AYON_EXECUTABLE"]
     executable_filename = os.path.basename(executable_path)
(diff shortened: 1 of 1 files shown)


client/ayon_core/lib/ayon_info.py
```diff
@@ -75,7 +75,10 @@ def is_using_ayon_console():
         return True
     executable_path = os.environ["AYON_EXECUTABLE"]
     executable_filename = os.path.basename(executable_path)
-    return "ayon_console" in executable_filename
+    return (
+        "ayon_console" in executable_filename
+        or "python.exe" in executable_filename
```
iLLiCiTiT, line 80:
```suggestion
        or "python" in executable_filename
```

iLLiCiTiT:
But maybe if that's the case then maybe the adobe hosts should use `is_running_from_build` to find out if console should be opened to be honest.

kalisp:
Maybe I misunderstood this function, I could definitely use in Adobe:
```
    if is_using_ayon_console() or is_running_from_build():
        kwargs.update({
            "creationflags": subprocess.CREATE_NEW_CONSOLE
        })
    else:
        kwargs.update({
            "creationflags": subprocess.CREATE_NO_WINDOW,
            "stdout": subprocess.DEVNULL,
            "stderr": subprocess.DEVNULL
        })
```

PR #1261, opened 2025-04-30
Filter launcher paths from `PATH` passed to host

antirotor:
## Changelog Description
Filter `PATH` environment variable so paths pointing to launcher are not passed to the launched hosts by default. Fixing mov writer issue in Nuke 16.

## Additional info
Passing those paths with PySide6 libraries (and others) can cause serious issues especially on Windows where DLL resolution is using `PATH` so wrong libraries can be pulled into runtime.

## Testing notes:
Run any DCC with suitable Python console and run:

```python
import os

for p in os.getenv("PATH").split(os.pathsep):
    print(p)
```
should print paths and there shouldn't be any pointing to location of your launcher. Your launcher path can be determined by `AYON_ROOT` environment variable so technically:

```python
import os

for p in os.getenv("PATH").split(os.pathsep):
    if p.startswith(os.getenv("AYON_ROOT")):
		print("All work and no play makes Jack a dull boy")
```

shouldn't print anything.

**OR**

run Nuke 16 via AYON, create Write node and set format to `mov` - it shouldn't crash

### DIFF
--- a/client/ayon_core/hooks/pre_remove_launcher_paths.py
+++ b/client/ayon_core/hooks/pre_remove_launcher_paths.py
@@ -1,0 +1,36 @@
+""""Pre launch hook to remove launcher paths from the system."""
+import os
+from ayon_applications import PreLaunchHook, LaunchTypes
+
+
+class PreRemoveLauncherPaths(PreLaunchHook):
+    """Remove launcher paths from the system.
+
+    This hook is used to remove launcher paths from the system before launching
+    an application. It is used to ensure that the application is launched with
+    the correct environment variables. Especially for Windows, where
+    paths in `PATH` are used to load DLLs. This is important to avoid
+    conflicts with other applications that may have the same DLLs in their
+    paths.
+    """
+
+    order = 1
+
+    platforms = {"linux", "windows", "darwin"}
+    launch_types = {LaunchTypes.local}
+
+    def execute(self):
+        # Remove launcher paths from the system
+        paths = []
+        try:
+            ayon_root = self.launch_context.env["AYON_ROOT"]
+        except KeyError:
+            self.log.warning("AYON_ROOT not found in environment variables.")
+            return
+
+        paths.extend(
+            path
+            for path in self.launch_context.env.get("PATH", "").split(os.pathsep)
+            if not path.startswith(ayon_root)
+        )
+        self.launch_context.env["PATH"] = os.pathsep.join(paths)


client/ayon_core/hooks/pre_remove_launcher_paths.py
```diff
@@ -0,0 +1,36 @@
...
+    def execute(self):
+        # Remove launcher paths from the system
+        paths = []
+        try:
+            ayon_root = os.path.normpath(self.launch_context.env["AYON_ROOT"])
+        except KeyError:
+            self.log.warning("AYON_ROOT not found in environment variables.")
+            return
+
+        paths.extend(
```
iLLiCiTiT, line 31: Why not just list comprehension?
```python
        paths = [
        ]
```

client/ayon_core/hooks/pre_remove_launcher_paths.py
```diff
@@ -0,0 +1,36 @@
...
+    the correct environment variables. Especially for Windows, where
+    paths in `PATH` are used to load DLLs. This is important to avoid
+    conflicts with other applications that may have the same DLLs in their
+    paths.
+    """
+
+    order = 1
+
+    platforms = {"linux", "windows", "darwin"}
+    launch_types = {LaunchTypes.local}
```
iLLiCiTiT, line 20: Shouldn't this happen all the time?
```suggestion
```

client/ayon_core/hooks/pre_remove_launcher_paths.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def execute(self):
+        # Remove launcher paths from the system
+        paths = []
+        try:
+            ayon_root = os.path.normpath(self.launch_context.env["AYON_ROOT"])
+        except KeyError:
+            self.log.warning("AYON_ROOT not found in environment variables.")
+            return
```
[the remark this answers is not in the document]
antirotor: there shouldn't be any difference, no? And I am glad GH is keeping history so I can say "I told you so" when we remove `AYON_ROOT` and this hook stops working 🤣

client/ayon_core/hooks/pre_remove_launcher_paths.py
antirotor: is it set on farm? I thought not

iLLiCiTiT: Well, `extractenvironments` is using AYON launcher, so yes...

iLLiCiTiT (approved):

iLLiCiTiT (approved):

PR #1115, opened 2025-01-30
Colorpsace: Update OCIO config options

jakubjezek001:
## Changelog Description
Added new ocio config options from new release of ayon-ocio addon.

## Additional information
- Introduced new OCIO config for ACES 1.3 and 2.0 Studio.
- Updated paths with relevant labels and descriptions.
- Ensured compatibility with OCIO v2 requirements.

With this PR ayon-ocio >= 1.2.0 is required.

PR #1268, opened 2025-05-09
Handles OCIO shared view token

jakubjezek001:
## Changelog Description
- The OCIO config can return a special token "<USE_DISPLAY_NAME>"
as the colorspace name for a display view. This commit
- implements handling for this token, replacing it with the display
name if the token is found. This is fixing compatibility for >= Aces 1.3 configs.
- updating colorspace conversion method to support source and target colorspace handling

## Additional info
- I had been doing tests in regards of implementation for new ocio configs https://github.com/ynput/ayon-core/pull/1115. Following issue had been blocking me from farther tests. 
- the `convert_colorspace` function was marked as deprecated and new function `oiiotool_transcode` with improved set of input args were added. 

## Testing notes:
1.  ExtractOIIOtrancode presets should work as they were before.
2. Use following snippet in any of your DCC or Tray Console
```
from ayon_core.pipeline.colorspace import get_display_view_colorspace_name

ocio_config_path = "<path to any of new aces1.3 configs from ayon-ocio addon >=1.2.1>"
display = "sRGB - Display"
view = "ACES 1.0 - SDR Video"

resolved_colorspace = get_display_view_colorspace_name(
    ocio_config_path,
    display,
    view,
)

print(resolved_colorspace)
```
3. This should be printing `sRGB - Display` but before it was returning `<USE_DISPLAY_NAME>`.
4.


moonyuet (approved):
It works in my side when speaking of ACESv1.3
image

jakubjezek001:
This has to be put on hold for while, it seems this is more workaround rather then proper solution. OCIO v2 seems to have some better way to resolve the colorspace from view. The token is actually returned only on shared views. So perhaps we need to first detect what is the version and then get the view colorspace.

BigRoy:
> 1. Implement proper handling of the `<USE_DISPLAY_NAME>` token as done in this PR

Note that this PR may potentially affect the Houdini fix here: https://github.com/ynput/ayon-houdini/pull/252
Tagging @MustafaJafar for visibility.

It looks like, if this PR is merged - that the fix implemented there could potentially be removed?

MustafaJafar:
> > 1. Implement proper handling of the `<USE_DISPLAY_NAME>` token as done in this PR
> 
> Note that this PR may potentially affect the Houdini fix here: [ynput/ayon-houdini#252](https://github.com/ynput/ayon-houdini/pull/252) Tagging @MustafaJafar for visibility.
> 
> It looks like, if this PR is merged - that the fix implemented there could potentially be removed?

I gave it a test run and It works. and when merging this PR we can remove https://github.com/ynput/ayon-houdini/blob/ba60bb3416397d7a8b0d638bba0ea94beaf75458/client/ayon_houdini/api/colorspace.py#L66-L67 
The mentioned PR also allows the color space validator to pass if display value is used instead of colorspace.

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -94,6 +94,8 @@ def process(self, instance):
 
             colorspace_data = repre["colorspaceData"]
             source_colorspace = colorspace_data["colorspace"]
+            source_display = colorspace_data.get("display")
+            source_view = colorspace_data.get("view")
```
BigRoy, line 98: As far as I know - any display/view that's currently collected does not represent the SOURCE display/view (as in, the input media does not currently have that embedded) but it represents the display/view that _should_ be applied to match what the user is currently seeing in their DCC.

I'm _quite sure_ that's the case at least inside Maya and Houdini integrations.

If we now start using those as source display/view I'm a bit lost as to where the instance/representation would specify the target display/view based on how the DCC is currently configured.

client/ayon_core/lib/transcoding.py
```diff
@@ -977,7 +1022,62 @@ def convert_colorspace(
...
+        source_colorspace,
+        target_colorspace=target_colorspace,
+        target_display=display,
+        target_view=view,
+        additional_command_args=additional_command_args,
+        logger=logger,
+    )
+
+
+def oiiotool_transcode(
```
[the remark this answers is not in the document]
jakubjezek001: oiiotool is used for more that just colorspace operations. It can be used for repositions too - so the name needs to be more generic.

client/ayon_core/plugins/publish/extract_color_transcode.py
jakubjezek001: You are right, I had forgotten about the case. Will add some changes for the case. But batchdelivery needs to have support for source and target display and view.

client/ayon_core/lib/transcoding.py
```diff
@@ -1024,23 +1133,95 @@ def convert_colorspace(
...
+                ])
+        else:
+            # Standard conversion from colorspace to display/view
+            oiio_cmd.extend([
+                "--iscolorspace",
+                source_colorspace,
+                "--ociodisplay:subimages=0",
+                target_display,
+                target_view,
+            ])
```
iLLiCiTiT, line 1224: This would make the conditions a lot easier to follow
```suggestion
        if source_view and source_display:
            color_convert_args = []
            if target_colorspace:
                # Use temporarty role space
                color_convert_args += ["scene_linear", target_colorspace]
            elif source_display != target_display or source_view != target_view:
                color_convert_args += [target_display, target_view]
            else:
                logger.debug(
                    "Source and target display/view pairs are identical. "
                    "No color conversion needed."
                )
    
            if color_convert_args:
                oiio_cmd.extend([
                    "--ociodisplay:inverse=1:subimages=0",
                    source_display,
                    source_view,
                    "--colorconvert:subimages=0",
                    *color_convert_args
                ])
    
        elif target_colorspace:
            # Standard color space to color space conversion
            oiio_cmd.extend([
                "--colorconvert:subimages=0",
                source_colorspace,
                target_colorspace,
            ])
        else:
            # Standard conversion from colorspace to display/view
            oiio_cmd.extend([
                "--iscolorspace",
                source_colorspace,
                "--ociodisplay:subimages=0",
                target_display,
                target_view,
            ])
```

client/ayon_core/lib/transcoding.py
```diff
@@ -1024,23 +1133,95 @@ def convert_colorspace(
...
     ])
 
-    if all([target_colorspace, view, display]):
-        raise ValueError("Colorspace and both screen and display"
-                         " cannot be set together."
-                         "Choose colorspace or screen and display")
-    if not target_colorspace and not all([view, display]):
-        raise ValueError("Both screen and display must be set.")
+    # Validate input parameters
+    if all([target_colorspace, target_view, target_display]):
```
iLLiCiTiT, line 1137: I would prefer `and` inbetween instead of `all([...])`.

client/ayon_core/pipeline/colorspace.py
```diff
@@ -1403,7 +1403,12 @@ def _get_display_view_colorspace_name(config_path, display, view):
 
     """
     config = _get_ocio_config(config_path)
-    return config.getDisplayViewColorSpaceName(display, view)
+    colorspace = config.getDisplayViewColorSpaceName(display, view)
+    # Special token. See https://opencolorio.readthedocs.io/en/latest/guides/authoring/authoring.html#shared-views # noqa
```
iLLiCiTiT, line 1407:
```suggestion
    # Special token. See https://opencolorio.readthedocs.io/en/latest/guides/authoring/authoring.html#shared-views  # noqa
```

client/ayon_core/lib/transcoding.py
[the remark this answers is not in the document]
BigRoy: > But this function is 100% colorspace related, you must pass in source and target colorspace otherwise it crashes... So it is not generic transcoding. The other arguments are "optional", so the primary usage of the function is "colorspace related", with possibility of doing other stuff, but not the other way around.

Correct - my comment was mostly about the `ExtractOIIOTranscode` plug-in behavior. The idea I had was just to make whatever it's calling as code would end up being less colorspace specific.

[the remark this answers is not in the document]
jakubjezek001: @iLLiCiTiT please suggest me the solution. I can name it whatever but I need to do it in the **depricated** way.

client/ayon_core/lib/transcoding.py
```diff
@@ -1024,23 +1133,95 @@ def convert_colorspace(
...
+        # Case 1: Converting to a named colorspace
+        if all([source_view, source_display]):
+            # First convert from source display/view to a role/reference space
+            # that can be used with colorconvert
+            # For example, converting to "scene_linear" or an appropriate
+            # intermediate space
+            # This is a two-step conversion process since there's no direct
+            # display/view to colorspace command
+            # This could be a config parameter or determined from OCIO config
+            tmp_role_space = "scene_linear"
```
BigRoy, line 1178: Is it actually required for an OCIO config to define a `scene_linear` role? :)

client/ayon_core/plugins/publish/extract_color_transcode.py
BigRoy: Let me know when the changes are made and what you ended up doing. We'll need a way to differentiate between source and target display/view in how we're doing that in those existing integrations.

client/ayon_core/lib/transcoding.py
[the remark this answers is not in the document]
BigRoy: To me `oiiotool_transcode` makes perfect sense if it's NOT specific to colorspace conversion which as to the point I made... is what I think it should be in the end, an `oiio` call that can be generalized to do color conversions, but also skip color conversions and do some plain old `oiiotool` magic or conversions... **Including supporting processing all images and maintaining those in e.g. output EXR** - for example a client requested being able to change EXRs into scanline - but leave the full EXR in tact, etc.

jakubjezek001:
https://github.com/ynput/ayon-core/pull/1268#discussion_r2142737987

it seems to me that the playblast is capturing actually source and target colorspace. It is rendering the viewport into what ever the display is set to, and then it is adding the display and view into the colorspaceData. We might consider the colorspace source and target in this case @BigRoy

jakubjezek001 (commented):
it looks ok to me

client/ayon_core/plugins/publish/extract_color_transcode.py
BigRoy: It seems the collected colorspace data `display` and `view` is still considered as the `source` here.

But it isn't in Maya and Houdini.

It's collected in this order:

- maya: https://github.com/ynput/ayon-maya/blob/06aa2445b4df511225b6795b283dd974040c0b5f/client/ayon_maya/plugins/publish/collect_render.py#L286-L287
- houdini: https://github.com/ynput/ayon-houdini/blob/5bd6349430dff213c2323213efd5566827bb01e0/client/ayon_houdini/plugins/publish/collect_render_colorspace.py#L41-L42
- blender (not in develop yet): https://github.com/ynput/ayon-blender/blob/016b44d670e76d6d309abb7128ab7fd1a0c68d58/client/ayon_blender/plugins/publish/collect_render.py#L109-L113

And then goes to `colorspaceData` via:
- https://github.com/ynput/ayon-core/blob/599ed3ab0478e87b4047cede073d8932fbafc15d/client/ayon_core/pipeline/farm/pyblish_functions.py#L593-L598
- https://github.com/ynput/ayon-core/blob/599ed3ab0478e87b4047cede073d8932fbafc15d/client/ayon_core/pipeline/farm/pyblish_functions.py#L938-L946

However, when rendering e.g. from Maya or Houdini - the default behavior is that the rendered output image renders in the given `colorspace` (the image is that colorspace) but the `display` and `view` is not BAKED in the image. (The image does not contain that display/view) The display/view however _is_ what is enabled/applied in the Maya/Houdini render views so it is what the user is seeing as they work, the rendered image however does not contain it.

As such, it's not the **source** display/view? Or am I misunderstanding what we're doing?

jakubjezek001:
> > [#1268 (comment)](https://github.com/ynput/ayon-core/pull/1268#discussion_r2142737987)
> > it seems to me that the playblast is capturing actually source and target colorspace. It is rendering the viewport into what ever the display is set to, and then it is adding the display and view into the colorspaceData. We might consider the colorspace source and target in this case @BigRoy
> 
> A Maya playblast, yes. (I believe we're also doing no OIIO transcode on those because we're not collecting actually any colorspace data on playblasts?)
> 
> Maya and Houdini renders, no. They render raw output - we pass the display and view from the DCC to the publish data for `ExtractOIIOTranscode` to transcode **to that display view**. In which it's not the source, but the target.

In that case it is not using the workfow as it should. Colorspace data are defining current state of colorspace in the representation and not future state. The target colorspace display and view in this case should be collected to instance.data and then used later as target display and view. We do use target colorspaces in other dccs in settings - but never they are kept in representation colorspaceData as this is not carrier for downstream target processing.

BigRoy:
> In that case it is not using the workfow as it should. Colorspace data are defining current state of colorspace in the representation and not future state. The target colorspace display and view in this case should be collected to instance.data and then used later as target display and view. We do use target colorspaces in other dccs in settings - but never they are kept in representation colorspaceData as this is not carrier for downstream target processing.

I think we should start converting this `instance.data`:
```python
{
	"colorspaceConfig": ocio_path,
	"colorspaceDisplay": display,
	"colorspaceView": view,
}
```
To:
```python
{
	"colorspaceConfig": ocio_path,
	"colorspaceTargetDisplay": display,
	"colorspaceTargetView": view,
}
```
And when referring to source in `instance.data` we should use:
```python
{
	"colorspaceConfig": ocio_path,
	"colorspaceSourceDisplay": display,
	"colorspaceSourceView": view,
}
```
And then for backwards compatibility
```python
instance.data["colorspaceTargetDisplay"] = instance.data.pop("colorspaceDisplay, None)
instance.data["colorspaceTargetView"] = instance.data.pop("colorspaceView, None)
```
Or something along those lines?
Do you have an example, from before this PR where `instance.data["colorspaceDisplay"]` is actively used as "source display" currently instead of target display?

> Here is proposed solution. I would rather add OCIO into the name of keys just to make sure it is explicit enought.

Agreed - for both source and target.



> Current use is following (after [2228037](https://github.com/ynput/ayon-core/commit/2228037656b8eeaa400fcaab1329c6416dbc050f)):
> 
> 1. check if instance data is having set target colorspace, display, view keys and set them as default value
> 2. check if settings are having set any value in fields if yes then override the value set in 1.
> 3. if nothing is set from 1. and 2. then fill the values from repre colorspaceData and use source and target as the same

What if "source" is not set? Which is the case for e.g. maya and houdini render publishes? Only target is defined, and there is no 'source' display view because no display view is ever applied to the source? 🗡️ (I've rarely seen a source with display view embedded on purpose; maybe only from comp?)

> Curretnly user might feel that any value has to always be added to the colorspace fields since it is not explicit enough to navigate him to not to add anything. We could solve it similarly like it is done in Batch delivery tools or Nuke presets.

It's explained in the tooltips for the attributes. @MustafaJafar worked on that I believe. Your provided screenshot does not make this clearer - the fact that it states "Target" does not elaborate to me that it can be empty. There should just be a label instead of tooltip "That these can be empty" OR an extra Enum option that just states: "Use expected target display/view from host" or whatever if you want to make it very explicit.

client/ayon_core/lib/transcoding.py
```diff
@@ -1024,23 +1133,82 @@ def convert_colorspace(
...
-    if all([target_colorspace, view, display]):
-        raise ValueError("Colorspace and both screen and display"
-                         " cannot be set together."
-                         "Choose colorspace or screen and display")
-    if not target_colorspace and not all([view, display]):
-        raise ValueError("Both screen and display must be set.")
+    # Validate input parameters
+    if target_colorspace and target_view and target_display:
+        raise ValueError(
+            "Colorspace and both view  and display cannot be set together."
```
antirotor, line 1139: remove extra space

```suggestion
            "Colorspace and both view and display cannot be set together."
```

antirotor (commented):
some quick notes about the code

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -85,6 +85,11 @@ def process(self, instance):
         new_representations = []
         repres = instance.data["representations"]
         for idx, repre in enumerate(list(repres)):
+            # target space, display and view might be defined upstream
+            target_colorspace = instance.data.get("targetOCIOColorspace")
+            target_display = instance.data.get("targetOCIODisplay")
+            target_view = instance.data.get("targetOCIOView")
```
MustafaJafar, line 91: where are these collected from?

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -94,6 +99,8 @@ def process(self, instance):
 
             colorspace_data = repre["colorspaceData"]
             source_colorspace = colorspace_data["colorspace"]
+            source_display = colorspace_data.get("display")
+            source_view = colorspace_data.get("view")
```
MustafaJafar, line 103: In addition to BigRoy's comment above: 
In Houdini, Colorspace data are not collected in `colorspaceData`. we have them directly inside `instance.data`
```python
instance.data["colorspaceConfig"]  
instance.data["colorspaceDisplay"]
instance.data["colorspaceView"]  
instance.data["colorspace"]             
```
where 
- colorspace is the colrospace of the rendered image.
- view & display the configuration that users should do in their DCCs to view the rendered image correctly.

client/ayon_core/lib/transcoding.py
```diff
@@ -977,7 +1022,62 @@ def convert_colorspace(
     additional_command_args=None,
     logger=None,
 ):
-    """Convert source file from one color space to another.
+    """DEPRECATED function use `oiio_color_convert` instead
```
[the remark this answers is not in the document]
MustafaJafar: I guess this is correct as `convert_colorspace` is deprecated infavor of `oiio_color_convert`.

client/ayon_core/plugins/publish/extract_color_transcode.py
MustafaJafar: Things started to be confusing. 
Maybe we can create a visual a or flowchart that shows plugins that work on these data and when they are triggered.

For reference, this plugin has two modes either it recieves 
- A `target_colorspace`
- **OR** `target_view` and `target_display`

I've already reworked the settings to make it easy for admins but the plugin code can be confusing. 
I'd really appreciate if we can reflect that on code to make it clear how it works.
Also, passing unnecessary arguments to the `oiio_color_convert` function can make it hard to the code the flow. I know the function is smart and handles differnet cases if certain arguments is `None`.

----

FYI, in Houdini we don't set `colorspaceData` in representations and I guess we depend on [ExtractColorspaceData](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_colorspace_data.py) which runs after `ExtractOIIOTranscode`, I doubt it can result in the source colorspace to be wrong.

```python
colorspace_data = repre["colorspaceData"]
source_colorspace = colorspace_data["colorspace"]
```

BigRoy: > FYI, in Houdini we don't set colorspaceData in representations and I guess we depend on [ExtractColorspaceData](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_colorspace_data.py) which runs after ExtractOIIOTranscode, I doubt it can result in the source colorspace to be wrong.

The colorspace data does get set for rendered images from Houdini - so beware of that. It's working in Houdini... for farm renders at least, no idea whether local renders actually set that data correctly too.

client/ayon_core/plugins/publish/extract_color_transcode.py
jakubjezek001: @antirotor and @BigRoy will know, I am not sure myself and it had been suggested to me. Please if there is any code using the key names refers to it here or if there is any issue created for implementing those.

BigRoy: > Shouldn't we do the same thing as with `colorspaceData` and wrap them under one key?
> 
> ```
> {
>     "targetColorspaceData": {
>         "colorspace": "...",
>         "display": "...",
>         "view": "...",
>     }
> }
> ```

And perhaps apply what @antirotor stated [here](https://github.com/ynput/ayon-core/pull/1268#issuecomment-3031448506).

Anyway, to get back to this. Looks like we're still lacking the backwards compatibility mentioned here: https://github.com/ynput/ayon-core/pull/1268#issuecomment-3031352566

So that essentially:
```python
instance.data["colorspaceTargetDisplay"] = instance.data.pop("colorspaceDisplay, None)
instance.data["colorspaceTargetView"] = instance.data.pop("colorspaceView, None)
```
These get converted.

MustafaJafar (approved):
image

Also, it works  in Houdini, making a PR to remove the redundant code.

PR #1275, opened 2025-05-16
Workfiles: Add basic workfiles api

iLLiCiTiT:
## Changelog Description
Moved functionality from workfiles tool and moved a lot of responsibility to host integration.

## Additional info
First of all we did not have functions that would mimic what workfiles tool did, which made it difficult to use workfiles in automations (for example). At the same time the functionality in workfiles tool was "generic" which worked for most of DCC integrations, but not all of them -> it was necessary to more the responsibility to host integration.

### Hosts responsibility
Host integration now has more targeted functions that are related to what it really does, and is context acknowledged. Instead of just open workfile and save workfile it has "Open workfile in context", "Copy workfile to context", "Copy representation as workfile to context", "List workfiles for context" and "List published workfiles for context". These functions are pre-implemented on `IWorkfileHost` interface to do what was done before this PR, but they can be overriden by host integration if necessary.
- NOTE: This causes tha `IWorkfileHost` has pre-implemented methods that should be abstract, but that would break all existing host integrations and most of them would implement the exact same logic.

### API functions
The logic in workfiles tool did trigger events before/after workfile was saved and on context change. This functionality was kept, BUT based on my (quick) research it was used only by host integrations to run additional functionality on context change/file change -> with the new methods introduced that could be potentially ommited over time. The event triggers are not reliable as it has to go through the utils functions in workfiles API now.

Please write down comments what is good/bad, should be moved elsewhere, should happen in different order etc.

## Testing notes:
1. Workfiles tools still does it's job.
2. Using workfiles api functions does the same job.


client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -278,42 +273,141 @@ def get_last_workfile_with_version(
...
         if version is None or file_version > version:
-            output_filenames[:] = []
+            output_filepaths.clear()
             version = file_version
 
         if file_version == version:
-            output_filenames.append(filename)
+            output_filepaths.append(filepath)
+
+    output_filepath = None
```
BigRoy, line 295: This really warrants a comment I noticed.
```suggestion
    # Among matching (highest) version numbers 
    # take the last modified file
    output_filepath = None
```

BigRoy:
> Host integration now has more targeted functions that are related to what it really does, and is context acknowledged. Instead of just open workfile and save workfile it has "Open workfile in context", "Copy workfile to context", "Copy representation as workfile to context", "List workfiles for context" and "List published workfiles for context". These functions are pre-implemented on `IWorkfileHost` interface to do what was done before this PR, but they can be overriden by host integration if necessary.

There being dedicated entry points for a save to another context - I like, it may help simplifying some logic that we currently trigger when opening a workfile from another context.

However, from the methods I'm not entirely sure how one would identify that the context is actually a **different context** than the one the user was/is currently operating in?

Say I wanted to reimplement that method and do something like "if saving to another context then do X". Actually, I'd **love** it if the event system would actually emit that too.



> * NOTE: This causes tha `IWorkfileHost` has pre-implemented methods that should be abstract, but that would break all existing host integrations and most of them would implement the exact same logic.

I think this is absolutely fine actually. Perfect.

> The logic in workfiles tool did trigger events before/after workfile was saved and on context change. This functionality was kept, BUT based on my (quick) research it was used only by host integrations to run additional functionality on context change/file change -> with the new methods introduced that could be potentially ommited over time. The event triggers are not reliable as it has to go through the utils functions in workfiles API now.

Please do keep the events - they are the easiest way to 'install' any custom studio behavior on save/load or any event that does not require actually changing host integrations. I think actually it'd be welcome if we'd also start emitting custom events for "save to different context", "open from different context" (incl. before/after files)

> Please write down comments what is good/bad, should be moved elsewhere, should happen in different order etc.

This is a massive PR, doubles the line count - but lacking the production examples / use cases to actually TEST with and confirm the changes against too. I think we should find a few cases that warrant these changes and try (as for testing the PR) to implement those.

Also, I couldn't directly find - API-wise, how to easily set e.g. a workfile "note". Was that API already existing so that could be easily done? Do you have an example snippet perhaps?

iLLiCiTiT:
> However, from the methods I'm not entirely sure how one would identify that the context is actually a different context than the one the user was/is currently operating in?

Host is responsible for holding current context, most of them are using default implementation (using env variables). So you can compare current context with context passed in. I don't think, API wise, that you should pass to host it's previous context, the host know that.

> Also, I couldn't directly find - API-wise, how to easily set e.g. a workfile "note". Was that API already existing so that could be easily done? Do you have an example snippet perhaps?

True, I focused on moving what we had in workfiles tool. Is there more api functions that comes to you mind?


I would say that the host workfiles implementation should be also responsible for change of current context (because it is host's responsibility), which is not true now. The reason why is because that triggers the events -> we would have to move the responsibility to trigger the events to hosts.

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +142,625 @@ def should_open_workfiles_tool_on_launch(
...
+            open_workfile=True,
+            src_representation_path=representation_path,
+            anatomy=anatomy,
+        )
+    else:
+        host.save_workfile_with_context(
+            workfile_path,
+            folder_entity,
+            task_entity,
+            open_workfile=True,
```
MustafaJafar, line 623:
```suggestion
            task_entity
```

MustafaJafar (commented):
Got this when switching switching context/opening a workfile from another context inside Houdini. 

```python
string indices must be integers, not 'str'
==============================
  File "E:\Ynput\ayon-core\client\ayon_core\tools\workfiles\models\workfiles.py", line 115, in open_workfile
    self._open_workfile(folder_id, task_id, filepath)
  File "E:\Ynput\ayon-core\client\ayon_core\tools\workfiles\models\workfiles.py", line 675, in _open_workfile
    open_workfile(filepath, folder_entity, task_entity)
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\workfile\utils.py", line 285, in open_workfile
    change_current_context(
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\context_tools.py", line 552, in change_current_context
    folder_path = folder_entity[ "path" ]
                  ~~~~~~~~~~~~~^^^^^^^^
TypeERROR: string indices must be integers, not 'str'
```

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -278,42 +273,141 @@ def get_last_workfile_with_version(
...
+    return filtered_paths, dotted_extensions
+
+
+def get_last_workfile_with_version(
+    workdir, file_template, fill_data, extensions
+):
+    """Return last workfile version.
+
+    Usign workfile template and it's filling data find most possible last
+    version of workfile which was created for the context.
```
antirotor, line 371: We already changed in this PR `fill_data` to `template_data`, shouldn't we do the same here?

```suggestion
def get_last_workfile_with_version(
    workdir: str, file_template: str, fill_data: dict[str, Any], extensions: Iterable[str]
) -> tuple[Optional[str], Optional[int]]:
    """Return the last workfile version.

    Using the workfile template and its filling data to find the most possible last
    version of the workfile which was created for the context.
```

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -278,42 +273,141 @@ def get_last_workfile_with_version(
...
 
 
 def get_last_workfile(
-    workdir, file_template, fill_data, extensions, full_path=False
+    workdir: str,
+    file_template: str,
+    fill_data: dict[str, Any],
+    extensions: set[str],
+    full_path: bool = False
 ):
```
antirotor, line 411:
```suggestion
) -> str:
```

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -324,17 +418,23 @@ def get_last_workfile(
         file_template (str): Template of file name.
         fill_data (Dict[str, Any]): Data for filling template.
         extensions (Iterable[str]): All allowed file extensions of workfile.
-        full_path (Optional[bool]): Full path to file is returned if
+        full_path (bool): Full path to file is returned if
```
antirotor, line 421:
```suggestion
        full_path (bool): The full path to the file is returned if
```

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -562,3 +662,100 @@ def create_workdir_extra_folders(
...
+            ("<<ext>>", any_extension),
+        )
+        for src, dest in replacements:
+            fname_pattern = fname_pattern.replace(re.escape(src), dest)
+
+        # Match from beginning to end of string to be safe
+        self._fname_regex = re.compile(f"^{fname_pattern}$")
+
+    def parse_comment(self, filename: str) -> Optional[str]:
+        """Parse the {comment} part from a filename"""
```
antirotor, line 716:
```suggestion
        """Parse the {comment} part from a filename."""
```

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -562,3 +662,100 @@ def create_workdir_extra_folders(
...
+    """Collect comments from workfile filenames.
+
+    Based on 'current_filename' is also returned "current comment".
+
+    Args:
+        filepaths (list[str]): List of filepaths to parse.
+        extensions (set[str]): Set of file extensions.
+        file_template (StringTemplate): Workfile file template.
+        template_data (dict[str, Any]): Data to fill the template with.
+        current_filename (str): Filename to check for current comment.
```
antirotor, line 740:
```suggestion
        current_filename (str): Filename to check for the current comment.
```

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +142,624 @@ def should_open_workfiles_tool_on_launch(
...
         return default_output
     return output
+
+
+def _get_event_context_data(
+    project_name: str,
+    folder_entity: dict[str, Any],
+    task_entity: dict[str, Any],
+    host_name: str,
+):
```
antirotor, line 152: In my ideal unicorn world this would also return dataclass. Also for this kind of function where project name, folder entity, task entity is always mandatory, shouldn't we wrap it to dataclass too? It would simplify things.

```suggestion
def _get_event_context_data(
    project_name: str,
    folder_entity: dict[str, Any],
    task_entity: dict[str, Any],
    host_name: str,
) -> dict[str, Any]:
```

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +142,624 @@ def should_open_workfiles_tool_on_launch(
...
+    project_name: str,
+    task_id: str,
+    rootless_path: str,
+    host_name: str,
+    version: Optional[int],
+    comment: Optional[str],
+    description: Optional[str],
+    username: Optional[str] = None,
+    workfile_entities: Optional[list[dict[str, Any]]] = None,
+):
```
antirotor, line 173: This could do with docstring and also "dataclassed" workfile entity that it returns :)

```suggestio
def save_workfile_info(
    project_name: str,
    task_id: str,
    rootless_path: str,
    host_name: str,
    version: Optional[int],
    comment: Optional[str],
    description: Optional[str],
    username: Optional[str] = None,
    workfile_entities: Optional[list[dict[str, Any]]] = None,
) -> dict[str, Any]:
```

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +142,624 @@ def should_open_workfiles_tool_on_launch(
...
+        )
+
+    data = {}
+    for key, value in (
+        ("host_name", host_name),
+        ("version", version),
+        ("comment", comment),
+    ):
+        if value is not None:
+            data[key] = value
```
antirotor, line 212: List comprehension is in this case readable too and a little bit faster.

```suggestion
    data = {
        key: value
        for key, value in (
            ("host_name", host_name),
            ("version", version),
            ("comment", comment),
        )
        if value is not None
    }
```

antirotor (changes requested):
This is mostly about the code itself, sorry about it. You can dismiss it if you want but I have few bigger topics there:

- it would be great if we could formalize all those values we are passing between functions as `dict[Any, str]` to dataclass. This is mostly about many types of entities.  If we can do that, that would really help with mypy and stuff.
- another thing is that we should use `@deprecated` decorator.

Other than that just please skim through the comments, apart of some typos and stuff that you'll change anyway are some notes that might help. I'll do another review just for the logic.

client/ayon_core/pipeline/workfile/utils.py
iLLiCiTiT: Less readable to me...

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +142,624 @@ def should_open_workfiles_tool_on_launch(
...
+    task_id: str,
+    rootless_path: str,
+    host_name: str,
+    version: Optional[int],
+    comment: Optional[str],
+    description: Optional[str],
+    username: Optional[str] = None,
+    workfile_entities: Optional[list[dict[str, Any]]] = None,
+):
+    # TODO create pipeline function for this
```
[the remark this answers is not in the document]
iLLiCiTiT:
```suggestion
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,654 @@
...
+
+        Args:
+            dst_path (str): Where the current scene should be saved. Or use
+                current path if 'None' is passed.
+
+        """
+        pass
+
+    @abstractmethod
+    def open_workfile(self, filepath: str):
```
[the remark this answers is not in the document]
iLLiCiTiT: Nope, I won't change existing function return type hint.

BigRoy:
> I would say that the host workfiles implementation should be also responsible for change of current context (because it is host's responsibility), which is not true now. The reason why is because that triggers the events -> we would have to move the responsibility to trigger the events to hosts.

I understand the reasoning - I'm just wondering what we'd gain (in actual use cases as of now) by doing so. Why couldn't some of that just be relying on an event getting triggered?

In that, core does same "base logic", but if you want additional things to happen, then you connect to the events?
The upside of events is that you can hook into them even from custom addons, instead of only being able to access them from the host addon itself.
That way, it's just a set of 'things to respond to' instead of completely changing around how workfiles API (may work) in each DCC. Is it maybe adding too much code/maintenance for little gain (if any?) in actual production use cases?

antirotor:
What I am missing related to the discussion about Workfile API is the actual list of things it provides, so here it goes, extracted from the code:

`WorkfileInfo` dataclass - [🔗 link](https://github.com/ynput/ayon-core/blob/edb371fb38b6c0d7e57f3450608db833080cb4f7/client/ayon_core/host/interfaces/workfiles.py#L18)

`PublishedWorkfileInfo` dataclass [🔗 link](https://github.com/ynput/ayon-core/blob/edb371fb38b6c0d7e57f3450608db833080cb4f7/client/ayon_core/host/interfaces/workfiles.py#L83C29-L84C5)

`IWorkfileHost` interface - [🔗 link](https://github.com/ynput/ayon-core/blob/edb371fb38b6c0d7e57f3450608db833080cb4f7/client/ayon_core/host/interfaces/workfiles.py#L149)
 - `save_workfile(...)`
 - `open_workfile(...)`
 - `get_current_workfile(...) -> Optional[str]`
 - `workfile_has_unsaved_changes(...) -> Optional[bool]`
 - `get_workfile_extensions(...) -> list[str]`
 - `save_workfile_with_context(...)`
 - `open_workfile_with_context(...)`
 - `list_workfiles(...)`
 - `list_published_workfiles(...)`
 - `copy_workfile(...)`
 - `copy_workfile_representation(...)`

antirotor:
Maybe not for this PR, but I think we should add some type of workfile localization into the API. So far I think only Nuke has something native. But especially with massive workfiles being able to transparently localize the scene so it is opened much faster is really beneficial. But that adds a lot complexity (local storage, cleanup, etc.) So just an idea.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,654 @@
...
+            folder_entity (dict[str, Any]): Folder entity.
+            task_entity (dict[str, Any]): Task entity.
+            open_workfile (bool): Open workfile when copied.
+
+        """
+        # TODO We might need option to open file once copied as some DCC might
+        #   actually need to open the workfile to copy it.
+        dst_dir = os.path.dirname(dst_path)
+        if not os.path.exists(dst_dir):
+            os.makedirs(dst_dir, exist_ok=True)
```
MustafaJafar, line 502: When working on the AY<->alab automation, we faced a similar situation. 
Where @BigRoy mentioned that creating the work directory like this doesn't fully initialize the work directory and it should happen in a similar manner to https://github.com/ynput/ayon-core/blob/88647eee1ec3b23b177297412e44f290fc8bec96/client/ayon_core/pipeline/workfile/utils.py#L503

Also, when copying a workfile, we need to consider/check if there are any existing workfiles and upversion accordingly. 

It's hard for to track all the changes so do these happen already?

Also, why do I have to compute and pass dst path?
I expected AYON should be able to compute the dst path itself since it already knows the context I want.

-----

<details>
<summary>Here's a snippet from Houdini about what I expected <code>copy_workfile</code> to do</summary>

```python
import os
import hou
from ayon_core.pipeline.workfile import (
    get_last_workfile,
    get_workfile_template_key,
    create_workdir_extra_folders,
)
from ayon_core.pipeline.context_tools import change_current_context
from ayon_core.pipeline.template_data import get_template_data
from ayon_core.pipeline import registered_host, Anatomy
from ayon_core.lib import version_up


def init_work_dir_and_create_workfile(folder_entity, task_entity):
    """Initialize work directory and create workfile.
    This function: 
        - Gets next version workfile or the first version if it doesn't exist.
        - Creates work directory if it doesn't exist.
        - Creates/Saves the workfile.
        - Also creates extra workdir folders.

    This is a modified version of 
    `ayon_core.pipeline.context_tools.version_up_current_workfile()`

    Args:
        folder_entity
        task_entity
    """

    host = registered_host()
    host_name = host.name

    project_name = <some houdini call to get cached project name >
    project_entity = <some houdini call to get cached project entity>
    project_settings = <some houdini call to get cached project settings>
    
    anatomy = Anatomy(project_name)
    
    task_name = task_entity["name"]

    # Prepare template data
    template_key = get_workfile_template_key(
         project_name, task_type, host_name, project_settings
    )

    data = get_template_data(
        project_entity, folder_entity, task_entity, host_name, project_settings
    )
    data["root"] = anatomy.roots

    work_template = anatomy.get_template_item("work", template_key)

    # Define saving file extension
    extensions = host.get_workfile_extensions()
    current_file = host.get_current_workfile()
    if current_file:
        extensions = [os.path.splitext(current_file)[-1]]

    # Get work template
    work_root = work_template["directory"].format_strict(data)
    file_template = work_template["file"].template
    # Get last or first workfile path
    workfile_path = get_last_workfile(
        work_root, file_template, data, extensions, True
    )

    # Get next workfile path if last version exists.
    if os.path.exists(workfile_path):
        workfile_path = version_up(workfile_path)

    # Create work dir if it doesn't exist already.
    workdir = os.path.dirname(workfile_path)
    if not os.path.exists(workdir):
        os.makedirs(workdir)

    # Save workfile. Typically this fails if workdir doesn't exist.
    host.save_workfile(workfile_path)

    # Create extra folders
    create_workdir_extra_folders(
        workdir,
        host.name,
        task_type,
        task_name,
        project_name
    )

# ----------------------
workfile_template = <some houdini call to get cached workfile template path>
folder_entity = <some houdini call to get cached folder entity>
task_entity = <some houdini call to get cached task entity>

# Open the file
hou.hipFile.load(workfile_template,
                 suppress_save_prompt=True,
                 ignore_load_warnings=False)
        
changed = change_current_context(
    folder_entity,
    task_entity
)

init_work_dir_and_create_workfile(folder_entity, task_entity)
```

</details>

And, I anticipated that in my automation, I can just say
```python
import ayon_api

# ----- Get AYON Context
project_name = "Animal_Logic_ALab"
folder_path = "/assets/book_encyclopedia01"
task_name = "modeling" 

folder_entity = ayon_api.get_folder_by_path(project_name, folder_path, fields={"id"})
task_entity = ayon_api.get_task_by_folder_path(project_name, folder_path, task_name, fields={"id"})

# ----- Copy and Open workfile
workfile_template = "Path to my template file"

host = registered_host()
host.copy_workfile(
    workfile_template,
    folder_entity,
    task_entity,
    True
)
```

iLLiCiTiT:
> In that, core does same "base logic", but if you want additional things to happen, then you connect to the events?
The upside of events is that you can hook into them even from custom addons, instead of only being able to access them from the host addon itself.
That way, it's just a set of 'things to respond to' instead of completely changing around how workfiles API (may work) in each DCC. Is it maybe adding too much code/maintenance for little gain (if any?) in actual production use cases?

It's not "additional", host integration is responsible for "current context" and with IWorkfileHost for workfiles handling. And if you "open workfile" you also "change context", it is not because someone called method to change context, but because the workfile is related to some context. It is not 2 step operation, it is 1 step operation.

client/ayon_core/pipeline/workfile/path_resolving.py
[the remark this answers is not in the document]
iLLiCiTiT: Please not in this PR, there is already bzillion changes and it's hard to follow which are actually related to changes here.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1177 @@
...
+
+        Notes:
+            Should this method care about context change?
+
+        Args:
+            filepath (str): Where the current scene should be saved.
+            folder_entity (dict[str, Any]): Folder entity.
+            task_entity (dict[str, Any]): Task entity.
+            version (Optional[int]): Version of the workfile.
+            comment (Optional[str]): Comment for the workfile.
```
BigRoy, line 298:
```suggestion
            comment (Optional[str]): Comment for the workfile.
                Usually used in the filename template.
```
Or is this data also in the workfile info on server?

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1177 @@
...
+            description (Optional[str]): Description for the workfile.
```
BigRoy, line 299: Bit confused that this is now suddenly description instead of Artist Note. I felt like artist note was clearer on what it actually is...

Anyway, can we get the docstring to clarify a bit more what comment versus description is. Like explain that e.g. this is some "artist note about what the workfile contains" or whatever?

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1177 @@
...
+            rootless_path (Optional[str]): Rootless path of the workfile.
+            workfile_entities (Optional[list[dict[str, Any]]]): Workfile
```
BigRoy, line 301: What's this? What type of workfile entities would I pass to `save_workfile`... and why is it plural?

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1177 @@
...
+            Should this method care about context change?
+
+        Args:
+            filepath (str): Where the current scene should be saved.
+            folder_entity (dict[str, Any]): Folder entity.
+            task_entity (dict[str, Any]): Task entity.
+            version (Optional[int]): Version of the workfile.
+            comment (Optional[str]): Comment for the workfile.
+            description (Optional[str]): Description for the workfile.
+            rootless_path (Optional[str]): Rootless path of the workfile.
```
BigRoy, line 300: I assume this is optional, just so you could pass it as an optimization to avoid the function querying this data again if you already have it. I wonder if those arguments in the docstring are best marked like that. Like have it decscribe: "Optimizes the logic if you have this data prefetched" or whatever.

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +186,408 @@ def should_open_workfiles_tool_on_launch(
...
+    task_id: str,
+    rootless_path: str,
+    host_name: str,
+    version: Optional[int] = None,
+    comment: Optional[str] = None,
+    description: Optional[str] = None,
+    username: Optional[str] = None,
+    workfile_entities: Optional[list[dict[str, Any]]] = None,
+):
+    if workfile_entities is None:
```
BigRoy, line 202: Wouldn't it be nicer if this would just take `WorkfileInfo` dataclass?

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +186,408 @@ def should_open_workfiles_tool_on_launch(
...
+        project_settings=project_settings,
+        anatomy=anatomy,
+    )
+
+
+def save_current_workfile_to(
+    workfile_path: str,
+    folder_entity: dict[str, Any],
+    task_entity: dict[str, Any],
+    *,
```
BigRoy, line 308: Why do we need the `*`?

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -329,10 +331,13 @@ def _on_ok_pressed(self):
...
             "filename": self._filename,
             "workdir": self._workdir,
+            "rootless_workdir": self._rootless_workdir,
             "folder_id": self._folder_id,
             "task_id": self._task_id,
             "template_key": self._template_key,
-            "artist_note": self._artist_note_input.toPlainText(),
+            "version": self._version_value,
+            "comment": self._comment_value,
+            "description": self._description_input.toPlainText(),
```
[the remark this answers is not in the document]
iLLiCiTiT: Because it is stored as `"description"` in workfile entity.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,654 @@
...
+        Notes:
+            Should this method care about context change?
+
+        Args:
+            filepath (str): Where the current scene should be saved.
+            folder_entity (dict[str, Any]): Folder entity.
+            task_entity (dict[str, Any]): Task entity.
+
+        """
+        self.save_workfile(filepath)
```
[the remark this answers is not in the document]
iLLiCiTiT: Outdated

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,654 @@
...
+        self.save_workfile(dst_path)
+
+    def open_file(self, filepath):
+        """Deprecated variant of 'open_workfile'.
+
+        Todo:
+            Remove when all usages are replaced.
+        """
+
+        return self.open_workfile(filepath)
```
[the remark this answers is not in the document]
iLLiCiTiT: > open_workfile() doens't return anything so this should neither.

Really? Have you checked all host integrations?

client/ayon_core/host/interfaces/workfiles.py
BigRoy: I'm still a bit confused by this one actually - I see WHY you're passing it the list, because `ayon_api` allows you to query it that way. However, functionality wise I'd actually expect to query the workfile entity for a particular file (instead of ALL workfile entities for the task) just by the filepath or rootless filepath.

So that it'd just be `ayon_api.get_workfile_entity(project_name, rootless_path)` or alike?

It just looks confusing to pass this list even though it really only ever just 'guesses' and takes only ONE of the entries. It's never about taking multiples from that list. It's like passing an optimized lookup instead of saying, here take this workfile entity to set it on (which is what I'd have expected).

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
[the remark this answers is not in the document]
BigRoy: I see - thanks.

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: It should either expect rootless path, or what is there now, not a combination.

BigRoy: @antirotor @philippe-ynput thoughts?

iLLiCiTiT: I just realized it is on the interface, and my statement that it should expect only one is probably wrong. I think that current state is conveniently better as api function. The method should be used instead of current `save_workfile` if it should also change context.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1179 @@
...
+        """
+        pass
+
+    @abstractmethod
+    def get_current_workfile(self) -> Optional[str]:
+        """Retrieve a path to current opened file.
+
+        Returns:
+            Optional[str]: Path to the file which is currently opened. None if
+                nothing is opened.
```
BigRoy, line 240:
```suggestion
                nothing is opened or the current workfile is unsaved.
```

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: True, but it does not support it now. We would have to wait for server PR, make server release, then change ayon api, make release, then use new ayon api in AYON launcher, make release, and then we can use it... 🤔

BigRoy: At the same time - if we don't... and release the API like this instead - we'll have to maintain that backwards compatibility forever. Whereas if we clean up this one bit, the calls would be faster and the API we want to release much simplified and have less technical debt.

It _could_ be worth it if we want to produce a clean API for this. (and perhaps rely on graphql in the meantime?)

client/ayon_core/host/host.py
```diff
@@ -141,6 +150,82 @@ def get_current_context(self):
...
+                of the arguments.
+
+        Args:
+            folder_entity (Optional[dict[str, Any]]): Folder entity.
+            task_entity (Optional[dict[str, Any]]): Task entity.
+            reason (Optional[str]): Reason for context change.
+            project_entity (Optional[dict[str, Any]]): Project entity data.
+            anatomy (Optional[Anatomy]): Anatomy instance for the project.
+
+        """
```
antirotor, line 180:
```suggestion
            anatomy (Optional[Anatomy]): Anatomy instance for the project.
            
        Returns:
            dict[str, Optional[str]]: context information about project name, folder path and task name.

        """
```

client/ayon_core/host/host.py
```diff
@@ -141,6 +150,82 @@ def get_current_context(self):
...
         }
 
+    def set_current_context(
+        self,
+        folder_entity: dict[str, Any],
+        task_entity: dict[str, Any],
+        *,
+        reason: Optional[str] = None,
+        project_entity: Optional[dict[str, Any]] = None,
+        anatomy: Optional["Anatomy"] = None,
```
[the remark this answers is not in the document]
antirotor: or is it because of circular imports?

client/ayon_core/host/host.py
```diff
@@ -187,3 +272,113 @@ def maintained_selection(self):
             yield
         finally:
             pass
+
+    def _emit_context_change_event(
```
antirotor, line 276: This should annotate and document what it returns

client/ayon_core/host/host.py
```diff
@@ -187,3 +272,113 @@ def maintained_selection(self):
...
+                os.environ[key] = value
+
+    def _before_context_change(
+        self,
+        project_entity: dict[str, Any],
+        folder_entity: Optional[dict[str, Any]],
+        task_entity: Optional[dict[str, Any]],
+        reason: Optional[str],
+        anatomy: "Anatomy",
+    ):
```
antirotor, line 349:
```suggestion
    ) -> None:
```

client/ayon_core/host/host.py
```diff
@@ -187,3 +272,113 @@ def maintained_selection(self):
...
+        pass
+
+    def _after_context_change(
+        self,
+        project_entity: dict[str, Any],
+        folder_entity: dict[str, Any],
+        task_entity: dict[str, Any],
+        reason: Optional[str],
+        anatomy: "Anatomy",
+    ):
```
antirotor, line 371:
```suggestion
    ) -> None:
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+
+@dataclass
+class WorkfileInfo:
+    """Information about workfile.
+
+    Host can open, copy and use the workfile using this information object.
+
+    Attributes:
+        filepath (str): Path to the workfile.
+        rootless_path (str): Path to the workfile without root. And without
```
antirotor, line 32:
```suggestion
        rootless_path (str): Path to the workfile without the root. And without
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+    available: bool
+
+    @classmethod
+    def new(
+        cls,
+        filepath: str,
+        rootless_path: str,
+        available: bool,
+        workfile_entity: dict[str, Any],
+    ):
```
antirotor, line 67: These are hard to annotate in Python versions without `Self` but I believe it can be done like this
```suggestion
    ) -> "WorkfileInfo":
```
But if inheritance is issue, `TypeVar` will help and it works in Python 3.7 too:

```python
Self = TypeVar("Self", bound="MyClass")
```

antirotor, line 67:
```suggestion
    def new(
        cls,
        filepath: str,
        rootless_path: str,
        workfile_entity: dict[str, Any],
        *,
        available: bool,
    ):
```

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: > At the same time - if we don't... and release the API like this instead - we'll have to maintain that backwards compatibility forever. Whereas if we clean up this one bit, the calls would be faster and the API we want to release much simplified and have less technical debt.
> 
> It could be worth it if we want to produce a clean API for this. (and perhaps rely on graphql in the meantime?)

So you suggest to wait with this PR until it's available, which might be few weeks?

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+            file_created=file_created,
+            file_modified=file_modified,
+            workfile_entity_id=workfile_entity.get("id"),
+            description=attrib.get("description") or "",
+            created_by=workfile_entity.get("createdBy"),
+            updated_by=workfile_entity.get("updatedBy"),
+            available=available,
+        )
+
+    def to_data(self) -> dict[str, Any]:
```
antirotor, line 95: This can make use of *TypedDict* because you know exactly what data you need to build `WorkfileInfo` obejct. Basically `TypedDict` with signature from the constructor. The same applies to `from_data()` below.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,654 @@
...
+
+        """
+        return WorkfileInfo(**data)
+
+
+class IWorkfileHost:
+    """Implementation requirements to be able use workfile utils and tool."""
+
+    @abstractmethod
+    def save_workfile(self, dst_path: Optional[str] = None):
```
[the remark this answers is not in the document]
antirotor: Why? If this doesn't change the functionality at all? It helps static type checkers and that is just some better developer experience.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+                'list_published_workfiles' are re-implemented with different
+                logic.
+
+        Returns:
+            list[str]: List of extensions that can be used for saving.
+
+        """
+        return []
+
+    def save_workfile_with_context(
```
antirotor, line 273: I don't like public functions with some many arguments (and most of them optional). Could we find a better way  to pass these commonly passed things? I am touching again the `folder_entity`, `task_entity` as I've mentioned before. There are several ways to solve it:

- split it into more functions - way more easier to write unit test.
- builder pattern
- group data together in defined structures

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+            project_entity (Optional[dict[str, Any]]): Project entity.
+            anatomy (Optional[Anatomy]): Project anatomy.
+
+        """
+        self._before_workfile_save(
+            filepath,
+            folder_entity,
+            task_entity,
+        )
+        project_name = self.get_current_project_name()
```
antirotor, line 323: This is calling `self.get_current_project_name()` that is not implemented in `IWorkfileHost` so it assumes that this interface is implementing something else too that is not defined here?

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+
+        workdir = os.path.dirname(filepath)
+
+        # Set 'AYON_WORKDIR' environment variable
+        os.environ["AYON_WORKDIR"] = workdir
+
+        if project_entity is None:
+            project_entity = ayon_api.get_project(project_name)
+
+        self.set_current_context(
```
antirotor, line 340: This isn't implemented on `IWorkfileHost` either

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+    def open_workfile_with_context(
+        self,
+        filepath: str,
+        folder_entity: dict[str, Any],
+        task_entity: dict[str, Any],
+        *,
+        project_entity: Optional[dict[str, Any]] = None,
+        project_settings: Optional[dict[str, Any]] = None,
+        anatomy: Optional["Anatomy"] = None,
+    ):
```
antirotor, line 379:
```suggestion
    ) -> None:
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,654 @@
...
+        workfile_entities_by_path = {
+            workfile_entity["path"]: workfile_entity
+            for workfile_entity in workfile_entities
+        }
+
+        workdir_data = get_template_data(
+            project_entity,
+            folder_entity,
+            task_entity,
+            host_name=self.name,
```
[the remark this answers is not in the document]
antirotor: Shouldn't it then inherit from the `HostBase`?

client/ayon_core/host/interfaces/workfiles.py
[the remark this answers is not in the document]
antirotor: No need to check. This is declaration of API - it won't change how this works. It saying that we don't expect it to return any value. And if we do expect, we should say so. With this change, developers will see somethings is wrong while they work on their integrations and change the code when needed.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+        repre_entities: Optional[list[dict[str, Any]]] = None,
+    ) -> list[PublishedWorkfileInfo]:
+        """List published workfiles for the given folder.
+
+        The default implementation looks for products with the 'workfile'
+            product type.
+
+        Pre-fetched entities have mandatory fields to be fetched.
+         -  Version: 'id', 'author', 'taskId'
+         -  Representation: 'id', 'versionId', 'files'
```
antirotor, line 568: Just so the docstring generator will handle this.

```suggestion
        Pre-fetched entities have mandatory fields to be fetched:
            - Version: 'id', 'author', 'taskId'
            - Representation: 'id', 'versionId', 'files'
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1214 @@
...
+        project_settings: Optional[dict[str, Any]] = None,
+        project_entity: Optional[dict[str, Any]] = None,
+        anatomy: Optional["Anatomy"] = None,
+        open_workfile: bool = True,
+        src_anatomy: Optional["Anatomy"] = None,
+        src_representation_path: Optional[str] = None,
+    ):
+        """Copy workfile representation.
+
+        Use representation as source for the workfile.
```
antirotor, line 783:
```suggestion
        Use representation as a source for the workfile.
```

antirotor (changes requested):
Another batch of some notes

client/ayon_core/host/host.py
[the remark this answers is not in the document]
iLLiCiTiT: It is hard to typehint entities that have fields defined by `fields` argument passed to the functing fetching them. I really don't have the intention to do it in this PR.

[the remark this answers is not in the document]
iLLiCiTiT: Circular imports...

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: Why? No reason to require kwargs here...

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: Do it if you want to do it. I don't know how, and I don't want to spend few day tweaking it to make you satisfied.

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: It is expected the Iterface is combined with `HostBase`. Don't know how to achieve that to not break hosts because of how python implement abstractions.

client/ayon_core/host/interfaces/workfiles.py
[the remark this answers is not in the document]
iLLiCiTiT: It would not be interface anymore, or?

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+
+    def _after_workfile_copy(
+        self, copy_workfile_data: WorkfileCopyData
+    ) -> None:
+        """After workfile is copied.
+
+        This method is called after the workfile is copied by host
+            integration.
+
+        Can be overriden to implement host specific logic.
```
antirotor, line 1252:
```suggestion
        Can be overridden to implement host specific logic.
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+
+    def _before_workfile_copy(
+        self, copy_workfile_data: WorkfileCopyData
+    ) -> None:
+        """Before workfile is copied.
+
+        This method is called before the workfile is copied by host
+            integration.
+
+        Can be overriden to implement host specific logic.
```
antirotor, line 1235:
```suggestion
        Can be overridden to implement host specific logic.
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+        pass
+
+    def _after_workfile_save(
+        self, save_workfile_data: WorkfileSaveData
+    ) -> None:
+        """After workfile is saved.
+
+        This method is called after the workfile is saved in the host.
+
+        Can be overriden to implement host specific logic.
```
antirotor, line 1213:
```suggestion
        Can be overridden to implement host specific logic.
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+        pass
+
+    def _before_workfile_save(
+        self, save_workfile_data: WorkfileSaveData
+    ) -> None:
+        """Before workfile is saved.
+
+        This method is called before the workfile is saved in the host.
+
+        Can be overriden to implement host specific logic.
```
antirotor, line 1197:
```suggestion
        Can be overridden to implement host specific logic.
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+        pass
+
+    def _after_workfile_open(
+        self, open_workfile_data: WorkfileOpenData
+    ) -> None:
+        """After workfile is opened.
+
+        This method is called after the workfile is opened in the host.
+
+        Can be overriden to implement host specific logic.
```
antirotor, line 1181:
```suggestion
        Can be overridden to implement host specific logic.
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+        }
+
+    def _before_workfile_open(
+        self, open_workfile_data: WorkfileOpenData
+    ) -> None:
+        """Before workfile is opened.
+
+        This method is called before the workfile is opened in the host.
+
+        Can be overriden to implement host specific logic.
```
antirotor, line 1165:
```suggestion
        Can be overridden to implement host specific logic.
```

client/ayon_core/host/interfaces/workfiles.py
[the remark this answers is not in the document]
antirotor: Interface inheritance is a real thing in most languages, so I see no issue there.

client/ayon_core/host/interfaces/workfiles.py
antirotor: It is not to satisfy me. I just think that if we are introducing new API, it should be nice to use and test. Maybe @philippe-ynput can decide here. I don't want to be human based "linter" here. But try to write unit tests for functions like these and it is major PITA.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1320 @@
...
+    def new(
+        cls,
+        project_name: str,
+        folder_id: str,
+        task_id: Optional[str],
+        repre_entity: dict[str, Any],
+        filepath: str,
+        author: str,
+        available: bool,
+        file_size: Optional[float],
```
antirotor, line 208:
```suggestion
        author: str,
        *,
        available: bool,
        file_size: Optional[float],
```

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -278,64 +280,184 @@ def get_last_workfile_with_version(
...
 
 
 def get_last_workfile(
-    workdir, file_template, fill_data, extensions, full_path=False
-):
+    workdir: str,
+    file_template: str,
+    template_data: dict[str, Any],
+    extensions: set[str],
+    full_path: bool = False,
```
antirotor, line 421:
```suggestion
    workdir: str,
    file_template: str,
    template_data: dict[str, Any],
    extensions: set[str],
    *,
    full_path: bool = False,
```

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -1,19 +1,81 @@
...
-    task_name,
-    task_type,
-    default_output=False,
-    project_settings=None,
+    project_name: str,
+    host_name: str,
+    task_name: str,
+    task_type: str,
+    default_output: bool = False,
+    project_settings: Optional[dict[str, Any]] = None,
```
antirotor, line 78: If we are adding type hinting here, we could hint the return value too.

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +186,429 @@ def should_open_workfiles_tool_on_launch(
...
+            workfile entities related to the task.
+        project_entity (Optional[dict[str, Any]]): Project entity used for
+            rootless path calculation.
+        project_settings (Optional[dict[str, Any]]): Project settings used for
+            rootless path calculation.
+        anatomy (Optional[Anatomy]): Project anatomy used for rootless
+            path calculation.
+
+    Returns:
+        dict[str, Any]: Workfile info entity.
```
antirotor, line 417: This function doesn't return anything

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +186,429 @@ def should_open_workfiles_tool_on_launch(
...
```
antirotor, line 486: This function doesn't return anything

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -124,3 +186,429 @@ def should_open_workfiles_tool_on_launch(
...
+        src_anatomy=src_anatomy,
+        src_representation_path=representation_path,
+        open_workfile=open_workfile,
+    )
+
+
+def find_workfile_rootless_path(
+    workfile_path: str,
+    project_name: str,
+    folder_entity: dict[str, Any],
```
antirotor, line 521: this isn't used down the line

antirotor (changes requested):
some additional notes

client/ayon_core/host/host.py
```diff
@@ -92,37 +110,38 @@ def log(self):
...
-    def name(self):
+    @property
+    @abstractmethod
+    def name(self) -> str:
         """Host name."""
 
         pass
 
-    def get_current_project_name(self):
+    def get_current_project_name(self) -> str:
```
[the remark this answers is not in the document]
iLLiCiTiT: In that case the host would fail to be registered. Plese don't do these in this PR.

client/ayon_core/host/interfaces/workfiles.py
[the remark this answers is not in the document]
iLLiCiTiT: In most languages, not in python.

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: I don't understand why it's PITA?

It can require `filepath`, `rootless_path`, `project_entity`, `folder_entity`, `task_entity`, `anatomy`, `workfile_entity_id` and optional `version`, `comment` and `description`.
But that would be PITA to use.

client/ayon_core/host/interfaces/workfiles.py
antirotor: Sure, but it should be probably higher-level function than this one? And if so, should it be part of this PR? @iLLiCiTiT

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: I really don't want to do that. This is interface that is pre-implementing stuff because we're adding ad-hoc functionality to base interface that should be in ideal case implemented by the host itself, or we should have prepared pre-implemented interface for that elsewhere -> but that would still use the methods that are defined on `HostBase`. Adding `HostBase` to inheritance would probably break more things based on how I know python class inheritance works.

I am afraid of what adding this might cause, breaking addons just because "this is coded by a book" doesn't sound good to me.

client/ayon_core/host/interfaces/workfiles.py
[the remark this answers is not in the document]
iLLiCiTiT: Declaration of deprecated api that was still used and the suggestion would break it.

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: It already is part of this PR (that's one of reasons why the methods require so much arguments).

BigRoy (changes requested):
```python
# ==============================
# IWorkfileHost.save_workfile_with_context() takes 4 positional arguments but 7 positional arguments (and 5 keyword-only arguments) were given
# ==============================
#   File "F:\dev\ayon-core\client\ayon_core\tools\workfiles\models\workfiles.py", line 154, in save_as_workfile
#     save_current_workfile_to(
#   File "F:\dev\ayon-core\client\ayon_core\pipeline\workfile\utils.py", line 363, in save_current_workfile_to
#     host.save_workfile_with_context(
TypeError: IWorkfileHost.save_workfile_with_context() takes 4 positional arguments but 7 positional arguments (and 5 keyword-only arguments) were given
```

Can't save in workfiles tool in Maya.

antirotor:
> So the remaining question is if the optional arguments for the new methods are ok, or we should wrap them into some object to make our lives easier for future (talking about project entity, anatomy, workfile entities, etc.).

I'd say if they belong together (ie. they need to be passed together) they should be grouped. Otherwise no.

iLLiCiTiT:
I've wrapper optional speed enhancement arguments into a wrappers. All the values are then transformed into a dataclasses that are used for rest of the logic.

antirotor (commented):
it looks fine, need to be tested

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -1,19 +1,87 @@
...
+
+    if workfile_entities is None:
+        workfile_entities = list(ayon_api.get_workfiles_info(
+            project_name,
+            task_ids=[task_id],
+        ))
+
+    if platform.system().lower() == "windows":
+        workfile_path = workfile_path.replace("\\", "/")
+    workfile_path = workfile_path.lower()
```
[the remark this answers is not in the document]
iLLiCiTiT: Even though you're right for linux, from pipeline point of view it would cause a lot of other issues. This is used to find matching workfile entity and from that point of view "auto-fix" change of case it always better option from UX point of view. If you've changed `sh0010` to `SH0010` and you're on linux, you're probably screwed on multiple other levels that are more harmfull.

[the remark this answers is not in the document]
iLLiCiTiT: BTW created issue for unique folder names https://github.com/ynput/ayon-backend/issues/642

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -278,64 +280,185 @@ def get_last_workfile_with_version(
...
+
+    """
+    dotted_extensions = set()
+    for ext in extensions:
+        if not ext.startswith("."):
+            ext = f".{ext}"
+        dotted_extensions.add(ext)
+    filtered_paths = [
+        os.path.join(dirpath, filename)
+        for filename in os.listdir(dirpath)
```
MustafaJafar, line 367: what if `dirpath` doesn't exist yet? 
I stumbled on this error while testing the alab project setup.

```python
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\workfile\path_resolving.py", line 367, in _filter_dir_files_by_ext
    for filename in os.listdir(dirpath)
                    ^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'H:\\AYON\\projects/AY_ALAB_Demo/assets/chemistry_beaker03/work/lookdev'
```

BigRoy (commented):
Trying to give this a go I found that:

It's weird that it's 'this hard' to just save to the current context a new version and have the author set. I seem to be required to first query the current folder entity and task entity for this? I'd have expected I could just `save` a new file and still have it save the file as a new version with the _author_ metadata. I'm still a bit surprised that `save_file` wouldn't just assume you'd be saving into the current context by default.

I actually think `host.save_file(filepath, workfile_info)` would've made more sense, where you could attach for example:
```
metadata = WorkfileInfo(
    description="Hello, this is the artist note.",
    author="userx"
)
```
Where if e.g. author is not passed, it defaults to the current user?

I'm also still a bit confused by the fact that we can pass `version` and `comment` metadata. Because it seems weird I can pass it an _arbitrary_ version and completely take control over the filename, hence not even matching the version info. The same goes for `comment` which is the "in-filename" comment string - which seem weird to pass a different comment for than what is actually imprinted in the filename. I wonder if that `comment` and `version` should even be arguments for metadata to begin with?

Also, I have my doubts about the `OptionalWorkfileData` design. It's quite unclear to me whether I _can_ initialize that with only _some_ of its arguments, so that the rest still gets auto-populated?

I changed the maya increment plug-in to:
```python
import pyblish.api

from ayon_core.lib import version_up
from ayon_core.host import IWorkfileHost
from ayon_core.host.interfaces import SaveWorkfileOptionalData
from ayon_maya.api import plugin


class IncrementCurrentFileDeadline(plugin.MayaContextPlugin):
    """Increment the current file.

    Saves the current maya scene with an increased version number.

    """

    label = "Increment current file"
    order = pyblish.api.IntegratorOrder + 9.0
    families = ["workfile"]
    optional = True
    targets = ["local"]

    def process(self, context):
        current_filepath = context.data["currentFile"]
        new_filepath = version_up(current_filepath)

        host: IWorkfileHost = registered_host()
        host.save_workfile_with_context(
            filepath=new_filepath,
            folder_entity=context.data["folderEntity"],
            task_entity=context.data["taskEntity"],
            description="Incremented by publishing.",
            # Optimize the save by not reducing needed queries for context
            prepared_data=SaveWorkfileOptionalData(
                project_entity=context.data["projectEntity"],
                project_settings=context.data["project_settings"],
                anatomy=context.data["anatomy"],
            )
        )
```
That seemed to work. But admittedly - all I'm _really trying to do_ is save a new workfile version + set a description. Feels a bit complex/verbose for an API.

Other than that - it does seem to work, but it's far from a user friendly API in my eyes.

Also, it's a bit odd that `get_current_context` doesn't proivde results you can use for e.g. `set_current_context` even though they seem 1:1 related? 

And furthermore, the `before_context_change` and `after_context_change` methods - I'm unsure why that isn't just an emitted event that the host integration can install callbacks against. Why does this need the methods on the Host class itself?

For a PR that designs an API it would also go miles if the PR description can please have some good use case examples of the API. Like examples for:
- Increment current scene file with a custom description.
- Save file into another asset.

I think once you'll start making the examples you'll see that it may make more sense to have the generalized "save" API and its higher level method for it to not take `filepath` at all? But have it compute it based on templates, etc?

I'd expect e.g.
```python
host.save_versioned_workfile(version=15, description="Fixed the assets", comment="automatic")
host.save_workfile(filepath, description="Saving explicitly")
```

With each taking optional context parameters, like:
```python
# Providing no `version` argument will increment +1 from latest?
host.save_versioned_workfile(description="Hello world", task_entity=task, folder_entity=folder)
```

BigRoy:
> To be honest I never liked that it is actually done with events as there is no window for host to affect the logic in any way. But the logic is sometimes mandatory to happen in sequence and to e.g. not allow save in certain context, or ask user for some information, or store context metadata to the file metadata etc. **That is just impossible to achieve with events.**

Maya uses an event callback system for saving, opening, etc. and in the pre-* events, like pre-save you can set a value on the event (essentially returning a true-false state) whether the save should continue or not. As such, an event callback system _can_ definitely be made to interrupt and stop the current process if needed.

The same for having the events registered in an ordered manner - that's totally possible and we do have an ordered event system inside AYON too.

> The main reason is that we should always create workfile entity when we work with workfiles, to have infromation about workfiles on server, even if you save them out of workfiles too.

If this is the case - we should strive to make it **absolutely clear** that for workfiles, `save_file` should **NOT** be called and all existing callers of that method (of which I suspect _all_ are about workfiles) should hence be refactored, similar to: https://github.com/ynput/ayon-maya/pull/311 - However, I still see _no reason_ why the replacement method requires more arguments. We know sufficiently about the current context to always compute them from the current context, including the user/author, etc.

> We can change save_current_workfile_to to expect folder path and task name for example

I think just having more arguments is worse, no? I think perhaps it should just not _require_ folder entity and task entity, and if not passed assume to keep the current context.

> There are no OptionalWorkfileData? I guess you mean ...OptionalData classes? All of them are optional each passed argument would make the logic faster tho.

For example, `SaveWorkfileOptionalData`, my IDE seems to be having a really hard time presenting me with the input arguments needed.

image

It does seem that I'm able to pass only partial info to the class - however, the fact that the info I get from my IDE about what arguments it takes I'd say this is worse now than having additional arguments on the method itself.

> version and comment are metadata about the file stored to workfile entity. The logic really can't parse that information from the path, and the same path does not have to follow our templates, so the only information about the metadata can come from the place which created the path.

Then I think we're missing the method that instead takes the individual components, like context, version and comment and constructs that filepath and saves it?

> Method save_file is existing method that can't be easily changed to do more stuff. This must be backwards compatible to some degree and changing save_file is not an option from my point of view (first argument is optional).

I'd say that due to how this negatively impact the simplicity of the API - it may have been better to monkeypatch or in another way work around existing subclasses of the host interface so that we can just keep simple `host.save_workfile(path)` working, but still register with an author. The _only_ reason I can see to NOT do that is if we want to be able to save a file from a host integration **without** workfile info (e.g. save a temporary file) with a unified host API.

Anyway, at this point it feels like a hard debate to continue - but I just want to remark again that I think `host.save_workfile_with_context` (or whatever we'll end up naming it) should be able to just save into my current context without requiring the context arguments - it should just default to my current context?

BigRoy:
> Even system is not written that way, and it would make sense only if the host would be the only one who is listening to events. Still the logic should be part of host, not as events handling.

Actually - I'd argue the opposite actually. The "workfile lock" profiles could e.g. be completely build using events, from another addon - if it were implemented like Maya does it. With how it's now up to each host addon - there's no way for an addon to customize this behavior from another addon + they'd need to make DCC specific changes instead.

> Again, that;s what helper functions are for. The implementation of what it does on host class, must be 100% defined this way.

I see - fine with me. Just saying, from an API standpoint I'd expect those helps to be available on the host itself, like `host.save()` instead of `save_file_current_context(host, version=1)` since it looks off.

> That would force to use the workfile template, which is also something that is not necessary. I do understand what are you trying to say, but you can have automation that is e.g. creating 2 different workfiles without using workfile path template. It has use-cases when you don't have to fill it. Again, can be simplified to create helper functions, when you'll give me reasonable usecase I'll do it.

Correct - I'm saying the low-level `save` functionality we have now is fine. But admittedly from an API standpoint, I think `save_next_version()` is something 99% of the use cases would most likely be... especially looking at our own codebase.

Anyway, I figured the API was to _make it easy to work with workfiles_. Like define a nice dev-facing interface to interact with the workfiles in a given host. Now it seems heavily focused on being the "technical interface" that _makes it work_ - which is fine in a way, but still means you'd end up seeing lots of boilerplate code around all the usage everywhere because the majority of the use cases will save e.g. a new version in current context or save a specific version in another context. If it wouldn't need to deal with how the filepath looks like, I can imagine that actually being much easier to use for what covers the most prominent use cases of this API.

It's like if `host.load_container` would need to take tons of additional arguments, even though in the majority of times I just want to load a particular representation with a particular loader. Yes - under the hood there may be much lower level functionality, but I still think most of the use cases want that simple entry point.

This API is lacking in a similar manner as that the publishing API currently is. E.g. you'd expect you could just do `publish()` on e.g. the `CreateContext` or whatever interface we have for the publishing logic - but we do not have an API that does so, so we need to push it through `pyblish` iterators, etc. meaning everyone will still need to go low-level. I just thought this PR would include that higher level (and more importantly, clear examples on how to use it - which seem lacking.)

MustafaJafar (commented):
Faced an error. 
I'ce create a new folder in my project but I never launched in any app in that context from the launcher, so the workfile directory is not created yet. 

when using the workfile tool to save my file in another context where its workfile directory isn't created yet on disk, I get this error: 

```python
==============================
ERROR: Cannot open file H:\AYON\projects\Experiments\assets\suzanne\work\modeling\Expr_suzanne_modeling_v001.blend@ for writing: No such file or directory

==============================
  File "E:\Ynput\ayon-core\client\ayon_core\tools\workfiles\models\workfiles.py", line 169, in save_as_workfile
    self._host.save_workfile_with_context(
  File "E:\Ynput\ayon-core\client\ayon_core\host\interfaces\workfiles.py", line 915, in save_workfile_with_context
    self.save_workfile(filepath)
  File "E:\Ynput\ayon-blender\client\ayon_blender\api\pipeline.py", line 95, in save_workfile
    save_file(dst_path if dst_path else bpy.data.filepath)
  File "E:\Ynput\ayon-blender\client\ayon_blender\api\workio.py", line 49, in save_file
    result = bpy.ops.wm.save_as_mainfile(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Blender Foundation\Blender 4.2\4.2\scripts\modules\bpy\ops.py", line 109, in __call__
    ret = _op_call(self.idname_py(), kw)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeERROR: ERROR: Cannot open file H:\AYON\projects\Experiments\assets\suzanne\work\modeling\Expr_suzanne_modeling_v001.blend@ for writing: No such file or directory
```

antirotor:
One thing to add there: it would help if we could define here the list of helper/higher-level API functions that would help developers.

client/ayon_core/pipeline/workfile/path_resolving.py
iLLiCiTiT: Fixed

iLLiCiTiT:
> I agree and some of it was part of my initial objections (about the need of so many arguments, etc.) I consider this a low-level API that would really benefit from what we are now calling helper functions but from my perspective are just higher-level API that will be probably used in the world out there more than the low level?

To avoid confusion, what do you mean? Because I'm now confused AF. Can you tell what should happen where? What exactly, no abstract concepts, exact logical steps.

How we want to make sure that we do handle stuff that is considered as mandatory from now on. There are multiple requests that are forcing us to "make sure all workfiles have entry in AYON", "know which host created which workfiles (and possibly which version in future)" etc. Because I have no idea how that would be possible if it would not happen on host without breaking compatibility of every host. The order in which the stuff is happening is also super important BTW.

> With the workfile locking - I think the AYON event system could be definitely used for that, maybe better than the file locks. The only real advantage of file locks is that anyone without pipeline but with access to filesystem knows that the file is opened (but that's IMO edge case we shouldn't really consider). If the event handler can stop scene opening in the host or not is up to the implementation and capabilities of individual hosts - but even if it cannot interupt loading, just displaying generic warning dialog is enough. The only thing I am missing from events is TTL that would automatically clear the lock after certain time. We are already emitting event in case of opening or not? For time tracking/etc.

What? Current events logic is there to tell "hey this happened", that's it. What you described is completelly of the scope of its and our capabalities (or at least mine). I have no idea how that would be even partially possible, especially if it would not be part of host logic.

> One thing to add there: it would help if we could define here the list of helper/higher-level API functions that would help developers.

Currently there are 2 `save_current_workfile_to` and `save_workfile_with_current_context`. As I wrote before, I don't have more use-cases that would benefit from having helper function, if you have more use-cases please write it here. Only reasonable use-case since then was `save_next_version` which would be a "guess" helper function as it has to parse the filename to get version and comment. But other than that now other use-case came out.

## Dev note
Just to give you an idea what was reason of the workfile api need: "We have to be able to do, what happens in workfiles tool, but from code so it is not tight to workfiles tool." What is now implemented here, is moving all the code from workfiles tool to host. I did few additional changes to make it a little bit better (e.g. change of context), but other than that, this is what happens in workfiles tool.

I've tried to implement all the stuff you're telling me that should not happen in host outside, that was first week I did spend on this PR, and I was not able to do it. At the same the the API for workfiles handling would be even more confusing, and it would not solve the issues we have in hosts like harmony or silhouette. Current state is the best I can do with all requirements I know I have to be able to achieve or be prepared to achieve.

philippe-ynput:
@iLLiCiTiT @BigRoy @antirotor @MustafaJafar @kalisp  Please, let's merge this ASAP. 
This is blocking other tasks and even though it's not as pretty as some would like, it does the job. Feel free to open PRs for clever helper functions, etc.
Thanks

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -207,12 +214,94 @@ def get_workdir(
...
+        for pattern, replacement in (
+            # Replace `.{ext}` with `{ext}` so we are sure dot is not at the end
+            (r"\.?{ext}", ext_replacement),
+        ):
+            file_template = re.sub(pattern, replacement, file_template)
+
+        file_template = StringTemplate(file_template)
+        comment_template = re.escape(str(file_template.format_strict(data)))
+        data.pop("comment")
+        file_template = re.escape(str(file_template.format_strict(data)))
```
BigRoy, line 254: What's this doing? Why pop the comment? Bit confused here.

client/ayon_core/pipeline/workfile/path_resolving.py
```diff
@@ -207,12 +214,94 @@ def get_workdir(
...
-    workdir, file_template, fill_data, extensions
-):
+@dataclass
+class WorkfileParsedData:
+    version: Optional[int] = None
+    comment: Optional[str] = None
+    ext: Optional[str] = None
+
+
+class WorkfileDataParser:
```
BigRoy, line 224: Just wanted to mention that something along these lines does exist in the workfiles tool logic. Or at least, e.g. getting the "subversion" or "comment" from the filename.

It's around here: https://github.com/ynput/ayon-core/blob/b0ac87b7b12006dc8b3fffe441b91a884cc65ce5/client/ayon_core/tools/workfiles/models/workfiles.py#L360-L366
Which uses this [`CommentMatcher`](https://github.com/ynput/ayon-core/blob/b0ac87b7b12006dc8b3fffe441b91a884cc65ce5/client/ayon_core/tools/workfiles/models/workfiles.py#L31-L83).

I wonder if this should get aligned more and maybe have the code there use this? or? seems like a lot of overlap.

pyproject.toml
```diff
@@ -19,6 +19,7 @@ python = ">=3.9.1,<3.10"
 pytest = "^8.0"
 pytest-print = "^1.0"
 ayon-python-api = "^1.0"
+arrow = "0.17.0"
```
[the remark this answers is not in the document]
iLLiCiTiT: Yes, arrow is already part of dependencies because it is dependency of a dependency, I just added it explicitly.

There was a comment saying that, but disappeared.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1792 @@
...
+            _data = workfile_entity["data"]
+            version = _data.get("version")
+            comment = _data.get("comment")
+            if version is None:
+                filename = os.path.basename(rootless_path)
+                parsed_data = data_parser.parse_data(filename)
+                version = parsed_data.version
+                comment = parsed_data.comment
+
+            filepath = prepared_data.anatomy.fill_root(rootless_path)
```
MustafaJafar, line 1157: `prepared_data` is optional in the `list_workfiles`.
However, this line raise an error if I didn't pass `prepared_data`.
```python
  File "E:\Ynput\ayon-core\client\ayon_core\host\interfaces\workfiles.py", line 1157, in list_workfiles
    filepath = prepared_data.anatomy.fill_root(rootless_path)
               ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'anatomy'
```

MustafaJafar: this error happens as well in a different method.

```python
  File "E:\Ynput\ayon-core\client\ayon_core\host\interfaces\workfiles.py", line 1206, in list_published_workfiles
    for version_entity in prepared_data.version_entities
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'version_entities
```

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -0,0 +1,1792 @@
...
+            items.append(workfile_item)
+
+        return items
+
+    def copy_workfile(
+        self,
+        src_path: str,
+        dst_path: str,
+        folder_entity: dict[str, Any],
+        task_entity: dict[str, Any],
```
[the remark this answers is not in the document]
BigRoy: > but, now I've to figure out the path myself.

according to earlier comments by @iLLiCiTiT that would be 'helper' functionality to define the path?

client/ayon_core/host/interfaces/workfiles.py
iLLiCiTiT: fixed

iLLiCiTiT:
> The workfile is greyed out.

That means there is a workfile entity but you don't have the file.

client/ayon_core/pipeline/workfile/path_resolving.py
iLLiCiTiT: > Which uses this CommentMatcher.

`CommentMatcher` is not used anymore, I've marked it as deprecated.

client/ayon_core/pipeline/workfile/path_resolving.py
iLLiCiTiT: Added comment to the code, not sure if it helps?

MustafaJafar:
> That means there is a workfile entity but you don't have the file. Which is suprisingly a requested feature, if you have e.g. remote users, they don't all create the same versions.

But the file exists.  and I launcher is able to open it if this greyed workfile is the latest.

MustafaJafar:
Still works https://github.com/ynput/ayon-core/pull/1275/commits/6b452e49291db234a0bc8d37105baf4c5571139e and I can no longer reproduce the issue mentioned above https://github.com/ynput/ayon-core/pull/1275#issuecomment-3052671313

---

Here's another issue. 
So, when copying/saving the files manually (outside of workfiles tool), the workfiles tool works properly and shows the availbale workfiles. and when deleting these workfiles (also manually), it disappears from the workfiles list. (and this behaviour is what I'm used to, becasue I clean up my workspace occasionally)

on contrast, when saving a new version using the workfiles tool, once again it shows propely within workfiles list in the workfiles tool. and when deleting that workfile (manually), it appears as greyed out in the workfiles list. 

image

I've to admit, I think this is finally working propely as I can see these workfiles as expected on AYON. 
But, I wonder how can I clean up my workspace ? anyways I think this can be done in later PRs either in ayon-core or on the front end side of AYON itself.

image

iLLiCiTiT:
> Shouldn't these workfiles be registered if they are actually available on the AYON server where there should be a button to retrieve this deleted file?

I don't understand this question. You can't retrieve the workfile, because server only knows that the workfile exists/existed, doesn't have the file itself (it does not have to be file).

client/ayon_core/host/interfaces/workfiles.py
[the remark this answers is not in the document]
iLLiCiTiT: Is this what you've asked for https://github.com/ynput/ayon-core/pull/1275/commits/260aad8c1fb19db519f8f64814b6e20030b36b1b ?

BigRoy (commented):
Seems to work as intended - still some remarks, but I think we're at a point where we'll just not be solving the API design itself in a few commits. I'll leave the API design notes to @philippe-ynput because I feel he's done in control of that power anyway.

PR #1276, opened 2025-05-16
Fixes frame string formatting for transcoding

jakubjezek001:
## Changelog Description
Ensures the frame string used for transcoding includes the correct padding,
based on the collection's padding attribute. This resolves issues where
the output file sequence name was not correctly formatted, leading to
transcoding failures.

## Additional info
I had found out about the issue during batchdelivery tests

## Testing notes:
1. only able to test this on higher then 4padding image sequence

### DIFF
--- a/client/ayon_core/plugins/publish/extract_color_transcode.py
+++ b/client/ayon_core/plugins/publish/extract_color_transcode.py
@@ -283,7 +283,11 @@ class ExtractOIIOTranscode(publish.Extractor):
             if collection.holes().indexes:
                 return files_to_convert
 
-            frame_str = "{}-{}#".format(frames[0], frames[-1])
+            # Get the padding from the collection
+            # This is the number of digits used in the frame numbers
+            padding = collection.padding
+
+            frame_str = "{}-{}%0{}d".format(frames[0], frames[-1], padding)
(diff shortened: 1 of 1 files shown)


BigRoy:
Isn't this exactly what the `--framepadding` argument is for in `oiiotool`. 

I understand it's optional, but since it's a simple override optional why not just use that?

See: https://openimageio.readthedocs.io/en/latest/oiiotool.html#frame-sequences

rdelillo (approved):
Approving this PR as this code works and fixes the issue.
If that's a urgent/critical bugfix, I believe this patch is good enough.

Now as flagged, there is a specific `--framepadding` argument that could be used here.
Just wanted to add that from my tests, `--framepadding` needs to be there along `--frames`.

```shell
# Works
oiiotool.exe --info -v -i:infoformat=xml C:/Users/robin/output.948845-948869%07.exr

# Works
oiiotool.exe --info -v -i:infoformat=xml C:/Users/robin/output.#.exr --frames 948845-948869 --framepadding 7

# Does NOT work
oiiotool.exe --info -v -i:infoformat=xml C:/Users/robin/output.948845-948869.exr --framepadding 7
````

PR #1281, opened 2025-05-21
Publisher: Set minimum amount of characters required for a comment before publishing is allowed

BigRoy:
## Changelog Description

Add setting to define a minimum amount of characters required for a comment on every publish through the publisher UI before being allowed to trigger the publish.

## Additional info

image

image

I'm pretty sure someone will start to hate me some day.

## Testing notes:

1. Upload package
2. Set minimum required characters
3. Try to publish with fewer characters in the comment

### DIFF
--- a/client/ayon_core/tools/publisher/window.py
+++ b/client/ayon_core/tools/publisher/window.py
@@ -245,6 +245,12 @@ class PublisherWindow(QtWidgets.QDialog):
         show_timer.setInterval(1)
         show_timer.timeout.connect(self._on_show_timer)
 
+        comment_invalid_timer = QtCore.QTimer()
+        comment_invalid_timer.setInterval(2500)
+        comment_invalid_timer.timeout.connect(
+            self._on_comment_invalid_timeout
+        )
+
         errors_dialog_message_timer = QtCore.QTimer()
         errors_dialog_message_timer.setInterval(100)
         errors_dialog_message_timer.timeout.connect(
@@ -395,6 +401,7 @@ class PublisherWindow(QtWidgets.QDialog):
         self._app_event_listener_installed = False
 
         self._show_timer = show_timer
+        self._comment_invalid_timer = comment_invalid_timer
         self._show_counter = 0
         self._window_is_visible = False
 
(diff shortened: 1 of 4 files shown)


antirotor (approved):
It works fine. Just stupid little comment about the validation itself - if you set the minimum count to 5 and then just press space 5 times, you'll pass :) I understand that at that point it's pure sabotage. Just saying that maybe some further validation might be added.

antirotor:
> I could do a `comment.strip()` at least?

That is probably good idea. Even for stripping blanc characters at the end of the string.

client/ayon_core/tools/publisher/window.py
```diff
@@ -823,15 +830,44 @@ def _set_publish_comment(self):
...
 
     def _on_publish_clicked(self):
-        if self._save_changes(False):
+        if self._start_publish_preflight() and self._save_changes(False):
             self._set_publish_comment()
             self._controller.publish()
 
+    def _start_publish_preflight(self) -> bool:
+        # Validate comment length
+        comment_def = self._controller.get_comment_def()
```
iLLiCiTiT, line 844: You can cache the value in `_on_controller_reset`.

client/ayon_core/tools/publisher/window.py
```diff
@@ -823,15 +830,44 @@ def _set_publish_comment(self):
...
             self._set_publish_comment()
             self._controller.validate()
 
     def _on_publish_clicked(self):
-        if self._save_changes(False):
+        if self._start_publish_preflight() and self._save_changes(False):
             self._set_publish_comment()
             self._controller.publish()
 
+    def _start_publish_preflight(self) -> bool:
```
iLLiCiTiT, line 842: Would rename this to `_validate_comment`.

client/ayon_core/tools/publisher/window.py
```diff
@@ -823,15 +830,44 @@ def _set_publish_comment(self):
...
+            char_count = len(self._comment_input.text())
+            if char_count < comment_def.minimum_chars_required:
+                self._overlay_object.add_message(
+                    "Please enter a comment of at least "
+                    f"{comment_def.minimum_chars_required} characters",
+                    message_type="error"
+                )
+                self._invalidate_comment_field()
+                return False
+        return True
```
iLLiCiTiT, line 855: No need to optimise str lenght calculation.
```suggestion
        char_count = len(self._comment_input.text())
        if (
            comment_def.minimum_chars_required
            and char_count < comment_def.minimum_chars_required
        ):
            self._overlay_object.add_message(
                "Please enter a comment of at least "
                f"{comment_def.minimum_chars_required} characters",
                message_type="error"
            )
            self._invalidate_comment_field()
            return False
        return True
```

client/ayon_core/tools/publisher/window.py
```diff
@@ -823,15 +830,44 @@ def _set_publish_comment(self):
...
+
+    def _invalidate_comment_field(self):
+        self._comment_invalid_timer.start()
+        self._comment_input.setStyleSheet(
+            "border-style: solid;"
+            "border-width: 1px;"
```
iLLiCiTiT, line 861: This is just guess, but I think it already has set 1px and solid.
```suggestion
```

client/ayon_core/tools/publisher/window.py
BigRoy: > You can cache the value in _on_controller_reset.

I could - but then any performance hit would already happen on reset - which usually is already the slow part ;) so I wanted to avoid that. But we could 'cache' it on first query and clear cache on reset. However, the "time taken" for the response is literally very fast (because the settings used are coming from the create context cache). So it'd be a micro-optimization currently.

Can we avoid it until it's an actual issue?

client/ayon_core/tools/publisher/window.py
```diff
@@ -245,6 +245,12 @@ def __init__(
         show_timer.setInterval(1)
         show_timer.timeout.connect(self._on_show_timer)
 
+        comment_invalid_timer = QtCore.QTimer()
+        comment_invalid_timer.setInterval(2500)
```
iLLiCiTiT, line 249:
```suggestion
        comment_invalid_timer.setSingleShot(True)
        comment_invalid_timer.setInterval(2500)
```

client/ayon_core/tools/publisher/window.py
```diff
@@ -823,15 +830,42 @@ def _set_publish_comment(self):
...
+            )
+            self._invalidate_comment_field()
+            return False
+        return True
+
+    def _invalidate_comment_field(self):
+        self._comment_invalid_timer.start()
+        self._comment_input.setStyleSheet("border-color: #DD2020")
+        # Set focus so user can start typing and is pointed towards the field
+        self._comment_input.setFocus()
```
iLLiCiTiT, line 863:
```suggestion
        self._comment_input.setFocus(QtCore.Qt.MouseFocusReason)
```

BigRoy: It works fine for me already. Seems odd to set MouseFocusReason - I'd feel like a liar! :)
Can you confirm it fails on your end, or?

iLLiCiTiT: > I'd feel like a liar! :)

Well, yes 🙂 

> Can you confirm it fails on your end, or?

It works, but the cursor position is not at the end.
```python
        self._comment_input.setCursorPosition(
            len(self._comment_input.text())
        )
```

iLLiCiTiT (approved):

PR #1282, opened 2025-05-22
Refactor: Hacky render frames/handles management caused by Harmony

Tilix4:
## Changelog Description
Refactor of hacky render frames/handles management due to AYON's Harmony implementation.

## Additional info
Dropped `ignoreFrameHandleCheck` which was a exception for Harmony.

Made when working on but not required: https://github.com/ynput/ayon-harmony/pull/50

## Testing notes:
1. Regular render publish


iLLiCiTiT:
> In fact no it isn't, you're right. I've tested it and each change completes without the other one being required. You can merge and release without caring about it.

So merging this PR won't break older harmony addon versions?

Tilix4:
> So merging this PR won't break older harmony addon versions?

That's what I ended up with my tests, but you cannot rely only on my words, another person has to give it a try.

iLLiCiTiT:
@kalisp can you validate this works with latest harmony changes?

PR #1292, opened 2025-05-27
Templated Workfile Build: support building from an AYON Entity URI instead of filepath

### LINKED ISSUE #742: Enhancement: Workfile template builder using publishes instead of direct filepath

The workfile template builder currently uses just some scene files saved in a specific folder for a particular host for which the path is set in the Studio Settings. I would like to propose to refactor this functionality into using published workfile templates instead.

**Pros**

- Workfile templates are versioned and tracked
- Workfile templates could then sync automatically via site-sync
- Workfile templates could also easily be "loaded manually" to test a specific workfile template.
- Additional studio-specific validators could be deployed upon the generation of workfile templates.
- The 'loaded' workfile template could containerise into the workfile as a 'loaded container' and thus a publish from the workfile can then also report which template was used to build that workfile due to "input links"

**Cons**

- It'd require a publish instead of just a "drop a file there" which might be trickier to use a single template across multiple projects (unless maybe it'd be possible to load from another project like library loader too)

In settings instead of defining a path to the file to load you'd instead pass it the `asset > subset` to load for particular profile.

---

Note that the API for the workfile builder should still have an interface through which you could import a particular template `builder.import_template` and also build the current scene `builder.populate_scene_placeholders`. It's only the automatic "build workfile for current task" feature which would then find the relevant publish and load it from the filepath. 

_I'd even argue that the current logic is already failing there since you can't just build the current scene in maya using `builder.populate_scene_placeholders` but require the `import_path` to have been triggered so a `PLACEHOLDER_SET` exists. The only workaround for that currently is:_

```python
from openpype.hosts.maya.api.workfile_template_builder import MayaTemplateBuilder, PLACEHOLDER_SET
from openpype.pipeline import registered_host

host = registered_host()
builder = MayaTemplateBuilder(host)
if not cmds.objExists(PLACEHOLDER_SET):
    cmds.sets(name=PLACEHOLDER_SET, empty=True)
builder.populate_scene_placeholders(keep_placeholders=False)
```

An alternative would be to, e.g. for maya, publish a maya scene and have the published path set in the studio settings. However it'd disable some of the cons of not loading an actual representation directly (e.g. site sync, etc).

### Additional context:

_No response_

[cuID:[OP-6162](https://app.clickup.com/t/861mxk3ug)]

BigRoy:
## Changelog Description

Allow Templated Workfile Build support building from an AYON Entity URI instead of filepath or templated filepath.

## Additional info

Fix #742 

image

Example URIs:
```
ayon+entity://test/apps/maya?product=mayaSceneMain&version=latest&representation=ma
ayon://test/apps/maya?product=mayaSceneMain&version=latest&representation=ma
ayon+entity://template_library/workfiles/maya?product=workfileModeling&version=latest&representation=ma
ayon://template_library/workfiles/maya?product=workfileModeling&version=latest&representation=ma
```

Easiest way I know to get an entity URI currently is to go through AYON frontend browser, and copy the top URL when selecting a version.

image

Gives me:
```
ayon+entity://CB_sandbox/asset/lookdev_kit?product=mayaSceneLookdev_kit_mtoa&version=v003
```
I just replace the version with `latest` and attach the representation to load `&representation=ma` to make:
```
ayon+entity://CB_sandbox/asset/lookdev_kit?product=mayaSceneLookdev_kit_mtoa&version=latest&representation=ma
```

## Testing notes:

1. Templated workfile build should continue to work in supported hosts:
    - [ ] houdini
    - [x] maya
    - [ ] nuke
    - [ ] after effects
    - [ ] silhouette
2. Templated workfile build should now work using an ayon entity URI


client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -835,103 +857,136 @@ def get_template_preset(self):
...
+                )
+            return resolved_path
+
+        # If the path is set and it's found on disk, return it directly
+        if path and os.path.exists(path):
+            return path
+
+        # Otherwise assume a path with template keys, we do a very mundane
+        # check whether `{` or `<` is present in the path.
+        if "{" in path or "<" in path:
```
iLLiCiTiT, line 940: NOTEs:
- Before this change it didn't carsh.
- This condition should check only for `"{"` and revert the logic -> return `path` if there is no `"{"` and fill it if it's not there.

BigRoy: Ah you mean - not to raise an error.

The error was previously raised here: https://github.com/ynput/ayon-core/pull/1292/files#diff-96b9bf4cef8323707aa5d129a3c9a93849242fd5b3c7c71ce0d2931f3fa3a46cL902

BigRoy: Changed with https://github.com/ynput/ayon-core/pull/1292/commits/7b5ff1394c1684a9a1740ece36e5d343adf853b9

iLLiCiTiT: We still should rever the condition to first return path, and to check only for `{` (there is no reason to check for `<`).

BigRoy: > (there is no reason to check for <).

Aren't those "optional keys" in an anatomy template, or?

iLLiCiTiT: They're not if `{` is not between them.

- `<{folder[name]}>` -> optional
- `<br>` -> not optional

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -835,102 +861,124 @@ def get_template_preset(self):
...
+                path = solved_path
+
+            solved_path = os.path.normpath(solved_path)
+            if not os.path.exists(solved_path):
+                return path
+
+            result = StringTemplate.format_template(path, fill_data)
+            if result.solved:
+                path = result.normalized()
+            return path
```
BigRoy, line 980: @iLLiCiTiT looking at the old code I have a feeling this should be above the `while True:` part... however, I'm not entirely sure why the recursive `anatomy.path_remapper` calls are needed. Do you know? If the string template is filled - why would it still need that recursive resolving?

Am I having a bad morning and am I just not seeing it? 🤯

BigRoy: Ah! The fact that it maps full path to rootless makes sense - that als means the format template should indeed be after it.

However, does that mean that the:
```python
        if "{" not in path:
            return path
```
Should also come after the `path_remapper` because it may introduce root keys again, etc?

LiborBatek (approved):
I have tested it with ```maya``` host and all seems working fine. I have also tested if URI works with latest version so I have produced new template publish and build from it...all seems working nicely as the latest file template was used then...

image

```ayon+entity://Ayon_Tests/maya/templates/template_lookdev?product=mayaSceneMain&version=latest&representation=ma```

If I have just used the URI pasted from the webUI like e.g the one below, it didnt worked straight away. Not sure if it should work without defining the representation, probably not ...so all good I guess.

```ayon+entity://Ayon_Tests/maya/templates/template_lookdev?product=mayaSceneMain&version=v002```

BigRoy:
@iLLiCiTiT ready to merge?

iLLiCiTiT (approved):

Issue #1289, opened 2025-05-23
Add validator in folder creation when provided resolution does not match existing one.


antirotor:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

This PR https://github.com/ynput/ayon-core/pull/1255 added ability to override the resolution settings. But that needs to be done in the settings. You can either not-override or select the override.

There should be third mode - "use source resolution" and that will override the resolution using resolution data coming directly from the source. It might be that this cannot be implemented in ayon-core and must be done in hosts themselves and if so, we should think how to combine functionality so it doesn't live in two places.

User story:
Artist wants to use whatever resolution is on the input. For example in Nuke, you want to use whatever is set on Read to be used on Write.

### How would you imagine the implementation of the feature?

When option on `Choose Explicit Resolution` is set to *use source resolution*, that one should be used. Maybe unified collector collecting instance source resolution can be used.

Maybe we should also change the plugin name because using source resolution is sort of implicit override?

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

BigRoy:
I'm so lost here. Isn't using the source resolution the default behavior? I can throw whatever into tray publisher or from any DCC at any resolution (regardless of how my shot or task is set up) and have it output the resolution I want. (I might need to disable an optional resolution validation, but still)

So what's different here?

JoseCovaOS:
Hi guys, trying to shed some light on this topic as it's a request from us (Orca). 

So the issue the editorial department is facing is that they want to be able to set the ftrack attributes **Resolution Width** and **Resolution Height**  of the Shot they create when publishing a plate to  be the resolution of that plate (the main one, other plates shouldn't be considered). 

This is because in some projects, we've had to deliver some EXR's to the same resolution as it's shot main plate.  What we were doing right now is: 
- Manually change this value in ftrack (shot level).
- Ask the comper to run Nuke action `set resolution`. 
- Comper renders (we've also disabled the validator for resolution so it doesn't fail).

It would be good if we could have a project setting that if enabled, it would gather resolution from the plateMain and during Hiero publish, set these attributes to the created shot.  Also, we would like to have a per instance override in case we do not need to do this for all the shots in a project but only to a few. 

Hope my explanation helps.

PR #1293, opened 2025-05-28
Implement ValidateFolderCreationResolution.

rdelillo:
## Changelog Description

resolve: #1289 

Implement a new validator that check existing shot resolution instead of blindly update.
image

If a mismatch is detected, 2 repair actions are suggested:
* Force update to new resolution
* Ignore new resolution and leave as-is

image

## Testing notes:
1. Try to re-create an existing shot with this option enabled (Resolve, Hiero, Flame or traypublisher)
2a. Ensure disabling the validator let the shot product to go through with no resolution changes
2b. Ensure enabling the validator does nothing if the resolution match
2c. Ensure the validator raises if the resolution mismatches with 2 suggested repairs:
    * force resolution update
    * ignore, which goes back to 2a

### DIFF
--- a/client/ayon_core/plugins/publish/validate_asset_docs.py
+++ b/client/ayon_core/plugins/publish/validate_asset_docs.py
@@ -1,5 +1,17 @@
 import pyblish.api
-from ayon_core.pipeline import PublishValidationError
+
+from ayon_api.entity_hub import EntityHub
+
+from ayon_core.lib import BoolDef
+from ayon_core.pipeline.publish import (
+    PublishValidationError,
+    RepairAction,
+    get_errored_instances_from_context,
+    AYONPyblishPluginMixin
+)
+
+
+_RESOLUTION_ATTRIBS = ("resolutionHeight", "resolutionWidth", "pixelAspect")
 
 
 class ValidateFolderEntities(pyblish.api.InstancePlugin):
@@ -37,3 +49,129 @@ class ValidateFolderEntities(pyblish.api.InstancePlugin):
                 "Instance \"{}\" doesn't have folder entity "
                 "set which is needed for publishing."
             ).format(instance.data["name"]))
+
+
+class RepairOverrideResolution(RepairAction):
+    """ Repair, force new resolution onto existing shot.
+    """
+    label = "Force new shot resolution."
+
+    def process(self, context, plugin):
+        entity_hub, entity, shot_data = ValidateFolderCreationResolution.get_shot_data(
+            context.data["hierarchyContext"]
+        )
+
+        for attrib in _RESOLUTION_ATTRIBS:
+            entity.attribs.set(attrib, shot_data[attrib])
+
+        entity_hub.commit_changes()
+
+
+class RepairIgnoreResolution(RepairAction):
+    """ Repair, disable resolution update in problematic instance(s).
+    """
+    label = "Do not update resolution."
+
+    def process(self, context, plugin):
+        create_context = context.data["create_context"]
+        for inst in get_errored_instances_from_context(context, plugin=plugin):
+            instance_id = inst.data.get("instance_id")
+            created_instance = create_context.get_instance_by_id(instance_id)
+            attr_values = created_instance.data["publish_attributes"].get(
+                "ValidateFolderCreationResolution", {})
+            attr_values["updateExistingFolderResolution"] = False
+
+        create_context.save_changes()
+
+
+class ValidateFolderCreationResolution(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+    """ Validate resolution values before updating an existing folder.
+    """
+
+    label = "Validate new folder resolution"
+    order = pyblish.api.ValidatorOrder
+    families = ["shot"]
+    actions = [RepairIgnoreResolution, RepairOverrideResolution]
+
+    @classmethod
+    def get_shot_data(self, hierarchy_context):
+        """ Retrieve matching entity and shot_data from hierarchy_context.
+        """
+        project_name = tuple(hierarchy_context.keys())[0]
+        entity_hub = EntityHub(project_name)
+        entity_data = {project_name: entity_hub.project_entity}
+        entity_to_inspect = [(entity_data, hierarchy_context)]
+
+        while entity_to_inspect:
+            entity_data, data = entity_to_inspect.pop(0)
+
+            for name, value in data.items():
+                entity = entity_data.get(name)
+                child = value.get("children", None)
+
+                if entity and child:
+                    entity_children = {child.name: child for child in entity.children}
+                    entity_to_inspect.append((entity_children, child))
+
+                # Destination shot already exists return for validation.
+                elif value.get("folder_type") == "Shot" and entity and not child:
+                    shot_data = value.get("attributes", {})
+                    return entity_hub, entity, shot_data
+
+        return None
+
+    def process(self, instance):
+        """ Validate existing shot resolution.
+        """
+        try:
+            hierarchy_context = instance.context.data["hierarchyContext"]
+
+        except KeyError:
+            self.log.info("No hierarchy context defined for shot instance.")
+            return
+
+        validation_data = self.get_shot_data(hierarchy_context)
+        if not validation_data:
+            self.log.info("Destination shot does not exist yet, nothing to validate.")
+            return
+
+        values = self.get_attr_values_from_data(instance.data)
+        _, entity, shot_data = validation_data
+
+        # Validate existing shot resolution is matching new one
+        # ask for confirmation instead of blind update, this prevents mistakes.
+        if values.get("updateExistingFolderResolution", True):
+            for resolution_attrib in _RESOLUTION_ATTRIBS:
+                shot_value = shot_data.get(resolution_attrib)
+                entity_value = entity.attribs.get(resolution_attrib)
+                if shot_value and shot_value != entity_value:
+                    raise PublishValidationError(
+                        "Resolution mismatch for shot."
+                        f"{resolution_attrib}={shot_value} but already existing "
+                        f"shot is set to {entity_value}."
+                    )
+
+        # If update existing shot is disabled, remove any resolution attribs.
+        else:
+            for resolution_attrib in _RESOLUTION_ATTRIBS:
+                shot_data.pop(resolution_attrib, None)
+
+            self.log.info(
+                "Ignore existing shot resolution validation "
+                "(update is disabled)."
+            )
+
+    @classmethod
+    def get_attr_defs_for_instance(
+        cls, create_context, instance,
+    ):
+        if instance.product_type not in cls.families:
+            return []
+
+        return [
+            BoolDef(
+                "updateExistingFolderResolution",
+                default=True,
+                label="Update existing shot resolution.",
+            ),
+        ]


client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+        cls, create_context, instance,
+    ):
+        if instance.product_type not in cls.families:
+            return []
+
+        return [
+            BoolDef(
+                "updateExistingFolderResolution",
+                default=True,
+                label="Update existing shot resolution.",
```
BigRoy, line 188: Avoid the dot on the label to match other attributes in the UI
```suggestion
                label="Update existing shot resolution",
```

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+                "Ignore existing shot resolution validation "
+                "(update is disabled)."
+            )
+
+    @classmethod
+    def get_attr_defs_for_instance(
+        cls, create_context, instance,
+    ):
+        if instance.product_type not in cls.families:
+            return []
```
BigRoy, line 182: You might want to use `cls.instance_matches_plugin_families(instance)` instead?

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+        except KeyError:
+            self.log.info("No hierarchy context defined for shot instance.")
+            return
+
+        validation_data = self.get_shot_data(hierarchy_context)
+        if not validation_data:
+            self.log.info(
+                "Destination shot does not exist yet, "
+                "nothing to validate."
+            )
```
BigRoy, line 148: An info level log is "artist-facing" in the publisher. If that's the case, this log should be _very clear_ what it's about and should be information they really want to know about. I'd say since this isn't that critical to know, this is better off as debug log.

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+        if values.get("updateExistingFolderResolution", True):
+            for resolution_attrib in _RESOLUTION_ATTRIBS:
+                shot_value = shot_data.get(resolution_attrib)
+                entity_value = entity.attribs.get(resolution_attrib)
+                if shot_value and shot_value != entity_value:
+                    raise PublishValidationError(
+                        "Resolution mismatch for shot."
+                        f"{resolution_attrib}={shot_value} but "
+                        f"already existing shot is set to {entity_value}."
+                    )
```
BigRoy, line 165: The downside to _raising_ here directly is that you may see that the Resolution Width is off, but even if Resolution Height is also off.. the user is not made aware of that. I'd instead log each invalid value. And then have the publish validation error be about "Resolution for X mismatches shot Y". The logs can then be `self.log.warning()` so they are still quite prominent to the user.

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+
+        # If update existing shot is disabled, remove any resolution attribs.
+        else:
+            for resolution_attrib in _RESOLUTION_ATTRIBS:
+                shot_data.pop(resolution_attrib, None)
+
+            self.log.info(
+                "Ignore existing shot resolution validation "
+                "(update is disabled)."
+            )
```
BigRoy, line 175: A validator should usually not **change** the data - it's really about validating, not by adjusting how the rest of publishing behaves. @iLLiCiTiT any pointers?

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+        return None
+
+    def process(self, instance):
+        """ Validate existing shot resolution.
+        """
+        try:
+            hierarchy_context = instance.context.data["hierarchyContext"]
+
+        except KeyError:
+            self.log.info("No hierarchy context defined for shot instance.")
```
BigRoy, line 140: So ... if this happens, what can we do about it? This sounds like an "oopsie" but the logged message doesn't really describe what is going on. Since it's an artist facing report log (`info` or higher) I think that's a bad thing currently.

This should likely be a debug log.

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+                        child.name: child
+                        for child in entity.children
+                    }
+                    entity_to_inspect.append((entity_children, child))
+
+                # Destination shot already exists return for validation.
+                elif (
+                    value.get("folder_type") == "Shot"
+                    and entity and not child
+                ):
```
BigRoy, line 127: Where does the requirement come from that it MUST be a Shot folder type? Also, even if a child already exists with the same name but different folder type - shouldn't we still consider it?

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+            for name, value in data.items():
+                entity = entity_data.get(name)
+                child = value.get("children", None)
+
+                if entity and child:
+                    entity_children = {
+                        child.name: child
+                        for child in entity.children
+                    }
+                    entity_to_inspect.append((entity_children, child))
```
BigRoy, line 121: Can you explain why it needs this traversal accross children? I'm a bit lost to what's this doing?

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,142 @@ def process(self, instance):
...
+
+            for name, value in data.items():
+                entity = entity_data.get(name)
+                child = value.get("children", None)
+
+                if entity and child:
+                    entity_children = {
+                        child.name: child
+                        for child in entity.children
+                    }
```
BigRoy, line 120: Isn't this `child` variable overriding/shadowing the outer `child = value.get("children", None)`. Seems unintentional?

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: Adjusted to debug.

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: good call, this is done.

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: I'm move this one to debug.
I'm personally not aware of any publisher that does not associated `hierarchyContext` with `shot` products, most of them are going through `collect_hierarchy` (Resolve, Flame, Hiero) and traypublisher is using its own logic.
If missing, then `extract_hierarchy_to_ayon` will do nothing, I just found it a bit dangerous to leave this one with a potential `KeyError`.

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: Totally, the `shot` instance can require to construct a whole hierarchy (parent folder, sequence then shot).
We rely on the `hierarchyContext` nested dict values to figure out the path to the shot to be created.

This code is doing a top-bottom parsing of the hierarchy dict value (technically from project level to the final shot) entity and check at each step if the folder exists or not. If the folder exist and it has children, then it needs to pursue its inspection deep down.

Hope that makes sense ?

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: I don't think so, but maybe I'm not following you. 
This validators only validate that the leaf shot folder within the hierarchy, if it already exists, matches current instance's resolution.

So to determine that we've reached the shot folder:
```python
  elif (
      value.get("folder_type") == "Shot"  # current folder is a shot
      and entity  # entity exists, so it already exists in AYON
      and not child  # no child, this is the leaf shot folder we are interested about
  ):
    # there is a validation against existing data to perform
```

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: Yeah, true !
I've rework this a bit as you suggested. Now each mismatch comes in its own warning log, then final error sum-up the info.

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,156 @@ def process(self, instance):
...
+        # Validate existing shot resolution is matching new one
+        # ask for confirmation instead of blind update, this prevents mistakes.
+        if values.get("updateExistingFolderResolution", True):
+            similar = True
+            for resolution_attrib in _RESOLUTION_ATTRIBS:
+                shot_value = shot_data.get(resolution_attrib)
+                entity_value = entity.attribs.get(resolution_attrib)
+                if shot_value and shot_value != entity_value:
+                    self.log.warning(
+                        "Resolution mismatch for shot."
```
BigRoy, line 163:
```suggestion
                        "Resolution mismatch for shot. "
```

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,156 @@ def process(self, instance):
...
+                        f"{resolution_attrib}={shot_value} but "
+                        f"already existing shot is set to {entity_value}."
```
BigRoy, line 165:
```suggestion
                        f" existing shot is set to {entity_value}."
```
I think shorter is better in these logs, if it's just as clear.

client/ayon_core/plugins/publish/validate_asset_docs.py
rdelillo: Maybe some concret examples to help understand how various settings/attributes/input might interfer the shot resolution present in `hierarchyContext` data as I'd need some guidance if we want to streamline that.

* Hiero publisher:
    * shot resolution computed in `ayon-hiero/collect_shots` plugin
        * either from timeline  if `sourceResolution` creator attribute is disabled
        * or read from plate source if `sourceResolution` creator attribute is enabled 
    * shot resolution then potentially overriden by `ayon_core/collect_explicit_resolution` if configured via settings
    * shot resolution is set in `hierarchyContext` via `ayon-core/collect_hierarchy` plugin

* CSV ingest (Traypublisher)
    * shot resolution is read from input CSV file if exists
    * shot resolution then potentially overriden by `ayon_core/collect_explicit_resolution` if configured via settings
    * shot resolution is reported to `hierarchyContext` via `ayon-traypublisher/collect_shot_instances`

rdelillo:
> I tested with hiero and get confused on the testing notes. Is the validator supposed to be optional? image I can't spot the toggle to turn/turn off the validator, but it is actually doing the check.

Thanks @moonyuet for the feedback !
The toogle to skip/perform the validation is here. Should I rephrase the label a bit or detail more in the testing steps ?
image

client/ayon_core/plugins/publish/validate_asset_docs.py
iLLiCiTiT: I'm sorry but I'm lost in this PR. Is the goal to not update attributes on folders based on checkbox, and have option to validate the attribute values (based on the same checkbox)?

If that's the case then the checkbox value changes behavior of the validator, the validator should not change anything, but currently I really don't understand how it works, so I can't tell.

The same value should be used in extract plugin to actually do what was set to do. Validator should only compare and raise error (repair actions can change).

antirotor (approved):
with the code changes I think this looks good. Can't test it right now, though.

client/ayon_core/plugins/publish/validate_asset_docs.py
```diff
@@ -37,3 +49,121 @@ def process(self, instance):
...
+            instance_id = inst.data.get("instance_id")
+            created_instance = create_context.get_instance_by_id(instance_id)
+            attr_values = created_instance.data["publish_attributes"].get(
+                "ValidateFolderCreationResolution", {})
+            attr_values["validateExistingFolderResolution"] = False
+
+        create_context.save_changes()
+
+
+class ValidateFolderCreationResolution(
```
iLLiCiTiT, line 71: I think this plugin should be Context plugin to run only once, or?

rdelillo:
Closing this PR as it should be more globally handled by https://github.com/ynput/ayon-core/pull/1550

Issue #1297, opened 2025-05-28
🏛️Product base types: Creation and creator plugins


antirotor:
## Intro

*Product base type* is solving the problem with *product type* being too tightly bound to the code and workflow logic but also used as a categorizing element in UIs. Having product type **image** is fine until I also want to differentiate between **reference image** and **concept image** but without changing any code whatsoever.
In this case, you would have to duplicate the creator plugin (best case scenario) or even tweak publishing plugins and loaders to support this.

We think that pipeline should define some core, basic types for workflows and studios can customize the naming as they want without affecting any logic and code - well except sorting and filtering. For that, we decided to introduce new property of every Product - `product base type`. It holds the information that the product is **image** no matter how we call it. With that, existing `product type` purpose switches to the customizable part - you can name it as you like and it will still behave as product base type defines.

## Changes

This poses some challenges because `product type` is currently mandatory and used on so many places.

### Server Support
We need to store new `product base type` on the server somehow. This is handled by https://github.com/ynput/ayon-backend/pull/575 - it is adding necessary fields and graphql support. `Product base type` can be NULL - this is intentional to easily identify all products created by legacy code without `product base type` support. As long as `product type` is there (and it must be as it is not optional) existing features should work without change. For example REST API endpoint for product creation `POST /api/projects/{project_name}/products` still needs `productType` but it doesn't require `productBaseType`.

### AYON Python API support
Is introduced by https://github.com/ynput/ayon-python-api/pull/255 - allows you to get `product base types`, it allows you to query them using GraphQL and you can add them to new product entities. Affected functions are:
- `new_product_entity()`
- `create_product()`
- `update_product()`
- `get_product_base_types()`
- `get_project_product_base_types()`
- `get_product_base_type_names()`

### AYON Core support
Multiple changes are made in ayon-core:
- https://github.com/ynput/ayon-core/pull/1299 is introducing check for feature support used in all subsequent PRs there
- https://github.com/ynput/ayon-core/pull/1301 this adds option to use `product base type` in the Loader plugins and it will show it in the Loader Tool.

In Loader plugins you can use `product_base_types` to limit the loader plugin to specific `product base types` the same way you can do it now for `product types`, `representations`, etc. like:

```python
class SomeLoader(Loader):
    product_base_types = {"image", "icon"}
```
This is something that needs to be refactored in existing Loader plugins across the pipeline, to unify and transition them from using `product types` as they will become too dynamic.

[...]

Issue #1295, opened 2025-05-28
🏛️Product base types: feature support


antirotor:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

We need function in ayon-core that will tell the rest of the pipeline that product base types are supported. This should simply return True/False and base on that other logic should find out when legacy logic needs to be used or when to use fallback values.

### How would you imagine the implementation of the feature?

 Implement `product_base_types_support() -> bool:` in **ayon-core**

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

PR #1299, opened 2025-05-28
🏛️Product base types: feature support check function

antirotor:
## Changelog Description
Adding simple function to handle transitional period until product base types are implemented and current use of product types is switched completely through the pipeline.

## Additional info
This function will return False, until all the breaking changes are properly implemented across the whole pipeline. For more information see description of #1297.

## Testing notes:
There is really nothing to test.

> [!NOTE]
> closing #1295


antirotor:
Direct use as `from ayon_core.pipeline import is_supporting_product_base_type` will most likely fail because of circular imports, but `from ayon_core.pipeline.compatibility import is_supporting_product_base_type` will work.

PR #1302, opened 2025-06-01
Fix maya plugin loading in case of remote is sftp

Aiacos:
Following this Bug:
[https://github.com/ynput/ayon-core/issues/1303](https://github.com/ynput/ayon-core/issues/1303)

### DIFF
--- a/client/ayon_core/host/dirmap.py
+++ b/client/ayon_core/host/dirmap.py
@@ -190,10 +190,17 @@ class HostDirmap(ABC):
             if remote_provider != "local_drive":
                 remote_site = "studio"
             for root_name, active_site_dir in active_roots_overrides.items():
-                remote_site_dir = (
-                    remote_roots_overrides.get(root_name)
-                    or sync_settings["sites"][remote_site]["root"][root_name]
-                )
+                # If my remote is SFTP or other kind of remote use work dir as path
+                if remote_provider == "SFTP":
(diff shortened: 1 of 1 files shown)


client/ayon_core/host/dirmap.py
```diff
@@ -190,10 +190,17 @@ def _get_local_sync_dirmap(self):
             if remote_provider != "local_drive":
                 remote_site = "studio"
             for root_name, active_site_dir in active_roots_overrides.items():
-                remote_site_dir = (
-                    remote_roots_overrides.get(root_name)
-                    or sync_settings["sites"][remote_site]["root"][root_name]
-                )
+                # If my remote is SFTP or other kind of remote use work dir as path
+                if remote_provider == "sftp":
```
iLLiCiTiT, line 194: Shouldn't we handle only those overrides that actually can affect this? That is question for @kalisp .

kalisp:
As mentioned in parent issue, it is unfortunately problem of explicitly setting `local` and `studio` sites in `ayon+settings://sitesync/sites`.

Singling out `sftp` provider wouldnt be good fix either way, it should be `!= 'local_drive'`, nevertheless it would be just patching wrong configuration and it shouldn't be in official `core`.

Root overrides might need to be rethought if there will be specified non working workflow and budget for it.

PR #1306, opened 2025-06-03
🏛️Product base types: Support in Creator logic

### LINKED ISSUE #1297: 🏛️Product base types: Creation and creator plugins

antirotor:
## Changelog Description
Add support for product base types to creation logic in **ayon-core**.

## Additional info
This is the first part of gradient transition of `product type` to optional alias for `product base type`. To keep the backwards compatibility, `product base type` is introduced as an optional argument to most of the existing functions. If it is not passed, warning is emitted - this will in most cases comes from the Creator plugins not implementing product base types.

So far - on the data model - `product type` is mandatory and `product base type` is nullable field. This will change in the future when `product_base_type` is consistently used across all the addons out there. 

Reflecting `product_base_type` in product name won't work if creator plugin overrides `get_product_name()` function. If it does, it needs to be refactored.

Also, the way how Creators create `CreatedInstance` differs so this must be considered too. The logic goes like this:

If `product_base_type` is passed to the constructor, it will be used. If not and the creator (it needs to be passed there too) itself defines `product_base_type`, it will  use that one.

There are some questions lingering:

How to handle `product_base_type` in Anatomy and Product name templates until it is made mandatory? If you have `{product[baseType]}` token in your templates:

| product_type  | product_base_type   |   |
|:---:|:---:|:---|
| ✔️ | ❌ | use `product_type`  |
| ❌ | ✔️ | use `product_base_type` |
| ❌ | ❌ | is invalid state |
| ✔️ | ✔️ | use `product_base_type` |

> [!NOTE]
> Closes #1297

> [!WARNING]
> This will break Maya integration if https://github.com/ynput/ayon-maya/pull/282 is not resolved and maybe other hosts.

## Testing notes:
For more information see description of #1297

### DIFF
--- a/client/ayon_core/pipeline/create/product_name.py
+++ b/client/ayon_core/pipeline/create/product_name.py
@@ -54,7 +54,7 @@ def get_product_name_template(
     profiles = tools_settings["creator"]["product_name_profiles"]
     filtering_criteria = {
         "product_types": product_type,
-        "host_names": host_name,: host_name,
+        "host_names": host_name,
         "task_names": task_name,
         "task_types": task_type,
         "product_base_types": product_base_type,


client/ayon_core/pipeline/create/context.py
```diff
@@ -1194,6 +1196,22 @@ def create(
             "productType": creator.product_type,
             "variant": variant
         }
+
+        # Add product base type if supported.
+        # TODO (antirotor): Once all creators support product base type
+        #   remove this check.
+        if is_supporting_product_base_type():
+
+            if hasattr(creator, "product_base_type"):
```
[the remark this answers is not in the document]
antirotor: your are absolutely right, no need for that check.

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -543,8 +554,12 @@ def get_product_name(
                 for which is product name updated. Passed only on product name
                 update.
             project_entity (Optional[dict[str, Any]]): Project entity.
+            product_base_type (Optional[str]): Product base type.
 
         """
+        if is_supporting_product_base_type() and (instance and hasattr(instance, "product_base_type")):  # noqa: E501
+            product_base_type = instance.product_base_type
```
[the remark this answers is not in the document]
antirotor: yes, but it works in this case as `product_base_type` is set only when `instances` is present. Otherwise it uses the value passed to the function. What I am wondering is what if the instance value should override the one explicitly sent to the function.

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -297,23 +285,40 @@ def register_callbacks(self):
...
         Default implementation returns plugin's product type.
-        """
 
-        return self.product_type
+        """
+        identifier = self.product_type
+        if is_product_base_type_supported():
+            identifier = self.product_base_type
+            if self.product_type:
+                identifier = f"{identifier}.{self.product_type}"
```
iLLiCiTiT: This affects how instances of the create plugin are discovered. We have to keep it as it is without using `.`.
```suggestion
        identifier = self.product_base_type
        if not identifier:
            identifier = self.product_type
        return identifier
```

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -385,18 +389,36 @@ def _create_instance(
...
 
         """
         if product_type is None:
             product_type = self.product_type
+
+        if (
+                is_product_base_type_supported()
+                and not product_base_type
+                and not self.product_base_type
+        ):
```
iLLiCiTiT:
```suggestion
        if not product_base_type:
```

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -574,6 +589,7 @@ def get_product_name(
             dynamic_data=dynamic_data,
             project_settings=self.project_settings,
             project_entity=project_entity,
+            product_base_type=product_base_type
```
iLLiCiTiT:
```suggestion
            product_base_type=self.product_base_type
```

client/ayon_core/pipeline/create/product_name.py
```diff
@@ -92,14 +114,20 @@ def get_product_name(
...
     Todos:
         Find better filtering options to avoid requirement of
             argument 'family_filter'.
 
     Args:
         project_name (str): Project name.
-        task_name (Union[str, None]): Task name.
-        task_type (Union[str, None]): Task type.
+        task_name (str): Task name.
+        task_type (str): Task type.
```
iLLiCiTiT: Task is optional
```suggestion
        task_name (Optional[str]): Task name.
        task_type (Optional[str]): Task type.
```

client/ayon_core/pipeline/create/product_name.py
```diff
@@ -46,6 +58,15 @@ def get_product_name_template(
         "task_types": task_type
     }
 
+    if is_product_base_type_supported():
```
iLLiCiTiT: The `is_product_base_type_supported` is to determine if
1. Server does support it.
2. ayon api does support it.

This logic can use the product base type filtering even if none of those are available. So we can add `"product_base_types"` all the time.

client/ayon_core/pipeline/create/product_name.py
```diff
@@ -58,14 +57,16 @@ def get_product_name_template(
...
-                "the future.", DeprecationWarning, stacklevel=2)
+    if not product_base_type:
+        warn(
+            "Product base type is not provided, please update your"
+            "creation code to include it. It will be required in "
+            "the future.",
+            DeprecationWarning,
+            stacklevel=2
+        )
+    filtering_criteria["product_base_types"] = product_base_type
```
iLLiCiTiT:
```suggestion
        "task_types": task_type,
        "product_base_types": product_base_type,
    }
    if not product_base_type:
        warn(
            "Product base type is not provided, please update your"
            "creation code to include it. It will be required in "
            "the future.",
            DeprecationWarning,
            stacklevel=2
        )

File: client/ayon_core/pipeline/create/creator_plugins.py [hunk 9]
```diff
@@ -543,6 +547,10 @@ def get_product_name(
             project_entity (Optional[dict[str, Any]]): Project entity.
 
         """
+        product_base_type = None
+        if hasattr(self, "product_base_type"):  # noqa: E501
+            product_base_type = self.product_base_type
+
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/pipeline/create/context.py
```diff
@@ -752,13 +753,12 @@ def _reset_creator_plugins(self) -> None:
...
-                self.log.debug(
-                    "Skipping abstract Creator {}".format(str(creator_class))
-                )
-                continue
+        for creator_class in report.abstract_plugins:
+            self.log.debug(
+                "Skipping abstract Creator '%s'", str(creator_class)
+            )
 
+        for creator_class in report.plugins:
```
iLLiCiTiT:
```python
            if not creator.product_base_type:
                warn(
                    f"Provided creator {creator!r} doesn't have"
                    " product base type attribute defined. This will be"
                    " required in future.",
                    DeprecationWarning,
                    stacklevel=2
                )
```
EDITED:
This should be after create plugin initialization before storing it by identifier (I can't put suggestion there).

client/ayon_core/pipeline/create/product_name.py
```diff
@@ -159,14 +180,29 @@ def get_product_name(
...
+    }
+    if not product_base_type and "{product[basetype]}" in template.lower():
+        product["baseType"] = product_type
+        warn(
+            "You have Product base type in product name template, "
+            "but it is not provided by the creator, please update your "
+            "creation code to include it. It will be required in "
+            "the future.",
+            DeprecationWarning,
+            stacklevel=2)
```
iLLiCiTiT:
```suggestion
    if not product_base_type and "{product[basetype]}" in template.lower():
        warn(
            "You have Product base type in product name template, "
            "but it is not provided by the creator, please update your "
            "creation code to include it. It will be required in "
            "the future.",
            DeprecationWarning,
            stacklevel=2)
        
    product: dict[str, str] = {
        "type": product_type,
        "baseType": product_base_type or product_type,
    }
```

client/ayon_core/pipeline/create/structures.py
```diff
@@ -562,6 +576,10 @@ def __init__(
         self._data["id"] = item_id
         self._data["productType"] = product_type
         self._data["productName"] = product_name
+
+        if is_product_base_type_supported():
+            self._data["productBaseType"] = product_base_type
```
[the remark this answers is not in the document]
iLLiCiTiT: If it's to make sure it is added only when it is supported, then we probably should store it only if the create plugin does support it?

In that case it would be:
```python
        if not product_base_type:
            product_base_type = creator.product_base_type

        ...
        if product_base_type:
            self._data["productBaseType"] = product_base_type
```
So if create plugin does define product base type, it is already available... But I don't think it is necessary as the product base type is what current product type is. I can't imagine a use-case when you'd want to know if the instance was created using pre-base type create plugin or not.

client/ayon_core/pipeline/create/product_name.py
```diff
@@ -192,16 +193,20 @@ def get_product_name(
...
-                "but it is not provided by the creator, please update your"
-                "creation code to include it. It will be required in "
-                "the future.", DeprecationWarning, stacklevel=2)
+    product: dict[str, str] = {
+        "type": product_type,
+        "baseType": product_base_type
+    }
+    if not product_base_type and "{product[basetype]}" in template.lower():
+        product["baseType"] = product_type
+        warn(
```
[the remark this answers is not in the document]
antirotor: Now I don't exactly know what you propose. These warnings will show only when you enable development mode in Python or set corresponding env vars. So no artist will probably ever see it. I think the bigger issue is that almost no dev will see it either so maybe we should actually `log.warning`?

PR #1317, opened 2025-06-10
Scene inventory tool: Update error more visible

tanh609:
## Changelog Description
When there is an exception in the `_update_containers` function, only AssertionError was caught. This lead to silent error for the other exception type. This pull request aims at providing visual feedback on the error when the other types of exceptions are caught. 

## Additional info

image
[src](https://community.ynput.io/t/ayon-manager-silent-error/2586/3?u=bigroy)

## Testing notes:
1. Use a loader to load an asset into the scene
2. Modify the update code of the plugin to raise a test exception
3. Use the manager to update the asset in the scene

### DIFF
--- a/client/ayon_core/tools/sceneinventory/view.py
+++ b/client/ayon_core/tools/sceneinventory/view.py
@@ -959,11 +959,13 @@ class SceneInventoryView(QtWidgets.QTreeView):
             remove_container(container)
         self.data_changed.emit()
 
-    def _show_version_error_dialog(self, version, item_ids):
+    def _show_version_error_dialog(self, version, item_ids, exception=None):
         """Shows QMessageBox when version switch doesn't work
 
         Args:
             version: str or int or None
+            item_ids (Iterable[str]): List of item ids to run the
+            exception (Exception, optional): Exception that occurred
         """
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -959,11 +959,13 @@ def _show_remove_warning_dialog(self, item_ids):
             remove_container(container)
         self.data_changed.emit()
 
-    def _show_version_error_dialog(self, version, item_ids):
+    def _show_version_error_dialog(self, version, item_ids, exception=None):
```
iLLiCiTiT, line 962:
```suggestion
    def _show_version_error_dialog(self, version, item_ids, exception):
```

client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -987,11 +989,14 @@ def _show_version_error_dialog(self, version, item_ids):
 
         dialog.addButton(QtWidgets.QMessageBox.Cancel)
 
+        exception = exception or "Unknown error"
+
```
BigRoy, line 993:
```suggestion
```

PR #1321, opened 2025-06-12
Extract review: Correct default value and typehint

iLLiCiTiT:
## Changelog Description
Use correct default factory for `filled_files` and use `dict` instead of `Dict` for typehint.

## Additional info
The differnce in `list` vs. `dict` was overlooked in https://github.com/ynput/ayon-core/pull/1319.

## Testing notes:
1. Validate code changes.


BigRoy (approved):
Thanks - makes sense. Seems I indeed missed updating the default.

Does the `dict[int, str]` work in Python 3.7 with `from __future__ import annotations`?

Issue #1309, opened 2025-06-04
Loader tool: Add tags filtering


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

There is a request for tags filtering in loader tool.

### How would you imagine the implementation of the feature?

Add another filter for tags.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

PR #1323, opened 2025-06-13
Loader tool: Add filtering bar with more filtering options

iLLiCiTiT:
## Changelog Description
Loader tool has filter bar where is possible to add/modify/remove filters for versions and products.

## Additional info
There was a request to add another filtering options which would require to expand the window more without option to make it smaller, adding a filter bar where filters can be added dynamically we don't need to worry about the width that much.

### Screenshots
image

image

image

image

### Notes
Currently the displayed filters have "less" visual information as we're not showing icons of set filters and the filter values are merged into smaller text.

The height of filter bar does not match height of projects combobox and refresh button, not sure if/how it should be matched.

Another thing is that e.g. task tags does not filter tasks, only versions which might be better to be changed maybe? We might also move folder name filter to the filters too in that case?

## Testing notes:
1. Open loader in a DCC or browser from tray.
2. Select a project.
3. Add filters. Remove filters. Change filters.
4. Validate they do filter correctly.
5. Report issues if there are any.

Resolves https://github.com/ynput/ayon-core/issues/1309


BigRoy (changes requested):
Visually this looks amazing, UX wise I'm not entirely convinced yet. Some thing that is _very_ common on our end is to quickly 'search' for a product by name... a lot. The other filters are more 'persistent' in that you tend to search with similar filters active in a particular DCC or task - but product name filtering should really be at your finger tips at any moments notice without too much hassle.

The fact that you now need to click twice is really getting in the way there.

Important:
- [x] Would it be possible that if you start typing when you do not select the second entry, e.g. just click the magnifying glass - that it defaults to "Product Name" filter searching?
    - [ ] Possibly even that if you click the magnifying glass and start typing and a filter exists already, that it starts 'editing' the product name filter too. The only reasoning here is that you can consistently click the magnifying glass to filter product name, instead of needing to click there if there's no filter yet and otherwise click further to the right if there is a filter.
- [x] We've lost the **Enable Grouping** combobox.
- [x] We've lost support for `ayon+settings://core/tools/loader/product_type_filter_profiles` - I don't get automatically applied Product Type filters based on the profile.

Nice-to-haves:
- [x] Any chance we can have CTRL+F in Loader tool activate this filter bar? :)
- [ ] The "2 items" filter would be nicer if similar to before it could hint at more of the entries of what's selected.
- [x] When selecting one of the filters, like e.g. Product Type, can we make it so that the user can type to "filter" the dropdown similar to web frontend.
- [x] When a selection is made of dropdown or product name filter, you can click outside of it or press ESC to close the dropdown... but should pressing enter close it too?
- [ ] Before this PR if you'd toggle e.g. Product Type filters you'd directly see the result... it also directly updates now too, but because it's over about 50% off the actual table you can hardly see the result. I wonder if should somehow make it less wide or do something else so that we can still see the changes live as you're editing instead of it being covered by the filter you're currently manipulating. (This is not an issue with product name filter because it's small and above the table)

---

> The height of filter bar does not match height of projects combobox and refresh button, not sure if/how it should be matched.

I'm less bothered by this than I expected - so will leave it up to others to decide.

> Another thing is that e.g. task tags does not filter tasks, only versions which might be better to be changed maybe? We might also move folder name filter to the filters too in that case?

Not too bothered by this myself - but maybe I'm not using tags enough to be the one to decide here. I'd say, visually it's clear that it filters the Products/Versions table.

---

Also, this PR may be good to test by @LiborBatek using his High DPI screen setups? ;)

LiborBatek (commented):
I have tested multiple scenarios and various filter settings and all seems working well and without any glitches.

So in this respect all is working and could be seen as good to go aka for approval.

But I have to agree with some of @BigRoy comments/suggestions as now its a bit cumbersome that typing directly does nothing and one have to click the filter type category first... the former way was better even tho limitating compared to this new system.

Also the fact that ```product type``` need to be picked from dropdown list (too loong imho) is not optimal as user need to go through all of the types and find a proper one.

In fact the already present filter system combined those 2 types (product name / type) together even if limitting the granularity ...in most cases its fast and efficient enough - now Im not complaining about introducing more filter types (its for the good for sure) but the way user interacts with those filters atm.

I could easily imagine just moving the filter window to each category would make it much more easy/straightforward to operate such filters....

starting typing into each could offer the available type etc.  e.g. statuses could have dropdown as atm, but product name category could accept full text... and together it could form the whole filter conditions (as now been stacked next to each if selected multiple filter types)


Screenshot 2025-06-13 153908


Just my 2 cents...again its functional as is atm but not ideal in operation for the user - more steps to do to get desired filtering but still offering more granularity then before - which is good!

LiborBatek:
...maybe just introducing ```CTRL+F``` command key or so for direct fulltext filtering as @BigRoy suggested too (for using present filter style) could do without need of refactoring the new filter tools too

iLLiCiTiT:
You can do CTRL+F and start typing..

BigRoy (changes requested):
Nice work, predefined product type filters from settings works (it does not in Tray Browser, but that was the case before this PR as well it seems).

I'm wondering if **Statuses** filter label should be **Version Statuses**

### Issue

The CTRL+F key fails for me in Fusion:
```python
  File "F:\dev\ayon-core\client\ayon_core\tools\loader\ui\window.py", line 352, in keyPressEvent
    combination = event.keyCombination()
AttributeError: 'PySide2.QtGui.QKeyEvent' object has no attribute 'keyCombination'
```
PySide2 with Qt `5.15.2`.

---

### Nice-to-have

**Nice-to-haves (not critical at all)**

Can we get backspace to also trigger 'editing' the product name filter if there's also a value, instead of only supporting pressing a character key to start _adding characters_

Workaround for now is to press a character on keyboard, and then backspace twice.

---

When filtering the 'enum types' like "Version Tags" can we get the up/down keys to go up/down the filtered list whilst still being in the text field.

BigRoy (approved):
This works, also in Fusion with older PySide2 releases now. ✅ I'd say, no showstoppers, just some mostly cosmetic remarks.

Tested in Tray, Maya, Fusion and Houdini.

- [x] I did come across this quirk which isn't a showstopper, but when pressing CTRL+F twice you're basically typing a special character into the text box:

image

Still some pending nice-to-haves as can be seen across my comments. I'll list them here:

- [ ] The "2 items" filter would be nicer if similar to before it could hint at more of the entries of what's selected.
- [ ] Before this PR if you'd toggle e.g. Product Type filters you'd directly see the result... it also directly updates now too, but because it's over about 50% off the actual table you can hardly see the result. I wonder if should somehow make it less wide or do something else so that we can still see the changes live as you're editing instead of it being covered by the filter you're currently manipulating. (This is not an issue with product name filter because it's small and above the table)
- [ ] When filtering the 'enum types' like "Version Tags" can we get the up/down keys to go up/down the filtered list whilst still being in the text field.
- [ ] Would still love to be able to see the product list better as I switch my filtering around, because it's currently often obscured by the 'dropdown dialog' as I'm making edits.

And some new ones:

- [ ] When making edits to the filtering, the only way to apply your filter is by getting the popup dialog to disappear manually by clicking outside of it. I wonder if we need a "Confirm" button at the bottom just for the sake of it... A potential upside to that is that it may make it match more with the web frontend design too?
image
_web frontend reference_

- [ ] In the off chance that you have many filters applied then you can't remove the _furthest right entries_ because you can't navigate to the X icon on the filter search bar.

image

iLLiCiTiT:
>  In the off chance that you have many filters applied then you can't remove the furthest right entries because you can't navigate to the X icon on the filter search bar.

Added mouse wheel scrolling, but there is no visual indicator for it.

Issue #1322, opened 2025-06-12
Tray Launcher: comments for new launcher action design


BigRoy:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

With https://github.com/ynput/ayon-core/pull/1226 merged there's now a new design in the Launcher UI for how actions and applications are shown and listed.
But there are some bits that feel less usable than before.

#### Disable auto expand

- [ ] Auto-expand on however of app group is really annoying

If I want to choose Silhouette, if I come in from the left with my mouse I'd be opening in Houdini instead.
If I were to come from above, but slightly nudge over the card left of it... then of course, I can't start Silhouette either.

![Image](https://github.com/user-attachments/assets/329b01ea-69f2-4118-bc22-397f3865273f)
![Image](https://github.com/user-attachments/assets/5eeac6f1-d7b6-4d85-82d9-24b2fb814473)
![Image](https://github.com/user-attachments/assets/ca2a8381-7b6a-48da-899f-a024e3ab28c7)

I'd recommend "click" to open the "group" (like in most other known UX out there; grouped icons on Ipad; grouped channels on discord, etc.) 
Worst case scenario we can also opt for a longer delay before the slide out happens, but that would make it slower to work with the tool + would still be largely problematic with many applications enabled.

Just want to remark that the new 'slideout' _looks nice_ but to me personally doesn't necessarily feel that much more navigatable. Within an application group all the icons will always be the same, the icons take up the majority of the space so suddenly I just need to move further with my mouse to click AND have less focus on what information I'm actually looking for in the expanded list - which is the _name_ I'm picking from (unless the big icon would hint it too).

#### Show full labels

- [ ] Labels on variants are cut off, I can't see which is Houdini Core or Houdini FX

![Image](https://github.com/user-attachments/assets/f3e4d68e-077a-40a3-b5e1-21b9d4377f2a)

In comparison, before:

![Image](https://github.com/user-attachments/assets/580e3485-3b86-43df-9495-1b1391baeddb)

![Image](https://github.com/user-attachments/assets/915c59fc-be9b-4220-8c0c-a28b4089736e)

#### Action name on action config

- [ ] Can we get the Action name on the Action Config title?

Since the action config pops up quite far from the clicked action it's easily missed which action it's related to.

![Image](https://github.com/user-attachments/assets/35b53569-ad50-4e60-943f-c4c69b4ec991)

In comparison, before:

[...]

PR #1331, opened 2025-06-20
Launcher tool: Enhance UX with actions

iLLiCiTiT:
## Changelog Description
Grouped actions now require a click to show subactions. Config dialog should popup under mouse. Added shadow frame around subactions popup. The font size of action labels is smaller as a quickfix for long labels.

## Additional info
I did test it only on windows with 100% and 150% scaling.

### Screenshot
image

## Testing notes:
1. Open launcher tool.
2. Right click should pop config dialog under mouse.
3. Visual check of all requested changes on as many platforms as possible.

Resolves https://github.com/ynput/ayon-core/issues/1322


BigRoy:
If I happen to enable "Show all applications" and disable "Show only available applications" of course it shows me all apps with all variants. But with many variants, the pop out gets broken.

image

Not really a production scenario, but this may hint at other issues with this particular UI code - so sharing it here.

BigRoy (approved):
Technically this is usable. Just still wish the Houdini variant label for `20.5.522 Core` would not be cut off.

MilaKudr (approved):
Screenshot 2025-06-24 090027

PR #1338, opened 2025-06-24
Feature: added source_resolution_* template keys for burnins

kalisp:
## Changelog Description
`{source_resolution_width}` and `{source_resolution_height}` can be used in burnins to highlight source resolution which might be quite different from review resolution.

## Additional info
Where we should add these to documentation. There is no complete list, just https://help.ayon.app/articles/3005275-core-addon-settings#trzd1epqsrr

Wondering if setting them in `ExtractReview` isnt too late, as this is build for requirement with `to visualize the and catch possible issues`. I can envision some potential validator to actually check values. But I dont think that would be trivial or even possible to create this generic enough for all DCCs. 

Maybe just validator for Tray/Web Publisher where it might be most useful? 
But thats question for different PR.

## Testing notes:
1. you will need to add `-vf scale=iw*0.5:ih*0.5` (scale 50%) to used `ayon+settings://core/publish/ExtractReview/profiles/1/outputs`
image
2. Use `{source_resolution_width}` or `{source_resolution_height}` in used burnin profile
`ayon+settings://core/publish/ExtractBurnin/profiles/0/burnins`
image
3. publish image sequence with higher resolution
4. check printed out resolution in burnin mp4

### DIFF
--- a/client/ayon_core/tools/launcher/models/actions.py
+++ b/client/ayon_core/tools/launcher/models/actions.py
@@ -352,9 +352,6 @@ class ActionsModel:
         )
 
     def _get_webaction_request_data(self, selection: LauncherActionSelection):
-        if not selection.is_project_selected:
-            return None
-
         entity_type = None
         entity_id = None
         entity_subtypes = []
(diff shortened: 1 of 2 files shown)


iLLiCiTiT:
This is just note: I'm not sure about the `"source"` in the key. Maybe `"origin"` would make more sense? Please do not just this is just me thinking.

MilaKudr:
Work as expected, published from Nuke
Screenshot 2025-06-25 151249

kalisp:
> This is just note: I'm not sure about the `"source"` in the key. Maybe `"origin"` would make more sense? Please do not just change it, this is just me thinking... Because source extract review can be plate exrs or mov renderer (but it should be the exr's resolution), for burnins is source extract review output.

It was key name provided by client, so I guess its fine. Naming is hard :D.

PR #1351, opened 2025-07-01
Add explicit_cleanup_dirs to cleanup.py

sjt-rvx:
## Changelog Description
To allow gaffer to publish to multiple contexts in one go, I wanted to be able to specify explicit cleanup directories per-instance, otherwise all the context's paths-to-cleanup would be removed once the first of the publishes would go through. Plus I needed to clean up more folders than just the `stagingDir` so cleanup_farm.py does not do it for me (plus, it's a context plugin).

### DIFF
--- a/client/ayon_core/plugins/publish/cleanup.py
+++ b/client/ayon_core/plugins/publish/cleanup.py
@@ -84,6 +84,14 @@ class CleanUp(pyblish.api.InstancePlugin):
                 f"type is excluded from cleanup: {product_type}")
             return
 
+        explicit_cleanup_dirs = instance.data.get("explicit_cleanup_dirs", [])
+        for _dir in explicit_cleanup_dirs:
+            self.log.debug("Removing explicit cleanup dir {}".format(_dir))
+            try:
+                shutil.rmtree(_dir)
+            except Exception as err:
+                self.log.debug(f"Could not remove dir {_dir}: {err}")
+
         temp_root = tempfile.gettempdir()
         staging_dir = instance.data.get("stagingDir", None)
 


client/ayon_core/plugins/publish/cleanup.py
```diff
@@ -84,6 +84,14 @@ def process(self, instance):
                 f"type is excluded from cleanup: {product_type}")
             return
 
+        explicit_cleanup_dirs = instance.data.get("explicit_cleanup_dirs", [])
+        for _dir in explicit_cleanup_dirs:
+            self.log.debug("Removing explicit cleanup dir {}".format(_dir))
+            try:
+                shutil.rmtree(_dir)
```
BigRoy, line 91: Although I can see the use case - I'm a bit worried that some day, accidentally a wrong path ends up in `explicit_cleanup_dirs` :') and it ends up wiping a full project or asset folder. :)

Additionally, I feel like it should _not be_ part of this plug-in. As far as I can see this plug-in is focused on handling `stagingDir` - and explicitly does not apply to _other files_. This logic is better kept in either a separate plug-in, or actually part of the explicit cleanup (even though that is `ContextPlugin` it can iterate over the instances just fine.)

sjt-rvx:
> Looks like duplicty of [ExplicitCleanUp](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/cleanup_explicit.py) . Just use `"cleanupFullPaths"` instead of `"explicit_cleanup_dirs"`.

ExplicitCleanup is a context plugin, what I needed was per-instance cleanup. But I agree with Roy about maybe having this inside ExplicitCleanup - and just have it also iterate over the instances being published to fish out `cleanupFullPaths` from there. 

The issue with just having it a context plugin is the case where publishing happens to multiple shots on the farm. So if I collect the paths in the publish context - all the cleanup paths for the different shots get added to the `cleanupFullPaths` then render 1 finishes and cleans all paths, even for renders 2, 3 and 4. This is a result of the farm-generation-jobs being in one publish context and the actual farm-image-publish being another, so transferring the publish context data to the per-render-context data one-to-one is the thing that's getting in my way.

PR #1352, opened 2025-07-01
Hacky version freezing

sjt-rvx:
## Changelog Description
This allows us to pin loaded versions in a scene; so that they don't change even if 'update all' is pressed. This requires a loader plugin that sets the 'version_freeze' data on the containers themselves - currently those live in their respective addons (nuke, gaffer). 

As a plus, I'm just hijacking the 'count' column from the inventory, since I wanted to test this out quickly, so this would need it's own column if it were to be a proper thing.

So this is more for discussion rather than anything else.

## Example loader plugin (the nuke one)
```python
class ToggleVersionFreeze(InventoryAction):

    label = "Toggle version freeze"
    icon = "snowflake-o"
    color = "#d8d8d8"

    def process(self, containers):

        for container in containers:
            node = container["node"]

            if "avalon:version_freeze" in node.knobs().keys():
                value = not node["avalon:version_freeze"].getValue()
            else:
                value = True
            data = {("avalon:version_freeze", "version_freeze"): value}
            for knob in create_knobs(data, None):
                if knob.name() in node.knobs().keys():
                    node[knob.name()].setValue(knob.value())
                else:
                    node.addKnob(knob)

            h = host_tools.get_tool_by_name("sceneinventory")
            h.refresh()
```

### DIFF
--- a/client/ayon_core/tools/sceneinventory/view.py
+++ b/client/ayon_core/tools/sceneinventory/view.py
@@ -1123,6 +1123,8 @@ class SceneInventoryView(QtWidgets.QTreeView):
         If at least one item is specified this will always try to refresh
         the inventory even if errors occurred on any of the items.
 
+        This will skip items that have 'version_freeze' set to True.
+
         Arguments:
             item_ids (Iterable[str]): Items to update
             version (Union[int, HeroVersion]): Version to set to.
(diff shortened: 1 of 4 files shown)


client/ayon_core/tools/sceneinventory/view.py
```diff
@@ -1131,8 +1133,18 @@ def _update_containers_to_version(self, item_ids, version):
...
+        non_frozen_item_ids = []
+        containers_by_id = self._controller.get_containers_by_item_ids(
+            item_ids
+        )
+        for item_id in item_ids:
+            container = containers_by_id[item_id]
+            if container.get("version_freeze", False):
+                continue
+            else:
+                non_frozen_item_ids.append(item_id)
```
iLLiCiTiT, line 1145:
```suggestion
        containers_by_id = self._controller.get_containers_by_item_ids(
            item_ids
        )
        non_frozen_item_ids = []
        for item_id in item_ids:
            container = containers_by_id[item_id]
            if not container.get("version_freeze", False):
                non_frozen_item_ids.append(item_id)

        if not non_frozen_item_ids:
            return
```

BigRoy:
Just want to raise the discussion whether "freezing" is the best correct terminology here? @mkolar @philippe-ynput 
Should this be **pinning** or **locking** instead?

Also, with the way loaded containers and there listing currently works across DCCs we can't abstractly imprint a container with additional metadata, like e.g. this "frozen" state - making the way it currently works to implement it quite DCC specific (e.g. adding the extra attribute in Maya will need to use the maya API to do so.)  This is, unlike Creators in CreateContext the Loaders do not have abstracted away the setting of data on containers on the Loader class itself. There is no `set_data` or `update` equivalent on Loaders that persists data to the scene from some sort of base `Container` object. Not very problematic, but definitely something we could gain from if this didn't need to be DCC specific.

PR #1354, opened 2025-07-01
Allow merging of file sequence entries in files_widgets

sjt-rvx:
## Changelog Description
There came up the issue a while back that when very long sequences (6000 frames or longer) were dragged and dropped onto the file widgets QT would freak out and not handle it (behaving as if nothing was dropped). The fix was to add this merge function and drop 2 x 3000 frames and then combine them. Stupid? Yes. Did it work? Yes.

### DIFF
--- a/client/ayon_core/tools/attribute_defs/files_widget.py
+++ b/client/ayon_core/tools/attribute_defs/files_widget.py
@@ -892,6 +892,28 @@ class FilesWidget(QtWidgets.QFrame):
             self._add_filepaths(new_items)
         self._remove_item_by_ids(item_ids)
 
+    def _on_merge_request(self):
+        if self._multivalue:
+            return
+
+        item_ids = self._files_view.get_selected_item_ids()
+        if not item_ids:
+            return
+
+        all_paths = []
+        for item_id in item_ids:
+            file_item = self._files_model.get_file_item_by_id(item_id)
+            if not file_item:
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/attribute_defs/files_widget.py
```diff
@@ -900,18 +900,20 @@ def _on_merge_request(self):
...
-        self._remove_item_by_ids(item_ids)
-        new_items = FileDefItem.from_value(unique_all_pahts, True)
+            if file_item is None:
+                continue
+            merged_item_ids.add(item_id)
+            all_paths |= {
+                os.path.join(file_item.directory, filename)
+                for filename in file_item.filenames
+            }
+
```
iLLiCiTiT, line 914:
```suggestion

```

iLLiCiTiT (approved):

client/ayon_core/tools/attribute_defs/files_widget.py
```diff
@@ -892,6 +892,29 @@ def _on_split_request(self):
             self._add_filepaths(new_items)
         self._remove_item_by_ids(item_ids)
 
+    def _on_merge_request(self):
```
iLLiCiTiT, line 895: BTW This should be ignored if the definition does not allow sequences.

And the action should also not be shown.

Issue #1176, opened 2025-03-10
AY-7575_Make instances in publisher mandatory


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

Instances in the Publisher have checkbox next to them so users can enable/disable them. But if there is instance that shouldn't be disabled at all, there is no way to mark such instance.

User story is: enforce publishing of workfiles

Solution would be to follow some flag on the instance data like optional = False and based on that allow disabling/enabling of the instance in the UI

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7575](https://app.clickup.com/t/86c2f64tw)_

iLLiCiTiT:
I don't think that information should be stored to scene, so I don't think `"optional"` on instance data is what should be used.

BigRoy:
Valid point - so transient data or an explicit attribute on the `CreatedInstance` may be more relevant (since it's not really relevant to the `pyblish.api.Instance` either.

PR #1360, opened 2025-07-02
Publisher: Mandatory instances

iLLiCiTiT:
## Changelog Description
Instances can be marked as mandatory, in that case they are always active and won't show checkbox in UI.

## Additional info
There was a discussion if we should use "is mandatory" or "is optional". Mandatory directly affects `"active"` state whereas Optional would only disable the checkbox for user. I don't have strong opinion, we might support both if necessary in future by introducing state as enum of states.

## Testing notes:
1. The easiest way would be to test with maya's PR https://github.com/ynput/ayon-maya/pull/312 .

Resolves https://github.com/ynput/ayon-core/issues/1176

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -872,7 +872,7 @@ class CreateContext:
         Event is triggered when instances are already available in context
             and have set create/publish attribute definitions.
 
-        Data structure of event::
+        Data structure of event:
 
             ```python
             {
@@ -899,7 +899,7 @@ class CreateContext:
 
         Event is triggered when instances are already removed from context.
 
-        Data structure of event::
+        Data structure of event:
 
             ```python
             {
@@ -927,7 +927,7 @@ class CreateContext:
         Event is triggered when any value changes on any instance or
             context data.
 
-        Data structure of event::
+        Data structure of event:
 
             ```python
             {
@@ -965,7 +965,7 @@ class CreateContext:
         Create plugin can trigger refresh of pre-create attributes. Usage of
             this event is mainly for publisher UI.
 
-        Data structure of event::
+        Data structure of event:
 
             ```python
             {
@@ -994,7 +994,7 @@ class CreateContext:
 
         Create plugin changed attribute definitions of instance.
 
-        Data structure of event::
+        Data structure of event:
 
             ```python
             {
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
It works, however - on first glance it feels odd that `mandatory` is a "state" but `active` is not - even though it just as much feels like "the state of the instance" but maybe state conceptually is quite similar to what the attribute encompasses.

Anyway, there may be some confusion going foward on "instance_state_change" and alike due to this. Changing `active` may seem like something that would emit a state change (even though it's an attribute change).

philippe-ynput:
@BigRoy @iLLiCiTiT To encapsulate the notion of [mandatory|optional] I suggest "requirement".

client/ayon_core/pipeline/create/context.py
```diff
@@ -1049,6 +1054,35 @@ def add_publish_attr_defs_change_callback(
...
         )
 
+    def add_instance_requirement_change_callback(
+        self, callback: Callable
+    ) -> "EventCallback":
+        """Register callback to listen instance requirement changes.
+
+        Create plugin changed attribute definitions of instance.
+
+        Data structure of event::
```
BigRoy, line 1064: Is the double `::` intentional?

```suggestion
        Data structure of event:
```

client/ayon_core/pipeline/create/context.py
```diff
@@ -1049,6 +1054,35 @@ def add_publish_attr_defs_change_callback(
             PUBLISH_ATTR_DEFS_CHANGED_TOPIC, callback
         )
 
+    def add_instance_requirement_change_callback(
+        self, callback: Callable
+    ) -> "EventCallback":
+        """Register callback to listen instance requirement changes.
```
BigRoy, line 1060:
```suggestion
        """Register callback to listen to instance requirement changes.
```

client/ayon_core/pipeline/create/context.py
```diff
@@ -1049,6 +1054,35 @@ def add_publish_attr_defs_change_callback(
...
+
+        Create plugin changed attribute definitions of instance.
```
BigRoy, line 1062: This seems unrelated actually here? Or plain wrong?

BigRoy (approved):
Works in Maya 2026 with https://github.com/ynput/ayon-maya/pull/312

The only slightly confusing thing is that in this mode the checkbox is still visible, yet clicking it does nothing. Maybe it should also be gone or disabled, but it's far from critical.

image

PR #1362, opened 2025-07-03
Chore: Define logging config for AYON processes

iLLiCiTiT:
## Changelog Description
Define default logging config for AYON processes.

## Additional info
CLI initializes logging config.

## Testing notes:
1. Logs should be visible in terminal.

### DIFF
--- a/client/ayon_core/cli.py
+++ b/client/ayon_core/cli.py
@@ -2,6 +2,7 @@
 """Package for handling AYON command line arguments."""
 import os
 import sys
+import logging
 import code
 import traceback
 from pathlib import Path
@@ -22,6 +23,9 @@ from ayon_core.lib.env_tools import (
     compute_env_variables_structure,
     merge_env_variables,
 )
+
+logging.basicConfig()
+log = logging.getLogger()
 
 
 @click.group(invoke_without_command=True)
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
I guess this would fix the issue with `acre` import having been doing this originally all this time. (didn't test it)

However, it does remind me of people always getting upset about `logging.basicConfig` in a 'library' or dependency, like [here](https://www.reddit.com/r/Python/comments/wt6325/comment/il29hal/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button).

> Under NO CIRCUMSTANCES use logging.basicConfig or use the base logger in a library. You will pollute the peoples logging who use your library and it is very bad form.
> 
> The standard best practices pattern is to request a logger with a specific name tied to your library, many people prefer to just ask for the logger called double underscore name, because python will provide the library file name that way.****

But maybe this `cli.py` isn't _that_ particular scenario since it's _just_ for the CLI + the CLI doesn't get imported elsewhere. I do wonder whether we want to add the `basicConfig` at start of `main_cli` so that if one were to, for whatever reason import the file then it wouldn't enforce the basic config.

Looks good to me. Not sure if it's the _right_ way however.

BigRoy (approved):
Seems fine to me.

PR #1363, opened 2025-07-04
Add "blender" to hosts list for Validate File Saved

BigRoy:
## Changelog Description

Trigger validation whether workfile is saved also in Blender
## Additional info

image

## Testing notes:

1. Validation should work in Blender.


iLLiCiTiT (approved):

BigRoy:
> @BigRoy Could you elaborate more on the testing steps? As I m not really sure whats changed as this validator been already used...
> 
> Is there new `ayon-core` settings for it? as Im slightly confused which category it is...
> 
> ideally provide ayon URI and testing steps (should I try to disable it and enable it too?)
> 
> thx

Start blender, don't save your scene.
Try and publish, it should get to the validation provided here.

Best to try with this PR https://github.com/ynput/ayon-blender/pull/137 to then also get rid of that very spammy validation inside Blender itself.

This way, the validation now shares the logic also used by other integrations, benefits from nicer actions on it (like the "Save current file" without having to open the workfiles tool).

Does that help?

PR #1366, opened 2025-07-10
Photoshop: new defaults for ExtractReview

kalisp:
## Changelog Description
As Photoshop was refactored to abandon its own `ExtractReview` these additional defaults for global `ExtractReview` mimics existing implementation as close as possible. (As `review` output for PS was `jpg` and not `png`, and `mov` instead of `mp4`.)

This allows to control reviews for Photoshop in same place as for other DCCs.

## Additional info
Should be tested with https://github.com/ynput/ayon-photoshop/pull/36

This PR is not necessary IF difference between produced data types is communicated with current clients and they will be ok with it OR they will set this config by themselves.

## Testing notes:
1. experiment with `ayon+settings://core/publish/ExtractReview`
2. publish from PS with different instances created


MilaKudr (approved):

64qam (approved):
Is there a reason why PS profile is created in the settings? It works with the "common" profile for all apps.

PR #1376, opened 2025-07-21
Chore: Detached process has own console

iLLiCiTiT:
## Changelog Description
New ayon console process has new console window.

## Additional info
The logic is handled in `run_detached_process` which is usually used for launch of new AYON launcher process.

## Testing notes:
1. Use `ayon_console.exe` executable.
2. Any new process of AYON launcher should spawn a console, for example traypublisher.


BigRoy:
Should this also work in dev mode and/or running from live source code?

iLLiCiTiT:
> Should this also work in dev mode and/or running from live source code? Should this be using [is_using_ayon_console](https://github.com/ynput/ayon-core/blob/698fef28330ce277ec9310535c1bb9277443dc5c/client/ayon_core/lib/ayon_info.py#L55) function instead?

The issue with that is that the function, where it is handled, is not used only for ayon launcher process.

iLLiCiTiT:
> Note that "Admin > Publish Report Viewer" does NOT show a console at all... just the actions do for a very brief second.

That tool is not showed in subprocess.

> Anyway, I get a very brief console window popping up - but it almost immediately shuts down. So it does seem to launch with a console, but whatever it executes 'finished' rapidly.

Aaah, I didn't test it with webactions. Interesting.

PR #1377, opened 2025-07-22
Use Workfiles API `save_next_version` in `version_up_current_workfile`

BigRoy:
## Changelog Description

Use the new workfiles API for version up in menus, etc.

## Additional info

After https://github.com/ynput/ayon-core/pull/1378 and https://github.com/ynput/ayon-core/pull/1384 now merged this should now be ready for review.

## Testing notes:
1. Increment workfile from menu (if enabled in settings should work as intended)
2. Save Workfile action for validator if current file is completely unsaved (has no filename) should work as intended and save a first workfile.

### DIFF
--- a/client/ayon_core/pipeline/context_tools.py
+++ b/client/ayon_core/pipeline/context_tools.py
@@ -16,7 +16,6 @@ from ayon_core.host import HostBase
 from ayon_core.lib import (
     is_in_tests,
     initialize_ayon_connection,
-    version_up
 )
 from ayon_core.addon import load_addons, AddonsManager
 from ayon_core.settings import get_project_settings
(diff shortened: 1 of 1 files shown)


BigRoy:
> Should it keep the comment? I would say that was side effect of version_up functionality. Question is if was expected.

I think for "Version Up" from e.g. Maya menu I'd expect it to preserve it - yes. Essentially just being an increment of version number to a free spot - disregarding anything else? For versioning up from publishing, not sure... I personally think it should there as well. However, when 'versioning up to another context' then suddenly preserving the comment makes less sense.

@philippe-ynput @antirotor @m-u-r-p-h-y @iLLiCiTiT thoughts?

iLLiCiTiT (approved):

PR #1378, opened 2025-07-23
Workfile: Fill extension in helper functions

iLLiCiTiT:
## Changelog Description
Fill `"ext"` key in template data in helper functions.

## Testing notes:
1. Functions `save_next_version` and `copy_workfile_to_context` do work.

### DIFF
--- a/client/ayon_core/pipeline/workfile/utils.py
+++ b/client/ayon_core/pipeline/workfile/utils.py
@@ -411,8 +411,8 @@ def save_next_version(
 ) -> None:
     """Save workfile using current context, version and comment.
 
-    Helper function to save workfile using current context. Last workfile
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Works as expected!

>     # 1. Use the current file extension.
>     # 2. Use the last known workfile extension.
>     # 3. Use the first extensions from 'get_workfile_extensions'.

These cases all behaves as they should.

---

It's good to note that if current workfile has a comment/subversion string - then `save_next_version` will make it lose that comment.

iLLiCiTiT:
> It's good to note that if current workfile has a comment/subversion string - then save_next_version will make it lose that comment.

Opened to a discussion with this one.

PR #1380, opened 2025-07-23
Collect Versions Loaded in Scene: Use `ILoadHost.get_containers()`

### LINKED ISSUE #1379: Use 'ILoadHost' methods in 'CollectSceneLoadedVersions'

### Current Behavior:

`CollectSceneLoadedVersions` plugin should collect loaded containers but is still using deprecated `ls` method which is not required by host implementation.

### Expected Behavior:

There should be used `get_containers` and check for `ILoadHost` inheritance.

The `ls` method should be marked as deprecated.

### Version

1.4.1

Windows

### Steps To Reproduce:

1. No step to repreduce.

### Relevant log output:

```shell

```

### Additional context:

_No response_

BigRoy:
## Changelog Description

Collect Versions Loaded in Scene: Use `ILoadHost.get_containers()`

## Additional info

Fix #1379 

## Testing notes:

1. Collecting versions loaded in scene should work using the new `get_containers` method on the registered host if it is present.

### DIFF
--- a/client/ayon_core/plugins/publish/collect_scene_loaded_versions.py
+++ b/client/ayon_core/plugins/publish/collect_scene_loaded_versions.py
@@ -1,7 +1,9 @@
 import ayon_api
 import ayon_api.utils
 
+from ayon_core.host import ILoadHost
 from ayon_core.pipeline import registered_host
+
 import pyblish.api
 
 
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/publish/collect_scene_loaded_versions.py
```diff
@@ -27,16 +29,27 @@ class CollectSceneLoadedVersions(pyblish.api.ContextPlugin):
...
-            self.log.warn("No registered host.")
+            self.log.warning("No registered host.")
             return
 
-        if not hasattr(host, "ls"):
+        if isinstance(host, ILoadHost):
+            containers = list(host.get_containers())
+        elif hasattr(host, "ls"):
+            # Backwards compatibility for legacy host implementations
+            containers = list(host.ls())
```
iLLiCiTiT, line 39: Is there any host from the hosts list that would not inherit from `ILoadHost`?
```suggestion
```

PR #1382, opened 2025-07-23
Collect Loaded Scene Versions: Enable for more hosts

BigRoy:
## Changelog Description

Collect Loaded Scene Versions: Enable for more hosts

## Additional info

I'm not sure if the hosts filtering is still valid and should be maintained. Likely it's there solely to exclude e.g. apps like Tray Publisher? Anyway, this PR enables some more of the hosts that are available in AYON.

## Testing notes:

1. Publishing in the added DCCs should work with Collecting Loaded Version in Scene.

- [ ] Cinema4D
- [ ] Flame
- [ ] Max
- [ ] Motionbuilder
- [ ] Silhouette
- [ ] Substance Painter
- [ ] Substance Designer
- [ ] Zbrush

### DIFF
--- a/client/ayon_core/plugins/publish/collect_scene_loaded_versions.py
+++ b/client/ayon_core/plugins/publish/collect_scene_loaded_versions.py
@@ -13,15 +13,23 @@ class CollectSceneLoadedVersions(pyblish.api.ContextPlugin):
         "aftereffects",
         "blender",
         "celaction",
+        "cinema4d",
+        "flame",
         "fusion",
         "harmony",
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
> I agree - I'd rather allow it for all hosts. I think it's solely there to exclude the Tray Publisher perhaps? Note that that check is actually already in the process method - it just logs a warning in that case (which may not be nice for tray publisher). So changing that to debug level would essentially resolve that.

I think that would be the best.

PR #1384, opened 2025-07-24
Workfiles: Save next version re-use comment

iLLiCiTiT:
## Changelog Description
Helper function `save_next_version` is re-using comment from current workfile (or last workfile if current is not set).

## Additional info
This behavior should cover most of function use-cases. It is still possible to pass `comment=""` to avoid that behavior. 

Also added helper functions to workfile's `__init__.py` for easier import.

## Testing notes:
1. Using `save_next_version` re-uses comment from current workfile.

### DIFF
--- a/client/ayon_core/pipeline/workfile/utils.py
+++ b/client/ayon_core/pipeline/workfile/utils.py
@@ -417,7 +417,9 @@ def save_next_version(
     Args:
         version (Optional[int]): Workfile version that will be used. Last
             version + 1 is used if is not passed in.
-        comment (optional[str]): Workfile comment.
+        comment (optional[str]): Workfile comment. Pass '""' to clear comment.
+            The last workfile comment is used if it is not passed in and
+            passed 'version' is 'None'.
         description (Optional[str]): Workfile description.
         prepared_data (Optional[SaveWorkfileOptionalData]): Prepared data
             for speed enhancements.
@@ -512,6 +514,9 @@ def save_next_version(
                 task_type=task_entity["taskType"],
                 product_type="workfile"
             )
+
+    if comment is None and last_workfile is not None:
+        comment = last_workfile.comment
 
     template_data["version"] = version
     template_data["comment"] = comment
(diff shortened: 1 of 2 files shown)


client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -417,7 +417,9 @@ def save_next_version(
     Args:
         version (Optional[int]): Workfile version that will be used. Last
             version + 1 is used if is not passed in.
-        comment (optional[str]): Workfile comment.
+        comment (optional[str]): Workfile comment. Pass '""' to clear comment.
+            The last workfile comment is used if it is not passed in and
+            passed 'version' is 'None'.
```
BigRoy, line 422: I wonder if that's really the most sensible behavior - even if I were to pass a version, why wouldn't it also preserve the comment? Not a big hurdle, but sounds confusing? good to see it mentioned in the docstring though.

client/ayon_core/pipeline/workfile/utils.py
```diff
@@ -513,8 +525,18 @@ def save_next_version(
...
             )
 
+    # Re-use comment if is not set
+    if comment is None:
+        if current_workfile is not None:
+            # Use 'comment' from the current workfile if is set
+            comment = current_workfile.comment
+        elif last_workfile is not None:
+            # Use 'comment' from the last workfile
+            comment = last_workfile.comment
```
BigRoy, line 535: I wouldn't have expected this myself, anyway - fine with me. Just want to point out that this is new behavior, different from how the Version Up logic previously worked.

BigRoy (changes requested):
Pretty much works.

However, there's a bug that if my current file is unsaved (no saved filepath) and I specify a version that somehow it gets a double dot before the extension. This only happens if I specify the version argument, e.g. `version=15`.

image


Also, I'd like to point out that when specifying a version there's no check that version either already exists or not. When combining that with a comment then you may be saving a version multiple times with a different comment:

image

I think the behavior is _fine_ in that regard, just wanted to point that out.

client/ayon_core/pipeline/workfile/utils.py
iLLiCiTiT: I actually don't have reason for this. Probably should be removed...

client/ayon_core/pipeline/workfile/utils.py
iLLiCiTiT: Maybe doesn't make sense. My logic was "if you pass in `version`, you define the rest too", maybe that's not right...

PR #1388, opened 2025-07-24
Chore: Add deprecations

iLLiCiTiT:
## Changelog Description
Mark few functions as deprecated. Some of them are considered as deprecated for a long time, some are new.

## Additional info
Function `get_current_project_settings` was moved to `ayon_core.pipeline.context_tools` -> requires access to current project name.
Marked registered root as deprecated, it is considered as deprecated for few years now. With that `get_representation_path` is now using current project anatomy to fill the path, making it slower.
Added host type validation to make sure we don't have some forgotten host without conversion to `HostBase`.
Marked `version_up_current_workfile` as deprecated with `save_next_version` helper function as replacement.

## Testing notes:
1. Validate code changes made here.

### DIFF
--- a/client/ayon_core/pipeline/load/utils.py
+++ b/client/ayon_core/pipeline/load/utils.py
@@ -720,11 +720,13 @@ def get_representation_path(representation, root=None):
         str: fullpath of the representation
 
     """
-
     if root is None:
-        from ayon_core.pipeline import registered_root
+        from ayon_core.pipeline import get_current_project_name, Anatomy
 
-        root = registered_root()
+        anatomy = Anatomy(get_current_project_name())
+        return get_representation_path_with_anatomy(
+            representation, anatomy
+        )
 
     def path_from_representation():
         try:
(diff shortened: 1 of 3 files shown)


client/ayon_core/pipeline/load/utils.py
```diff
@@ -720,11 +720,13 @@ def get_representation_path(representation, root=None):
...
-
     if root is None:
-        from ayon_core.pipeline import registered_root
+        from ayon_core.pipeline import get_current_project_name, Anatomy
 
-        root = registered_root()
+        anatomy = Anatomy(get_current_project_name())
+        return get_representation_path_with_anatomy(
+            representation, anatomy
+        )
```
BigRoy, line 729: I can't find any `get_representation_path()` query where we are passing the `root`. Which means that essentially with this change this now always would go through this logic.

Which makes me think, is all the code of "switching between how we're detecting the filepath" below still relevant in any way?

BigRoy (approved):
LGTM - just not sure about [this change](https://github.com/ynput/ayon-core/pull/1388/files#r2228592617)?

client/ayon_core/pipeline/load/utils.py
BigRoy: Fair point.

PR #1394, opened 2025-07-28
Ensure OTIO video triming is duration accurate (for re-encoding to avoid keyframe rounding issue).

rdelillo:
## Changelog Description

resolve: https://github.com/ynput/ayon-hiero/issues/77

When trimming video from OTIO timeline, we use ffmpeg with `-c copy` settings.
This means copy the streams (audio + video) without re-encoding, which is very performant but only allows to cut at video keyframes.
The issue with this approach is that the resulting trimmed video duration is not safely guaranteed: 
* https://superuser.com/questions/924223/ffmpeg-no-cutting-accurately
* https://video.stackexchange.com/questions/16750/using-ffmpeg-to-cut-videos-with-more-precision-than-key-frames-allow

## Additional info

## Testing notes:
1. Create a timeline in Hiero or Resolve
2. Create a clip based on video media which requires trimming with no handles
3. Publish the resulting plate with review
4. Ensure the plate + review videos duration match the source clip in the timeline

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,6 +35,7 @@ body:
       label: Version
       description: What version are you running? Look to AYON Tray
       options:
+        - 1.8.6
         - 1.8.5
         - 1.8.4
         - 1.8.3
--- a/client/ayon_core/hooks/pre_global_host_data.py
+++ b/client/ayon_core/hooks/pre_global_host_data.py
@@ -85,6 +85,12 @@ class GlobalHostDataHook(PreLaunchHook):
             return
 
         folder_entity = get_folder_by_path(project_name, folder_path)
+        if not folder_entity:
+            raise ApplicationLaunchFailed(
+                f"Folder '{folder_path}' not found"
+                f" in project '{project_name}'."
+            )
+
         self.data["folder_entity"] = folder_entity
 
         task_name = self.data.get("task_name")
@@ -100,4 +106,10 @@ class GlobalHostDataHook(PreLaunchHook):
         task_entity = get_task_by_name(
             project_name, folder_entity["id"], task_name
         )
+        if not task_entity:
+            raise ApplicationLaunchFailed(
+                f"Task '{task_name}' not found"
+                f" in '{project_name}' > '{folder_path}'."
+            )
+
         self.data["task_entity"] = task_entity
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -20,6 +20,7 @@ from ayon_core.pipeline import (
 from ayon_core.pipeline.create import get_product_name
 from ayon_core.pipeline.farm.patterning import match_aov_pattern
 from ayon_core.pipeline.publish import KnownPublishError
+from ayon_core.pipeline.publish.input_versions import serialize_input_versions
 
 log = Logger.get_logger(__name__)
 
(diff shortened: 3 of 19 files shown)


client/ayon_core/plugins/publish/extract_otio_trimming_video.py
```diff
@@ -89,7 +89,6 @@ def _ffmpeg_trim_seqment(self, input_file_path, otio_range):
             "-ss", str(sec_start),
             "-t", str(sec_duration),
             "-i", video_path,
-            "-c", "copy",
```
rdelillo, line 92: @jakubjezek001 @iLLiCiTiT @philippe-ynput @antirotor 

I'm torn by this change. Using the `-c copy` stream option when trimming the video get us codec accurate sub-video with potential duration mismatch (see PR description). 
While removing this as I do force a re-encoding, which gives frame-accurate video but through different codec than source (ffmpeg default is x264) which might not be what we want here.

iLLiCiTiT: I guess the codec copy is important.

> I could go with a more sophisticated approach is and detect codec from source than attempt to re-encode in same codec but not all codec variations are natively supported by FFmpeg and we could still end-up with subtle variations (color, gamma, bitrate...) compare to copy.

There is `get_ffmpeg_codec_args` to do so.

philippe-ynput: The accuracy of the cut is probably more important than the image quality, assuming the re-encoding will not totally destroy the image quality.

client/ayon_core/plugins/publish/extract_otio_trimming_video.py
```diff
@@ -89,10 +88,29 @@ def _ffmpeg_trim_seqment(self, input_file_path, otio_range):
...
+        # https://video.stackexchange.com/questions/16750/
+        ffprobe_data = get_ffprobe_data(input_file_path, self.log)
+        video_codec_args = get_ffmpeg_codec_args(ffprobe_data)
+        if not video_codec_args:
+            self.log.warning(
+                "FFmpeg could not identify the video codec for %s. "
+                "Use '-c copy' which might lead in duration mismatches.",
+                input_file_path,
+            )
+            video_codec_args = ["-c", "copy"]
```
rdelillo: @jakubjezek001 I'm wondering what's best here.
If video codec cannot be identified by FFmpeg, should we attempt a `-c copy` and make it backward compatible or shall we just raise with an not-supported input media ?

client/ayon_core/plugins/publish/extract_otio_trimming_video.py
```diff
@@ -89,10 +88,29 @@ def _ffmpeg_trim_seqment(self, input_file_path, otio_range):
...
+        # the same codec. This is the only way to ensure a precise
+        # output duration.
+        # '-c copy' make FFmpeg rely on key I-frames, not 100% accurate.
+        # https://video.stackexchange.com/questions/16750/
+        ffprobe_data = get_ffprobe_data(input_file_path, self.log)
+        video_codec_args = get_ffmpeg_codec_args(ffprobe_data)
+        if not video_codec_args:
+            self.log.warning(
+                "FFmpeg could not identify the video codec for %s. "
+                "Use '-c copy' which might lead in duration mismatches.",
```
BigRoy: Incorrect english? 

What do you mean with "Use '-c copy' which might lead in duration mismatches."?

Did you mean something like:
```suggestion
                "Falling back to '-c copy' which may lead to"
                " duration mismatches.",
```

client/ayon_core/plugins/publish/extract_otio_trimming_video.py
jakubjezek001: if we use PNGs with some solid  high specs as intermediate only for those cases where codecs is ungnown, we should be covered, right? 

To be honest - not sure how known codec with higher GOP frames will be doing either. I would rather convert all but it might be slowing down the process lot.

Issue #874, opened 2024-09-06
Publisher editorial linked instances with grouping view


jakubjezek001:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Traypublisher's editorial instances can be quite challenging to manage, especially when dealing with multiple instances for different shots. For instance, ShotA might have associated instances for Plate, Audio, and Reviewable. To add to the complexity, we also need to create an additional 'parent' instance for ShotA, categorized as a product type shot. This parent instance allows for user-configurable attributes related to the shot, which are then distributed to other connected instances like Plate or Audio during publishing.

Some productions might work on a single layer with roughly 100 shots, leading us to manage around 200 instances. However, other productions may use a multi-track workflow with vertically aligned clips, potentially increasing the number of instances to 500 or more.

### How would you imagine the implementation of the feature?

Let's categorize the instances and their connections:

*   **Shot Instance (SI)** - This holds metadata, some of which may be editable and should be shared with other linked instances.
*   **Main Layer Linked Clip Instance (MLLCI)** - These are Plate/Audio/StillFrame instances that do not have editable shot-related attributes.
*   **Vertically Aligned Linked Clip Instances (VALCI)** - Similar to MLLCI, these are Plate/StillFrame instances.

Suggested UX improvements:

*   An enhanced view for instances that includes a grouped mode. In this mode, MLLCI and VALCI would be organized under the SI, similar to the image below.
![image](https://github.com/user-attachments/assets/ec31be7d-16a6-404b-a195-34f44689e9a6)

*   Enhanced instance context view for MLLCI and VALCI types. Users can currently override Folder and Task, but it would be better to make these non-editable.

*   Differentiate linked instances visually using color, icons, or widget size.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

iLLiCiTiT:
@jakubjezek001 @dee-ynput @philippe-ynput

jakubjezek001:
- Shot Instance (SI) - This holds metadata, some of which may be editable and should be shared with other linked instances.
- Main Layer Linked Clip Instance (MLLCI) - These are Plate/Audio/StillFrame instances that do not have editable shot-related attributes.
- Vertically Aligned Linked Clip Instances (VALCI) - Similar to MLLCI, these are Plate/StillFrame instances.

```
[...]

PR #1395, opened 2025-07-30
AY-7790 Create: Parenting of instances

iLLiCiTiT:
## Changelog Description
Create plugins can parent an instance. The parented instances can define flags like share disabled state or share lifetime.

## Additional info
Parenting of instance is set on instance object with flags that define their behavior related to the parent. By default it is just shown as children in UI. If shared disabled state is set then disabling parent will also disable the instance. If lifetime is shared then removement of the parent will also remove the instance from context.

Instances that do have set parent id but parent is not available are considered as "invalid" and are disabled.

## Technical details
This PR is introducing an api and UI modifications required to achieve the final result. The source issue is dedicated to editorial parenting, but the parenting can be used for multiple different use-cases that might help to visually organize instances (e.g. render layer -> render pass instances). It is meant as backend first and UI second.

I did add an option to share disabled state, but at the same time I didn't want to loose the value of active state of the children -> If you disable and then enable a Shot instance you probably want to keep previous active states of all children (plate, audio, render etc.). That has a downside that to get "real" active state, one would have to go top to bottom to get it. There is no helper function for that right now. Let me know if should be added and how it should work. NOTE Because it is dynamic data and instance object is meant as dataholder it would be hard to maintain that using `CreatedInstance` object directly.

The parenting is shown only in list view, card view only disables respected instances by parenting logic. The view needed to be changed as the previous look (and it's implementation) was tightly connected to fact that they don't have children instances.

NOTE I didn't test all possible use-cases but what I tested was working ok. For example I didn't test change of parent id in a single session as I didn't have a use-case. Let me know if you hit something.

WARNING Both views were heavily re-written so we have to validate existing functionality too.

Opened to suggestions and comments.

## Screenshots
image

## Testing notes:
- First of all, all existing functionality should be working. You can work with instances as before: create/remove, change label, context, enable/disable them.
- You can test with Simple editorial workflow in traypublisher https://github.com/ynput/ayon-traypublisher/pull/98 .

resolves https://github.com/ynput/ayon-core/issues/874


jakubjezek001 (approved):
works perfect

PR #1399, opened 2025-07-31
Addons: Plugin paths interface

iLLiCiTiT:
## Changelog Description
First step to remove `get_plugin_paths` method from `IPluginPaths`.

## Additional info
The method was over the time expended to separate methods for each plugin type, now the specific methods do expect more arguments (usually host name) and the `get_plugin_paths` did become deprecated at that moment and also is complicating the readability with related `AddonsManager` method `collect_plugin_paths`.

First of all the method is not abstract anymore, so it is not required to be implemented. Added warning everytime the method is called and returns paths.

Added warning if the output of plugin path getters is string instead of list of strings.

## How to prepare addon for full deprecation
First of all keep `get_plugin_paths` to keep support with older core addon but return empty dictionary. Reimplement `get_create_plugin_paths`, `get_load_plugin_paths`, `get_inventory_action_paths` or `get_publish_plugin_paths` if you want to register paths for the plugin type.

Method `get_launcher_action_paths` is in grey zone now as webactions should be used instead, but some features can't be achieved with webactions at this moment, but is still available.

## Testing notes:
1. Validate the code and logic changes.
2. All plugins in your addons are registered and available.

### DIFF
--- a/client/ayon_core/addon/interfaces.py
+++ b/client/ayon_core/addon/interfaces.py
@@ -61,7 +61,8 @@ class IPluginPaths(AYONInterface):
         return {}
 
     def _get_plugin_paths_by_type(
-            self, plugin_type: str) -> list[str]:
+        self, plugin_type: str
+    ) -> list[str]:
         """Get plugin paths by type.
 
         Args:
--- a/client/ayon_core/addon/interfaces.py
+++ b/client/ayon_core/addon/interfaces.py
@@ -1,6 +1,7 @@
 """Addon interfaces for AYON."""
 from __future__ import annotations
 
+import warnings
 from abc import ABCMeta, abstractmethod
 from typing import TYPE_CHECKING, Callable, Optional, Type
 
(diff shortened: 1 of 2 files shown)


kalisp (approved):
Seems legit. Tested with SiteSync addon which implemented `get_plugin_paths`, works with just `get_publish_plugin_paths`.

client/ayon_core/addon/interfaces.py
```diff
@@ -78,6 +82,25 @@ def _get_plugin_paths_by_type(
...
+        elif plugin_type == "inventory":
+            new_function_name = "get_inventory_action_paths"
+
+        warnings.warn(
+            f"Addon '{self.name}' returns '{plugin_type}' paths using"
+            " 'get_plugin_paths' method. Please implement"
+            f" '{new_function_name}' instead.",
+            DeprecationWarning,
+            stacklevel=2
+
```
iLLiCiTiT, line 102:
```suggestion
```

PR #1400, opened 2025-08-01
Editorial: Added safe importing of opentimelineio

iLLiCiTiT:
## Changelog Description
Don't import otio module if it is not necessary.

## Additional info
Moved imports that do need OTIO into the functions that do use them.

## Testing notes:
1. No crashed plugins because of missing `opentimelineio` in any DCC that does not have it (e.g. maya).


MustafaJafar (approved):
The import errors vanished.
image

Issue #1408, opened 2025-08-11
AY-8079_Hero versioning


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

Create a Loader action/web action that will take selected version and make hero version from it. There could be settings like "approve and create hero version" to automate more the workflow.

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-8079](https://app.clickup.com/t/86c4z20x3)_

PR #1415, opened 2025-08-15
Supports hero versioning in the loader action

moonyuet:
## Changelog Description
This PR is to add the loader action of `create hero version` in browser tool.
Fixes https://github.com/ynput/ayon-core/issues/1408
image

## Additional info
Should we allow creating hero version only for certain status?

## Testing notes:
1. Right-click launch tool action -> Browser
2. Right click the instance
3.  Select `create hero version`
image

4. Once it is created it would show [your instance version] in browser and Hero in browser in ayon server.
image


moonyuet:
Tested with updated commit, looks good
image

BigRoy:
Can we update the PR title to be more descriptive, so that the changelog becomes clearer about what this feature is? It doesn't hint to that it adds a "Create hero version button in the tray loader"?

> This PR is to add the loader action of create hero version in browser tool.

(This is what the PR description states which I think is much clearer than the PR title)

PR #1417, opened 2025-08-18
Multiple products could be pushed to another project

kalisp:
## Changelog Description
This PR implements support for pushing (copying) multiple product items from source project into destination (library).
It introduces new toggle "Use original product names" which bypasses calculation of new product name.

It expects products only from single folder!

## Testing notes:
1. try to push multiple products to library or any other project
2. toggle `Use original product names` to bypass folder name and variant field validations.

### DIFF
--- a/client/ayon_core/tools/push_to_project/ui/window.py
+++ b/client/ayon_core/tools/push_to_project/ui/window.py
@@ -501,9 +501,8 @@ class PushToContextSelectWindow(QtWidgets.QWidget):
         if self._controller.original_names_required():
             use_original_names = True
 
-        if use_original_names:
-            self._variant_input.setEnabled(not use_original_names)
-            self._invalidate_variant(not use_original_names)
+        self._variant_input.setEnabled(not use_original_names)
+        self._invalidate_variant(not use_original_names)
 
         self._controller._use_original_name = use_original_names
         self._use_original_names_checkbox.setChecked(use_original_names)


client/ayon_core/tools/push_to_project/main.py
```diff
@@ -4,28 +4,28 @@
...
     """Run PushToProject tool to integrate version in different project.
 
     Args:
         project (str): Source project name.
-        version (str): Version id.
+        versions (str): comma separated versions for same context
     """
 
-    main_show(project, version)
+    main_show(project, versions)
```
iLLiCiTiT, line 28:
```suggestion
    main_show(project, versions.split(","))
```

client/ayon_core/tools/push_to_project/control.py
```diff
@@ -27,18 +27,18 @@ def __init__(self, project_name=None, version_id=None):
...
+        self._src_product_entities = []
+        self._src_version_entities = []
         self._src_label = None
 
         self._submission_enabled = False
         self._process_thread = None
         self._process_item_id = None
 
-        self.set_source(project_name, version_id)
+        self._use_original_name = False
```
iLLiCiTiT, line 41:
```suggestion
        self._use_original_name = False
        
        self.set_source(project_name, version_ids)
```

client/ayon_core/tools/push_to_project/control.py
```diff
@@ -51,38 +51,44 @@ def emit_event(self, topic, data=None, source=None):
...
-        if project_name and version_id:
-            version_entity = ayon_api.get_version_by_id(
-                project_name, version_id
-            )
-
-        if version_entity:
-            product_entity = ayon_api.get_product_by_id(
-                project_name, version_entity["productId"]
+        product_entities = None
+        version_entities = None
```
iLLiCiTiT, line 75:
```suggestion
        product_entities = []
        version_entities = []
```

client/ayon_core/tools/push_to_project/control.py
```diff
@@ -97,23 +103,23 @@ def set_source(self, project_name, version_id):
 
         self._src_folder_entity = folder_entity
         self._src_folder_task_entities = task_entities
-        self._src_product_entity = product_entity
-        self._src_version_entity = version_entity
-        if folder_entity:
+        self._src_product_entities = product_entities
+        self._src_version_entities = version_entities
+        if folder_entity and len(list(version_entities)) == 1:
```
iLLiCiTiT, line 108: Why only when one version is selected?
```suggestion
        if folder_entity:
```

client/ayon_core/tools/push_to_project/control.py
```diff
@@ -207,22 +217,34 @@ def wait_for_process_thread(self):
...
+        if not self._src_project_name or not self._src_version_ids:
             return "Source is not defined"
 
         folder_entity = self._src_folder_entity
         if not folder_entity:
             return "Source is invalid"
 
+        no_of_products = len(self._src_product_entities)
+        no_of_versions = len(self._src_version_entities)
+        if no_of_products != no_of_versions:
```
iLLiCiTiT, line 229: This doesn't make any sense? Why would you require to have the same number of versions and products? And how do you know the order does match? You should find product entity based on `"productId"` stored on version entity.

client/ayon_core/tools/push_to_project/control.py
```diff
@@ -338,13 +363,9 @@ def _invalidate(self):
         )
 
     def _submit_callback(self):
-        process_item_id = self._process_item_id
-        if process_item_id is None:
-            return
-        self._integrate_model.integrate_item(process_item_id)
+        for process_item_id in self._process_item_ids:
+            self._integrate_model.integrate_item(process_item_id)
         self._emit_event("submit.finished", {})
```
iLLiCiTiT, line 368:
```suggestion
        process_item_ids = self._process_item_ids
        for process_item_id in process_item_ids:
            self._integrate_model.integrate_item(process_item_id)
        
        self._emit_event("submit.finished", {})

        if process_item_ids is self._process_item_ids:
            self._process_item_ids = []
```

LiborBatek (changes requested):
Now it correctly validates if multiple folders been selected and fails due to that which is correct.

Screenshot 2025-08-22 104220

However the issue is this happens also when just single folder been selected but multiple products of it...it shouldnt fail and user could push to library without any issues (like commits before)


Screenshot 2025-08-22 104315

LiborBatek:
Speaking of the actual pushing action...I would prefer to have  ```Use original product name``` as ```ENABLED``` by defaults instead of not.

So changing the default settings might do too IMHO.
Screenshot 2025-08-22 104727

LiborBatek (approved):
Works for me now under all circumstances...

LGTM

PR #1421, opened 2025-08-25
Loader: Actions plugin

iLLiCiTiT:
## Changelog Description
Implement of loader actions -> actions that can be used in loader tool. The implementation is meant as replacement for all loader plugins that are not used for loading.

## Additional info
This is one of steps for future load API enhancements. A loader action plugin can define multiple actions that are returned for a selected context. The selected context can be enhanced in future to allow more features, e.g. more nested grouping of actions, selection on folder or task etc.

Loader actions can have multiple steps for their execution, the result can contain form that is shown to user, or a message that is shown as popup message in loader window (or both).

## Testing notes:
1. Validate code changes.
2. Existing actions do work.
- [x] Delete versions
- [x] Calculate delete versions
- [x] Copy file
- [x] Copy file path
- [x] Delivery
- [x] Export OTIO
- [x] Push to project

### DIFF
--- a/server/settings/conversion.py
+++ b/server/settings/conversion.py
@@ -1,7 +1,32 @@
+import re
 import copy
 from typing import Any
 
 from .publish_plugins import DEFAULT_PUBLISH_VALUES
+
+PRODUCT_NAME_REPL_REGEX = re.compile(r"[^<>{}\[\]a-zA-Z0-9_.]")
+
+
+def _convert_imageio_configs_1_6_5(overrides):
+    product_name_profiles = (
+        overrides
+        .get("tools", {})
+        .get("creator", {})
+        .get("product_name_profiles")
+    )
+    if isinstance(product_name_profiles, list):
+        for item in product_name_profiles:
+            # Remove unsupported product name characters
+            template = item.get("template")
+            if isinstance(template, str):
+                item["template"] = PRODUCT_NAME_REPL_REGEX.sub("", template)
+
+            for new_key, old_key in (
+                ("host_names", "hosts"),
+                ("task_names", "tasks"),
+            ):
+                if old_key in item:
+                    item[new_key] = item.get(old_key)
 
 
 def _convert_imageio_configs_0_4_5(overrides):
--- a/server/settings/tools.py
+++ b/server/settings/tools.py
@@ -25,16 +25,27 @@ class ProductNameProfile(BaseSettingsModel):
     _layout = "expanded"
 
     product_types: list[str] = SettingsField(
-        default_factory=list, title="Product types"
+        default_factory=list,
+        title="Product types",
     )
-    hosts: list[str] = SettingsField(default_factory=list, title="Hosts")
+    host_names: list[str] = SettingsField(
+        default_factory=list,
+        title="Host names",
+    )
     task_types: list[str] = SettingsField(
         default_factory=list,
         title="Task types",
-        enum_resolver=task_types_enum
+        enum_resolver=task_types_enum,
     )
-    tasks: list[str] = SettingsField(default_factory=list, title="Task names")
-    template: str = SettingsField("", title="Template")
+    task_names: list[str] = SettingsField(
+        default_factory=list,
+        title="Task names",
+    )
+    template: str = SettingsField(
+        "",
+        title="Template",
+        regex=r"^[<>{}\[\]a-zA-Z0-9_.]+$",
+    )
 
 
 class FilterCreatorProfile(BaseSettingsModel):
@@ -433,27 +444,27 @@ DEFAULT_TOOLS_VALUES = {
         "product_name_profiles": [
             {
                 "product_types": [],
-                "hosts": [],
+                "host_names": [],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "{product[type]}{variant}"
             },
             {
                 "product_types": [
                     "workfile"
                 ],
-                "hosts": [],
+                "host_names": [],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "{product[type]}{Task[name]}"
             },
             {
                 "product_types": [
                     "render"
                 ],
-                "hosts": [],
+                "host_names": [],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "{product[type]}{Task[name]}{Variant}<_{Aov}>"
             },
             {
@@ -461,11 +472,11 @@ DEFAULT_TOOLS_VALUES = {
                     "renderLayer",
                     "renderPass"
                 ],
-                "hosts": [
+                "host_names": [
                     "tvpaint"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": (
                     "{product[type]}{Task[name]}_{Renderlayer}_{Renderpass}"
                 )
@@ -475,65 +486,65 @@ DEFAULT_TOOLS_VALUES = {
                     "review",
                     "workfile"
                 ],
-                "hosts": [
+                "host_names": [
                     "aftereffects",
                     "tvpaint"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "{product[type]}{Task[name]}"
             },
             {
                 "product_types": ["render"],
-                "hosts": [
+                "host_names": [
                     "aftereffects"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "{product[type]}{Task[name]}{Composition}{Variant}"
             },
             {
                 "product_types": [
                     "staticMesh"
                 ],
-                "hosts": [
+                "host_names": [
                     "maya"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "S_{folder[name]}{variant}"
             },
             {
                 "product_types": [
                     "skeletalMesh"
                 ],
-                "hosts": [
+                "host_names": [
                     "maya"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "SK_{folder[name]}{variant}"
             },
             {
                 "product_types": [
                     "hda"
                 ],
-                "hosts": [
+                "host_names": [
                     "houdini"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "{folder[name]}_{variant}"
             },
             {
                 "product_types": [
                     "textureSet"
                 ],
-                "hosts": [
+                "host_names": [
                     "substancedesigner"
                 ],
                 "task_types": [],
-                "tasks": [],
+                "task_names": [],
                 "template": "T_{folder[name]}{variant}"
             }
         ],


iLLiCiTiT:
> now there is only Hero action.

That action was created in between of this PR creation and current time. The plugin is too complex to convert it in this PR, deservers separate PR and testing.

kalisp:
I couldn't test `Delivery` and `OTIO` actions as they weren't shown. Maybe I am missing required configuration to show them?

client/ayon_core/plugins/loader/copy_file.py
```diff
@@ -0,0 +1,115 @@
...
+
+class CopyFileActionPlugin(LoaderActionPlugin):
+    """Copy published file path to clipboard"""
+    identifier = "core.copy-action"
+
+    def get_action_items(
+        self, selection: LoaderActionSelection
+    ) -> list[LoaderActionItem]:
+        repres = []
+        if selection.selected_type == "representations":
```
BigRoy, line 23: Why is "represenstations" type plural with trailing `s` but "version" isn't?

iLLiCiTiT: Typo...

iLLiCiTiT: Fixed.

iLLiCiTiT:
> I couldn't test Delivery and OTIO actions as they weren't shown. Maybe I am missing required configuration to show them?

There was a bug ther, should be fixed now.

client/ayon_core/plugins/loader/copy_file.py
```diff
@@ -0,0 +1,113 @@
...
+        self,
+        identifier: str,
+        selection: LoaderActionSelection,
+        data: dict,
+        form_values: dict[str, Any],
+    ) -> Optional[LoaderActionResult]:
+        from qtpy import QtWidgets, QtCore
+
+        repre_ids = data["representation_ids"]
+        repre = next(iter(selection.entities.get_representations(repre_ids)))
```
BigRoy, line 80: Won't this falsely give the promise that this will actually work with multiple representations when it doesn't.
Also, if we know we'll only really be using the last entity only - why still query them all? Why not just `repre_ids[-1]`?

client/ayon_core/plugins/loader/delete_old_versions.py
```diff
@@ -0,0 +1,390 @@
...
+                    title="Calculated versions size",
+                    fields=[
+                        UILabelDef(
+                            f"Total size of files: {format_file_size(size)}"
+                        ),
+                    ],
+                    submit_label=None,
+                    cancel_label="Close",
+                ),
+            )
```
BigRoy, line 252: I'm unsure myself about abstracting this away into custom Form classes, etc. What benefit do we really get from enforcing user prompts to be pushed out as a return value here? Why can't this just be a custom Qt dialog? I feel like as soon as we get to only being able to use strict Attribute Defs we're limiting what an action could essentially show as a UI to the user... and most devs will know Qt in our industry, but _everyone_ would need to read and learn this 'custom form designing API' we're introducing as  a return value here. Also, what if I want the form to respond dynamically, etc.

BigRoy (commented):
Can someone explain to me why the `get_action_items` system is needed? Is this an optimization aspect as we don't want to clutter with many plug-ins each representation a single action? Or do we expect highly dynamic lists of actions in some way or another?

Similarly with the `LoaderActionResult` return values with a `LoaderActionForm` class, etc. - Why can't an action just build a simple Qt UI to prompt the user if needed?

I'm just wondering whether the whole complexity of needing to write code with all these `action_items` aspects, etc. is really worth the hassle of not just writing a few more much smaller plug-ins instead (which can still inherit from one another to share logic, etc.)

What do we really gain from adding this _extra_ layer which seems to obfuscate.

Don't get me wrong, I like the fact that actions run over all selected entries, and that they get the data class as input, etc. but why all the surrounding mumbo jumbo?

It takes me a bit back to what @antirotor said in the call today - where is the documentation? Which admittedly, if half of the layers weren't there I hardly feel that question is as relevant since it's just actions running Qt dialogs that everyone in our industry (or at least devs) know a bit about or at least has plenty of documentation, AI can even code pretty well and allows for way more control then our forms. Why do we need to abstract that away?

iLLiCiTiT:
Can someone explain to me why the get_action_items system is needed? Is this an optimization aspect as we don't want to clutter with many plug-ins each representation a single action? Or do we expect highly dynamic lists of actions in some way or another?

Similarly with the LoaderActionResult return values with a LoaderActionForm class, etc. - Why can't an action just build a simple Qt UI to prompt the user if needed?

I'm just wondering whether the whole complexity of needing to write code with all these action_items aspects, etc. is really worth the hassle of not just writing a few more much smaller plug-ins instead (which can still inherit from one another to share logic, etc.)

What do we really gain from adding this extra layer which seems to obfuscate.

Don't get me wrong, I like the fact that actions run over all selected entries, and that they get the data class as input, etc. but why all the surrounding mumbo jumbo?

It takes me a bit back to what @antirotor said in the call today - where is the documentation? Which admittedly, if half of the layers weren't there I hardly feel that question is as relevant since it's just actions running Qt dialogs that everyone in our industry (or at least devs) know a bit about or at least has plenty of documentation, AI can even code pretty well and allows for way more control then our forms. Why do we need to abstract that away?

It feels more complicated than it needs to be. I think this more explanation of what it solves or tried to solve and why the changes are for the better.

It seems the data of those forms are really for the action itself - they are not relevant elsewhere? Why can't it be a part of the action? etc.

1. This implementation allows you to use (and import) the loader actions without having Qt available.
2. You can still implement Qt dialog and handle it on your own. But most of the dialogs we have implemented in loader pugins are simple dialogs working exactly the same way as the form does. For example delivery action still has custom dialog. In some actions we actually are not using dialogs because it it actually too much work to create, compared to the form definition.
3. Using `get_action_items` allows your to explicitly tell what actions are actually shown, per product/version/representation or different variants of the action. For me this is the most benefial as I can explicitly say, with a code, if the action is shown or not, and how it is shown, what variants are shown, based on selection.

Currently we don't have way how to define when the loader plugin is shown. There is `LoaderPlugin` and `ProductLoaderPlugin`. `LoaderPlugin` is tight to representation, but is also shown on products/versions and `ProductLoaderPlugin` is tight to versions (even when is called product loader) and is shown on product/versions.

We do lack "real" `VersionLoaderPlugin` -> version related action. But that still means you can't define if action is shown when representation is selected, but not show the action if version is selected -> that's why site sync has different behavior in the controller, because we want to show them only when showing actions in representations widget. With this implementation the site sync actions can be moved to site sync addon. We have more use-cases like this, but all the time when someone asks for it we just say "Yeah, we can't do that.".

We do want to add folder and task actions too, what would we do then? Add `LoaderFolderPlugin`? If that would share logic with version action and representation action how would you handle the inheritance?

This is actions handling, not load plugin handling. The way how it is implemented is 1:1 how webaction do work BTW. This same logic is also used in launcher (but I had to wrap it more to keep backwards compatibility there).

BigRoy:
> Currently we don't have way how to define when the loader plugin is shown. There is `LoaderPlugin` and `ProductLoaderPlugin`. `LoaderPlugin` is tight to representation, but is also shown on products/versions and `ProductLoaderPlugin` is tight to versions (even when is called product loader) and is shown on product/versions.

Sure - I'm fine with changing that. But why can't it just be an `is_compatible` check that takes the current selection and defines whether the current plug-in will show. Why does it need builds lists of action manifests, etc. Why can't they each be plug-ins if you need multiple actions?

So - yes, I'm all with you here on allowing it. I'm just saying, what makes it so that it's required to be implemented in this complex fashion. Why can't it just be:
```python
class MyAction(...):
    def is_compatible(self, selection: LoaderActionSelection) -> bool:
        """Return whether action shows for current selection."""
        if selection.selected_type() == "version":
            return True

    def process(self, selection: LoaderActionSelection):
        """Run the action over all needed entities in the selection."""
        for version_entity in selection.versions:
            self._process_version(version_entity)
```

> So you can have multistep actions, like delete/calculate versions. It first shows form asking you for more options, and when finishes shows different form with the result. Again the same way how webactions do work, and how ftrack actions do work. You can use them or you can skip it if you don't need it.

Why can't the action's process just prompt the user 'multi-step' using Qt UIs? What does the whole layer of "form results" etc really help to do here? Do we really need that data anywhere else than just for displaying something to the user?

Or is this all with the idea that all these actions instead will be running ON the web frontend? (Which seems overkill?)

> What this might allow is to change how load plugins are shown on representations, and it can be dynamically changed based on user's preferences, but it can be changed only for load plugins and not for others.

Not entirely sure what you mean here. What preferences, where?

---

Also instead of needing to write this:
```python
            repres = selection.entities.get_versions_representations(
                selection.selected_ids
            )
```

I think a default helper like this would go a long way:
```python
            if selection.has_selected_versions():
                repres = selection.get_selected_versions()
```
Or:
```python
            if selection.selected_type() == "version":
                repres = selection.get_selected_versions()
```
But admittedly I'm not to keen on the direct 'string' comparison there.
Perhaps this is better or something along these lines:
```python
            if selection.selected_type() == selection.VersionType:
                repres = selection.get_selected_versions()
```

I think that we're passing `selected_ids` which can be ids for any arbitrary type just makes it a bit confusing because you first need to check type, then know which function to use, etc.

client/ayon_core/plugins/loader/copy_file.py
iLLiCiTiT: > Won't this falsely give the promise that this will actually work with multiple representations when it doesn't.

I didn't change the logic, I just converted the logic to action plugin. This is exactly what happened before.

BigRoy:
> Everytime I have to show a simple dialog showing some information or inputs I actually do spend a lot of time doing it correctly, for me the form simplifies a lot. Using attribute definitions is a lot simpler and like I wrote, YOU DON"T HAVE TO USE IT! What I don't understand is why it bothers you so much? You prefer custom Qt dialog, I get it, but I don't if I don't have to. It simplified delete/calculate version action code a lot and the dialog looks exactly the same.

I'm mostly saying that due to this being there it makes this whole loader/action plug-in structure much more complicated visually. It really feels like _I need to read up on how this works._ Whereas if it's not there, it's so light weight and I can rely on existing knowledge instead. You _like_ to use it, mostly because you implemented it and know all the functionality. But admittedly from my perspective it's just more stuff I need to learn - and more stuff you need to maintain over time?

> Whatever, e.g. site settings or local user settings to change how load plugins are shown for example, if are shown grouped or if are shown flat. Or hide actions because it is artist user. That's up to the implementation of the actions and what we decide to do.

I may have missed it here - but how can I define an action that's non-flat structured? I don't see how it allows that now?

> Yes, which is what I was talking about, some actions should be shown only if actions are shown on versions, or only on representations, or can be shown at both. By adding task and folder this extends a lot. For example delivery action or delete old versions can be shown on all of the views. But copy path does not make sense on folder or task. It might even be valid to show an action if only one entity is selected or if combination of entities is available on selection.

So it's merely an optimization that e.g. two plug-ins can't both just run `is_compatible` with one being for versions and one for representations. Or even just one plug-in, returning True for both types and the `process` method can change how it applies based on the selection. The whole "action identifier"-middle man seems redundant and just added complexity.

---

Again I'm merely saying there's a lot of bloat here. If the "time to make a Qt dialog" is where you often spend a lot of time _why does that need loader specific forms logic_? Why can't we just use some generic attribute definitions dialog that's much more separate from loader actions?

It just feels like it's trying to do too much in the one system where it doesn't need to.

BigRoy:
> Did tou actually look at the load plugins code, or test it?

I did not test it. I did review the code.

iLLiCiTiT:
> The whole "action identifier"-middle man seems redundant and just added complexity.

So if I want to show per-representations actions on versions selection. How to define multiple actions to show? And how would you then in process know which of the actions was triggered if you have only selection (where is information about version, not about representations).

> Again I'm merely saying there's a lot of bloat here. If the "time to make a Qt dialog" is where you often spend a lot of time why does that need loader specific forms logic? Why can't we just use some generic attribute definitions dialog that's much more separate from loader actions?

I've write it multiple times. This is 1:1 to what webactions do. The exact same logic is implemented in launcher, it is NOT loader specific. If gives more options on how to handle the result of an action in UI. Is it that you don't want to use it? Or you don't like it? What's the problem? I still don't understand why it bothers you. Yes it make it more complex, which is what is usually used in actions, and it is optional...

> I may have missed it here - but how can I define an action that's non-flat structured? I don't see how it allows that now?

You can add `group_label` to action items, for example.

client/ayon_core/plugins/loader/open_file.py
```diff
@@ -0,0 +1,132 @@
...
+        if not repres:
+            return []
+
+        repre_ids = {repre["id"] for repre in repres}
+        versions = selection.entities.get_representations_versions(
+            repre_ids
+        )
+        product_ids = {version["productId"] for version in versions}
+        products = selection.entities.get_products(product_ids)
+        fitlered_product_ids = {
```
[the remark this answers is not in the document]
iLLiCiTiT: You don't like fitlering? ❤️

iLLiCiTiT:
I just realized that the `action_identifier` is probably not necessary.

BigRoy:
> Workflow is: Trigger action > action returns response that contains form > UI shows dialog with the form > when user confirms the form the very same action is triggered again but with the form values from the UI dialog > ... Now the action can do something with the values, AND it can return result with another form to get even more additional data (multi-step action).

So `execute_action` just turns recursive all of a function? :P Admitttedly it feels so very very weird. We're returning an action result with form values, making it rerun the same action with these new values. Why can't the execute action just trigger a form get the results and do what it wants with it.. instead of having to return, get re-executed, etc. It feels very inconvenient?

> What? In result, you define form to show, it can be dynamic and based on the selection and for each run of the action the form can be different. The form_values in arguments contains values that were filled by user when he confirms the form.

I understand what data flows there. But why can't execute action, in the code just do:

```python
form = Form(attribute_defs)
form_values = form.get_input()
```

And we if need additional values, or additional forms, etc. then either we retrigger the same form? `form.get_input()` or just spawn another form instance. I'm  still lost why we need it go through `execute_action` with all this bloat around.

> Because the code is not used anywhere else now. Are you complaining becuase it is named `LoaderActionForm` instead of `ActionForm`? If yes then you should have write that... Because I didn't get that until now.

I didn't write it before because I hadn't thought of it nor did I know that it was intended to being generic without tight coupling to the load action. Admittedly, why is an `AttributeDefinitions` dialog not sufficient here? Isn't it the exact same thing?

```python
from ayon_core.tools.attribute_defs.dialog import AttributeDefinitionsDialog
dialog = AttributeDefinitionsDialog(attr_defs)
values = dialog.get_values()
```
What's unique about the loader form?

This ☝️ is my point. It feels like you need to use the forms for the loaders in the code - because it's so tightly coupled to it, with supporting it in return values and the input arguments... even though I see NO reason why it needs to flow through those.

The input arguments can take `options` though, similar to what loaders currently do.. which in essence is equivalent to passing custom form data in a way in this implementation.

Does that make sense? Or am I really saying stupid things and am I misunderstanding your point of view? Maybe @antirotor or @philippe-ynput can weigh in.

PR #1422, opened 2025-08-25
Implement upload retries for reviewables and add user-friendly timeout error; update validation help

FuzzkingCool:
While we don't use AYON reviews, its baked into the publishes so when I ran into numerous timeouts today, which almost always were fine the 2nd attempt to publish, I created this pull request.

## Changelog Description
- Adds exponential backoff retries (3) when uploading reviewable files to AYON.
- Raises a clear, user-facing validation error on repeated timeouts with actionable help text.
- Updates validation help docs to guide users when review uploads fail.

## Additional info
Retry policy: up to 3 attempts with backoff (2s, 4s, 8s). On exhaustion, raises PublishXmlValidationError with key="upload_timeout" and formatting data for file and error.

## Testing notes:
Honestly not sure if you do not have a way to simulate the timeout. The timeouts were common today so I was able to test.

### DIFF
--- a/client/ayon_core/plugins/publish/integrate_review.py
+++ b/client/ayon_core/plugins/publish/integrate_review.py
@@ -1,11 +1,14 @@
 import os
+import time
 
-import pyblish.api
 import ayon_api
+import pyblish.api
 from ayon_api.server_api import RequestTypes
-
 from ayon_core.lib import get_media_mime_type
-from ayon_core.pipeline.publish import get_publish_repre_path
+from ayon_core.pipeline.publish import (
+    PublishXmlValidationError,
+    get_publish_repre_path,
+)
 
 
 class IntegrateAYONReview(pyblish.api.InstancePlugin):
@@ -82,11 +85,12 @@ class IntegrateAYONReview(pyblish.api.InstancePlugin):
             headers = ayon_con.get_headers(content_type)
             headers["x-file-name"] = filename
             self.log.info(f"Uploading reviewable {repre_path}")
-            ayon_con.upload_file(
+            # Upload with retries and clear help if it keeps failing
+            self._upload_with_retries(
+                ayon_con,
                 endpoint,
                 repre_path,
-                headers=headers,
-                request_type=RequestTypes.post,
+                headers,
             )
 
     def _get_review_label(self, repre, uploaded_labels):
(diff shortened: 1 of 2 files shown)


client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -100,3 +104,55 @@ def _get_review_label(self, repre, uploaded_labels):
...
+        for attempt in range(1, max_retries + 1):
+            try:
+                ayon_con.upload_file(
+                    endpoint,
+                    repre_path,
+                    headers=headers,
+                    request_type=RequestTypes.post,
+                )
+                return
+            except Exception as exc:  # noqa: BLE001 - bubble after retries
```
iLLiCiTiT, line 132: Don't we know the exception?

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -100,3 +104,55 @@ def _get_review_label(self, repre, uploaded_labels):
...
+                last_error = exc
+                # Log and retry with backoff if attempts remain
+                if attempt < max_retries:
+                    wait = backoff_seconds * (2 ** (attempt - 1))
```
iLLiCiTiT, line 136: Don't understand this wait time calculation... I know how it works but don't understand why we need this complex calculations?

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -100,3 +104,55 @@ def _get_review_label(self, repre, uploaded_labels):
...
+                        time.sleep(wait)
+                    except Exception:  # Sleep errors are highly unlikely; continue
+                        pass
+                else:
+                    # Exhausted retries - raise a user-friendly validation error with help
+                    raise PublishXmlValidationError(
+                        self,
+                        (
+                            "Upload of reviewable timed out or failed after multiple attempts. "
+                            "Please try publishing again."
```
iLLiCiTiT, line 151:
```suggestion
                            "Upload of reviewable timed out or failed after multiple attempts."
                            " Please try publishing again."
```

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -100,3 +104,55 @@ def _get_review_label(self, repre, uploaded_labels):
...
+        max_retries: int = 3,
+        backoff_seconds: int = 2,
+    ):
+        """Upload file with simple exponential backoff retries.
+
+        If all retries fail we raise a PublishXmlValidationError with a help key
+        to guide the user to retry publish.
+        """
+        last_error = None
+        for attempt in range(1, max_retries + 1):
```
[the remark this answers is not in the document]
FuzzkingCool: Good call. I kept it 1-based because the value is used directly in human-readable logs and in the backoff computation. I am fine switching to for attempt in range(max_retries): and then deriving attempt_no = attempt + 1 for logs and the wait math

client/ayon_core/plugins/publish/integrate_review.py
FuzzkingCool: Agreed that we should not retry on “anything.” The intent was to catch transient upload failures only. I will narrow the except clause to network or server-side cases that are safe to retry and let everything else bubble immediately. Concretely something like:

from requests import exceptions as req_exc

retryable = (
    req_exc.Timeout,
    req_exc.ConnectionError,
)

client/ayon_core/plugins/publish/integrate_review.py
```diff
@@ -1,11 +1,14 @@
 import os
+import time
 
-import pyblish.api
 import ayon_api
+import pyblish.api
```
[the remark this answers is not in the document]
FuzzkingCool: These were reordered by Ruff. Does Ynput use the Ruff: Format Imports?

PR #1423, opened 2025-08-25
Disable vertical and horizontal scrollbars in report page and log ite…

FuzzkingCool:
The publish report can create extreme flickering due to nested scrollbars rendering it useless and providing a poor user experience even after a successful publish. This pull request fixes that issue in most cases.


iLLiCiTiT:
Even tho it fix the flickering, now the issues is that it doesn't show the scroll bar...

PR #1426, opened 2025-09-03
OCIO: Use "max" for host group to get the correct ocio environment variable

moonyuet:
## Changelog Description
This PR is to edit "max" instead of "3dsmax" in host group so that the OCIO-related environment variable being colllected.
This fixes the ocio misconfiguring issue in 3dsmax.

image

## Additional info
n/a

## Testing notes:
1. Build Core Addon with this PR
2. Launch 3dsMax
3. The error shown in the screenshot should be gone


LiborBatek (approved):
Works for me...ocio config file been correctly set and available.

Screenshot 2025-09-03 171252

Issue #1229, opened 2025-01-08
AY-1284_show Application used on task


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

We need to know what application was used already on a task before launching a DCC in Launcher.
It can be easily missed there was a work done on a task already in other DCC.

Launcher should visually indicate somehow, that there is a workfile already present under the task

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-1284](https://app.clickup.com/t/861n1gre1)_

PR #1434, opened 2025-09-12
Launcher: Show task workfiles

iLLiCiTiT:
## Changelog Description
Show task workfiles in launcher UI.

## Additional info
The request is to show which hosts were used on a task. By showing workfiles user gets 2 information at a time. The workfiles must be saved using workfiles tool, and for icon it has to be saved using new workfiles api.

To see applications actions you must have https://github.com/ynput/ayon-applications/pull/98 and the workfile entities must have stored host name (using new workfiles api).

### Notes
It might make sense to hide it and keep window size as was and expand it only if user "enables" that.

### Screenshot
image

## Testing notes:
1. You can see workfiles when task is selected in launcher.

Resolves https://github.com/ynput/ayon-core/issues/1229

### DIFF
--- a/client/ayon_core/hooks/pre_add_last_workfile_arg.py
+++ b/client/ayon_core/hooks/pre_add_last_workfile_arg.py
@@ -38,18 +38,20 @@ class AddLastWorkfileToLaunchArgs(PreLaunchHook):
     launch_types = {LaunchTypes.local}
 
     def execute(self):
-        if not self.data.get("start_last_workfile"):
-            self.log.info("It is set to not start last workfile on start.")
-            return
+        workfile_path = self.data.get("workfile_path")
+        if not workfile_path:
+            if not self.data.get("start_last_workfile"):
+                self.log.info("It is set to not start last workfile on start.")
+                return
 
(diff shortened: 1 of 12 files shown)


iLLiCiTiT:
> I think 'local only' entries don't show a file icon - could that be? It's a bit confusing why it doesn't.

Yes, icons are shown only for workfiles that have filled `host_name` in workfile entity metadata.

> Also, my first intuition was to start clicking on the workfiles, but you can't. They are essentially only listed - but you can't pick which file to launch with? I think this will be the very first question newcomers will start asking - I can't click it.

Not supported, should be added, but in future, because actions on workfiles are not currently supported, and they would have to be "dynamic actions". Agreed it is confusing.

> Aside of that, starting the launcher now starts a very wide launcher... which feels out of proportion until you click into a project.

Agreed, it might be worth to show more information on project's page. But my first impression was to not show workfiles by default as I might not be interested in that at all, and show it only if user "checks" something?

iLLiCiTiT:
So, now the workfiles with filled host name can be used to launch application.

PR #1435, opened 2025-09-12
Loader: Product type icons

iLLiCiTiT:
## Changelog Description
Use project config to visually show product types the same way AYON frontend does.

## Additional info
The feature is available with https://github.com/ynput/ayon-backend/pull/622 but it should work as it was before without it.

## Testing notes:
1. Make sure you have AYON server with https://github.com/ynput/ayon-backend/pull/622 .
2. Setup product types looks in a project's anatomy.
3. Loader should visually show the products the way you define it in anatomy.

### DIFF
--- a/client/ayon_core/tools/common_models/__init__.py
+++ b/client/ayon_core/tools/common_models/__init__.py
@@ -10,6 +10,7 @@ from .projects import (
     PROJECTS_MODEL_SENDER,
     FolderTypeItem,
     TaskTypeItem,
+    ProductTypeIconMapping,
 )
 from .hierarchy import (
     FolderItem,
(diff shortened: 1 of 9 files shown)


BigRoy (approved):
Works in browser standalone, Silhouette, Maya 2026, Houdini 20.5, Blender 4.4 and Blender 4.5 and Fusion Studio 20.

image


Filter shows them as well in browser:
image

I wonder if bottom right the representations should also display the color and icon (even though it is not the product but a representation):
image


It does not work in Scene Inventory however:
image

---

I now start to wonder whether bottom left should also display task type colors instead of only icons.

iLLiCiTiT:
> I now start to wonder whether bottom left should also display task type colors instead of only icons.

Uuu, I didn't realize that's missing. Creating issue for that.

BigRoy (approved):
Scene Inventory now also works. I like it.

image


I actually like the icon there at product type more than how it's on the left in the browser UI.
So all good for me.

MilaKudr (commented):
Screenshot 2025-09-15 140211

Issue #1437, opened 2025-09-15
Task icon color in tools


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Tasks shown in UI show correct color but does not use color from project anatomy.

### How would you imagine the implementation of the feature?

Use color from anatomy for task items.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

PR #1438, opened 2025-09-15
Tasks widget: Use icon color from project anatomy

iLLiCiTiT:
## Changelog Description
Use color from task types in project anatomy in UIs.

## Testing notes:
1. Change color of task type in project anatomy.
2. Look at the task in pipeline UIs.
3. The color should be the same which is set in anatomy.

Resolves https://github.com/ynput/ayon-core/issues/1437


MilaKudr (approved):
Screenshot 2025-09-15 132001

PR #1439, opened 2025-09-15
Loader: Elide status name text

iLLiCiTiT:
## Changelog Description
Elide status name text.

## Additional info
Use-case is that status without an icon and short name is just hidden from the UI if the text does not fit the width of column. 2 things were changed, if short name is not available then use full name. If text does not fit into the text rectangle then elide the text to the width. Elided text must have at least one of the text characters.

### Screenshot
image

## Testing notes:
1. Set short name of status to empty string `""`.
2. Open loader, make the status column smaller.
3. It should keep the text filled but should be elided (added 3 dots at the end).


BigRoy (approved):
Works!

Image

No short name:

Image

Image

MilaKudr (approved):
image

PR #1459, opened 2025-10-02
Bugfix/add back OpenPype compatibility

jrsndl:
## Changelog Description
Adds compatibility for older OpenPype context keys.

The change ensures that 'subset' and 'asset' keys present in older OpenPype representations are correctly mapped to the new 'product' and 'folder' keys used in Ayon. This allows older representations to be loaded correctly.

Also adds a fallback mechanism when project anatomy is unavailable. This ensures the application remains functional by using the registered root path in cases where anatomy retrieval fails, enhancing OpenPype compatibility.

## Additional info

## Testing notes:
1. Open project migrated from OpenPype
2. Load product in Maya or Resolve

### DIFF
--- a/client/ayon_core/pipeline/load/utils.py
+++ b/client/ayon_core/pipeline/load/utils.py
@@ -752,12 +752,18 @@ def get_representation_path(representation, root=None):
 
     """
     if root is None:
-        from ayon_core.pipeline import get_current_project_name, Anatomy
+        try:
+            from ayon_core.pipeline import get_current_project_name, Anatomy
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
I'm afraid we won't be adding OpenPype support back.

I would recommend to change the data in AYON database to match AYON naming at all places to avoid these issues.

antirotor:
I am closing this as it was at the end resolved on database level (old OpenPype respresentation migrated to AYON).

PR #1467, opened 2025-10-07
using correct func for view display conversion

timsergeeff:
for some reason in previeus code mistake was made and files will never be converted to target display view since colorconvert cant recieve display and view


client/ayon_core/lib/transcoding.py
```diff
@@ -1167,6 +1167,12 @@ def oiio_color_convert(
     # Handle the different conversion cases
     # Source view and display are known
     if source_view and source_display:
+        oiio_cmd.extend([
```
timsergeeff, line 1170: i think i miss color_convert_args = None ocio_display_args = None at the start so it fails

PR #1471, opened 2025-10-08
Thumbnail: Fallback to OIIO if ffmpeg fails

iLLiCiTiT:
## Changelog Description
Allow to create thumbnail using OIIO if ffmpeg fails.

## Additional info
For some reason (I guess historical reasons) we use ffmpeg instead of OIIO to create the thumbnail, this adds fallback to be able to  use OIIO if ffmpeg fails to consume the input file. Also do not crash if the thumbnail creation fails completely, blocking production because of failed thumbnail creation is not nice.

## Testing notes:
1. Creating thumbnail from "weird" exrs works as expected.

### DIFF
--- a/client/ayon_core/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_core/plugins/publish/extract_thumbnail.py
@@ -471,6 +471,9 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
         try:
             resolution_arg = self._get_resolution_arg("oiiotool", src_path)
         except RuntimeError:
+            self.log.warning(
+                "Failed to create thumbnail using oiio", exc_info=True
+            )
             return False
 
         input_info = get_oiio_info_for_input(src_path, logger=self.log)
@@ -501,6 +504,9 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
         try:
             resolution_arg = self._get_resolution_arg("ffmpeg", src_path)
         except RuntimeError:
+            self.log.warning(
+                "Failed to create thumbnail using ffmpeg", exc_info=True
+            )
             return False
 
         ffmpeg_path_args = get_ffmpeg_tool_args("ffmpeg")


client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -229,17 +240,16 @@ def validate_string_against_patterns(input_str, patterns):
...
-                    self.log.debug(
-                        "Converting with FFMPEG because input"
-                        " can't be read by OIIO."
-                    )
-
                 repre_thumb_created = self._create_thumbnail_ffmpeg(
                     full_input_path, full_output_path
                 )
 
             # Skip representation and try next one if  wasn't created
```
BigRoy, line 247:
```suggestion
            # Skip representation and try next one if wasn't created
```

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -210,6 +215,12 @@ def validate_string_against_patterns(input_str, patterns):
             full_output_path = os.path.join(dst_staging, jpeg_file)
             colorspace_data = repre.get("colorspaceData")
 
+            # NOTE We should find out what is happening here. Why don't we
+            #   use oiiotool all the time if it is available? Only possible
+            #   reason might be that video files should be converted using
+            #   ffmpeg, but other then that, we should use oiio all the time.
+            #   - We should also probably get rid of the ffmpeg settings...
```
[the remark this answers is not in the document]
iLLiCiTiT: I think it is for @jakubjezek001 who is not available for next 2 weeks. Does not have to be resolved in this PR.

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -455,9 +465,44 @@ def _create_thumbnail_oiio(
...
+            )
+            return False
+
     def _create_thumbnail_ffmpeg(self, src_path, dst_path):
-        self.log.debug("Extracting thumbnail with FFMPEG: {}".format(dst_path))
-        resolution_arg = self._get_resolution_arg("ffmpeg", src_path)
+        try:
+            resolution_arg = self._get_resolution_arg("ffmpeg", src_path)
+        except RuntimeError:
+            return False
```
BigRoy, line 504: Do we lack a logged message that FFMPEG conversion failed? Just so that when reading the logs it's clear that both OIIO and FFMPEG may have failed or not?

BigRoy (approved):
Didn't test it but looks good to me. It does change that now if both FFMPEG and OIIO would fail you'd end up succeeding and passing the code instead of failing the publish.

Which I'd say that for something as mundane as a thumbnail is actually the right thing to do instead of holding production from continuing.

This may need some thorough testing though.

Issue #1450, opened 2025-09-24
Hardcoded template paths


TobiasPharos:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

The code base contains several places where get_template_item is hard-coded with "default" as template name, where instead methods like ayon_core.pipeline.publish.get_publish_template_name() (there seems to be no equivalent for the work template yet) should be used to identify the template name linked in the Ayon settings.

This will error for example, when one part of the code uses get_publish_template_name to store data and another part looks at the default path to load it, like in replace_with_published_scene_path.
It would also create files in locations where the user doesn't expect them.

Some of those places are:
- [x] [pipeline/farm/tools.py#L57](https://github.com/ynput/ayon-core/blob/da93a0a3644cadb238aef0c667514aebdd7af765/client/ayon_core/pipeline/farm/tools.py#L57C5-L57C75)

- [x] [pipeline/publish/lib.py#L806](https://github.com/ynput/ayon-core/blob/da93a0a3644cadb238aef0c667514aebdd7af765/client/ayon_core/pipeline/publish/lib.py#L806C4-L806C71): PR [#1472 ](https://github.com/ynput/ayon-core/pull/1472)

- [ ] ~~[open_file_explorer.py#L68](https://github.com/ynput/ayon-core/blob/da93a0a3644cadb238aef0c667514aebdd7af765/client/ayon_core/plugins/actions/open_file_explorer.py#L68C27-L70C10)~~ and ~~[open_file_explorer.py#L85](https://github.com/ynput/ayon-core/blob/da93a0a3644cadb238aef0c667514aebdd7af765/client/ayon_core/plugins/actions/open_file_explorer.py#L85C27-L87C10)~~ It is necessary to have host name to be able to get correct work template. There is probably not better solution than using `"default"`.

- [x] [collect_resources_path.py#L85](https://github.com/ynput/ayon-core/blob/da93a0a3644cadb238aef0c667514aebdd7af765/client/ayon_core/plugins/publish/collect_resources_path.py#L85C1-L87C10)

### Expected Behavior:

Identify the template name linked in the Ayon settings under "workfile template profiles" and "template name profiles" (for publish)

### Version

1.6.1

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Setup a path template in Anatomy, which is not the default (e.g. for Nuke workfile)
2. Link it in ayon_core settings (e.g. workfile template profile for host nuke)
[...]

PR #1472, opened 2025-10-08
fix hardcoded template path for "replace_with_published_scene_path"

TobiasPharos:
replace_with_published_scene looks up the actual publish path defined in anatomy and core settings.

## Changelog Description
Before replace_with_published_scene always took the default publish path, which caused tools looking into the wrong location if the publish path was not configured in "default", leading to errors.
Requires the productType of the scene to be "workfile", as the method is looking for the scene path, not the publish instance, otherwise will still fallback to default.

## Additional info
Fixes one of the five hardcoded template paths mentioned in [#1450](https://github.com/ynput/ayon-core/issues/1450).


TobiasPharos:
Should be fixed now

PR #1480, opened 2025-10-10
Feature: Fix frames on Deadline

kalisp:
## Changelog Description
Changes here are required for new feature where Artist could re-rendered only specific frames and rest of them will be reused for creation of new version.

## Additional info
Same logic is used in `ExtractReview` so it has been extracted out.
Used by https://github.com/ynput/ayon-deadline/pull/205

## Testing notes:
1. start with this step

### DIFF
--- a/client/ayon_core/plugins/publish/collect_resources_path.py
+++ b/client/ayon_core/plugins/publish/collect_resources_path.py
@@ -13,6 +13,8 @@ import copy
 
 import pyblish.api
 
+from ayon_core.pipeline.publish import get_publish_template_name
+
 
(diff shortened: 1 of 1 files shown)


kalisp:
> @kalisp does this work only if creator have `Fix Frames` settings e.g. in `nuke` host or does it work even in `maya`?

This is only meant for Deadline publishes where it should work on all DCCs. `Fix Frames` is Nuke local functionality.
(Both fields shouldnt be visible, but until https://github.com/ynput/ayon-deadline/issues/57 is not solved, they are.)

PR #1485, opened 2025-10-16
Extends audio to sibling review instances under editorial publishing

jakubjezek001:
## Changelog Description

This change extends the audio extraction process to include sibling instances that require audio for reviewable media. This ensures that all related instances, especially those used for review purposes, have access to the necessary audio tracks.

This addresses an issue where reviewable instances that are siblings to the main audio instance were not receiving the extracted audio, leading to incomplete review packages.

## Additional info

- The code iterates through the context to find sibling instances (instances with the same `parent_instance_id` but different IDs). It then appends the audio file path and offset to the `audio` attribute of these sibling instances. A check prevents the same instance from being processed multiple times.
- Also improved the temporary file and folder management. The process does not need to create multiple temp directories and rather store all within one per publishing session.
- Single plugin module file is holding two plugins for better organization of related processes. The file contains shared functions which can be used from Collector and Extractor.
- The host filtering was also removed, because the process will always work only with editorial related audio instances and those are checked at any plugin process top.

## Testing notes:
1. Publish a shot with multiple instances, including an instance that generates audio and at least one sibling instance intended for review.
2. Verify that the sibling review instance includes the extracted audio file in its instance data after publishing. Simply by observing web reviewable version is with correct audio.

closes https://github.com/ynput/ayon-hiero/issues/93

### DIFF
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -1045,7 +1045,9 @@ def get_resources(project_name, version_entity, extension=None):
             filtered.append(repre_entity)
 
     representation = filtered[0]
-    directory = get_representation_path(representation)
+    directory = get_representation_path(
+        project_name, representation
+    )
     print("Source: ", directory)
     resources = sorted(
         [
--- a/client/ayon_core/pipeline/load/__init__.py
+++ b/client/ayon_core/pipeline/load/__init__.py
@@ -25,8 +25,8 @@ from .utils import (
     get_loader_identifier,
     get_loaders_by_name,
 
-    get_representation_path_from_context,
     get_representation_path,
+    get_representation_path_from_context,
     get_representation_path_with_anatomy,
 
     is_compatible_loader,
@@ -85,8 +85,8 @@ __all__ = (
     "get_loader_identifier",
     "get_loaders_by_name",
 
-    "get_representation_path_from_context",
     "get_representation_path",
+    "get_representation_path_from_context",
     "get_representation_path_with_anatomy",
 
     "is_compatible_loader",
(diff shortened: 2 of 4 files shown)


client/ayon_core/plugins/publish/collect_audio.py
```diff
@@ -46,6 +46,19 @@ class CollectAudio(pyblish.api.ContextPlugin):
     audio_product_name = "audioMain"
 
     def process(self, context):
+        # Make sure Editorial related products are excluded
```
iLLiCiTiT: I don't think this belongs here. This is generic plugin for all cases, but now editorial is affecting all of them. If editorial needs different behavior for audio, it should have own collect plugin to do so.

iLLiCiTiT: Now you must have at least one instance with `"audio"` in families.
1. That is change of the logic.
2. Doesn't make sense, because the instances might not be related to any other instances.

I'm still not sure I understand what is the goal of this logic? Was this added to skip audio collection if there is no audio instance?

jakubjezek001:
changes are implemented @iLLiCiTiT

client/ayon_core/plugins/publish/extract_otio_audio_tracks.py
```diff
@@ -9,6 +9,56 @@
...
+            list: list of selected instances
+        """
+        return [
+            _i for _i in context
+            # filter only those with audio product type or family
+            # and also with reviewAudio data key
+            if bool("audio" in (
+                _i.data.get("families", []) + [_i.data["productType"]])
+            ) or _i.data.get("reviewAudio")
+        ]
```
iLLiCiTiT:
```suggestion
        audio_instances = []
        for instance in context:
            if (
                instace.data["productType"] == "audio"
                or instace.data.get("reviewAudio")
            ):
                audio_instances.append(instance)
        return audio_instances
```

iLLiCiTiT (approved):

MilaKudr (approved):

PR #1486, opened 2025-10-16
Bundle: Project bundle addons version resolution

iLLiCiTiT:
## Changelog Description
Use settings summary to resolve which addon versions are used.

## Additional info
This approach uses the same approach which is used on AYON server, which might not be the same as was used in previous implementation (see https://github.com/ynput/ayon-backend/pull/771). But at this exact moment nothing should really change.

## Testing notes:
1. Resolution of project bundle addons work as expected (hard to say what is expected at this moment).


LiborBatek:
I cant make the `Project Bundle` to effect and it reads still as the ordinary `Prod Bundle` when launching in the project context with set `Proj Bundle` ...as here:

My Project Bundle having `ayon-core 1.6.4+dev` while when running (after host app launch) still shows the prod bundle `ayon-core 1.6.4`

Am I missing something (as I know it first runs with prod but then switches to proj bundle addons)?
Screenshot 2025-10-16 143828



Screenshot 2025-10-16 143855

PR #1498, opened 2025-10-21
Corrects file sequence frame offset

jakubjezek001:
## Changelog Description
Corrects how the frame offset is calculated for file sequences, ensuring accurate time remapping. The issue was that the frame_in was always calculated with offset even if the offset was zero, leading to calculation errors. This fix ensures that the offset is only applied when it is present, resolving discrepancies in the frame ranges shown in the publisher.

## Testing notes:
1. Open a sequence in a project.
2. Publish the sequence which frame number starts from `1` and has no head frames on clip.
3. Verify the the Loaded product is not having any offset.

closes https://github.com/ynput/ayon-hiero/issues/92

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,6 +35,7 @@ body:
       label: Version
       description: What version are you running? Look to AYON Tray
       options:
+        - 1.6.7
         - 1.6.6
         - 1.6.5
         - 1.6.4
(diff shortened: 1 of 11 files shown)


client/ayon_core/pipeline/editorial.py
```diff
@@ -256,10 +257,14 @@ def remap_range_on_file_sequence(otio_clip, otio_range):
         )
 
     src_offset_in = otio_range.start_time - media_in
-    frame_in = otio.opentime.RationalTime.from_frames(
-        media_ref.start_frame + src_offset_in.to_frames(),
+    # make sure that only if any offset is present
+    if media_ref.start_frame == src_offset_in.to_frames():
+        frame_in = src_offset_in.to_frames()
```
rdelillo: Quickly looking I'm not sure I understand this condition change.
Would you have a example to explain when is this fix required please ?
I might be wrong but I feel like this is compensating a weird media reference value (inconsistent available_range + start_frame).

jakubjezek001: @rdelillo is this something you want to discuss more in depth before I merge it?

rdelillo (approved):

Issue #1496, opened 2025-10-20
YN-0069: More user data in templates


iLLiCiTiT:
### Please describe the issue you have encountered?
Extend `user` key available in template data to support nested dictionary with `attrib`.

_This issue was manually created from AYON ticket [YN-0069](https://support.ayon.app/projects/Active_Clients/overview)_

PR #1501, opened 2025-10-22
Template data: More user data in templates

iLLiCiTiT:
## Changelog Description
Extend user data available in templates.

## Additional info
Because user in template data before these changes was a simple string and now it is a dictionary a little nasty change had to be made. Implemented `DefaultKeysDict` which can define default subkey to use when used for string conversion when used for template format.

I was not able to figure out where we do support to use `"{user}"` except for project anatomy templates. If you know about more usages, please test it and let me know.

We can't do auto-conversion of anatomy templates because they are set on project entity and in anatomy presets.

## Testing notes:
1. You should be able to use both `"{user}"` and `"{user[name]}"`
2. Published representations can be then loaded if any of them was used.

Resolves https://github.com/ynput/ayon-core/issues/1496

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,6 +35,8 @@ body:
       label: Version
       description: What version are you running? Look to AYON Tray
       options:
+        - 1.6.6
+        - 1.6.5
         - 1.6.4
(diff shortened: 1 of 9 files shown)


64qam (approved):

antirotor (approved):
Didn't test but it looks fine

Issue #1488, opened 2025-10-16
YN-0117: USD texture publish to different drive


ynbot:
### Please describe the issue you have encountered?

Hey, so this works fine for textures, but it's not working for the look

It half works, it publishes the look usd file to the T drive

but the relative textures remain on the Z drive

And the path to the lookMain file in the USD point towards the Z drive as well 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0117](https://support.ayon.app/projects/Active_Clients/overview)_

BigRoy:
This issue is still present here: https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/pipeline/publish/lib.py#L986

PR #1506, opened 2025-10-27
Use correct publish template in `get_instance_expected_output_path`

BigRoy:
## Changelog Description

Use correct publish template in `get_instance_expected_output_path`
## Additional info

Fix https://github.com/ynput/ayon-core/issues/1488

Without this fix USD files generated with USD contribution workflow will still reference/sublayer paths that are at the default anatomy template instead of a custom one e.g. if set for `look` or `usd` products.

## Testing notes:

1.  With this fix USD files generated with USD contribution workflow will still reference/sublayer paths that are the correct anatomy template instead of the default one, e.g. if set for `look` or `usd` products.


MustafaJafar (approved):
It works as before with default template.
and it also works when updating the template ❤️.

Issue #1504, opened 2025-10-24
YN-0141: My tasks not displaying in Workfiles App inside Blender Host


ynbot:
### Please describe the issue you have encountered?

 
While having `assignee` on task and it correctly been used when using `My Task` knob within the Launcher it doesnt work properly and instead of filtering the task just for the `My Tasks` it does not show any of those but empty browser instead which is not correct.

It happens and been spotted in `3dsmax` `maya` and `blender` host apps. So pretty much everywhere.

### How to replicate the issue?

Assign an assignee to particular task within active project and use filter tool `My Tasks` in the Launcher and the same when the host app been running but using `Workfiles` tool instead where the filtering been failing and show nothing instead...

<img width="1390" height="939" alt="Image" src="https://github.com/user-attachments/assets/77b7c905-cb63-4e62-bae9-5cd35a53e739" />

Been tested with latest `ayon-core 1.6.6` and `ayon-applications 1.2.4`

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1430849746020597851/1430849750160375910)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0141](https://support.ayon.app/projects/Active_Clients/overview)_

PR #1508, opened 2025-10-27
Workfiles tool: Fix my tasks filtering

iLLiCiTiT:
## Changelog Description
Fix filtering by my tasks in workfiles tool.

## Additional info
Workfiles window stored current project before refresh of the controller. Now is the project re-set after each refresh.

## Testing notes:
1. My tasks filter works as expected.

Resolves https://github.com/ynput/ayon-core/issues/1504

### DIFF
--- a/client/ayon_core/tools/workfiles/widgets/window.py
+++ b/client/ayon_core/tools/workfiles/widgets/window.py
@@ -411,7 +411,6 @@ class WorkfilesToolWindow(QtWidgets.QWidget):
             entity_ids = self._controller.get_my_tasks_entity_ids(
                 self._project_name
             )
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
Been tested in `blender` and `maya` in the `Workfiles` tool and also in the `Launcher` and all seems working.
Launcher02

note: There arent any host apps icons next to the workfiles in the list (right hand side panel with all available workfiles) and also those which arent eligible for launching arent visible too... which I guess is on purpose (aka those outdated workfiles simply not shown)

correct me if Im assuming it wrong... below is example in the Launcher

Launcher


LGTM

PR #1510, opened 2025-10-27
Library: Implemented copy of source folder thumbnail

kalisp:
## Changelog Description
Only product thumbnails were copied previously, now even folder base thumbnail gets copied.

## Testing notes:
1. pick some folder with attached thumbnail
2. push product of this folder to Library and use `Create New Folder`
3. check in Loader that destination folder was created with thumbnail (select only created folder, thumbnail should be different from product one)

### DIFF
--- a/client/ayon_core/addon/base.py
+++ b/client/ayon_core/addon/base.py
@@ -141,6 +141,9 @@ def _get_ayon_bundle_data() -> tuple[
 ]:
     studio_bundle_name = os.environ.get("AYON_STUDIO_BUNDLE_NAME")
     project_bundle_name = os.getenv("AYON_BUNDLE_NAME")
+    # If AYON launcher <1.4.0 was used
+    if not studio_bundle_name:
+        studio_bundle_name = project_bundle_name
     bundles = ayon_api.get_bundles()["bundles"]
     studio_bundle = next(
         (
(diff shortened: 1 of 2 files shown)


client/ayon_core/tools/push_to_project/models/integrate.py
```diff
@@ -733,6 +737,36 @@ def _create_folder(
...
+                with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
+                    tmp_file.write(thumbnail.content)
+                    temp_file_path = tmp_file.name
+
+                new_thumbnail_id = ayon_api.create_thumbnail(
+                    project_entity["name"], temp_file_path)
+            finally:
+                if temp_file_path and os.path.exists(temp_file_path):
+                    os.remove(temp_file_path)
+        return new_thumbnail_id
```
iLLiCiTiT:
```suggestion
        if not src_folder_entity["thumbnailId"]:
            return None

        thumbnail = ayon_api.get_folder_thumbnail(
            self._item.src_project_name,
            src_folder_entity["id"],
            src_folder_entity["thumbnailId"]
        )
        if not thumbnail.id:
            return None

        with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
            tmp_file.write(thumbnail.content)
            temp_file_path = tmp_file.name

        new_thumbnail_id = None
        try:
            new_thumbnail_id = ayon_api.create_thumbnail(
                project_entity["name"], temp_file_path)
        finally:
            if os.path.exists(temp_file_path):
                os.remove(temp_file_path)
        return new_thumbnail_id
```

PR #1513, opened 2025-10-30
Select instances in host using double click on publish instances

FuzzkingCool:
A new feature that allows users to double-click on instances in the publisher overview widget to trigger selection and focus of the corresponding nodes/objects in the host DCC's current file. The selection logic is integrated with Creator plugin specific hooks, so this pull request is only to enable the ability for a Creator to implement the selection logic.

**UI interaction improvements:**

* Updated double-click event handlers in `overview_widget.py` to use a new method `_on_instances_double_clicked`, enabling selection of instances in the host DCC on double-click.
* Implemented the `_on_instances_double_clicked` method to call the controller's `select_instances_in_host` and emit the publish tab navigation, ensuring UI responsiveness even if selection fails.

**Backend selection logic:**

* Added `select_instances_in_host` to the controller in `control.py`, delegating selection requests to the model.
* Implemented `select_instances_in_host` in the model (`create.py`), invoking Creator specific `select_in_host` hooks for each instance, with error handling to avoid UX interruptions.


iLLiCiTiT:
We'll have a discussion about this. The functionality to select something in the scene is good feature, but we're not convinced if double click is the way to go and if simple method to do so is enough (if this can cover all the use-cases that would came when added).

iLLiCiTiT:
> Sounds good. I implemented the Creator method in my Harmony creators which allow users to quickly find nodes and template backdrops in the Node View for instance, and it has proven useful in production.

Which covers one use-case but not all of them -> we'll need something more extendable, and with deterministic approach how to ask "if is possible" and "what options there are". I think right-click could show actions to run, so UI can ask controller and create plugins if they want to show actions to do on the selection.

PR #1521, opened 2025-10-31
Library: Copy over product grouping

kalisp:
## Changelog Description
This PR copies also product grouping that could be configured in source project.

## Additional info
Product grouping is stored on `attrib` of product entity.

## Testing notes:
1. group any products in Loader with `CTRL+G`
2. push them to Library project
3. check that group got copied too in destination project
image

### DIFF
--- a/client/ayon_core/tools/push_to_project/models/integrate.py
+++ b/client/ayon_core/tools/push_to_project/models/integrate.py
@@ -955,8 +955,9 @@ class ProjectPushItemProcess:
             "description",
             "productGroup",
         }:
-            if key in src_attrib:
-                dst_attrib[key] = src_attrib[key]
+            value = src_attrib.get(key)
+            if value:
+                dst_attrib[key] = value
 
         product_entity = new_product_entity(
             product_name,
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
LGTM, this will copy over e.g. attribs like `{'description': None, 'productGroup': 'renderLightingMain'}`.

It may hence also copy over a description.

iLLiCiTiT (approved):

client/ayon_core/tools/push_to_project/models/integrate.py
```diff
@@ -952,6 +952,7 @@ def _make_sure_product_exists(self):
             product_name,
             product_type,
             folder_id,
+            attribs=self._src_product_entity["attrib"]
```
iLLiCiTiT: I wonder if this would also copy attribute like `ftrackId`, `shotgridId`? Maybe it should be more limited instead of "copy all" > causing possible issues.

kalisp: Ok, copied only `description` and `productGroup`.

PR #1522, opened 2025-10-31
Library: Version up existing product

kalisp:
## Changelog Description
Added checkbox that if selected it versions up existing product. Former behavior it would overwrite the existing version.

## Additional info
Not completely sure if it makes sense to have in Library more versions, but it was customer requirement.

## Testing notes:
1. experiment with `Version up existing Product`
image
2. check destination project if product was version up or version was overwritten


client/ayon_core/tools/push_to_project/ui/window.py
```diff
@@ -424,6 +436,10 @@ def _on_original_names_change(self, state: int) -> None:
         use_original_name = bool(state)
         self._invalidate_use_original_names(use_original_name)
 
+    def _on_version_up_checkbox_change(self, state: int) -> None:
+        self._controller._version_up = bool(state)
```
iLLiCiTiT: Please add `set_version_up` to controller and call it. Never touch private attributes or methods.

PR #1526, opened 2025-11-04
Support to create Reviewables for Substance Painter

moonyuet:
## Changelog Description
This PR is to add the support of reviewable for Substance Painter
Resolve https://github.com/ynput/ayon-substance-painter/issues/58

## Additional info
Tested with https://github.com/ynput/ayon-substance-painter/pull/60

## Testing notes:
1. Create Texture Instance in SP and make sure review option set to True
2. Publish
3. You will find the reviewable shown in your browser
image

### DIFF
--- a/server/settings/publish_plugins.py
+++ b/server/settings/publish_plugins.py
@@ -251,6 +251,19 @@ class AyonEntityURIModel(BaseSettingsModel):
     )
 
 
+class ExtractUSDLayerContributionModel(AyonEntityURIModel):
+    enforce_default_prim: bool = SettingsField(
+        title="Always set default prim to folder name.",
+        description=(
+            "When enabled ignore any default prim specified on older "
+            "published versions of a layer and always override it to the "
+            "AYON standard default prim. When disabled, preserve default prim "
+            "on the layer and then only the initial version would be setting "
+            "the AYON standard default prim."
+        )
+    )
+
+
 class PluginStateByHostModelProfile(BaseSettingsModel):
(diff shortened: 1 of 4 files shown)


BigRoy:
After discussing some more about this it essentially boils down to:

- We need to be able to generate a reviewable from textures / UDIM sequences.
- These may be a single frame, but also may be e.g. 1001, 1002, 1011, 1012 (in that sense they'd be a sequence with holes).
- Preferably that'd generate a reviewable that contains the frame number 1001, 1002, 1011, 1012 burned in accordingly **without** the holes filled up with additional data.

@iLLiCiTiT @kalisp perhaps you know with the recent "holes/gaps" support from deadline publishes, etc. - how should we configure the instance from Substance Painter to make this possible?

server/settings/publish_plugins.py
```diff
@@ -1448,6 +1448,105 @@ class PublishPuginsModel(BaseSettingsModel):
...
+                            "ftrackreview",
+                            "kitsureview",
+                            "webreview"
+                        ],
+                        "burnins": [],
+                        "ffmpeg_args": {
+                            "video_filters": [],
+                            "audio_filters": [],
+                            "input": [
+                                "-apply_trc gamma22"
```
BigRoy: Do we want to still apply the 2.2 gamma to EXRs in this case?

moonyuet:
@LiborBatek can you test again with the build of this PR to ensure no error hits before the merge?

Issue #1530, opened 2025-10-31
Multi selections of Export Texture Sets and Channels fails to restore for an instance


sosoyan:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Description

In the publish dialog when saving multi selections of Export Texture Sets and Channels fails to restore after refresh or reopening the publish dialog.

![Image](https://github.com/user-attachments/assets/e1574bbb-c14c-4213-957c-1426004488c0)

### Relevant log output:

```shell

```

Issue #1194, opened 2025-03-13
Publisher: Multiselection EnumDef looses value in UI


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

Multiselection EnumDef sometimes does not display value in Publisher UI, even though the value on instance is set. We were able to replicate it on publish attributes, but it was random, some values worked, some did not work. The issue is probably caused by logic in `client/ayon_core/tools/publisher/widgets/product_attributes.py`

### Expected Behavior:

The multiselection `EnumDef` should show value stored on instance.

### Version

1.0.14

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Create testing publish plugin with multiselection EnumDef with e.g. 5 items.
2. Select instance on which is attribute shown.
3. Change value of the enum attribute > Save > Reset > Repeat until it shows empty value in UI.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell

```

PR #1531, opened 2025-11-05
Publisher: Multiselection EnumDef does not loose values

iLLiCiTiT:
## Changelog Description
Function `convert_value` does not convert value to set.

## Additional info
If value is converted to `set` then it can change order of the items. That can randomly cause that value comparison does not equal for list value of EnumDef.

## Testing notes:
1. Multiselection enum definitions do not randomly loose values.

Resolves https://github.com/ynput/ayon-core/issues/1194
Resolves https://github.com/ynput/ayon-core/issues/1530

### DIFF
--- a/client/ayon_core/pipeline/create/structures.py
+++ b/client/ayon_core/pipeline/create/structures.py
@@ -137,6 +137,7 @@ class AttributeValues:
             if value is None:
                 continue
             converted_value = attr_def.convert_value(value)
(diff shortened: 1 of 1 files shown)


moonyuet (approved):
Checked with SP Looks good.

https://github.com/user-attachments/assets/3b9f53f4-ba9b-45bd-a305-80353467dcda

PR #1533, opened 2025-11-07
Launcher: Deselect entities on change of focus

iLLiCiTiT:
## Changelog Description
Deselect task if folder view is focused and deselect workfile if task view is focused.

## Additional info
Helps to see task specific actions in case workfile entity does not support one.

Question is if tasks should be deselected too? That might be too much as deselection of tasks was not reported (easy to revert).

## Testing notes:
1. Open launcher tool.
2. Select folder > task > workfile.
3. Click to folder view. Task and workfile are deselected.
4. Select folder > task > workfile.
5. Click to task view. Workfile is deselected.

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -202,7 +202,7 @@ class ContextCardWidget(CardWidget):
     Is not visually under group widget and is always at the top of card view.
     """
 
-    def __init__(self, parent):
+    def __init__(self, parent: QtWidgets.QWidget):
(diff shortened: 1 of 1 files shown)


BigRoy:
It works, but...

> Question is if tasks should be deselected too? That might be too much as deselection of tasks was not reported (easy to revert).

**Yes, please revert.** Without this PR if I have many folders with a task named Lighting, if I go to the other folder, it will autoselect the Lighting task because I had it previously selected - making it quick to switch folders but remain in similar tasks...

With this PR that behavior is lost. That behavior we implemented intentionally originally so allow for quicker context switching. So please revert this one.

Issue #1525, opened 2025-10-31
YN-0158: USD contribution for shots starting with digit breaks USD


ynbot:
### Please describe the issue you have encountered?

The AYON USD contribution workflow will set the context's **folder name** as the default prim for the USD layer. If your folder name (e.g. for a shot) starts with a digit that may make the default prim e.g.

```
104_004_010
```

however, for USD this is a completely invalid prim specification because USD Prim names are not allowed to start with a number, making this essentially garbage data.

  

See e.g. [here](https://docs.omniverse.nvidia.com/usd/code-docs/usd-exchange-sdk/latest/api/group__names.html) stating that USD Prim names have the following limitations:

- No empty strings.
- Must start with characters in the set of [A-Za-z_].
- May continue with characters in the set of [A-Za-z0-9_].

Loading a USD file with such an invalid default prim set may have unexpected consequence, e.g. Houdini 21 will straight out crash and USD view the view may remain empty, etc.

Whenever dealing with the 'default name' for a prim we should always enforce it to be validly named, e.g. prefixing it with an allowed character if it starts with a digit, like an underscore. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1425438330048479303/1430830252665471069)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0158](https://support.ayon.app/projects/Active_Clients/overview)_

MustafaJafar:
> Whenever dealing with the 'default name' for a prim we should always enforce it to be validly named, e.g. prefixing it with an allowed character if it starts with a digit, like an underscore.

I wonder if some studios already have something like this handled on the AYON server level. e.g. when creating folders and tasks via API, always prefix folder names with `_` for add some letters `p`, `f`, `t`for project, folder and task respectively.

I like the idea of prefixing which also solves this issue as well https://github.com/ynput/ayon-houdini/issues/52 
But, I'm not sure what is the best place to solve it. I'm afraid if we fixed it here, it'll lead to inconsistency especially with automations.

PR #1535, opened 2025-11-08
USD contribution workflow: prefix folder names with _ if starting with a digit to avoid invalid prim names in USD data

BigRoy:
## Changelog Description

USD contribution workflow: prefix folder names with _ if starting with a digit to avoid invalid prim names in USD data

## Additional info

Prim names are not allowed to start with a digit in USD. Authoring them would mean generating essentially garbage data and may result in unexpected behavior in certain USD or DCC versions, like failure to refresh in usdview or crashes in Houdini 21.

The added `get_standard_default_prim_name` library function may need to get some adoption in other DCCs where we're trying to set the root names of hierarchies with the context's folder name as well.

Fix: https://github.com/ynput/ayon-core/issues/1525

## Testing notes:

1. Folder names starting with a digit should now generate default prims prefixed with `_`

### DIFF
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -181,6 +181,8 @@ def get_instance_uri_path(
             anatomy=context.data["anatomy"],
             **names
         )
+        if not path:
+            raise RuntimeError(f"Unable to resolve publish path for: {names}")
 
         # Ensure `None` for now is also a string
         path = str(path)


MustafaJafar:
maybe we need one fix for the validation to expect the new value

MustafaJafar (approved):
It works as expected. Tested in Houdini. Maybe I can test it in Maya as well.
I know that instance label is showing the product name. I wonder if some users may expect seeing the `_` prefix here.
image

MustafaJafar:
I got this error in Maya.
```python
  File "C:\Users\Mustafa Taher\AppData\Local\Ynput\AYON\dependency_packages\ayon_2510230030_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\Ynput\ayon-core\client\ayon_core\plugins\publish\extract_usd_layer_contributions.py", line 672, in process
    path = get_instance_uri_path(contribution.instance,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Ynput\ayon-core\client\ayon_core\plugins\publish\extract_usd_layer_contributions.py", line 180, in get_instance_uri_path
    path = get_representation_path_by_names(**query)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: get_representation_path_by_names() got an unexpected keyword argument 'project'
```

However, it seems the name space is named correctly.
image

MustafaJafar:
I guess this line 
https://github.com/ynput/ayon-core/blob/5bccc7cf2bcdc51e446cf7794f09d6081ff7255a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py#L179
should be 
```python
        path = get_representation_path_by_names(**names)
```

once I did this change the extractor no longer errors but I got another error on the integrate plugin
```python
  File "C:\Users\<USERNAME>\AppData\Local\Ynput\AYON\dependency_packages\ayon_2510230030_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\Ynput\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 143, in process
    filtered_repres = self.filter_representations(instance)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Ynput\ayon-core\client\ayon_core\plugins\publish\integrate.py", line 180, in filter_representations
    raise KnownPublishError(
ayon_core.pipeline.publish.publish_plugins.KnownPublishError: Instance usd has no representations to integrate
```

BigRoy:
@MustafaJafar it seems that your publish did not generate any USD publish files... :/ which seems odd. Anyway, I fixed that code path indeed - however, it doesn't solve the issue at hand.

Is the Extractor disabled by any chance?
@MustafaJafar my guess is that you have this version of the maya addon on your server: https://github.com/ynput/ayon-maya/pull/363

Which adds settings for `MayaUsdExtractor` which admittedly it shouldn't have, but it made the plug-in optional and active state False by default.. hence.. it would be "off" by default which would cause these issues you were facing. Can you confirm?

And if so, can you upload latest maya dev addon to your server and active in bundle, and then test again?

Issue #1288, opened 2025-05-16
AY-7805_MyTasks in Loader and Publisher


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

the same functionality we have in Launcher we need in Loader and Publisher

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7805](https://app.clickup.com/t/86c3hut8w)_

PR #1539, opened 2025-11-10
Tools: Add my tasks filters to loader and publisher

iLLiCiTiT:
## Changelog Description
Added my tasks filtering to loader and publisher tools.

## Additional info
Implemented helper widget `FoldersFiltersWidget` for filtering. And `get_current_username` to `UsersModel`. Use the `FoldersFiltersWidget` in launcher and workfiles tools too -> in workfiles tool it meant to change order of widgets.

## Testing notes:
1. Tools Launcher, Workfiles, Loader and Publisher have `My tasks` filtering.
2. The filtering works in all of the tools.

Resolves https://github.com/ynput/ayon-core/issues/1288

### DIFF
--- a/client/ayon_core/tools/launcher/ui/hierarchy_page.py
+++ b/client/ayon_core/tools/launcher/ui/hierarchy_page.py
@@ -9,7 +9,6 @@ from ayon_core.tools.utils import (
     TasksWidget,
 )
 from ayon_core.tools.utils.folders_widget import FoldersFiltersWidget
-from ayon_core.tools.utils.lib import checkstate_int_to_enum
 
 from .workfiles_page import WorkfilesPage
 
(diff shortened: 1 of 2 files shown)


BigRoy (approved):
Works in Tray publisher and for loader and publisher in Maya 2026, Fusion 20, Houdini 20.5 and Blender 4.5.

**However!** The only thing I'm not convinced about is in the Loader UI that the task filter bottom left also filters to YOUR tasks only - yet it still displays all products by default regardless. Hence, you suddenly can't filter to the other tasks but they are displayed if no filter is selected. Seems off?

Either we should skip filtering the tasks lists there, or always exclude the products not  from your tasks. I'd opt for skipping the tasks filtering there, but maybe best to check with client.
@m-u-r-p-h-y thoughts?

PR #1540, opened 2025-11-10
Experimental: Track CLI processes using Process Manager

antirotor:
## Changelog Description
Adds ability to track AYON itself as process in Process Manager.

## Additional info
Process Manager is a tool in [ayon-applications](http://github.com/ynput/ayon-applications) to track launched processes. This PR is using that tool to track AYON processes too.

> [!WARNING]
This is experimental and proof-of-concept. Not to be merged in this form

## Testing notes:
**This requires ayon-applications addon 1.2.4 and higher.**

When you run AYON, you should see `ayon-cli` process in the Process Manager window with output - the same as with DCCs. You can run Publisher, or other tools - they well create new ayon-cli process. Even launching apps will create it (with the output from pre-launch hooks).

While this might be useful, it is also very verbose, so without auto-clean function it is probably unusable.

Also there is some bug regarding colors in the output.


client/ayon_core/cli.py
```diff
@@ -59,9 +59,9 @@ def main_cli(ctx, *_args, **_kwargs):
         if os.getenv("AYON_HEADLESS_MODE") == "1":
             print(ctx.get_help())
             sys.exit(0)
-        else:
-            ctx.params.pop("project")
-            ctx.forward(tray)
+
+        ctx.params.pop("project")
+        ctx.forward(tray)
```
iLLiCiTiT: We should not start tray in headless mode.
```suggestion
        else:
            ctx.params.pop("project")
            ctx.forward(tray)
```

client/ayon_core/cli.py
```diff
@@ -386,6 +386,53 @@ def main(*args, **kwargs):
 
     _cleanup_project_args()
 
+    app_addon = addons_manager.get_enabled_addon("applications")
```
iLLiCiTiT: Do you realize that this cli is also used for small subprocesses like extract burnin, ffmpeg user service ocio functionality etc? Some of them are already outputing to parent process. On windows it also doesn't make sense to read output if stdout/stderr is not available at all (is set to `None`).

I don't think we should do it automatically for all processes TBH. It should be explicit and not enabled by default. Have one function in ayon-core to store the process info that can be easily called.

iLLiCiTiT: With that function we should move that logic to ayon-core.

antirotor:
I don't think we need this anymore. The idea of tracking is good but it can be achieved better with proper logging.

PR #1543, opened 2025-11-12
Tools: Use GraphQl to get projects

iLLiCiTiT:
## Changelog Description
Use GraphQl to get projects in tools.

## Additional info
There is a bug in ayon api which causes that `get_projects` uses `get_rest_projects` even if it is not necessary which might cause slow loading of UI in case there is a lot of projects.

## Testing notes:
1. Loading of projects in UIs should be faster.

### DIFF
--- a/client/ayon_core/tools/common_models/projects.py
+++ b/client/ayon_core/tools/common_models/projects.py
@@ -541,7 +541,7 @@ class ProjectsModel(object):
             self._projects_cache.update_data(project_items)
         return self._projects_cache.get_data()
 
-    def _fetch_projects_bc(self) -> list[dict[str, Any]]:
+    def _fetch_graphql_projects(self) -> list[dict[str, Any]]:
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/common_models/projects.py
BigRoy: Just noticed a typo here.. :) but may need more changes in code.
```suggestion
def _get_project_items_from_entity(
```

client/ayon_core/tools/common_models/projects.py
```diff
@@ -538,8 +541,32 @@ def _refresh_projects_cache(self, sender=None):
             self._projects_cache.update_data(project_items)
         return self._projects_cache.get_data()
 
+    def _fetch_projects_bc(self) -> list[dict[str, Any]]:
```
BigRoy: What does the `bc` suffix stand for?

PR #1547, opened 2025-11-13
Publish: Store host name to version entity data

iLLiCiTiT:
## Changelog Description
Store host name to version entity data during publishing.

## Additional info
It is important information about the version from where it was created and yet we don't store it already.

Adding it version in integrate also passes the information to hero version.

## Testing notes:
1. New versions should have information about host name.

### DIFF
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -591,22 +591,6 @@ def create_instances_for_aov(
     # AOV product of its own.
 
     log = Logger.get_logger("farm_publishing")
-    additional_color_data = {
-        "renderProducts": instance.data["renderProducts"],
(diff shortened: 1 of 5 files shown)


iLLiCiTiT:
> Code change looks good. Should I be able to see that data in any meaninful place currently?
 
 Nope, it was something I realized when I looked at `Create hero version` action code.

> I'm just wondering whether "host_name" suffices, or whether we'd also want to store the running application variant and metadata along those lines, so that you can backtrace a publish was made from maya/2026 for example?
And perhaps even store the currently applied tools as well. Not sure how critical, but I can imagine that may be worth for debugging certain things in productions.

I'm not against adding app information too.

iLLiCiTiT:
Made PR in applications addon https://github.com/ynput/ayon-applications/pull/114

Issue #1537, opened 2025-11-10
YN-0048: Opening documents in browser


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Option to open document files like PDF, CSV, etc in Tray Browser.

  

We have an `OpenFile` action already in core, but it applies only to `render2d` product types. We may want to expose settings for this plug-int o allow e.g. any product type, or a studio specific product type and say allow it only in standalone Browser perhaps instead of also inside DCCs. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1412709602805682277/1412710928499216424)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0048](https://support.ayon.app/projects/Active_Clients/overview)_

PR #1548, opened 2025-11-13
Extended open file possibilities

iLLiCiTiT:
## Changelog Description
Extended `Open file` action to support open of any files that has associated application.

## Additional info
This implementation should be working on each platform. Changed icon to be more generic.

## Testing notes:
1. You can see `Open file` action for all representations that have assigned application.

Resolves https://github.com/ynput/ayon-core/issues/1537

### DIFF
--- a/client/ayon_core/plugins/loader/open_file.py
+++ b/client/ayon_core/plugins/loader/open_file.py
@@ -26,7 +26,7 @@ class _Cache:
 
     Notes:
         The cache is cleared when loader tool is refreshed so it might be
-            moved to other place which is not cleared of refresh.
+            moved to other place which is not cleared on refresh.
 
     """
     supported_exts: set[str] = set()


client/ayon_core/plugins/loader/open_file.py
```diff
@@ -13,6 +15,235 @@
 )
 
 
+WINDOWS_USER_REG_PATH = (
+    r"Software\Microsoft\Windows\CurrentVersion\Explorer\FileExts"
+    r"\{ext}\UserChoice"
+)
+
+
+class _Cache:
```
BigRoy: Would it make more sense to move this cache out of this action file so that it doesn't reload on plug-in rediscovery? Because currently it would, right?

client/ayon_core/plugins/loader/open_file.py
```diff
@@ -13,6 +15,234 @@
...
+
+
+def _filter_supported_exts(
+    extensions: set[str], test_func: Callable
+) -> set[str]:
+    filtered_exs: set[str] = set()
+    for ext in extensions:
+        if not _Cache.already_checked(ext):
+            r = test_func(ext)
+            _Cache.set_ext_support(ext, r)
```
iLLiCiTiT:
```suggestion
            _Cache.set_ext_support(ext, test_func(ext))
```

client/ayon_core/plugins/loader/open_file.py
BigRoy: True - I figured you did this because it was slow, but it seemed fine on my machine on Windows at least so 🤷‍♂️ 
Can you add a `# TODO` to hint that moving it out may be helpful if it's slow because it currently reloads with the action reload.

Then we can leave it as is. Thanks.

PR #1550, opened 2025-11-14
(YN-0075) Publish shot attributes on demand with setting

jakubjezek001:
## Summary
This enhancement introduces conditional shot attribute publishing functionality to the `CollectHierarchy` plugin, allowing users to control when shot attributes (Frame Start/End, Clip In/Out, etc.) are updated during publishing.

User is able to toggle the updating by going into Publisher's Context item and toggle `Ignore shot attributes on update`
image

## Changelog Description
Shot attributes can now be updated on demand with Shot product instance attribute toggle. When enabled, shot attributes like Frame Start/End and Clip In/Out will be updated during publishing. When disabled, existing shot attributes remain unchanged.

### Core Functionality
- **Enhanced `CollectHierarchy` plugin** with conditional shot attribute updating
  - Added `ignore_shot_attributes_on_update` setting with per-instance control
  - Introduced `AYONPyblishPluginMixin` for settings integration
  - Added API call optimization for checking existing folder entities

### Key Features Added
- **Conditional Attribute Updates**: Shot attributes are only updated when:
  - The shot folder already exists AND
  - The `ignore_shot_attributes_on_update` toggle is disabled
- **Instance-level Control**: Each shot instance can independently control attribute updating
- **Performance Optimization**: Batch API calls to check existing folder entities
- **Settings Integration**: Full integration with AYON's settings system

## Settings
- **Location**: `Project Settings > Core > Collect Hierarchy`
- **Option**: "Ignore shot attributes on update" (default: `false`)
- **Behavior**: When `true`, prevents updating shot attributes on existing shots

## Shot Attributes Controlled
The following shot attributes are conditionally updated:
- `handleStart` / `handleEnd` - Shot handle frames
- `frameStart` / `frameEnd` - Shot frame range  
- `clipIn` / `clipOut` - Editorial clip timing
- `fps` - Frame rate
- `resolutionWidth` / `resolutionHeight` - Shot resolution
- `pixelAspect` - Pixel aspect ratio

## Testing Notes
1. **Enable Attribute Updates**: Publish a shot with specific cut and duration settings, keep the toggle **OFF** (default)
   - ✅ Shot attributes like Frame Start/End and Clip In/Out should be updated
2. **Disable Attribute Updates**: Publish the same shot again with the toggle **ON** 
   - ✅ Shot attributes should remain unchanged from previous publish
3. **New Shot Creation**: Create a new shot with toggle **ON**
   - ✅ Shot attributes should still be set (only affects existing shots)

## Compatibility
- ✅ Backward compatible - defaults maintain existing behavior
- ✅ No breaking changes to existing workflows

## Related
- **Client ticket**: YN-0075

### DIFF
--- a/client/ayon_core/plugins/publish/integrate_traits.py
+++ b/client/ayon_core/plugins/publish/integrate_traits.py
@@ -5,6 +5,7 @@ import copy
 import hashlib
 import json
 from pathlib import Path
+from pprint import pformat
 from typing import TYPE_CHECKING, Any
 
 import pyblish.api
@@ -25,6 +26,7 @@ from ayon_core.lib import source_hash
 from ayon_core.lib.file_transaction import (
     FileTransaction,
 )
+from ayon_core.pipeline import is_product_base_type_supported
 from ayon_core.pipeline.publish import (
     PublishError,
     get_publish_template_name,
@@ -391,6 +393,21 @@ class IntegrateTraits(pyblish.api.InstancePlugin):
 
         # 9) Create representation entities
         for representation in representations:
+            attributes = {
+                "path": transfers[0].destination,
+                "template": transfers[0].template,
+            }
+
+            data = {"context": self.get_template_data_from_representation(
+                representation, instance)}
+
+            # Original integrator at this moment took all additional data
+            # on the representation and added them into either attribs or data.
+            # This should be avoided - we need to identify anything that
+            # is broken by this and move it to traits. Representation
+            # context in data is already handled by TemplateData trait, so
+            # the line above and any usage should be removed in the future.
+
             representation_entity = new_representation_entity(
                 representation.name,
                 version_entity["id"],
@@ -398,8 +415,8 @@ class IntegrateTraits(pyblish.api.InstancePlugin):
                     transfers,
                     representation,
                     anatomy=instance.context.data["anatomy"]),
-                attribs={},
-                data="",
+                attribs=attributes,
+                data=data,
                 tags=[],
                 status="",
             )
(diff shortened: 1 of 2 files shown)


client/ayon_core/plugins/publish/collect_hierarchy.py
```diff
@@ -41,6 +67,65 @@ def process(self, context):
...
+
+        # then we loop by all folder paths and try to find existing entity
+        existing_entities = {}
+        for folder_path in folder_paths:
+            found_entity = None
+            for entity in existing_entities:
+                if entity["path"] == folder_path:
+                    found_entity = entity
+                    break
+            existing_entities[folder_path] = found_entity
```
iLLiCiTiT:
```suggestion
        existing_entities = {
            folder_entity["path"]: folder_entity
            for folder_entity in ayon_api.get_folders(
                project_name, folder_paths=folder_paths)
        }
        for folder_path in folder_paths:
            existing_entities.setdefault(folder_path, None)
```

server/settings/publish_plugins.py
```diff
@@ -49,6 +49,14 @@ class CollectAudioModel(BaseSettingsModel):
     )
 
 
+class CollectHierarchyModel(BaseSettingsModel):
+    _isGroup = True
+    ignore_shot_attributes_on_update: bool = SettingsField(
+        False,
+        title="Ignore shot attributes on update"
```
rdelillo: I'm wondering if it shouldn't be the other way around ?
Like a `Update shot attributes` toogle that is on by default. Makes more sense to me to disable a toggle for something you don't want to happen.

jakubjezek001: you are right, lets flip the logic

Issue #1291, opened 2025-05-27
AY-7846_ExtractOIIOTranscode: Convert EXR to scanline whilst keeping all channels


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

I'd like to render tiled EXRs from Houdini Arnold (HtoA), but convert them to scanline EXRs during publish to improve efficiency while reading them in Nuke.

ExtractOIIOTranscode can be used for this by running the image through oiiotool  with the --scanline  argument. However, ExtractOIIOTranscode  as traditionally been used in AYON to solely do colorspace conversion and be the main go to for transcoding media for reviewables, and in particular converting and generating an RGBA output. Currently there is no way to trigger --scanline  conversion WITHOUT losing all additional channels.

It would be great to have a toggle per ExtractOIIOTranscode that could be used to enable/disable the "find a reviewable channel" logic so that it can also run in a non-reviewable focused mode where it's more a raw oiiotool  call and you can e.g. use the custom arguments to do your own thing.

(This potentially also aligns with the idea of perhaps supporting a method where it does no colorspace conversion at all, so that you can use the profile, regardless of colorspace data, regardless of reviewable to just run --scanline over the file.)

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7846](https://app.clickup.com/t/86c3nn7qg)_

PR #1553, opened 2025-11-17
Publish: Implement generic ExtractOIIOPostProcess plug-in

BigRoy:
## Changelog Description

Add new publish plugin: "OIIO Post Process" plugin.

This can be used to take any image representation through `oiiotool` to process with settings-defined arguments, to e.g. resize an image, convert all layers to scanline, etc.

## Additional info

This is separate from `ExtractOIIOTranscode` because that one is completely designed around the fact that it only tries to process an "RGB(A)" layer, usually for reviewables only.

Fix https://github.com/ynput/ayon-core/issues/1291

- [x] **Pending issue:**

There is still an issue in this PR that it runs over all representations - and hence, if an instance has a thumbnail jpeg and an exr sequence it will start trying to output a new named representation for both of them, making it fail to integrate. E.g. see this report:
[publish-report-251117-14-44.json](https://github.com/user-attachments/files/23583882/publish-report-251117-14-44.json)

As such we'll need to further filter running only on certain representations, e.g. only input EXRs.

Here's an example profile I used:
image

```json
{
  "enabled": true,
  "profiles": [
    {
      "product_types": [],
      "hosts": [],
      "task_types": [],
      "task_names": [],
      "product_names": [],
      "representation_names": ["exr"],
      "representation_exts": ["exr"],
      "delete_original": true,
      "outputs": [
        {
          "name": "exr_scanline",
          "extension": "exr",
          "input_arguments": [
            "-a"
          ],
          "output_arguments": [
            "--scanline"
          ],
          "tags": [],
          "custom_tags": []
        }
      ]
    }
  ]
}
```

## Testing notes:

1. Extract OIIO Post Process should work when a profile is set.

### DIFF
--- a/client/ayon_core/plugins/publish/extract_oiio_postprocess.py
+++ b/client/ayon_core/plugins/publish/extract_oiio_postprocess.py
@@ -43,6 +43,7 @@ class ExtractOIIOPostProcess(publish.Extractor):
         if instance.data.get("farm"):
             self.log.debug("Should be processed on farm, skipping.")
             return
+
         if not self.profiles:
             self.log.debug("No profiles present for OIIO Post Process")
             return


client/ayon_core/plugins/publish/extract_oiio_postprocess.py
```diff
@@ -0,0 +1,338 @@
...
+
+    label = "OIIO Post Process"
+    order = pyblish.api.ExtractorOrder + 0.020
+
+    settings_category = "core"
+
+    optional = True
+
+    # Supported extensions
+    supported_exts = {"exr", "jpg", "jpeg", "png", "dpx"}
```
antirotor: any reason why not to use [IMAGE_EXTENSIONS](https://github.com/ynput/ayon-core/blob/5bccc7cf2bcdc51e446cf7794f09d6081ff7255a/client/ayon_core/lib/transcoding.py#L49) - too broad perhaps for OIIO?

antirotor (approved):
The code makes sense. It always hurts when I see the use of the sequence vs. file/list vs. string logic but there is still long way to get rid of it and no other way around yet.

So if I got it right, you need to specify explicit profile to match to trigger this. It avoids the problem with thumbnails - or rather it passes it to whomever defines the profile. I think we should mention this in documentation. Not that it behaves differently from other places just that I saw too many cases of broken workkflows just because the profile were wrong.

BigRoy:
> The code makes sense. It always hurts when I see the use of the sequence vs. file/list vs. string logic but there is still long way to get rid of it and no other way around yet.
> 
> So if I got it right, you need to specify explicit profile to match to trigger this. It avoids the problem with thumbnails - or rather it passes it to whomever defines the profile. I think we should mention this in documentation. Not that it behaves differently from other places just that I saw too many cases of broken workkflows just because the profile were wrong.

Agreed - and yes, sounds about right.

MustafaJafar:
Could we add this to the PR changelog description?
Add new publish plugin: "OIIO Post Process" plugin.

PR #1554, opened 2025-11-18
Workfile template builder: Consider Folder in filtering

MustafaJafar:
## Changelog Description
Add `folderpaths` for template profile filtering

## Additional info
Btw, this will make it work with other DCCs if they have the setting for folder paths added like here https://github.com/ynput/ayon-houdini/pull/318 which only adds this setting.

## Testing notes:
1. template workfile build should work as before in different DCCs.
2. In Houdini DCC with https://github.com/ynput/ayon-houdini/pull/318 PR, workfile templates selection should follow specified folder paths.

Resolve https://github.com/ynput/ayon-houdini/issues/315

### DIFF
--- a/client/ayon_core/pipeline/workfile/workfile_template_builder.py
+++ b/client/ayon_core/pipeline/workfile/workfile_template_builder.py
@@ -853,7 +853,6 @@ class AbstractTemplateBuilder(ABC):
             "folder_paths": folder_path,
         }
 
-
         build_profiles = self._get_build_profiles()
         profile = filter_profiles(
             build_profiles,


MustafaJafar:
I test building maya workfile template, I don't know why it doesn't build automatically on launching the app but it works as before when clicking `Build from template` action in AYON menu.

BigRoy:
@MustafaJafar while we're at it, can you please add Folder Types as well? I feel like it may be just as common or even more common to then start selecting by Folder type instead of Folder name.

BigRoy (approved):
Works - but **I'd love to see folder types filter added too.** I'd say, can be added to this PR.
In settings it can use the `folder_types_enum` from backend.

Also, I **wonder** if core should start exposing a `WorkfileTemplateBuilderProfileModel` on the backend for other addons to start inheriting from - so that the actual exposed settings attributes per addon don't need manual updating for each single addon over time. But I suppose it may be tricky to import those 'across addons' if nothing ensures e.g. core is in the addon 🤷‍♂️. Anyway, would be nice if we can avoid duplicating this maintenance everywhere.

iLLiCiTiT:
Ok, then please add folder type here, and to the houdini PR too.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -840,14 +840,23 @@ def get_template_preset(self):
...
+
+        filter_data = {
+            "task_types": task_type,
+            "task_names": task_name,
+            "folder_paths": folder_path
+        }
+
+        folder_entity = self.current_folder_entity
+        if folder_entity:
+            filter_data.update({"folder_types": folder_entity["folderType"]})
```
iLLiCiTiT:
```suggestion
        folder_path = self.current_folder_path
        folder_type = None
        folder_entity = self.current_folder_entity
        if folder_entity:
            folder_type = folder_entity["folderType"]

        filter_data = {
            "task_types": task_type,
            "task_names": task_name,
            "folder_types": folder_type,
            "folder_paths": folder_path,
        }

```

PR #1561, opened 2025-11-26
OCIO Color management: Allow profiles to also choose to disable OCIO management

BigRoy:
## Changelog Description

Allow OCIO color management profiles that have no matching profile or… do match a profile with "disabled" status to be considered as NOT color managed. So that you can specify a particular part of the project to NOT be OCIO color managed.

## Additional info

image

Fix https://github.com/ynput/ayon-core/issues/1563

Related to [YN-0245](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=task&id=38ee98a0cb2311f084cde5e8803d9fc9)

## Testing notes:
1. Color management should still work
2. Color management should now be able to allow to be enabled but have a context where no profile matches (so that that context has OCIO disabled)
3. Color management can now define profiles that are explicitly not color managed / OCIO.

### DIFF
--- a/client/ayon_core/pipeline/create/product_name.py
+++ b/client/ayon_core/pipeline/create/product_name.py
@@ -41,8 +41,8 @@ def get_product_name_template(
     profiles = tools_settings["creator"]["product_name_profiles"]
     filtering_criteria = {
         "product_types": product_type,
-        "hosts": host_name,
-        "tasks": task_name,
+        "host_names": host_name,
+        "task_names": task_name,
         "task_types": task_type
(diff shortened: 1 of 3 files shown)


rdelillo (approved):
Did not test it locally, but those changes makes sense to me.
Do you want me to test it locally as well or shall we wait for client feedback ?

BigRoy:
Client confirmed it works for them too.
@iLLiCiTiT can you do code review to see if you're ok with the changes? Then we can merge it.

PR #1566, opened 2025-11-29
Product name template: Fix profiles filtering

iLLiCiTiT:
## Changelog Description
Fix filtering keys used for profiltes filtering.

## Additional info
Changed with https://github.com/ynput/ayon-core/pull/1546 but not modified where it is being used.

## Testing notes:
1. Product name template profiling should work now.

Resolves https://github.com/ynput/ayon-substance-painter/issues/61

### DIFF
--- a/server/settings/conversion.py
+++ b/server/settings/conversion.py
@@ -164,5 +164,6 @@ def convert_settings_overrides(
 ) -> dict[str, Any]:
     _convert_imageio_configs_0_3_1(overrides)
     _convert_imageio_configs_0_4_5(overrides)
+    _convert_imageio_configs_1_6_5(overrides)
     _convert_publish_plugins(overrides)
     return overrides


BigRoy (approved):
It was the lack of settings conversion that bit me - and it **will** bite others if they update through core 1.6.10 or 1.6.11 from <1.6.10 because they'll lose host names and task names mappings in the product name profiles on copying settings.

This PR now fixes it. But we may want to somehow be wary for anyone going through those in-between releases.

Works fine with Substance Painter at least, and settings copy fine too.

moonyuet:
> @moonyuet it is the same issue, but likely you also have the 'broken' fixed overrides already (likely already had settings for this release version?)
> 
> If you remove those overrides, does it work?

Yes, it is working after removing overrides and deprecated attributes, as you see in the screenshot above.

Issue #1562, opened 2025-11-27
Version Locking does not work (Nuke, Maya tested)


LiborBatek:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

When loading some product version and use `Actions>Lock Version` it actually does not lock it and when performing `Update Version to latest` or so it also affect the locked ones...

Also worth noting that the `lock` icon not been visible inside Maya host but only in Nuke.

Im enclosing a vid with this issue from Nuke 16.0

https://github.com/user-attachments/assets/facb7516-48d5-4250-b797-bf8c4040c397

### Expected Behavior:

The `Lock Version` action should actually lock the product version so it wont change even when performing `Manage>Update version to lates` or other managing action related to versions.

### Version

1.6.11

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Load any suitable product version
2. Perform lock version via Scene Inventory actions
3. Perform `Update Version to latest` or similar action like `Set Version`
4. The locked product version still been updated no matter it should be locked.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the bug report.

### Relevant log output:

```shell

```

LiborBatek:
@BigRoy after small talk with @iLLiCiTiT  it seems that it actually does not "physically" lock the version to my surprise but only turn off validation errors for such locked (and possibly also outdated) versions. So the only issue been present in `maya` host where the lock icon is not shown when locking the version...


IMHO it should be more fledged feature as its name `lock version`  suggests and actually lock it - if so these versions wont be updated via manage actions at all and stays as locked no matter what until the user unlock them back to normal state...

Just my 2 cents.

PR #1570, opened 2025-12-01
Scene manager: Ignore locked containers when updating versions

iLLiCiTiT:
## Changelog Description
Containers with locked versions cannot be updated using scene manager UI.

## Testing notes:
1. Lock container that is not last from last version.
2. Rub update to last action.
3. It should not change.

Resolves https://github.com/ynput/ayon-core/issues/1562


LiborBatek (approved):
Yeah after using latest develop of `ayon-maya` together with this PR all been working fine!

LGTM

PR #1576, opened 2025-12-03
Chore: Skip base classes in plugin discovery

iLLiCiTiT:
## Changelog Description
Added option to mark base classes with `skip_discovery= True` to skip them during discovery.

## Additional info
This should help with marking custom base classes in other addons.

## Testing notes:
1. Mark base plugin with `is_base_class = True`.
2. Import the plugin in a file with "real" plugin.
3. It should not be "used" or logged out as abstract plugin.

NOTE: Do not use for plugins that do override all abstract attributes as those will be considered.

### DIFF
--- a/client/ayon_core/pipeline/load/plugins.py
+++ b/client/ayon_core/pipeline/load/plugins.py
@@ -21,6 +21,13 @@ from .utils import get_representation_path_from_context
 class LoaderPlugin(list):
     """Load representation into host application"""
 
+    # Attribute 'skip_discovery' is used during discovery phase to skip
+    #   plugins, which can be used to mark base plugins that should not be
+    #   considered as plugins "to use". The discovery logic does NOT use
+    #   the attribute value from parent classes. Each base class has to define
+    #   the attribute again.
+    skip_discovery = True
+
     product_types: set[str] = set()
     product_base_types: Optional[set[str]] = None
     representations = set()


iLLiCiTiT:
> But I've never seen anything like this used before. Making me wonder, is it too odd to do it this way? Would it make more sense to just mark it an abstract class with a decorator (even if it has no abstract properties) and we can filter it out that way? Or?

Well, the decorator would do exactly this 🙂. There is no way how to mark a class without affecting children, except this one.

BigRoy (approved):
If there's no default Python convention for marking something a base class, I guess this convention we come up with is fine.

I'd await review by @philippe-ynput however.

iLLiCiTiT:
> could this be solved by .mro() in the filter?

How?

The goal is that you explicitly mark a class as base class so it is ignored during discovery.

```python
# This should be always ignored during discovery
# - it might not be abstract anymore because the
#    base implementation is same for all create plugins
#    and differs only on class attributes
# - also we don't want to report this plugin as "abstract"
#   if is imported in a file
class MayaCreatePlugin(...):
    is_base_class = True
```
```python
# create_rig.py - with real plugin
from ayon_maya.api.plugins import MayaCreatePlugin


class CreateRig(MayaCreatePlugin):
    def create(...):
        ...
```

If `create_rig.py` is looked up in discovery it will find 2 classes that inherit from create plugin `CreateRig` and `MayaCreatePlugin`. Because there is `MayaCreatePlugin` it will be reported as "Hey there is abstract class, something is wrong".

client/ayon_core/pipeline/plugin_discover.py
```diff
@@ -138,7 +138,14 @@ def discover_plugins(
         for item in modules:
             filepath, module = item
             result.add_module(module)
-            all_plugins.extend(classes_from_module(base_class, module))
+            for cls in classes_from_module(base_class, module):
+                if cls is base_class:
+                    continue
+                # Class has defined 'is_base_class = True'
+                is_base_class = cls.__dict__.get("is_base_class")
```
philippe-ynput: Result should always be a bool ?
```suggestion
                is_base_class = cls.__dict__.get("is_base_class", False)
```

PR #1588, opened 2025-12-09
Extract Review: Substance Painter reviewables do not integrate

BigRoy:
## Changelog Description

With https://github.com/ynput/ayon-core/pull/1526 we got reviewables for substance painter textures, but they also got integrated as representations, like `png_png`.

## Additional info

With these changes it should only make the reviewable upload.

See ayon chat conversation; https://discord.com/channels/517362899170230292/563751989075378201/1447931139263168553

Community and @LiborBatek says that these are working settings:

image

https://discord.com/channels/517362899170230292/1447896320101580851/1447972537060298902

## Testing notes:

1. Reviewables should be uploaded, but no extra representation should be integrated.

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -285,6 +285,7 @@ class CoreSettings(BaseSettingsModel):
         "{}",
         title="Global environment variables",
         widget="textarea",
+        syntax="json",
         scope=["studio"],
     )
     update_check_interval: int = SettingsField(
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
As I have tested this does not work...when using tags `delete` and `need_thumbnail` which results into not having any sort of reviewable but just thumbnail...
image

so I guess this combo of `tags` is not working for having reviewable and must be adjusted.
image


On positive side of things, yes there is no additional repre integrated tho...

There must be probably different `tag`  used other than `need_thumbnail`

iLLiCiTiT:
I guess this deservers an issue. Right now there is a lot of compatibilities and checks that are out of date, or out of use-cases. We could go through all of that and change the thumbnail creation behavior.

PR #1596, opened 2025-12-10
USD Contributions: Allow specifying a strength ordering offset per instance 

### LINKED ISSUE #1595: YN-0304: USD Contributions customizable order strength per instance

When multiple instances are targeting the same department layer, e.g. like when the department layer will hold a "proxy", "lods" and the main contribution we may want to nudge e.g. the proxy to have a slightly lower layer strength that the other, etc.

  

For this, the approach may be to allow an instance to "nudge" itself in the ordering with an integer on the instance attributes. 

### Additional context:

Source [discussion](https://discord.com/channels/517362899170230292/1442506412046549123/1442870830114803795).

_This issue was automatically created from AYON ticket [YN-0304]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=573686f0d5ba11f08be7e949d558f8cd)

BigRoy:
## Changelog Description

USD Contributions: Allow specifying a strength ordering offset per instance.

When multiple instances are targeting the same department layer, e.g. like when the department layer will hold a "proxy", "lods" and the main contribution we may want to nudge e.g. the proxy to have a slightly lower layer strength than the others, etc.

For this, the approach here is to allow an instance to "nudge" itself in the ordering with an integer on the instance attributes.

## Additional info

image

This only affects the ordering within the department layer - not the department layer itself (because a full department layer may encompass more instances). 

### ~~Question! ⚠️~~ The attribute now defines an in-layer order, not an offset to department layer order. 

This raises the question whether INSIDE the department layer whether it even makes sense to order the layers initially with the department layer's order - or instead just have a dedicated order number of themselves.

So that e.g. department layer at order 100 may have individual contributions inside at e.g. order 0, 1 and 2 instead of 100, 101 and 102. The nice thing is by not having the department layer order included it means that if you were to switch department layer order during productions it does not affect the ordering inside the department layer.

So, do we want:
```
usdAsset
    usdAsset_model (order: 100)
        modelMain (order: 0)
        modelMainLODs (order: -1)
        modelMainProxy (order: -2)
```
Or:
```
usdAsset
    usdAsset_model (order: 100)
        modelMain (order: 100)
        modelMainLODs (order: 99)
        modelMainProxy (order: 98)
```
~~Currently it's the latter in this PR but I wonder if changing it around to be the first makes more sense. @antirotor thoughts?~~

With https://github.com/ynput/ayon-core/commit/ced9eadd3d3d182df5341cd3f60897058dfdf9b3 it is now the first one.

Fix #1595 

## Testing notes:
1. A new strength order offset attribute should be visible on USD publishes: "Strength order offset"
2. Changing it should affect how to overlapping contributions (e.g. multiple contributions to a department layer) should be ordered in strength.

### DIFF
--- a/client/ayon_core/lib/transcoding.py
+++ b/client/ayon_core/lib/transcoding.py
@@ -134,16 +134,29 @@ def get_transcode_temp_directory():
     )
 
 
-def get_oiio_info_for_input(filepath, logger=None, subimages=False):
+def get_oiio_info_for_input(
+    filepath: str,
+    *,
+    subimages: bool = False,
+    verbose: bool = True,
+    logger: logging.Logger = None,
+):
     """Call oiiotool to get information about input and return stdout.
+
+    Args:
+        filepath (str): Path to file.
+        subimages (bool): include info about subimages in the output.
+        verbose (bool): get the full metadata about each input image.
+        logger (logging.Logger): Logger used for logging.
 
     Stdout should contain xml format string.
     """
     args = get_oiio_tool_args(
         "oiiotool",
         "--info",
-        "-v"
     )
+    if verbose:
+        args.append("-v")
     if subimages:
         args.append("-a")
 
@@ -573,7 +586,10 @@ def get_review_layer_name(src_filepath):
         return None
 
     # Load info about file from oiio tool
-    input_info = get_oiio_info_for_input(src_filepath)
+    input_info = get_oiio_info_for_input(
+        src_filepath,
+        verbose=False,
+    )
     if not input_info:
         return None
 
@@ -1234,7 +1250,11 @@ def oiio_color_convert(
         for token in ["#", "%d"]:
             first_input_path = first_input_path.replace(token, first_frame)
 
-    input_info = get_oiio_info_for_input(first_input_path, logger=logger)
+    input_info = get_oiio_info_for_input(
+        first_input_path,
+        verbose=False,
+        logger=logger,
+    )
 
     # Collect channels to export
     input_arg, channels_arg = get_oiio_input_and_channel_args(input_info)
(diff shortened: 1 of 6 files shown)


BigRoy:
> @BigRoy I like what you've suggested. It sounds more flexible and easy to read what offset was chosen.
> 
> usdAsset usdAsset_model (order: 100) modelMain (order: 0) modelMainLODs (order: -1) modelMainProxy (order: -2)

Adjusted with https://github.com/ynput/ayon-core/pull/1596/commits/ced9eadd3d3d182df5341cd3f60897058dfdf9b3

moonyuet:
I have tested with LODs object and rig assets. Seems to work as expected
I am pretty confused by your screenshot when the maya has different name for the offset setting, but once I look at it again, I find out your screenshot is blender host.
image
image

MustafaJafar (approved):
It works..

---

Adding some info for documentation purposes... 

I don't think this has any effect on publishing variants. (in this case the order is added as `ayon_order`  int property).
However, this can be pretty useful when publishing with `Add as variant` disabled. either publishing an asset or shot. In both cases your publishes will be sublayered.
image
In this case the order is meaningful as it affects the layer order within the department layer. 
```
    subLayers = [
        @./..\..\..\model\modelMain\v001\Expr_houdini_modelMain_v001.usd:SDF_FORMAT_ARGS:layer_id=modelMain&order=1@,
        @./..\..\..\model\modelBig\v001\Expr_houdini_modelBig_v001.usd:SDF_FORMAT_ARGS:layer_id=modelBig&order=-1@,
        @./..\..\..\model\modelSmall\v001\Expr_houdini_modelSmall_v001.usd:SDF_FORMAT_ARGS:layer_id=modelSmall&order=-2@
    ]
```
---

image

When publishing with `Add as variant` enabled but the variant name is the same, your publishes will override each other where the last process instance will be the one present in the variant set.

```
    variantSet "model" = {
        "main" (
            prepend references = [
                @./..\..\..\model\modelBig\v002\Expr_houdini_modelBig_v002.usd@ (
                    customData = {
                        int ayon_order = -1
                        string ayon_uri = "ayon://Experiments//random/houdini?product=modelBig&version=2&representation=usd"
                    }
                )
            ]
        ) {

        }
    }
```

BigRoy:
> When publishing with Add as variant enabled but the variant name is the same, your publishes will override each other where the last process instance will be the one present in the variant set.

Ah - good to know - we may want to solve that at some point if there's an actual need for it.

Issue #1524, opened 2025-10-31
YN-0156: USD contribution workflow layer strength configured hierarchically


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

AYON USD contribution workflow configure department layer strengths/orders in a hierarchical manner instead of flat, so that e.g. shot "lookdev" can be ordered differently than asset "lookdev".

  

Currently AYON defines a customizable single list of department layers for all USD contributions.  Instead we want to have a tree of arrays, so they can define.

- Department strength order is configurable by hierarchy, e.g. Asset order, Sequence order, Shot order, Global order.

In particular it’s mostly due to difference in order of LOOK department layer in ASSET versus SHOTS.

  

A workaround for now is to use department layer names that are unique, like `asset_look` but it essentially means more clutter on the department layer names. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1422572883833651312)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0156](https://support.ayon.app/projects/Active_Clients/overview)_

PR #1597, opened 2025-12-10
USD Contributions: Allow department layers scoped to either "shot" or "asset"

BigRoy:
## Changelog Description

Allow to define department layers scoped only to a particular department layer type, e.g. "shot" versus "asset". This way, you can scope same layer names for both shot and asset at different orders if they have differing target scopes.

## Additional info

This adds a "Scope" for department layer orders so they only are valid for e.g. asset versus shot layers.

image

With that, the department layers in publisher UI also get filtered to department layers applicable only to the "init_as" state of asset versus shot.
image

image

Hence, this allows you to build more granular department layer orders between the different targets that still but share the same name.

**Note:** The scope also does allow multiselection if you want to essentially scope both at once.

Fix https://github.com/ynput/ayon-usd/issues/87
Fix https://github.com/ynput/ayon-core/issues/1524

## Testing notes:

1. Existing instances should adhere to their existing layer.
2. Layer should be preserved/saved and refreshed on recollect in the publisher UI on user customizations.
3. The department layer and its order should be correctly applied to the instance product on publish

### DIFF
--- a/server/settings/publish_plugins.py
+++ b/server/settings/publish_plugins.py
@@ -94,6 +94,7 @@ class ContributionLayersModel(BaseSettingsModel):
         # this default directly did not show any ill side effects.
         default=["asset", "shot"],
         title="Scope",
+        min_items=1,
         enum_resolver=usd_contribution_layer_types)
     order: int = SettingsField(
         default=0,


MustafaJafar:
When publishing a USD shot, the order seems not correct?
```
    subLayers = [
        @./..\..\usdShot_layout\v001\Expr_USD_Render_usdShot_layout_v001.usd:SDF_FORMAT_ARGS:layer_id=layout&order=0@
    ]
```

It's correct with in my USD Asset
```
    subLayers = [
        @./..\..\usdAsset_look\v002\Expr_usd_assets_usdAsset_look_v002.usd:SDF_FORMAT_ARGS:layer_id=look&order=200@,
        @./..\..\usdAsset_model\v002\Expr_usd_assets_usdAsset_model_v002.usd:SDF_FORMAT_ARGS:layer_id=model&order=100@
    ]
```

BigRoy:
> I wonder if this can be a dynamic enum.

That would possible, we could apply `enum_resolver`:

```
async def contribution_layers_enum(
    project_name: str | None = None,
    addon: BaseServerAddon = None,
    settings_variant: str = None,
):
    if project_name is None:
        settings = await addon.get_studio_settings(variant=settings_variant)
    else:
        settings = await addon.get_project_settings(
            project_name=project_name, variant=settings_variant
        )

    plugin_settings = settings.publish.CollectUSDLayerContributions
    layer_names = {
        contribution_layer.name for contribution_layer
        in plugin_settings.contribution_layers
    }
    result = []
    for layer_name in sorted(layer_names):
        result.append({
            "value": layer_name,
            "label": layer_name,
        })
    return result
```

It's just that as soon as you start changing the layer names and want to update below in profiles, you first need to save before the enum is refreshed it seems - and as you do that you've essentially lost your current selection state. Quite annoying?

> Also, I think these profiles are missing the initialize as.

Correct - I think it's still based on whether target product is `usdAsset` or `usdShot` :)
Definitely worth an issue, but no high priority.

BigRoy:
> When publishing a USD shot, the order seems not correct?
> 
> ```
>     subLayers = [
>         @./..\..\usdShot_layout\v001\Expr_USD_Render_usdShot_layout_v001.usd:SDF_FORMAT_ARGS:layer_id=layout&order=0@
>     ]
> ```
> 

I can't seem to reproduce this? How? :)

MustafaJafar:
> It's just that as soon as you start changing the layer names and want to update below in profiles, you first need to save before the enum is refreshed it seems - and as you do that you've essentially lost your current selection state. Quite annoying?

Correct. However, it'll also minimize typo errors. Anyways, users can copy/paste the department layer names.
I think it's better to share the feature with the client and wait their feedback.

MustafaJafar (approved):
works.

BigRoy:
> Just answering myself.. I think I'm mixing the order of department layer and the order of the contribution within a department layer. `usdMain` order's is `0` ()order within layout department layer. and it can be controlled by #1596 `usdShot_layout` order's is `200`
> 
> and the outcome I'm facing is the expected result..

Thanks so much - was almost about to go on a hunt. Happy to hear it works!

@MustafaJafar last question - how do you feel about perhaps, when "copying settings" from older bundles that all scopes are forced to both asset and shot. To ensure backwards compatibility? I'm worried that previously someone may have e.g. used a shot department layer in asset build (or the `init_as` value was wrong previously without knowing) and suddenly now it'd start going to a different layer because the different layer does not match the unintentional scope?

BigRoy:
> > @MustafaJafar last question - how do you feel about perhaps, when "copying settings" from older bundles that all scopes are forced to both asset and shot. To ensure backwards compatibility? I'm worried that previously someone may have e.g. used a shot department layer in asset build (or the `init_as` value was wrong previously without knowing) and suddenly now it'd start going to a different layer because the different layer does not match the unintentional scope?
> 
> I can't tell actually, I keep getting `loading...` image
> 
> I can only see this error in console.
> 
> ```
> Uncaught (in promise) TypeError: Cannot convert undefined or null to object
>     at Object.keys (<anonymous>)
>     at w (AddonSettings-pgoMJIiL.js:111:3237)
>     at async AddonSettings-pgoMJIiL.js:111:4783
> ```

@MustafaJafar are you copying from a very old bundle perhaps and see something like this in your docker server logs:
```
DEBUG api.logging | Environment's can't be parsed as json object
Extra data: line 1 column 4 (char 3)
```
?

MustafaJafar:
The issue with `loading...` is fixed.
image
However, Scope is left unset.
image

BigRoy:
> The issue with `loading...` is fixed. image However, Scope is left unset. image

Ah sorry, I commented this on Discord - not here. Let me copy paste.

> Ok, I'll skip the conversion - because it won't apply it if previously it didn't have any overrides... (because it only considers overrides). Then I'd have to essentially enforce some default overrides completely if it's an older release which I didn't really want to do.

I can implement it, but it'd make it so that when copying settings from an older release that it overrides it with a fully hardcoded list of overrides which I'm a bit wary of to be honest since that can easily go out of sync in the future.

MustafaJafar:
> I can implement it, but it'd make it so that when copying settings from an older release that it overrides it with a fully hardcoded list of overrides which I'm a bit wary of to be honest since that can easily go out of sync in the future.

there's an edge case in my mind, if someone copied settings and left scope empty. didn't test it yet.

BigRoy:
@MustafaJafar now copying overrides from before will transfer with asset + shot both as the default. Also, newly added entries in the list will default to it too. Had to debug a bit why `default_factory` did not come through to the frontend, but Martin explained to me that of course frontend can't interpret the python callable so had to resort to using `default=["asset", "shot"]` which does work fine.

---

Also added a warning for the case if there are no entries. The case of "scope" being empty is now captured by disallowing them to be empty with the `min_items=1` requirement on settings

Note that originally I tried:
```python
    @validator("scope")
    def validate_scope(cls, value):
        if not value:
            raise BadRequestException(
                "Scope must have at least one item assigned."
                " It can't be empty."
            )
```
But that somehow forced the values to be empty when loading the page. Luckily `min_items` was exactly what I needed here.

Edit: the validator issue described there was because it should return the original value.

> Martin: i think validators must return the value - they can modify it in-place

```python
    @validator("scope")
    def validate_scope(cls, value):
        if not value:
            raise BadRequestException(
                "Scope must have at least one item assigned."
                " It can't be empty."
            )
        return value
```
So that worked fine.
But the `min_items=1` is the better approach here.

---

The only case where it does not scope to both "asset" and "shot" is when copying from an older addon that did not have "scope" attribute  yet and it did not have any overrides previously. 
Then the only it's the one edge case and I don't want to bother storing a full 'override list' just for that. So I'd say, this is now good to go.

PR #1599, opened 2025-12-10
Publisher Create: Do not fallback to current context if no folder or task selected

### LINKED ISSUE #1598: Publisher: Create enforces 'current task' if no task is selected inside DCC

### Current Behavior:

When Create Context has a 'current task' in context, then deselecting a task left side still creates instances in the current task (if creating under the same folder).

### Expected Behavior:

Deselecting the task should generate the instance 'task-less'.

### Version

1.6.13

Windows

### Steps To Reproduce:

I can reproduce it in Maya, Houdini and Fusion using this plug-in:
```python
from ayon_core.pipeline import Creator


class TestCreator(Creator):
    identifier = "test"
    label = "Test"
    product_type = "render"
    product_base_type = "render"

    def collect_instances(self):
        return

        pass

    def update_instances(self, update_list):
        pass

    def remove_instances(self, instances):
        pass
```

I mean, create does nothing of course - but it can at least trigger it... and you can also see the product name pre-populated with current task.

image

Note that it's also reproducable with Maya and Fusion's own render creators - the above example snippet is just to show that it's not due to 'overrides' on the creator plug-in itself.

### Relevant log output:

```shell

```

### Additional context:

Reported here originally: https://github.com/ynput/ayon-core/pull/1462#issuecomment-3638455050

BigRoy:
## Changelog Description

Do not fallback to current task name if no task selected

## Additional info

Fix #1598

## Testing notes:

1. Deselecting task should not fall back to the active task name

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/create_widget.py
+++ b/client/ayon_core/tools/publisher/widgets/create_widget.py
@@ -310,9 +310,6 @@ class CreateWidget(QtWidgets.QWidget):
         folder_path = None
         if self._context_change_is_enabled():
             folder_path = self._context_widget.get_selected_folder_path()
-
-        if folder_path is None:
-            folder_path = self.get_current_folder_path()
         return folder_path or None
 
     def _get_folder_id(self):


client/ayon_core/tools/publisher/widgets/create_widget.py
```diff
@@ -328,9 +328,6 @@ def _get_task_name(self):
             folder_path = self._context_widget.get_selected_folder_path()
             if folder_path:
                 task_name = self._context_widget.get_selected_task_name()
-
```
iLLiCiTiT: BTW `_get_folder_path` also returns current folder path if nothing is selected. I wonder if there was a reason to do this?

iLLiCiTiT: Yes, but in that case `None` is passed to them (AFAIK). So it doesn't seem to have a reason...

BigRoy: Fixed - seems to work as expected.

PR #1607, opened 2025-12-15
Console interpreter: Allow to change registry name

iLLiCiTiT:
## Changelog Description
Constole interpreter controller allows to use different name of registry where metadata are stored.

## Additional info
This might be usefull when the UI is used at different place than tray.

## Testing notes:
Validate code changes.

### Custom name usage
This will use `my_dcc_console.json` to store tabs content instead of the default one.
```python
from ayon_core.tools.console_interpreter import InterpreterController
from ayon_core.tools.console_interpreter.ui import ConsoleInterpreterWindow

controller = InterpreterController("my_dcc_console")
window = ConsoleInterpreterWindow(controller)
window.show()
```


BigRoy (approved):
Works flawlessly!

PR #1608, opened 2025-12-15
Console Interpreter: Override full stdout and stderr

iLLiCiTiT:
## Changelog Description
Override `sys.stdout` and `sys.stderr` instead of their `write` method.

## Additional info
This approach should help to handle DCCs which are using own wrappers of stdout and stderr which don't allow to customize `write`, e.g. Maya.

## Testing notes:
1. Console Interpreter UI should work in tray (ideally for all OS).
2. It also does work in hosts (e.g. in Maya).

### DIFF
--- a/client/ayon_core/lib/transcoding.py
+++ b/client/ayon_core/lib/transcoding.py
@@ -134,16 +134,29 @@ def get_transcode_temp_directory():
     )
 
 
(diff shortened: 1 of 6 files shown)


BigRoy (approved):
Works flawlessly standalone - and also makes it work nicely in Maya. 👍

Issue #1470, opened 2025-10-08
YN-0067: Publisher crashed plugins


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Add strict validation of crashed plugins discovery to publisher.

  

Add a setting to enable stricter validation which is off by default. The setting should also contain a list of regexes to be able to skip the validation to filepaths that do match the regex, in case it is crashing but not relevant plugin that would only block prouction.

When publisher is opened it would show user a dialog with something like "Failed to load plugins that do prevent you from using publish tool. Please contact your administrator." with list of crashed files, and Details and Export button that would also store tracebacks (client suggested to show tracebacks directly in the dialog, but longer traceback may overflow the screen size and screenshot would not be helpful in that case).

That would also prevent user to continue to work in publisher. 

### Additional context:
link to discussion on [Discord]()
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0067](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=8b36f930992911f0a113d5f9fc86937d)_

PR #1612, opened 2025-12-16
Plugins discovery: Strict publish plugins discovery

iLLiCiTiT:
## Changelog Description
Added strict mode to plugins discovery. Strict mode won't let publishing to happen if there are plugin paths that failed to import.

## Additional info
There is an option to ignore crashed filespath using regex. Right now it validates only publish plugins, and does not handle abstract plugins.

### Screenshot
image

## Testing notes:
1. Enable `ayon+settings://core/publish/discover_validation/enabled`.
2. Open Publisher in a host.
3. If there are crashed plugins it won't let you to create or publish, and dialog is shown.
4. If you add regex to filter out the the filepath it should work on refresh.

Resolves https://github.com/ynput/ayon-core/issues/1470

### DIFF
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -2,6 +2,7 @@ from operator import attrgetter
 import dataclasses
 import os
 import platform
+from collections import defaultdict
 from typing import Any, Dict, List
 
 import pyblish.api
@@ -13,10 +14,11 @@ except ImportError:
 from ayon_core.lib import (
     TextDef,
     BoolDef,
+    NumberDef,
     UISeparatorDef,
     UILabelDef,
     EnumDef,
-    filter_profiles, NumberDef
+    filter_profiles,
 )
 try:
     from ayon_core.pipeline.usdlib import (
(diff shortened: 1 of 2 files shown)


BigRoy (approved):
Works well - with the regex used, it's also able to differentiate which are blocking and which aren't: 

Image

When the plug-in is fixed and you reset, it also correctly unblocks it again.

Tested in standalone, Blender and Maya. 👍 I do think we should move the setting like I commented - other than that all good.

---

Minor side note: if you run this PR without uploading the package to the server you get:
```python
C:\Users\User\AppData\Local\Temp\traypub_3l153sny.json
  File "F:\dev/ayon-core/client\ayon_core\tools\publisher\window.py", line 591, in _on_show_timer
    self.reset()
  File "F:\dev/ayon-core/client\ayon_core\tools\publisher\window.py", line 654, in reset
    self._controller.reset()
  File "F:\dev/ayon-core/client\ayon_core\tools\publisher\control_qt.py", line 86, in reset
    super().reset()
  File "F:\dev/ayon-core/client\ayon_core\tools\publisher\control.py", line 382, in reset
    self._publish_model.reset()
  File "F:\dev/ayon-core/client\ayon_core\tools\publisher\models\publish.py", line 970, in reset
    blocking_crashed_paths = filter_crashed_publish_paths(
  File "F:\dev/ayon-core/client\ayon_core\pipeline\publish\lib.py", line 283, in filter_crashed_publish_paths
    project_settings["core"]["publish"]["discover_validation"]
KeyError: 'discover_validation'
```
Not a big issue, but something to be aware of when running dev mode.

kalisp:
I tried it in Nuke, I introduced error in `process` part, which failed in runtime, but I couldn't then Validate nor Publish anymore after Refresh.
image
(This is not happening in develop.)

(I fixed it by introducing syntax error to the plugin (in import), this resulted in plugin to fail loading, I got a warning, I fixed import, refreshed and Validate/Publish were back.)

kalisp (approved):
This seems to be working.

Found different issue, reported to Kuba.

PR #1618, opened 2025-12-18
ExtractOIIOTranscode: Add filtering by representation name

jakubjezek001:
## Changelog Description

This enhancement adds representation filtering capability to ExtractOIIOTranscode, enabling selective processing of representations based on name patterns. The feature addresses the need for more granular control over which representations undergo color transcoding operations.

The implementation introduces a new `representations` field in profile settings, allowing users to specify regex patterns that determine which representations should be processed. When patterns are defined, only matching representations are transcoded; an empty list processes all representations, maintaining backward compatibility with existing workflows.

Additionally, the system now automatically adds the "review" family to instances when review outputs are generated, ensuring proper downstream processing of transcoded review files. This resolves workflow gaps where review files were created but not properly tagged for review systems.

## Additional info

**Key technical changes:**
- Added `representations` list field to `ExtractOIIOTranscodeProfileModel` in server settings
- Modified `_repre_is_valid()` method to accept profile parameter and evaluate representation name patterns using regex matching
- Implemented auto-tagging of "review" family when review representations are added

**Pattern matching behavior:**
```python
# Example profile configuration
representations: ["beauty", "diffuse.*", ".*_comp"]
# Matches: "beauty", "diffuse", "diffuse_raw", "main_comp"
# Skips: "shadow", "reflection"
```

## Testing notes:
1. Configure an OIIO transcode profile with specific representation patterns (e.g., `["beauty", "render.*"]`)
2. Publish an instance with multiple representations including matching and non-matching names
3. Verify only representations matching the patterns are transcoded
4. Test with empty `representations` list to confirm all representations are processed (backward compatibility)
5. Confirm review family is automatically added when review outputs are generated

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,6 +35,7 @@ body:
       label: Version
       description: What version are you running? Look to AYON Tray
       options:
+        - 1.7.2
         - 1.7.1
         - 1.7.0
         - 1.6.13
--- a/client/ayon_core/plugins/publish/integrate_hero_version.py
+++ b/client/ayon_core/plugins/publish/integrate_hero_version.py
@@ -429,7 +429,7 @@ class IntegrateHeroVersion(
             )
             mode = FileTransaction.MODE_COPY
             if self.use_hardlinks:
-                mode = FileTransaction.MODE_LINK
+                mode = FileTransaction.MODE_HARDLINK
 
             try:
                 for src_path, dst_path in itertools.chain(
--- a/client/ayon_core/version.py
+++ b/client/ayon_core/version.py
@@ -1,3 +1,3 @@
 # -*- coding: utf-8 -*-
 """Package declaring AYON addon 'core' version."""
-__version__ = "1.7.1+dev"
+__version__ = "1.7.2+dev"
--- a/package.py
+++ b/package.py
@@ -1,6 +1,6 @@
 name = "core"
 title = "Core"
-version = "1.7.1+dev"
+version = "1.7.2+dev"
 
 client_dir = "ayon_core"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@
 
 [tool.poetry]
 name = "ayon-core"
-version = "1.7.1+dev"
+version = "1.7.2+dev"
 description = ""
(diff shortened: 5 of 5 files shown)


client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -403,7 +410,23 @@ def _repre_is_valid(self, repre):
...
+        if representations:
+            result = []
+            for r_pattern in representations:
+                if re.match(r_pattern, repre_name):
+                    result.append(True)
+                    break
+            else:
+                result.append(False)
+
+        return any(result)
```
iLLiCiTiT:
```suggestion
        representations = profile["representations"]
        if not representations:
            return True
        repre_name = repre["name"]

        # check if any of representation patterns match in repre_name
        for r_pattern in representations:
                if re.match(r_pattern, repre_name):
                    return True
        return False
```

BigRoy: Isn't that just:
```suggestion
        regexes = profile["representation_name_regexes"]
        if not regexes:
            return True
        pattern = "|".join(regexes)
        return re.match(pattern, repre["name"])
```

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -279,6 +280,12 @@ def process(self, instance):
             if "delete" in tags and "thumbnail" not in tags:
                 instance.data["representations"].remove(repre)
 
+            if (
+                added_review
+                and "review" not in instance.data["families"]
+            ):
+                instance.data["families"].append("review")
```
BigRoy: Should not be part of this PR.

iLLiCiTiT: No, it is from https://github.com/ynput/ayon-core/pull/1617

client/ayon_core/plugins/publish/extract_color_transcode.py
```diff
@@ -403,7 +404,21 @@ def _repre_is_valid(self, repre):
...
+
+        repre_name = repre["name"]
+
+        # check if any of representation patterns match in repre_name
+        for r_pattern in representations_names:
+            if re.match(r_pattern, repre_name):
+                return True
+
+        return False
+
```
iLLiCiTiT:
```suggestion
```

PR #1620, opened 2025-12-18
Extract thumbnail from source: Safe collection of rescale arguments

iLLiCiTiT:
## Changelog Description
Don't hard fail if failed to collect rescale arguments for thumbnail creation.

## Additional info
If input file is not supported by oiio nor ffmpeg the extraction fails, but it should just continue.

## Testing notes:
1. Best is using traypublisher and pass .mov file to a render product type.
2. Publishing should not fail on Extract Thumbnail from source (which with ayon 3rd party addon 1.4.0 would fail on OIIO).


kalisp (approved):
Tested with .mp4.

PR #1621, opened 2025-12-18
Library: provide source version description

kalisp:
## Changelog Description
Added description of source version to destination version to provide additional information, mostly original `author` as that information is not passed over as it might require administrator permissions in case when user who is copying version is different from author.

image

## Additional info
It was decided that separate permission to overwrite author might not be safest option.

## Testing notes:
1. push any version to Library project
2. check its Description


iLLiCiTiT:
Does this mean we can remove `author` in `create_version`?

kalisp:
> Does this mean we can remove `author` in `create_version`?

It is removed here 
https://github.com/ynput/ayon-core/pull/1621/changes#diff-be5409636e4a4f60baef5ddc16adc8bbc8635054bd032fda5451bfa66881cd20L1051

PR #1626, opened 2026-01-05
Do not enforce frame padding minimum to match end frame length.

### LINKED ISSUE #1625: YN-0338: Anatomy formatting with padding of three does not work

We are trying to configure the project to use 3-digit frame padding (###.ext) for Nuke outputs.

  

However, we are facing an issue during publishing. We have updated the padding setting in the Project Anatomy template, and it is correctly reflected in the render output folder. However, when we publish the same file, the frame padding automatically switches back to 4 digits, even though the export is set to 3-digit padding. 

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0338]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=5e5bb390ea1b11f081d80f97afbc5bbb)

BigRoy:
## Changelog Description

Do not enforce frame padding minimum to match end frame length.

A sequence like `998-1010` should write 998, 999, 1000, 1001, ... - not 0998, 0999, 1000, 1001 if the anatomy template sets a frame padding of 3.

## Additional info

Fix #1625

## Testing notes:

1. Set frame padding to 3
2. Write frames 998-1002
3. The published frames 998 and 999 should **not** have a padded zero up front.
4. Set frame padding to 4
5. Write frames 998-1002
6. The published frames 998 and 999 should have a padded zero up front.

---

We're also testing whether DCCs load this correctly and play back how we expect it to:
The test run should test with a sequence that goes to beyond the padding, like with padding of three go from three digits to four digits.

So a sequence like: 
- 999, 1000, 1001

As opposed to:
- 0999., 1000, 1001

Because some applications may assume the frame number in a file sequence keeps a fixed width of padding.
So in this case you'll need a publish with this PR that generated the first 999, 1000, 1001 published frames.
Then load that, and then confirm the DCC correctly displays the exact frames you expect for each.


iLLiCiTiT:
> This is in the Integrator, it's not before? It's directly affecting the template data for formatting?

Sorry I jumped from ExtractOIIO PR to this and mismatched the plugin. Ok, but still should we use length of last frame to actually fill the padding? Nuke won't be able to recognize the sequence `999-1000` either (if I remember correctly).

iLLiCiTiT:
> The first_index_padded value should actually match the value set up with this change? So should be correct?

It should match, but it doesn't based on the code I've seen. It does add padding based on the last frame. So if you have padding `3` and frame range `999-1000` the `"frame"` in context would be `"0999"` which is not what would be used in the path (`"999"`).

> Works fine in Maya for file nodes and image planes, works fine to load and play in Nuke, Fusion and Blender (load to compositor).

That's great start, now Flame, Harmony, Houdini, TVPaint, Houdini, AE, OpenRV, etc. BTW did you use AYON loaders or just dropped them into the DCC?

I'm really scared about this and I'm tempted to say that this should be settings based and not enabled by default.

BigRoy:
> That's great start, now Flame, Harmony, Houdini, TVPaint, Houdini, AE, OpenRV, etc. BTW did you use AYON loaders or just dropped them into the DCC?

AYON loaders of course.

I now tested these - and these play fine:

- [x] Maya Image Plane
- [x] Maya File Node
- [x] Blender Compositor Image
- [x] Fusion Load Image
- [x] Nuke Load Clip
- [x] After Effects Load file
- [x] Premiere Pro Load file

---

These currently fail:

- [ ] Houdini Load Image (Copernicus)
- [ ] Houdini Load Image (COP2)

This is because our load logic loads it with `$F4` in the filenames instead of `$F3`. Once that's swapped it works fine too.
Those values are hardcoded and that's just a trivial bug: https://github.com/ynput/ayon-houdini/blob/a690f83e5945c16c3677c4fe21b0ce3124964242/client/ayon_houdini/plugins/load/load_image_copernicus.py#L146

It would likewise also fail if e.g. padding was 3 or 5 throughout the full sequence. So it's unrelated to this PR.

---

More to test would be:

- [ ] TVPaint
- [ ] OpenRV
- [ ] Harmony
- [ ] Flame

But so far so good.
@LiborBatek perhaps?

LiborBatek:
I guess speaking of testing the rest its more for @kalisp and @MustafaJafar (houdini)  and flame maybe @MilaKudr flame?

MustafaJafar:
Publishing in Houdini works.

---

For reference, `.$F4.` is hard coded in many places in Houdini addon. 
I believe we can skip the 
- creators as they affect only the staging/intermediate files.
- validators as `$F4` is only mentioned in doc strings.

For loaders, we have
- load_ass.py
- load_bgeo.py
- load_image_copernicus.py
- load_image.py
- load_redshift_proxy.py
- load_vdb.py

I tested bgeo where I get the same issue and using `$F3` instead worked. 
So, do we save the frame padding info some where or should we deduce it from the file name?

moonyuet:
Loaded the frame in OpenRV. seems ok
image

PR #1627, opened 2026-01-05
Integrate: don't fail on instance without representations

antirotor:
## Changelog Description
Do not stop publishing on instance without representations. Allow trait-based representations or any custom integrator running on instance even if there are no old-style representations present.

## Additional info
Integrator was raising exception when it encountered instance without representations. Since trait-based representations are not using this instance key and instances can be without representations (you can have any custom integrator processing it in some other way), there is no need to stop publishing.

## Testing notes:
Publishing should work as before. Instances with trait-based representations should continue to the correct integrator.

### DIFF
--- a/client/ayon_core/tools/push_to_project/models/integrate.py
+++ b/client/ayon_core/tools/push_to_project/models/integrate.py
@@ -1172,17 +1172,15 @@ class ProjectPushItemProcess:
         self, anatomy, template_name, formatting_data, file_template
     ):
         processed_repre_items = []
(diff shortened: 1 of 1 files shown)


iLLiCiTiT (approved):

BigRoy:
It'll still log a warning however - which may look annoying if you're correctly using trait based integrations?
Should that become a debug log? or can we somehow detect whether it's the _other_ way so that only if it isn't we log warning?

PR #1639, opened 2026-01-10
Profiling: fix `do_profile` decorator

MustafaJafar:
## Changelog Description
- Fix `do_profile` decorator
- Add support for controlling the profiling based on an environment variable and given tool name. (set env variable to `ALL` to enable the profiling regardless of the tool name)

## Testing notes:
In AYON console, running this should work:
_Note: `@do_profile` won't work without parentheses_
```python
import os
from ayon_core.lib.profiling import do_profile

@do_profile()
def get_bundle_settings(bundle_name):
    import ayon_api 

    return ayon_api.get_addons_settings(bundle_name)


bundle_name = os.getenv("AYON_BUNDLE_NAME")
settings = get_bundle_settings(bundle_name)
```

Also, 

```python
# Set `AYON_ENABLE_PROFILING` to `example`.
import os
from ayon_core.lib.profiling import do_profile

@do_profile(tool_name="example")
def get_bundle_settings(bundle_name):
    import ayon_api 

    return ayon_api.get_addons_settings(bundle_name)


bundle_name = os.getenv("AYON_BUNDLE_NAME")
settings = get_bundle_settings(bundle_name)
```


iLLiCiTiT:
I'm not sure I understand this PR? It is not fixinig anything, or? It adds `tool_name` to the decorator and removes option to use it without calling the decorator.

To be honest I don't think we should mark functions (and commit those markings) into codebase. I would say it is something you as a developer want to do on demand, in that case having tool name doesn't make sense because you do it, and changing what you're interested in using env is not ideal as you have to reset process anyways.

PR #1644, opened 2026-01-12
Core default settings: Update project folder structure to use AYON template keys

MustafaJafar:
## Changelog Description
- Update project folder structure to use AYON template keys

## Testing notes:
Create project structure web action should work as before when setting `ayon+settings://core/project_folder_structure`


BigRoy (approved):
Change looks good to me. @iLLiCiTiT can you confirm?

For what it's worth - note that the `__project_root__` key would've behaved slightly differently, because it'd have applied to EACH `root` key that may be in anatomy. Not sure that was a good thing, but it's a change I just wanted to note here.

PR #1656, opened 2026-01-22
Product name: Create can work with product type

iLLiCiTiT:
## Changelog Description
Methods in create api can work with product type being passed to create plugins and create context.

## Additional info
This PR is adding the option to pass the custom product type to create logic. These changes were extracted from https://github.com/ynput/ayon-core/pull/1647 as are not related to the issue it is solving.

## Testing notes:
At this exact moment nothing should really change. Other addons has to start using product types to be able to validate it, but other addons cannot use product types without these changes. Product type is defaulting to product base type.

### DIFF
--- a/client/ayon_core/pipeline/create/product_name.py
+++ b/client/ayon_core/pipeline/create/product_name.py
@@ -471,6 +471,11 @@ def get_product_name(
             "name": folder_entity["name"],
             "type": folder_entity["folderType"],
         }
+        # Folder label was added ad-hoc and might not be passed from create
+        #   plugins that do fake entities
+        folder_label = folder_entity.get("label")
+        if folder_label:
+            fill_pairs["folder"]["label"] = folder_label
 
     if dynamic_data:
         # Dynamic data may override default values


antirotor (approved):

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -508,16 +512,21 @@ def get_icon(self):
...
-        instance
-    ):
+        project_name: str,
+        folder_entity: Optional[dict[str, Any]],
+        task_entity: Optional[dict[str, Any]],
+        variant: str,
+        host_name: str,
+        instance: Optional[CreatedInstance] = None,
+        project_entity: Optional[dict[str, Any]] = None,
+        product_type: Optional[str] = None,
```
BigRoy: Shouldn't dynamic data technically rely more on product_base_type than being custom data based on the product_type?

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -530,6 +539,7 @@ def get_product_name(
         host_name: Optional[str] = None,
         instance: Optional[CreatedInstance] = None,
         project_entity: Optional[dict[str, Any]] = None,
+        product_type: Optional[str] = None,
```
BigRoy: Technically the product name could also be named by the product base type, right? As such, why doesn't it need that argument?

Is it just that because a single Creator may never deviate from the base type? I wonder if by stating that we're essentially disallowing more dynamic creators to exist or be created? E.g. a creator that may specify product_base_type dynamically based on some data or user choice on e.g. pre_create attributes or instance attributes?

PR #1657, opened 2026-01-22
Representation traits: change how trait classes are matched

antirotor:
## Changelog Description
Use more reliable and lightweight method to find trait classes for deserialized json traits.

## Additional info
This change should solve issues with different modules crashing import logic and so on.

## Testing notes:
Loading of trait-based representations should still work fine.

This should also address some concerns mentioned in comments here https://github.com/ynput/ayon-core/pull/1631

### DIFF
--- a/client/ayon_core/pipeline/traits/representation.py
+++ b/client/ayon_core/pipeline/traits/representation.py
@@ -511,9 +511,9 @@ class Representation(Generic[T]):  # noqa: PLR0904
                 continue
 
             t_ver = _get_version_from_id(t_id)
-            t_base = (
-                re.sub(r"\.v\d+$", "", t_id) if t_ver is not None else t_id
-            )
+            t_base = t_id
+            if t_ver is not None:
+                t_base = re.sub(r"\.v\d+$", "", t_id)
 
             if t_base == base_id:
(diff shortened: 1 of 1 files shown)


client/ayon_core/pipeline/traits/representation.py
```diff
@@ -470,159 +461,99 @@ def __eq__(self, other: object) -> bool:  # noqa: PLR0911
...
 
-        trait_candidates = cls._get_possible_trait_classes_from_modules(
-            trait_id
-        )
-        if not trait_candidates:
-            return None
+        def _get_subclasses(c):
+            return set(c.__subclasses__()).union(
+                [s for c in c.__subclasses__() for s in _get_subclasses(c)]
+            )
```
BigRoy: This really needs a comment or docstring, haha. What!? :)

But... isn't this infinite recursion? `_get_subclasses(c)` calling `_get_subclasses(c)`?

BigRoy: I think this is an invalid copy of code from here: https://stackoverflow.com/a/3862957

Note how there it is `def get_subclasses(cls):`

PR #1667, opened 2026-01-29
Chore: Versioning start uses product base type

iLLiCiTiT:
## Changelog Description
Use product base types in versioning start logic.

## Additional info
PR related to product base type changes. The end goal is to use product base type instead of product type, current implementation expects both to be passed in and `product_type` being deprecated over time. Added settings conversion.

## Testing notes:
1. Versions start works.

### DIFF
--- a/client/ayon_core/pipeline/create/README.md
+++ b/client/ayon_core/pipeline/create/README.md
@@ -6,7 +6,7 @@ Entry point of creation. All data and metadata are handled through create contex
 
 Discovers Creator plugins to be able create new instances and convert existing instances. Creators may have defined attributes that are specific for their instances. Attributes definition can enhance behavior of instance during publishing.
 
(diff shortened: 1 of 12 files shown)


client/ayon_core/pipeline/version_start.py
```diff
@@ -1,9 +1,14 @@
...
 from typing import Optional, Any
 
+from ayon_core.lib import Logger
 from ayon_core.lib.profiles_filtering import filter_profiles
 from ayon_core.settings import get_project_settings
 
+log = Logger.get_logger(__name__)
+
 
 def get_versioning_start(
```
BigRoy: I think this could be wrapped backwards compatible manner where we just remove `product_type` directly, swap it with `product_base_type` (on same place in argument list) and then if `product_type` kwarg was passed just pass...

Sorry, you know what I mean :P - i can elaborate, but no need. Just commenting here as reminder.

PR #1669, opened 2026-01-29
Chore: Create hero version loader plugin cleanup

iLLiCiTiT:
## Changelog Description
Cleanup of Create hero version loader plugin.

## Additional info
I need to do other changes in the plugin but the current code structure makes it really complicated, so I decided to frist do small cleanup before the real changes are applied.

The code was copied from integrate plugin which means that it contained a lot of unnecessary steps and ilogical code. Like get `"hostName"` from context or finding transfers and hard links in instance data.

Avoid of guessing rootless path, instead use the one with which the path is calulated.

## Testing notes:
1. Create hero version load plugin should work.


BigRoy (approved):
Didn't test - LGTM. ~~I wonder if it's something Libor could give a test run~~? Actually let me test it.

PR #1672, opened 2026-01-29
Chore: Staging dir functions expect product base type

iLLiCiTiT:
## Changelog Description
Expect product base types in `get_staging_dir_info` and `get_staging_dir_config` functions.

## Additional info
Removed deprecated `get_custom_staging_dir_info` which is not used anywhere anymore.

The functions changed signature a little bit more to require some arguments to be key-word arguments. The decorators can be removed when product base type changes position in arguments.

The functions seems to be used only in ayon-core and one private addon which is not maintained so it should be save to not bother much.

## Testing notes:
1. Staging dir functionality works as before.

### DIFF
--- a/client/ayon_core/plugins/load/create_hero_version.py
+++ b/client/ayon_core/plugins/load/create_hero_version.py
@@ -4,9 +4,8 @@ import os
 import copy
 import shutil
 import errno
-import itertools
 from concurrent.futures import ThreadPoolExecutor
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 from speedcopy import copyfile
 import clique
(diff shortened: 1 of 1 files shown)


client/ayon_core/pipeline/staging_dir.py
```diff
@@ -78,18 +126,30 @@ def get_staging_dir_config(
...
 
-    if not log:
-        log = Logger.get_logger("get_staging_dir_config")
+    if not logger:
+        logger = Logger.get_logger("get_staging_dir_config")
+
+    if not product_base_type:
+        msg = (
+            "Missing 'product_base_type' in 'get_staging_dir_config'."
+            " Will be soon required."
```
BigRoy:
```suggestion
            " Will soon be required."
```

PR #1674, opened 2026-01-30
Publisher: Use product base type

iLLiCiTiT:
## Changelog Description
Use product base type in publisher.

## Additional info
Publisher tool now uses product base type in variables and items. Also adds `product_base_type` property to `CreatedInstance`.

This PR interfere with https://github.com/ynput/ayon-core/pull/1647, but that's because this has to be merged earlier to make other PRs possible (has smaller effect on publisher logic).

## Testing notes:
1. Publisher still works.

### DIFF
--- a/client/ayon_core/pipeline/publish/publish_plugins.py
+++ b/client/ayon_core/pipeline/publish/publish_plugins.py
@@ -234,7 +234,7 @@ class AYONPyblishPluginMixin:
         if not cls.__instanceEnabled__:
             return False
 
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/publisher/models/create.py
BigRoy: Since Creators can split based on product type in the publish UI - should the `CreatorItem` represent that uniqueness by also having `product_type`, or why is it irrelevant?

PR #1682, opened 2026-02-04
Chore: Disable legacy product name calculation in defaults

iLLiCiTiT:
## Changelog Description
Disable `use_legacy_product_name` by default in settings.

## Additional info
This is controversial change that might break stuff. I don't know if there is anyone ever turning that setting off so this might break stuff and we should test it heavily.

## Testing notes:
1. Well, this affects all 3D DCCs so I guess it should be validated with all of them.


BigRoy (commented):
I think this is a good idea **however** - we need settings conversion.

We should:
- On settings conversion check the source version that is being copied from, if BEFORE the next release (as in prior to this change) and it has no opinion on `use_legacy_product_names_for_renders` it should set it to `True` or something alike? In project settings I'd do that (if we have that level of control.) So that existing projects will remain using the legacy product names even after the update.

Is that possible?

BigRoy (approved):
I think we'll just need to do this.

BigRoy:
I think we'll have to bite the bullet and merge this with a next bigger version release perhaps.

PR #1686, opened 2026-02-05
Workfile builders: Use product base type

iLLiCiTiT:
## Changelog Description
Use product base type in workfile builders.

## Additional info
There are 2 workfile builder logics in pipeline. Both should not support product base type instead of product type.

The products filtering by product base type will be available with https://github.com/ynput/ayon-backend/pull/848 merged with should be since 1.14.0 ayon-backend release. So technically this is preparation for the change, to store product base type in metadata, but it still filters products by product type.

Workfilte template builder is looking for `product_base_types` in loaders to suggest product base types in UI.

Not sure how much is workfile builder used (the one without templates). That added few more options in the settings. Added option to use `"linked_folders"` instead of `"linked_assets"`, `"product_base_types"` instead of `"product_types"` and `"task_names"` instead of `"tasks"`. That is affecting nuke and maya addons:
- `ayon-maya/server/settings/workfile_build_settings.py`
- `ayon-nuke/server/settings/workfile_builder.py`
PRs comming

## Testing notes:
1. Workfile building still works, both template and dynamic.

### DIFF
--- a/client/ayon_core/pipeline/workfile/path_resolving.py
+++ b/client/ayon_core/pipeline/workfile/path_resolving.py
@@ -104,7 +104,7 @@ def get_workfile_template_key(
 
     profile_filter = {
         "task_types": task_type,
-        "hosts": host_name
+        "host_names": host_name,
     }
     profile = filter_profiles(profiles, profile_filter)
     if profile:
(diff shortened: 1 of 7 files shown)


LiborBatek (changes requested):
Build workfile (legacy) gives me an error while triggered

Image

```
# Error: KeyError: file C:\Users\lbate\AppData\Local\Ynput\AYON\addons\core_1.7.2+dev\ayon_core\pipeline\workfile\build_workfile.py line 116: 'id'
```

PR #1701, opened 2026-02-12
Chore: Collect context entities plugin

iLLiCiTiT:
## Changelog Description
Lower order of collect context entities plugin and also collect anatomy in it.

## Additional info
The context should be set as soon as possible, I couldn't find a host where context would be set/changed other way than using `CollectCurrentContext` plugin which has -0.5` order.
Because the plugin is also collecting project entity it can easily fill anatomy too, which means we can get rid of collect anatomy plugin. Refactored the code a little.

## Testing notes:
1. All hosts should be able to publish and use correct "current cotext" data.


BigRoy (approved):
LGTM - but I didn't do a few test runs. Perhaps @LiborBatek can do so?

PR #1702, opened 2026-02-13
Farm: Keep product type from instances

iLLiCiTiT:
## Changelog Description
Keep product type from instance in `create_skeleton_instance`.

## Additional info
When product base type is being used, it is possible that product type is customized. Unfortunatelly the only way how to find out if product type is customized is that we do compare it to product base type.

This uncovered huge flaw in current situation with product types we have defined un host. Maya is using `"renderlayer"` as product base type, 3ds max is using `"maxrender"` and who knowh what houdini is using. The point being that the create plugins are using product base type "A" but when `create_skeleton_instance` is used, it is converted to `"render"`.

To fully resolve that issue we have to create alternative versions of each function, and alternative versions of farm plugins that operate on different level.

## Testing notes:
This change affects all hosts using `create_skeleton_instance`. Test farm publishin for all hosts using it.
- [ ] 3DsMax
- [x] After Effects
- [ ] Blender
- [ ] Cinema 4D
- [ ] Fusion
- [x] Houdini
- [ ] Maya
- [x] Nuke


jakubjezek001 (approved):
works in Nuke and on Farm ;)

PR #1705, opened 2026-02-16
Refactor profiles filtering: add rank_profile, get_matching_profiles, get_highest_score_profiles

vincentullmann:
## Changelog Description

Refactored profile filtering and added new methods so callers can match multiple profiles instead of only picking the single best one.

### Unchanged

- `filter_profiles` expected behaviour is unchanged and backwards compatible. (hopefully)

### New
- **`rank_profile`**:  ranks a single profile using a lexicographic binary score.
  - Each key contributes one bit (`1` = exact match, `0` = wildcard); results form a bitmask so earlier keys matter more.
  - e.g. `0b110 = 6` beats `0b101 = 5` (both have 2 value matches and 1 wildcard).
  - Total number of exact matches = number of `1` bits (e.g. `score.bit_count()`).
- **`rank_profiles`**: helper to rank multiple profiles
- **`get_matching_profiles`**: returns all profiles with a non‑negative score.
- **`get_highest_score_profiles`**: Returns all profiles with the highest score.
- Unit tests added for the new helpers and edge cases.

## Use case

Useful for addons that want to compose profiles dynamically (eg. merge multiple rules into one).

Example (eg. dynamic profiles for deadline):

```py
profiles = {
    # default priority
    ["host": ["*"], "priority": 50],

    # increase chunk size to run the entire playblast in one chunk
    ["host": ["maya"],  "productBaseTypes": ["playblast"], "chunk_size": 999, "overrides": ["chunk_size"]],

    # allow FX tasks to choose the machine pool + allow them to overwrite the default pool
    ["host": ["maya"],  "task": ["fx"], "pool": "high_memory", "overrides": ["pool"]],
}


def get_profile_for_instance(self, instance):
    values = {
        "host": "maya",
        "productBaseTypes": "playblast",
        "task": "fx",
    }
    # custom filter to only include profiles with at least two exact matches
    profiles = [profile for profile, score in rank_profiles(self.profiles, values) if score.bit_count() >= 2]

    # addon specific merge of profiles
    profile = self.combine_profiles(profiles)
    # {
    #     "priority": 50,
    #     "chunk_size": 999,
    #     "pool": "high_memory",
    #     "overrides": ["chunk_size", "pool"]
    # }
```

### DIFF
--- a/client/ayon_core/lib/profiles_filtering.py
+++ b/client/ayon_core/lib/profiles_filtering.py
@@ -1,5 +1,8 @@
+from __future__ import annotations
+
 import re
 import logging
+import typing
 
 log = logging.getLogger(__name__)
 
@@ -23,60 +26,30 @@ def compile_list_of_regexes(in_list):
     return regexes
 
 
-def _profile_exclusion(matching_profiles, logger):
-    """Find out most matching profile byt host, task and family match.
-
-    Profiles are selectively filtered. Each item in passed argument must
-    contain tuple of (profile, profile's score) where score is list of
-    booleans. Each boolean represents existence of filter for specific key.
-    Profiles are looped in sequence. In each sequence are profiles split into
-    true_list and false_list. For next sequence loop are used profiles in
-    true_list if there are any profiles else false_list is used.
-
-    Filtering ends when only one profile left in true_list. Or when all
-    existence booleans loops passed, in that case first profile from remainded
-    profiles is returned.
-
-    Args:
-        matching_profiles (list): Profiles with same scores. Each item is tuple
-            with (profile, profile values)
-
-    Returns:
-        dict: Most matching profile.
-    """
-    if not matching_profiles:
-        return None
-
-    if len(matching_profiles) == 1:
-        return matching_profiles[0][0]
-
-    scores_len = len(matching_profiles[0][1])
-    for idx in range(scores_len):
-        profiles_true = []
-        profiles_false = []
-        for profile, score in matching_profiles:
-            if score[idx]:
-                profiles_true.append((profile, score))
-            else:
-                profiles_false.append((profile, score))
-
-        if profiles_true:
-            matching_profiles = profiles_true
-        else:
-            matching_profiles = profiles_false
-
-        if len(matching_profiles) == 1:
-            return matching_profiles[0][0]
-
-    return matching_profiles[0][0]
-
-
 def fullmatch(regex, string, flags=0):
     """Emulate python-3.4 re.fullmatch()."""
     matched = re.match(regex, string, flags=flags)
     if matched and matched.span()[1] == len(string):
         return matched
     return None
+
+
+def sort_dict(data: dict, keys_order: typing.Iterable[str]):
+    """Sort dictionary by a given key order.
+
+    Keys not included in `keys_order` are added to the end of the dictionary.
+
+    Args:
+        data (dict): Dictionary to sort.
+        keys_order (Iterable[str]): Desired order of keys.
+
+    Returns:
+        dict: Sorted dictionary.
+
+    """
+    ordered = {k: data[k] for k in keys_order if k in data}
+    rest = {k: data[k] for k in data if k not in ordered}
+    return ordered | rest
 
 
 def validate_value_by_regexes(value, in_list):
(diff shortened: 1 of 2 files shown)


client/ayon_core/lib/profiles_filtering.py
```diff
@@ -23,60 +26,30 @@ def compile_list_of_regexes(in_list):
     return regexes
 
 
-def _profile_exclusion(matching_profiles, logger):
```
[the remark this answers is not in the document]
iLLiCiTiT: ~~No it does not, it does use the first matching profile in `filter_profiles`. The order of the keys in profile is ignored and first profile is used.~~ Poor testing

client/ayon_core/lib/profiles_filtering.py
```diff
@@ -116,23 +89,145 @@ def validate_value_by_regexes(value, in_list):
...
+
+    The score is calculated by the following rules:
+    - If any key does not match: return -1.
+
+    The score is a binary number where each bit represents a match or
+    no match for a key.
+
+    Examples:
+        failed match:
+        >>> rank_profile({"a": ["A"], "b": ["B"]}, {"a": "A", "b": "D"}),
```
iLLiCiTiT, line 107:
```suggestion
        >>> rank_profile({"a": ["A"], "b": ["B"]}, {"a": "A", "b": "D"})
```

client/ayon_core/lib/profiles_filtering.py
```diff
@@ -116,23 +89,145 @@ def validate_value_by_regexes(value, in_list):
...
+        >>> rank_profile({"a": ["A"], "b": ["B"]}, {"a": "A", "b": "B"}),
+        3  # => binary: 11
+
+        value match on first key, wildcard match on second key:
+        >>> rank_profile({"a": ["A"], "b": ["*"]}, {"a": "a", "b": "B"}),
+        2  # => binary: 10
+
+        implicit wildcard match on "a", value match on "b":
+        >>> rank_profile({"b": ["B"]}, {"a": "a", "b": "B"}),
+        1  # => binary: 01
```
iLLiCiTiT, line 120: My IDE is screaming that the output is not correct because of the commas as comments
```suggestion
        >>> rank_profile({"a": ["A"], "b": ["B"]}, {"a": "A", "b": "B"})
        3

        value match on first key, wildcard match on second key:
        >>> rank_profile({"a": ["A"], "b": ["*"]}, {"a": "a", "b": "B"})
        2

        implicit wildcard match on "a", value match on "b":
        >>> rank_profile({"b": ["B"]}, {"a": "a", "b": "B"})
        1
```
BTW `rank_profile({"a": ["A"], "b": ["*"]}, {"a": "a", "b": "B"})` does not return `2`.

client/ayon_core/lib/profiles_filtering.py
```diff
@@ -116,23 +89,145 @@ def validate_value_by_regexes(value, in_list):
...
+        score <<= 1     # shift score left by 1 bit
+        score |= match  # set the current match bit in the score
+
+    return score
+
+
+def rank_profiles(
+    profiles: list[dict[str, typing.Any]],
+    key_values: dict[str, typing.Any],
+    logger: logging.Logger,
```
iLLiCiTiT, line 151:
```suggestion
    logger: logging.Logger | None = None,
```

client/ayon_core/lib/profiles_filtering.py
```diff
@@ -116,23 +89,145 @@ def validate_value_by_regexes(value, in_list):
...
+    return [
+        (profile, rank_profile(profile, key_values, logger))
+        for profile in profiles
+    ]
+
+
+def get_matching_profiles(
+    profiles: list[dict[str, typing.Any]],
+    key_values: dict[str, typing.Any],
+    logger: logging.Logger,
```
iLLiCiTiT, line 166:
```suggestion
    logger: logging.Logger | None = None,
```

client/ayon_core/lib/profiles_filtering.py
```diff
@@ -116,23 +89,145 @@ def validate_value_by_regexes(value, in_list):
...
+        profile
+        for profile, score in rank_profiles(profiles, key_values, logger)
+        if score >= 0
+    ]
+
+
+def get_highest_score_profiles(
+    profiles: list[dict[str, typing.Any]],
+    key_values: dict[str, typing.Any],
+    logger: logging.Logger,
```
iLLiCiTiT, line 194:
```suggestion
    logger: logging.Logger | None = None,
```

client/ayon_core/lib/profiles_filtering.py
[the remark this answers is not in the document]
iLLiCiTiT: Not poor testing, it does not work
```python
profiles = [
    {
        "product_base_type": "model",
        "task": "modeling",
    },
    {
        "product_base_type": "model",
        "host": "maya",
    },
    # Profile that has 3 matching keys
    {
        "host": "maya",
        "task": "modeling",
        "test": "1",
    },
]
key_values = {
    "product_base_type": "model",
    "host": "maya",
    "task": "modeling",
    "test": "1",
}
matched = get_highest_score_profiles(profiles, key_values, log)
print(matched)
```
This should return the third profile, but it does return the first. By using byte operations you ignore the count of matches.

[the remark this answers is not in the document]
vincentullmann: Ah, yeah, that was indeed a bug.
I updated the code to take into account the number of matches

I also added this example as a test case + explained in the comments how each profile is expected to get ranked / filtered https://github.com/ynput/ayon-core/pull/1705/changes#diff-39cc39d083adab56dcc7c74e9411d702183e048e4deb8c6dd7b73724abf771e9R223

iLLiCiTiT:
I would like to point to this https://github.com/ynput/ayon-core/pull/1705#discussion_r2947927538 . I still don't understand the need to split the logic. From api point of view it does not make sence to split the logic it into separate functions.

PR #1712, opened 2026-02-20
Chore: Update qtmaterialsymbols

iLLiCiTiT:
## Changelog Description
Updated qtmaterialsymbols.

## Additional info
New version of qtmaterialsymbols does support font variable axes (fill, grade, weight and optical size). The axes are supported since PySide 6.9.x so they can't be used with current AYON launcher versions, but they might be used in DCCs.

## Testing notes:
1. Material symbol icons still do work (icons of folders, tasks and products).

### DIFF
--- a/server/settings/publish_plugins.py
+++ b/server/settings/publish_plugins.py
@@ -1595,7 +1595,11 @@ DEFAULT_PUBLISH_VALUES = {
                                 "-c:a aac",
                                 "-b:a 192k",
                                 "-g 1",
-                                "-movflags faststart"
+                                "-movflags faststart",
+                                "-color_range tv",
+                                "-colorspace bt709",
+                                "-color_primaries bt709",
+                                "-color_trc bt709",
                             ]
                         },
                         "filter": {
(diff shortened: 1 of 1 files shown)


client/ayon_core/vendor/python/qtmaterialsymbols/utils.py
```diff
@@ -41,6 +40,40 @@ def _load_font():
...
+
+
+def _load_fonts():
+    if QtWidgets.QApplication.instance() is None:
+        raise ApplicationNotRunning("No QApplication instance found.")
+
+    _load_font()
+    _load_font_filled()
+    print(_Cache.font_name)
+    print(_Cache.font_name_filled)
```
BigRoy: Apparently this is called _very_ often and these are printed MANY times.
The caches are in the calls above and hence this is always printed.

```suggestion
```

BigRoy (approved):
✅ Works fine in Maya 2026, Fusion 20, Blender 4.5 and Launcher standalone. However,
I did get this logged now:
```
Material Symbols Outlined
Material Symbols Outlined Filled
Material Symbols Outlined
Material Symbols Outlined Filled
Material Symbols Outlined
Material Symbols Outlined Filled
Material Symbols Outlined
Material Symbols Outlined Filled
```
And opening the loader in Maya I get that logged x10 (tons of prints of that) which I think at least we should fix. Other than that, things seem fine 🤔 (e.g. moving the mouse over the loader UI just seems to spam these messages.)

PR #1714, opened 2026-02-22
File transfers: Allow disabling use of speedcopy

BigRoy:
## Changelog Description

Allow all file copies to go through regular `shutil.copyfile` as opposed to `speedcopy.copyfile` via env var `AYON_COPY_FILE_DISABLE_SPEEDCOPY=1` to support arbitrary infrastructure that may not work nicely with the speedcopy - e.g. certain shares may not allow the required fstat calls speedcopy performs, resulting in "[Errno 121] Remote I/O error path" errors.

## Additional info

...

## Testing notes:

1. Check code changes.
2. Confirm setting `AYON_COPY_FILE_DISABLE_SPEEDCOPY=1` does disable use of speedcopy.


client/ayon_core/lib/file_transaction.py
```diff
@@ -2,12 +2,30 @@
...
+    calls that it does may fail. For those cases, environment variable
+    `AYON_COPY_FILE_DISABLE_SPEEDCOPY=1` could be set to avoid using speedcopy.
+
+    Args:
+        src (str): Source path.
+        dst (str): Destination path.
+    """
+    if os.getenv("AYON_COPY_FILE_DISABLE_SPEEDCOPY") != "1":
+        speedcopy.copyfile(src, dst)
+    else:
```
iLLiCiTiT:
```suggestion
    else:
        import speedcopy
```

PR #1723, opened 2026-02-24
Settings: Make project environments scoped to project settings

BigRoy:
## Changelog Description

Settings: Make project environments scoped to project settings

## Additional info

No idea whether this is actually a sensible change. It seems that this setting is focused solely on being project-specific settings, hence I wonder why is it listed in studio settings.

May need some good confirmation it doesn't break anything.

## Testing notes:

1. Project environments setting should work `ayon+settings://core/project_environments?project=projectname`

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -330,7 +330,8 @@ class CoreSettings(BaseSettingsModel):
         "{}",
         widget="textarea",
         title="Project environments",
-        section="---"
+        section="---",
+        scope=["project"],
     )
     filter_env_profiles: list[FilterEnvsProfileModel] = SettingsField(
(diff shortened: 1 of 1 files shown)


BigRoy:
> What is the reason? Why not allow to have "studio default"?

I was just confused that we had at the top: "Global Environments" and then at the bottom "Project Environments" which seemed to be project specific environments, which made me wonder why that latter one was exposed in Studio Settings. The fact that project settings may have studio defaults, sure - fine with me. I was mostly just confused by this distinction. Plus I had never seen that "Project Environments" to begin with, so wasn't sure of the intent - so this PR was mostly to ask about it.

PR #1724, opened 2026-02-25
Publish: Product entity updates

iLLiCiTiT:
## Changelog Description
Avoid update of product base type during publishing.

## Additional info
This is helper to avoid blocking artist from publishing, we can discuss if we'll keep all of it. This PR does not crash if product base type is going to be updated on product entity (which is not allowed). If product base type was not filled on entity (server version is lower than 1.14.0) then nothing happens, if is filled then warning is logged out.

## Testing notes:
1. Update of product entity should not cause crash.


BigRoy:
Is this only relevant for server version <=1.14.0. If so, I think we're reaching a point whre wee may as well ignore this now?

iLLiCiTiT:
> Is this only relevant for server version <=1.14.0. If so, I think we're reaching a point whre wee may as well ignore this now?

There are few enhancement related to `"data"` updates. The only validation is change of product base type, which should not happen.

PR #1736, opened 2026-03-05
Experimental: switch logging to loguru

antirotor:
## Changelog Description
[Loguru](https://github.com/Delgan/loguru) is a library which aims to bring enjoyable logging in Python. AYON was already customizing logging to handle some colorization and contextualization but using loguru will allow to do it with more consistency and in standard way. It also simplify connecting logging facility to some telemetry/log aggregation system.

### TODOs:
- fine-tune highlighting
- implement changes in logger from ynput/ayon-review-desktop
- add api to expose sink to addons

> [!NOTE]
> For more information see [PR in ayon-launcher](https://github.com/ynput/ayon-launcher/pull/291)

### DIFF
--- a/client/ayon_core/cli.py
+++ b/client/ayon_core/cli.py
@@ -24,6 +24,8 @@ from ayon_core.lib.env_tools import (
 )
 
 
+log = Logger.get_logger("CLI")
+
 @click.group(invoke_without_command=True)
 @click.pass_context
 @click.option(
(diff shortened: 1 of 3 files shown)


BigRoy (commented):
Aside of the API changes - I'm not convinced directly that we need `loguru` so this may need some more discussion.

What would you say for you is the selling point to use it?

I saw this on their repo:

> Although logging impact on performances is in most cases negligible, a zero-cost logger would allow to use it anywhere without much concern. In an upcoming release, Loguru's critical functions will be implemented in C for maximum speed.

I wonder if that means we may at some point start having issues with cross-OS binaries and having this in dependency package out of the box, or?

BigRoy:
`structlog` is the more viable candidate here. Closing this.

PR #1738, opened 2026-03-05
added Custom variables at the OCIO config profiles.

robertokker:
So you can use anatomy templates to populate variables that get converted to a string and used in the pre_ocio_hook so the ocio.config can pick it up and you can easily add show/shot luts

Priority also works, and if none of the variables are found a empty string is passed along so the ocio.config can revert to the fallback

image

## Testing notes:
1. set profile type to "path to OCIO config"
2. in your ocio config set your variable in the enviroment
```
environment:
  {
    AYON_OCIO_LUT: "fallback/grade.cube",
  }
```
3. set your search_path based on your enviroment
```
search_path: "${AYON_OCIO_LUT}"
```
4. add it to your look etc 

now you can use your .cube .cdl etc etc like you want in your OCIO config with dynamically loading the luts per shot, asset etc etc.
But there vars could also purpose other options.
Like a different var name could be used for an export lut etc etc

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -35,6 +35,7 @@ body:
       label: Version
       description: What version are you running? Look to AYON Tray
       options:
+        - 1.8.5
         - 1.8.4
         - 1.8.3
         - 1.8.2
--- a/client/ayon_core/pipeline/colorspace.py
+++ b/client/ayon_core/pipeline/colorspace.py
@@ -1738,6 +1738,11 @@ def get_representation_ocio_config_path(
     if not config_template:
         return None
 
+    # Fill template with environment variables
+    # - use StringTemplate to ignore unfilled values so anatomy can try
+    config_template = str(StringTemplate.format_template(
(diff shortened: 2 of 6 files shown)


robertokker:
I do agree on one part, and have thought about this.
And it's not that common to have an OCIO config file per Sequence/shot etc etc. (although it sometimes does happen)
But if implemented in the OCIO you can have different search path's per OCIO config (which is otherwise difficult or going to be cumbersome)

But we could offcourse move this to the tools (or any other) section so you can use it more "widespread" throughout ayon.
And (and it could be me) but i couldn't find a way to create custom variables, and one important thing is that they need to have a priority, otherwise it's still difficult to automatically switch to an episode, sequence, shot or asset lut .
So if a shot lut isnt found it should swiych to the next etc etc like is al;so working on the OCIO config profiles themselfes.

Attached is the OCIO config 
[cg-config-v1.0.0_aces-v1.3_ocio-v2.1 - PLX.yaml](https://github.com/user-attachments/files/25851686/cg-config-v1.0.0_aces-v1.3_ocio-v2.1.-.PLX.yaml)

PR #1743, opened 2026-03-10
Oiio input, process, output arguments sepperation

robertokker:
## Changelog Description
Additional OIIO arguments that can be inserted as a custom global, input, process or output argument.

As OIIO has a strictly left to right order of operations it's important to be able to set where the extra arguments can be inserted.
The previous arguments could only be inserted after the input path. (at the processing stage)
But as i experienced some issue's with converting png's that stayed linear i needed to add "-d uint8 --attrib oiio:ColorSpace sRGB" just before the -o.
But even then, now you have full control over the argument and can finetune the commend better.

This is now the new order of the OIIO operations.

```python
args = [
    "oiiotool",
    "custom_global_arg"  # NEW
    *input_arguments,
    "custom_input_arg"  # NEW
     input_path,
    "custom_process_arg"  # replaces the old arguments
    *output_arguments,
    "custom_output_arg"  # NEW
     "-o",
    output_path
]
```

## Additional info
image

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_core/hooks/pre_ocio_hook.py
+++ b/client/ayon_core/hooks/pre_ocio_hook.py
@@ -75,3 +75,17 @@ class OCIOEnvHook(PreLaunchHook):
             f"Setting OCIO environment to config path: {ocio_path}")
 
         self.launch_context.env["OCIO"] = ocio_path
+
+        custom_vars = config_data.get("custom_vars", {})
+        for var_name, var_value in custom_vars.items():
+            if var_value:
+                # Normalize path to fix mixed slashes like `W:\/AY1_Ayon/foo`
+                import os
+                var_value = os.path.normpath(var_value)
+                if self.host_name in ["nuke", "hiero"]:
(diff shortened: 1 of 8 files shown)


iLLiCiTiT:
Hi, we do put a lot of effort to avoid being able to define these arguments via settings. Usually these arguments are context based -> you have to know what the input is, which is not something general plugin should handle. And they might be doing the oposite of the logic of the plugin. (We burned with Extract Review settings.)

It is also very hard to explain to people that they have to split the arguments correctly.

I'm not responsible for this part of code so I won't be fighting for my opinion on this, but I do thing that this is not good idea. I believe there might be better ways how to solve the issue.

robertokker:
At first i used this to overcome an issue that was even resolved in alter stage.
And i do agree it can make it overly-complicated.

So i'll close this request (forgot it was still open )

iLLiCiTiT:
Great to hear the issue was resolved.

PR #1746, opened 2026-03-10
Chore: Use published template name based on settings

Liametc:
## Changelog Description
Use published template based on instance data instead of `"default"`.

## Additional info
This fixes an issue where "default" was hardcoded as the template name in `ayon_core.pipeline.farm.tools.from_published_scene`.
Added the function `get_template_name_from_instance` to consolidate a few instances in the code base that had duplicated functionality and use in the function mentioned above.

## Testing notes:
To test the `from_published_scene` fix
1. Create a template that is not called "default"
2. Set it as a template on some usd publish
3. Publish something belonging to that type

### DIFF
--- a/client/ayon_core/pipeline/publish/lib.py
+++ b/client/ayon_core/pipeline/publish/lib.py
@@ -31,7 +31,7 @@ from ayon_core.lib import (
 )
 from ayon_core.settings import get_project_settings
 from ayon_core.addon import AddonsManager
-from ayon_core.pipeline import get_staging_dir_info, CreatedInstance
+from ayon_core.pipeline import get_staging_dir_info
 from ayon_core.pipeline.plugin_discover import DiscoverResult
 from ayon_core.lib.file_transaction import copyfile
 from .constants import (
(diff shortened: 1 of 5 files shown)


client/ayon_core/pipeline/publish/lib.py
```diff
@@ -228,6 +228,53 @@ def get_publish_template_name(
     return template or default_template
 
 
+def get_template_name_from_instance(
```
iLLiCiTiT:
```suggestion
def get_template_name_for_instance(
```

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -228,6 +228,53 @@ def get_publish_template_name(
...
+    instance: CreatedInstance,
+    mapped_product_base_type: Optional[dict[str, str]] = None,
```
iLLiCiTiT: I would say this is reason why we can't have the generic function. It is bigger issue that is being on track for some time.

If we consider keeping the function then I'd rather allow to pass `product_base_type` instead of mapping, so the mapping happens outside of this function.

PR #1766, opened 2026-03-27
Representation traits: hero version integrator

### LINKED ISSUE #1685: YN-0470: Integrating hero version not using all representations

 

Adding the files exported from Marvelous Designer but the issue is that when publish model creator it writes out 5 Files (XML, OBJ, FBX, MTL, ABC).Added the publish representations which was missing and when it runs the `IntegrateHeroVersion` it runs those 5 files 1 by 1 so the xml goes into the hero folder then when obj needs to be moved it clears everything in the hero folder removing the xml file before moving over the obj.

  
> Would it be possible to override by file extension to only replace specific files rather than clearing the entire directory or another way to performing the hero portion of publish.

  

This looks like bug processing the individual representation, possibly related to trait-based representation not processed correctly by Hero Version Integrator.

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0470]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=3f9a1dd0027511f1bcc54f6b8bb44fb0)

antirotor:
## Changelog Description
Adding hero version integrator and re-organizing code for publishing traits. Added possibility to integrate file paths sharing common root but living under different leaf folders where the anatomy template is used just for the common part. Added more unit tests for publishing traits.

## Additional info
This adds hero version integrator for trait-based representations. It is a little bit simpler, basically replacing the whole hero version with the new one, It tries to follow the logic of the "standard" integrators but during writing this I realized, we should really simplify the logic, get rid of all the exceptions and make it more api (and therefor testing) friendly.

One additional change adds possibility to integrate representation made up by paths pointing to files in different directories but sharing the same root. For example:

```
Representation "foo":
    - file path 1: /foo/bar/baz/file.1
    - file path 2: /foo/bar/goo/file.2
```
in this case the template would be applied to `/foo/bar` and the `baz/file.1` and `goo/file.2` will be copied there as they are. Note that this won't change the frame padding, or do re-numbering of the files when needed - so in that sense it is not functionally on par with how the representations are integrated right now, that functionality has to be added later when needed.

## Testing notes:
1. Publish something using trait-based representation with enabled hero version.
2. Check the hero version exists and can be loaded.

*Note: I'll link simple [testing branch for Maya](https://github.com/ynput/ayon-maya/tree/experimental/pointcache-trait-repre) that I used to publish pointcaches as trait-based representations. Currently only Marvelous Designer is using them.*


moonyuet:
Tested in my side, successfully created the hero version
image
image

[publish-report-260331-13-17.json](https://github.com/user-attachments/files/26368486/publish-report-260331-13-17.json)

moonyuet (approved):

BigRoy (commented):
The amount of code here is massive :')
But overall seems ok, commented on some code bits mostly.

BigRoy (approved):
Didn't test but aside of the still pending comments, code changes seem fine.

BigRoy:
Can someone resolve the conflicts.
@iLLiCiTiT can you give it your review please?

client/ayon_core/pipeline/anatomy/anatomy.py
```diff
@@ -17,7 +20,12 @@
 
 from .exceptions import RootCombinationError, ProjectNotSet
 from .roots import AnatomyRoots
-from .templates import AnatomyTemplates
+from .templates import AnatomyTemplates, PLACEHOLDER
```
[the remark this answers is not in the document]
iLLiCiTiT: Then rename it, `PLACEHOLDER` is a "placeholder", if it's being made public it has to make sense, something like `NOT_SET`.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -1413,3 +1417,223 @@ def _get_last_version_files(
...
+
+    # Include instance.data[versionData] directly
+    version_data_instance = instance.data.get("versionData")
+    if version_data_instance:
+        version_data.update(version_data_instance)
+
+    return version_data
+
+
+def get_rootless_path(anatomy: "Anatomy", path: str) -> str:
```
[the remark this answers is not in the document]
iLLiCiTiT: To be honest I don't see any advantage of having this function. I see it being used to get source, I would rather have get_instance_source instead of this.

EDITED:
I see it is also being used for representation entity files, but that one probably should not gracefully use path but force to use rootless path.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -1413,3 +1417,223 @@ def _get_last_version_files(
...
+        families.append(family)
+
+    for _family in (instance.data.get("families") or []):
+        if _family not in families:
+            families.append(_family)
+
+    return families
+
+
+def get_version_data_from_instance(
```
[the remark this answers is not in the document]
iLLiCiTiT: Me being technicall, the `get_` probably should be `prepare_` as it also changes data on the instance.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -1413,3 +1417,223 @@ def _get_last_version_files(
...
+        path = rootless_path
+    else:
+        log.warning((
+            'Could not find root path for remapping "%s".'
+            " This may cause issues on farm."
+        ), path)
+    return path
+
+
+class TemplateItem:
```
iLLiCiTiT: This is very confusing a `TemplateItem` holding `AnatomyTemplateItem`, which in fact is `TemplateItem`... I wonder if you ever tried to find out where classes are being used... This is maintanance hell.

I would move it to traits as it has no other use, and rename it to something like `IntegrateTemplateItem`.

Is there a reason to have both `template` and `template_object`? You could just have the object and get the template from it...

client/ayon_core/pipeline/anatomy/anatomy.py
```diff
@@ -104,7 +112,13 @@ def templates_obj(self):
...
         return self._templates_obj
 
-    def get_template_item(self, *args, **kwargs):
+    def get_template_item(
+            self,
+            category_name: str,
+            template_name: str,
+            subkey: Optional[str] = None,
+            default: Any = PLACEHOLDER,
+    ) -> Union[TemplateItem, str, None]:
```
iLLiCiTiT: How it can return `str`? It either returns `TemplateItem` or `Any` if `default` is passed in.
```suggestion
    ) -> TemplateItem | Any:
```

client/ayon_core/pipeline/publish/__init__.py
```diff
@@ -52,6 +52,15 @@
     get_trait_representations,
     has_trait_representations,
     set_trait_representations,
+
+    get_template_name,
+    get_publish_template,
+    get_publish_template_object,
+    get_instance_families,
+    get_version_data_from_instance,
+    get_rootless_path,
```
iLLiCiTiT, line 61: I would put `instance` to each of the functions (`get_instance_template_name`, `get_instance_publish_template`). We should easily distinct these helper from "the" implementations.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -1413,3 +1417,223 @@ def _get_last_version_files(
...
+    Returns:
+        str: Anatomy template name
+
+    """
+    # Anatomy data is pre-filled by Collectors
+    template_name = get_template_name(instance)
+    anatomy = instance.context.data["anatomy"]
+    publish_template = anatomy.get_template_item("publish", template_name)
+    path_template_obj = publish_template["path"]
+    return path_template_obj.template.replace("\\", "/")
```
iLLiCiTiT:
```suggestion
    publish_template = get_publish_template_object(instance)
    return publish_template["path"].template.replace("\\", "/")
```

client/ayon_core/pipeline/traits/publishing.py
```diff
@@ -0,0 +1,697 @@
...
+        Returns:
+            str: Checksum of the file.
+
+        """
+        return hashlib.sha256(
+            file_path.read_bytes()
+        ).hexdigest()
+
+
+def get_publish_template_object(
```
iLLiCiTiT: Do we have the same function twice now?

client/ayon_core/pipeline/traits/publishing.py
```diff
@@ -0,0 +1,697 @@
...
+        self.destination = destination
+        self.size = size
+        self.checksum = checksum
+        self.template = template
+        self.template_data = template_data
+        self.representation = representation
+        self.related_trait = related_trait
+
+    @staticmethod
+    def get_size(file_path: Path) -> int:
```
iLLiCiTiT: Why are these helper methods on the class? If it's because it is directly related to them then rename it to `get_file_size` and `get_file_checksum` as I would expect that `get_size` would return what `self.size` value is. We should also have a way how to get infromation about which checksum type was used (`sha256`).

client/ayon_core/pipeline/traits/content.py
```diff
@@ -483,3 +566,22 @@ class Fragment(TraitBase):
     id: ClassVar[str] = "ayon.content.Fragment.v1"
     persistent: ClassVar[bool] = True
     parent: str
+
+
+@dataclass
+class OriginalFilename(TraitBase):
```
iLLiCiTiT, line 620: I wonder what this trait is about? When do we set this? What is usage of it? Because we do define this based on template, and it might be partial, we technically do allow to have template like:
- directory: `{originalDirname}/publish`
- file: `{originalFilename}`

client/ayon_core/pipeline/traits/content.py
```diff
@@ -131,6 +132,47 @@ def get_files(self) -> Generator[Path, None, None]:
         for file_location in self.file_paths:
             yield file_location.file_path
 
+    def get_common_root(self) -> Path:
```
iLLiCiTiT, line 164: What is the use-case of this again? How can one representation have multiple files in different directories?

iLLiCiTiT:
Why we don't integrate traits within integrate hero plugin?

client/ayon_core/pipeline/traits/content.py
```diff
@@ -348,6 +391,46 @@ def _get_frame_info_with_handles(
...
+            paths (list[Path]): List of file paths.
+
+        Returns:
+            FrameRanged: FrameRanged trait.
+
+        Raises:
+            ValueError: If paths cannot be assembled into one collection
+
+        """
+        cols, rems = assemble([path.as_posix() for path in paths])
```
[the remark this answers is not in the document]
iLLiCiTiT: Agree, I though that Sequence would solve our current issues of knowing if we have a sequence, which path of filename is the frame and what the frame actually is, none of that is covered at this moment. We still don't support single frame sequences and we have to be carefull about naming as `renderMain_sh0010.1001.exr` can still endup as frame `10` instead of `1001`.

BigRoy:
> Why we don't integrate traits within integrate hero plugin?

@antirotor can you loop back here?

client/ayon_core/pipeline/traits/content.py
antirotor: With traits, it can. Under the common root. If that's the case, integrator will apply template formatting up to the common root. If we ever need to go further, we would need to add some "globing" functionality to templates..,

client/ayon_core/pipeline/traits/content.py
antirotor: The idea is that this trait is set by the integrator if the file names needs to remain in case where there are multiple files in forking paths. Until we support it directly in templates, this is at least flag to know that it happened.

client/ayon_core/pipeline/traits/publishing.py
```diff
@@ -0,0 +1,697 @@
...
+            # If representation has FileLocations trait (list of files)
+            # it can be a Sequence, UDIM tile set, or a group of related
+            # files that share a common root and preserve their hierarchy.
+            # Note: we do not support yet frame sequence of multiple UDIM
+            # tiles in the same representation.
+            get_transfers_from_file_locations(
+                representation, template_item, transfers
+            )
+        elif representation.contains_trait(FileLocation):
+            # This is just a single file representation
```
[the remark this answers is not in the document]
antirotor: This is now validated here https://github.com/ynput/ayon-core/blob/4792b1e18e19bcac7ed1ff16360760bf4ab4739c/client/ayon_core/pipeline/traits/content.py#L107 and here https://github.com/ynput/ayon-core/blob/4792b1e18e19bcac7ed1ff16360760bf4ab4739c/client/ayon_core/pipeline/traits/content.py#L236

but I think it would be better just to allow it. But for that we would have to re-think the integration process which is - by my opinion - a little bit complicated right now.

client/ayon_core/pipeline/traits/content.py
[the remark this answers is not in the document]
antirotor: It now supports single frame sequence. If we nned to support more complex naming, we'll need to pass the pattern supported by traits to these lower-level functions. For another issue once it arises.

client/ayon_core/pipeline/publish/lib.py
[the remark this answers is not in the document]
antirotor: > Me being technicall, the `get_` probably should be `prepare_` as it also changes data on the instance.

maybe I've changed it meanwhile, but I don't think it changes anything on the instance...

client/ayon_core/pipeline/traits/publishing.py
antirotor: not anymore

client/ayon_core/pipeline/traits/publishing.py
```diff
@@ -0,0 +1,697 @@
...
+            representation, instance)
+        # add instance based template data
+
+        template_data.update(instance_template_data)
+
+        # treat Variant as `output` in template data
+        with contextlib.suppress(MissingTraitError):
+            template_data["output"] = (
+                representation.get_trait(Variant).variant
+            )
```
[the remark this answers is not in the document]
antirotor: It is the second one. Variants in representation context are meant to replace `output`, which always felt as variant of a representation. Happy to discuss different terminology, but rather in separate issue.

antirotor:
> Why we don't integrate traits within integrate hero plugin?

they are integrated from the original representation here
https://github.com/ynput/ayon-core/blob/d9fc253969a24ef68ce5fc27bba410d71100fd52/client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py#L254

BigRoy:
@antirotor please request review if anyone needs to review latest state again.

antirotor:
> @antirotor please request review if anyone needs to review latest state again.

hmm, I thought I've done that, but it seems I haven't.

BigRoy (approved):
Didn't test but aside of the still pending comments, code changes seem fine.

Also, quite massive for something that seemingly should be trivial :')

image

antirotor:
> Also, quite massive for something that seemingly should be trivial :')

most of the changes were moves of the code to "library" as it is now used in two integrators instead of one.  I must admit that both integrators are breaking my brain now. I tried to follow the logic of the non-trait representations but I still feel there is something wrong about how complex it is.

BigRoy:
@antirotor is this now in a state that it solely requires a review from @iLLiCiTiT at this point? or should @moonyuet test again as well due all recent changes?

antirotor:
> @antirotor is this now in a state that it solely requires a review from @iLLiCiTiT at this point? or should @moonyuet test again as well due all recent changes?

I think I've addressed what I could, there is a lot of changes so any testing is welcomed. I'd like to have some test harness in place to easily running integration unit tests as what we have isn't enough.

moonyuet (approved):
Tested successfully in MD with Integrate hero version for representation traits.

[publish-report-260613-17-28.json](https://github.com/user-attachments/files/28910234/publish-report-260613-17-28.json)

BigRoy:
@antirotor I'm seeing some [internally reported issue here](https://discord.com/channels/517362899170230292/1531446888409206864/1531573899018506260) and I wonder why this integrator runs over ALL hero versions, and not only those ones that have representation traits. It seems to be killing/destroying those instances that are already integrated by the regular hero version integrator.

Meaning that this may result in just hero versions without any representations now.

antirotor:
It should be fixed by https://github.com/ynput/ayon-core/pull/1978

PR #1770, opened 2026-03-30
Attribute definitions: Button definition

### LINKED ISSUE #1730: YN-0534: Publisher: Add ButtonDef implementation

For some use cases the regular UI customizations that are possible in the publisher UI with attribute definitions just doesn't suffice, sometimes you need to manage more complex data to influence your publishing.

  

Here a dedicated "ButtonDef" would make sense. When clicked the dev can attach any custom callback that would do anything to the selected instance. It could affect other attributes on the instance, like set them to different values but the ButtonDef can also store it's own data dictionary of custom values.

  

The button callback won't be available in Publish Report Viewer, however we should still be able to investigate the "custom data" stored as part of that attribute. As such, I think e.g. right clicking the button or something along those lines would allow you to view the dictionary / JSON data that's stored in that attribute definition on the instance.

  

The ButtonDef can show a custom UI, but also can just be a custom action that it triggers - it does not even have to change the state of the instance itself (it could even be used for other things, it just triggers a Python callback with the create_context and the selected instances as arguments and awaits the response to define how to respond (e.g. with a refresh of attribute values or not) 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1475446042701856919/1477899058332041286)

_This issue was automatically created from AYON ticket [YN-0534]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=e68bc210161111f1bf4a450e7e78dbad)

iLLiCiTiT:
## Changelog Description
Added button to attribute definitions, and modified publisher UI to add the support.

## Additional info
Base implementation of buton definition support. Logic related to create and publish logic is a little bit more complex as there are 3 types of definitions: precreate, create and publish attributes.

Create and publish plugins can register a callback which can be a method that expects one argument where selection of instances is passed in (not for pre-create attributes -> there is no selection).

Right now the logic does not support that button definition of instance A has different callback from instance B. If the key of the button is the same then they are considered as one definition.

There is no api for trigger of the button definitions, not sure if is needed?

## Testing notes:
1. Use `ButtonDef` in a create plugin for pre-create attributes and instance attributes.
2. Registered callbacks should be triggered.
3. Use `ButtonDef` in publish plugin for instance or context.
4. Registered callback should be triggered.

Resolves https://github.com/ynput/ayon-core/issues/1730


moonyuet (approved):
I tested with the material library creator in 3dsmax by implementing the buttonDef as pre-create attribute definition. 
```python
    def get_pre_create_attr_defs(self):
        return [
            ButtonDef(
                key="my_action",
                label="Run Action",
                callback=self._on_my_action_clicked
            ),
        ]

    def _on_my_action_clicked(self):
        # Example action callback
        print("Action button clicked!")
```
It is working, it can trigger the callback function as expected.

https://github.com/user-attachments/assets/53334924-487e-4041-8200-1c570915c469

jakubjezek001 (changes requested):
Creating context works as @moonyuet was showing, but I could not manage to make this work in publishing context. I was trying it at clolectors and extractrors but button did not show.

moonyuet:
> Creating context works as @moonyuet was showing, but I could not manage to make this work in publishing context. I was trying it at clolectors and extractrors but button did not show.

Good catch. I have tested it again and find out the following error when I introduce the button definition in the collector.
```
Triggered button callback for non-button attribute got '<class 'NoneType'>' expected 'ButtonDef'
Triggered button callback for non-button attribute got '<class 'NoneType'>' expected 'ButtonDef'
Triggered button callback for non-button attribute got '<class 'NoneType'>' expected 'ButtonDef'

```
I would test the loader and see if the same issue spot.

moonyuet:
The attribute definition is also working in the loader. So it is only errored out in collector/extractor.
image

BigRoy:
I have a feeling we may need to right-align the button to fit in the right column of the form, or make it full width. It feels a bit odd being the only thing being randomly centered.

Also, when testing:
- It's important that the button can access the current context (and instance; if it's on an instance attribute definition).
- It's important that the callback _can_ change the attribute values as well. (So it can show a dialog, change some attributes as a result)
- The button may need to be able to store some internal data, but probably that's overkill and is better left to having it adjust actual other attributes, hidden attributes or even custom data on the instances.

For the Loader, I'm not sure what the button in the option box would really be able to do. Unless perhaps it can also set the other attributes in the pop-up... or on completion actually close the dialog itself and trigger the load itself. So you can click it to trigger some custom load logic. Can we think of some use cases there?

iLLiCiTiT:
> The button may need to be able to store some internal data, but probably that's overkill and is better left to having it adjust actual other attributes, hidden attributes or even custom data on the instances.

No, won't be possible. Especially not in publishing.

> For the Loader, I'm not sure what the button in the option box would really be able to do. Unless perhaps it can also set the other attributes in the pop-up... or on completion actually close the dialog itself and trigger the load itself. So you can click it to trigger some custom load logic. Can we think of some use cases there?

Question for future. Right now it can't change the existing attributes/form shown to user, or close it. That would require more changes. Not impossible tho.

> It's important that the button can access the current context (and instance; if it's on an instance attribute definition).
> It's important that the callback can change the attribute values as well. (So it can show a dialog, change some attributes as a result)

It can access it and can change things.

```python
# Callback on publish plugin
@classmethod
def _btn_callback(cls, info):
    print(info.create_context)  # Create context
    print(info.instance_ids)  # Selected instances and filtered for the button definition
    # ~~print(info.selected_instance_ids)~~ # Removed
```

iLLiCiTiT:
I might remove the `selected_instance_ids`, not sure if it is usefull?

Another thing is that currently the id of context is `None` so you'd do `if None in info.instance_ids` to check if context is selected, that might need change.

jakubjezek001:
> right clicking the button or something along those lines would allow you to view the dictionary / JSON data that's stored in that attribute definition on the instance.

Also, wasn't this one of requirements?

iLLiCiTiT:
> Also, wasn't this one of requirements?

It was an idea. I do try to make it work, considering that I also have to show it for multiple instances at once. I really have no idea how it should work if I should allow to store data to it. From my point of view not needed and it makes already complicated approach even more complicated (I actually don't know how that would work) for not much benfit of it.

iLLiCiTiT:
> I have a feeling we may need to right-align the button to fit in the right column of the form, or make it full width. It feels a bit odd being the only thing being randomly centered.

Any UX ideas about this? The middle is the less weird from all variants I did try. Full width is the most weird, left feels wrong, right feels wrong too. Maybe fill the right column (value column) at least? But it might end up as full width.

BigRoy:
> Any UX ideas about this? The middle is the less weird from all variants I did try. Full width is the most weird, left feels wrong, right feels wrong too. **Maybe fill the right column (value column) at least?** But it might end up as full width.

Yup, filling the right side of the form (value) may make most sense if full width feels off. But the current middle not aligning with anything also feels off.

> right clicking the button or something along those lines would allow you to view the dictionary / JSON data that's stored in that attribute definition on the instance.

Let's ignore that for now - especially since now the button itself doesn't store any data as far as I understand.

BigRoy:
> Right now the logic does not support that button definition of instance A has different callback from instance B. If the key of the button is the same then they are considered as one definition.

I assume this solely applies to multi-selection?

> There is no api for trigger of the button definitions, not sure if is needed?

Would be nice, but not critical. I'd say however usually that API should come first. The fact that the API can't but the UIs can (without any API) does sound wrong though.

iLLiCiTiT:
> I assume this solely applies to multi-selection?

Yes, but I'd say it is a general issue. I'm not sure how I would handle that TBH. Multiselection in the UI will always be a blocker for this feature.

> Would be nice, but not critical. I'd say however usually that API should come first. The fact that the API can't but the UIs can (without any API) does sound wrong though.

You can, but the api does not filter the instance ids.

> Let's ignore that for now - especially since now the button itself doesn't store any data as far as I understand.

It does not, it is UI definition.

iLLiCiTiT:
So...
- the button is in value column
- added api methods to trigger button callback on `CreateContext`
- the api methods are actually used by publisher
- removed selected instance ids from the info passed to the callback, as it should not have a value for processing logic of the button callback -> and it complicates the api methods

BigRoy (changes requested):
This seems to work, nice!

1. Let's make it so that the `ButtonDef` has two labels, the left side label and the button text.

This will make it so that it matches much more with the other attribute definitions.
I think if no button text is set (default None) then it should likely be "Trigger" or perhaps default to no text and a finger/click icon? 

Can we support button icons? If so then defaulting to an empty text and likely a click icon that would be nice.

2. Tooltip doesn't show on the button

Using `tooltip="Hello world"` does not show the tooltip on hover for me.

3. The button text is ellided, but should me made to fit?

Image

Minimum button width would likely need to the fit the text?

Here's an example of what I did:
```python
    @classmethod
    def get_attr_defs_for_instance(cls, create_context, instance):
        return [ButtonDef("button_def", label="Fix this thing!", tooltip="Hello world", callback=cls._on_callback)]

    @classmethod
    def _on_callback(cls, callback_info):
        for instance in callback_info.create_context.instances:
            if instance.id in callback_info.instance_ids:
                instance.publish_attributes["ValidateNodeIDsRelated"]["active"] = False
```

iLLiCiTiT:
> This will make it so that it matches much more with the other attribute definitions.

Does this actually make sense?

> Minimum button width would likely need to the fit the text?

It can't/should not resize publisher window. I think that was the reason I did put it to the middle.

> Can we support button icons?

Probably yes.

BigRoy (approved):
This is awesome! 🥇 

Image

Even opens door to build some custom instance configuration preset button.

Image

PR #1779, opened 2026-04-07
Chore: Icon definitions

iLLiCiTiT:
## Changelog Description
Add objected icon definitions.

## Additional info
Added strict definitions of icons for actions and UI purposes. Hopefully should help to unify how icons are defined all over the place. From dict -> defined object.

The objects are used in ayon-core's code. It is NOT used when the icon is not defined in ayon-core (e.g. load/create plugins).

To be able to start using these new object we first have to make new release of ayon-core and bump compatible version in other addons.

## Testing notes:
1. Validate code changes.
2. All tools should be working as before.


kalisp (approved):
Clicked around, tested publish in Blender, it seems that I see all icons as before.

BigRoy (approved):
Seems to work.

PR #1780, opened 2026-04-07
Users model: Use 'get_users' from ayon api

iLLiCiTiT:
## Changelog Description
Users model in tools use ayon api implementation from ayon api.

## Additional info
All the missing parts are already avaialable in ayon api so the custom function is not needed anymore. Also fixes an issue if `allAttrib` are used instead of `attrib`.

## Testing notes:
1. Workfiles tool works with https://github.com/ynput/ayon-launcher/pull/299 .


BigRoy (approved):
Works for me.

PR #1783, opened 2026-04-13
Implement a Recent Actions list in Launcher

### LINKED ISSUE #1782: YN-0617: Launcher: Show "Recent" / "Favorites" for recent launched apps/contexts

Add a Recent and a Favorites (Recent would be sufficient as a start) option added to the Launcher and potentially other menus.

  

This is most useful for leads/supervisors and TDs to jump between recently access tasks, filtering to my tasks isn't always the easiest. Just a simple dropdown that then populates the main menu is sufficient. Might want to be able to launch a previously used application from it as well for super quick usage.

  

It'd be either the favorites/recents are per context or per action launch to Favorites (or recents). So either:

1. /film/sh010/animation
1. /asset/char_hero/modeling

or:

1. /film/sh010/animation -> maya 2026
1. /asset/char_hero/modeling -> houdini 21

In the latter case I suppose you'd be clicking "favorite" on the actual application action?

In the latter scenario I guess the listed action could then still have a main area to click to launch the app but with e.g. an Error button to the side that when clicking there it just brings you to that context.

  

Currently assumption is we'll want to store these favorites and recents per user instead of per machine (local storage). 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0617]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=e228f05028ed11f19c28518cb9699a0b)

BigRoy:
## Changelog Description

The Launcher UI now keeps track of history of recent actions that were performed so you can quickly go back to a previous context or re-run an action without losing your current context.

## Additional info

image

Older designs/states:

https://github.com/user-attachments/assets/aa237b65-3027-4efe-8b24-f00d3a4f4cd8

Slightly tweaked design after video:
image

Fix part of https://github.com/ynput/ayon-core/issues/1782
(Another part is allowing to favorite entries)

Note that history is stored per user, but local to the machine. We may want to store that per user in the actual database so it's shared across machines perhaps but is still per user?

## Testing notes:

1. Launcher should work
2. Recent action history should work


client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,205 @@
...
+            action_type="webaction",
+            identifier=event["identifier"],
+            label=event.get("full_label") or event["identifier"],
+            icon=None,
+            addon_name=event.get("addon_name"),
+            addon_version=event.get("addon_version"),
+            project_name=event.get("project_name"),
+            folder_id=event.get("folder_id"),
+            task_id=event.get("task_id"),
+            task_name=None,
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/tools/launcher/ui/recent_actions_widget.py
```diff
@@ -0,0 +1,507 @@
...
+
+_TRANSPARENT_ICON_DEF = {"type": "transparent", "size": 256}
+_PLAY_ICON_DEF = {
+    "type": "material-symbols",
+    "name": "play_arrow",
+}
+
+
+def _get_play_icon():
+    return get_qt_icon(_PLAY_ICON_DEF)
```
iLLiCiTiT:
```suggestion
    return get_qt_icon({
        "type": "material-symbols",
        "name": "play_arrow",
    })
```

client/ayon_core/tools/launcher/ui/recent_actions_widget.py
```diff
@@ -0,0 +1,507 @@
...
+from ayon_core.tools.utils.delegates import pretty_timestamp
+
+RECORD_ID_ROLE = QtCore.Qt.UserRole + 1
+_QWIDGETSIZE_MAX = (1 << 24) - 1
+
+_TRANSPARENT_ICON_DEF = {"type": "transparent", "size": 256}
+_PLAY_ICON_DEF = {
+    "type": "material-symbols",
+    "name": "play_arrow",
+}
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/tools/launcher/ui/recent_actions_widget.py
```diff
@@ -0,0 +1,507 @@
...
+        scroll_area.setFrameShape(QtWidgets.QFrame.NoFrame)
+
+        rows_container = QtWidgets.QWidget(scroll_area)
+        rows_container.setObjectName("RecentActionsContainer")
+        rows_layout = QtWidgets.QVBoxLayout(rows_container)
+        rows_layout.setContentsMargins(4, 4, 4, 4)
+        rows_layout.setSpacing(2)
+
+        empty_label = QtWidgets.QLabel(
+            "No recent actions yet.", rows_container
```
iLLiCiTiT:
```suggestion
            "No recent actions...", rows_container
```

client/ayon_core/tools/launcher/abstract.py
```diff
@@ -67,6 +69,42 @@ class WorkfileItem:
     version: Optional[int]
 
 
+@dataclass
+class RecentActionItem:
```
iLLiCiTiT: As mentioned above, I believe this class is related to recent actions model for storing and loging, and here should be different item model for UI purposes.

client/ayon_core/tools/launcher/abstract.py
```diff
@@ -67,6 +69,42 @@ class WorkfileItem:
...
+    action_type: str
+    identifier: str
+    label: str
+    icon: Optional[dict]
+    addon_name: Optional[str]
+    addon_version: Optional[str]
+    project_name: Optional[str]
+    folder_id: Optional[str]
+    task_id: Optional[str]
+    task_name: Optional[str]
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/tools/launcher/control.py
```diff
@@ -246,3 +249,69 @@ def get_my_tasks_entity_ids(
...
+        """
+        item = self._recent_actions_model.get_recent_action_item(record_id)
+        if item is None:
+            return
+
+        self.set_selected_project(item.project_name)
+        self.set_selected_folder(item.folder_id)
+        self.set_selected_task(item.task_id, item.task_name)
+        self.set_selected_workfile(item.workfile_id)
+
```
iLLiCiTiT, line 331: I believe this is what UI should do.
```suggestion
```

iLLiCiTiT:
Shouldn't this be unified with frontend?

BigRoy:
> Shouldn't this be unified with frontend?

The frontend has no equivalent - consider this a draft prototype to discuss the design. I'd say at this point the exact implementation is slightly less of a concern then how we want to expose it to the user (that's the first thing to decide upon)

So consider this a UX design draft.

BigRoy:
Just noting here that this is somewhat 'paused' due to: https://github.com/ynput/ayon-core/pull/1816 meaning that this may need to be potentially use different sort of components.

mahesh-ynput:
@BigRoy, should I use new core where we moved ui_qt?

BigRoy:
Yes, the end goal is that this feature starts to live and use the AYON UI qt components - basing it on what comes into core. You may be best targeting this PR to your core qt ui branch so that the changes listed on Github remain minimal to where we'll be merging it into otherwise it'd end up showing all UI Qt component diffs as well if we keep it focused to `develop`.

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,283 @@
...
+    def log(self) -> Logger:
+        if self._log is None:
+            self._log = Logger.get_logger(self.__class__.__name__)
+        return self._log
+
+    def _get_current_username(self) -> str | None:
+        try:
+            return get_ayon_username()
+        except Exception:
+            return None
```
BigRoy: Any reason this needs the try except?

client/ayon_core/tools/launcher/ui/recent_actions_widget.py
```diff
@@ -0,0 +1,326 @@
...
+        ):
+            continue
+
+        icon_def = candidate.icon
+        if not icon_def:
+            return fallback_icon
+        try:
+            return get_qt_icon(icon_def)
+        except Exception:
+            return fallback_icon
```
BigRoy: @mahesh-ynput can you fix all these bare excepts and simplify the code? Same for the code above.

BigRoy (commented):
There's a lot of bare try..excepts and code that feels overly complex for what it's trying to do - can we consolidate the changes a bit so they aren't so "overdone". Can we clean up all these bare try.. except Exception cases - it'll just hide bugs otherwise.

client/ayon_core/tools/launcher/ui/recent_actions_widget.py
BigRoy: I think comment is outdated/irrelevant now.

client/ayon_core/tools/launcher/abstract.py
BigRoy: @mahesh-ynput

mahesh-ynput: https://github.com/ynput/ayon-core/pull/1783/commits/90f2618c0e0faf49a09d43b4717efef2c1a8df82

client/ayon_core/tools/launcher/models/recent_actions.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1783/commits/90f2618c0e0faf49a09d43b4717efef2c1a8df82

client/ayon_core/tools/launcher/ui/recent_actions_widget.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1783/commits/a0aa0df2bc05ff6c49ae7031f6b407dbfe568865

client/ayon_core/tools/launcher/abstract.py
```diff
@@ -72,6 +74,75 @@ class WorkfileItem:
...
+    label: str
+    timestamp: float
+    project_name: Optional[str] = None
+    folder_id: Optional[str] = None
+    task_id: Optional[str] = None
+    workfile_id: Optional[str] = None
+    addon_name: Optional[str] = None
+    addon_version: Optional[str] = None
+    task_name: Optional[str] = None
+    icon: Optional[dict] = None
```
BigRoy, line 143: Why do we need these separate data structures? They seem to be 1:1 the same?

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,274 @@
...
+            timestamp=record.timestamp,
+            project_name=record.project_name,
+            folder_id=record.folder_id,
+            task_id=record.task_id,
+            workfile_id=record.workfile_id,
+            addon_name=record.addon_name,
+            addon_version=record.addon_version,
+            task_name=None,
+            icon=record.icon,
+        )
```
BigRoy, line 93: Instead of storing, label and icon - can we store the action identifier instead?
So that if on addon updates or action changes over time the Recent Actions list the current label - not the one we 'stored' in the past.

I've faced some issue now where the `icon` values have changed to `get_qt_icon` and that before I pulled upstream the stored recent actions suddenly broke the launcher because they were invalid data. This made me think about why we're storing the icon data like that to begin with... :/ why not store the action itself instead so that get the icons and label from the action now?

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,274 @@
...
+        items = items[:RECENT_ACTIONS_MAX]
+        self.log.debug(
+            "After dedup+cap count=%d max=%d", len(items), RECENT_ACTIONS_MAX
+        )
+
+        self._save(items)
+        self._controller.emit_event(
+            "recent_actions.changed",
+            {},
+            "recent_actions.model",
```
BigRoy, line 202: Might be easier here to read when we pass the keywords arguments explicitly?
```suggestion
            topic="recent_actions.changed",
            data={},
            source="recent_actions.model",
```

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def _save_to_user_data(self, items: list[RecentActionRecord]) -> bool:
+        username = get_ayon_username()
+
+        try:
+            user = ayon_api.get_user()
+            user_data = self._normalize_user_data(user.get("data"))
+            user_data[_USER_DATA_KEY] = self._serialize_items(items)
+            response = ayon_api.raw_patch(
+                f"users/{username}", json={"data": user_data}
```
BigRoy, line 151: Does patching work if you patch by `f"data.{_USER_DATA_KEY}"` so that you don't need to pass the full data. I know where we have certain places we can use the dot notation, so just wondering here.

BigRoy (commented):
There's some weird things going on for me:

Image

It's very short (why doesn't it expand further down?) with a scrollbar.

Also, can we get the title label also have the same background color as the reset. No need to separate with a dark header. Match this design more: https://github.com/ynput/ayon-core/pull/1783#issue-4252413394

client/ayon_core/tools/launcher/abstract.py
mahesh-ynput: @BigRoy , https://github.com/ynput/ayon-core/pull/1783#discussion_r3071652839. My understanding wrong here?

BigRoy: Seems a bit verbose just maintaining these two of the same dataclasses @iLLiCiTiT ?

Also, bit confused as to why the attribute order differs 🤷‍♂️

LiborBatek:
I can see and navigate the history via menu in the Launcher...cool!

image

Out of curiousity, right now clicking some history entry will take me into that context but wont trigger the any action directly. Is this intended behaviour? I mean if it should perform that action right away or just take you to that context used before (aka place in history)

I was thinking it would trigger that action upon clicking it....but tell me!

Despite the above, very cool and handy feature!

BigRoy:
Thanks @LiborBatek actually - there should be two buttons on the right side (one to favorite them, and one to trigger the action) - but it seems because your context paths are so long it gets cut off. This is a perfect bug! 🗡️ Thanks, will patch it.

Thanks for the screenshot!

LiborBatek (approved):
Works now...I am able to see those btns even when having long project paths shown,,,

Image

Also clicking those recent paths perform the action and btns does its job too (favourite / go to context)

LGTM

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,496 @@
...
+        # Guards '_items' and the recording queue. Both are touched by the
+        # worker thread and by the thread loading the history.
+        self._lock = threading.Lock()
+        self._items: Optional[list[RecentActionItem]] = None
+        self._queued_triggers = collections.deque()
+        self._save_requested = False
+        self._worker: Optional[threading.Thread] = None
+        self._prewarmed = False
+
+        controller.register_event_callback(
```
moonyuet, line 96: consistency only, maybe there is some reason behind to straight up using `controller` as variable.
```suggestion
        self._controller.register_event_callback(
```

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,496 @@
...
+            "action.trigger.finished",
+            self._on_action_trigger_finished,
+        )
+        controller.register_event_callback(
```
moonyuet, line 100:
```suggestion
        self._controller.register_event_callback(
```

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,496 @@
...
+            for existing in items:
+                if (
+                    existing.identifier == item.identifier
+                    and existing.action_type == item.action_type
+                    and existing.addon_name == item.addon_name
+                    and existing.project_name == item.project_name
+                    and existing.folder_id == item.folder_id
+                    and existing.task_id == item.task_id
+                    and existing.workfile_id == item.workfile_id
+                ):
```
moonyuet, line 487: I am thinking if it is more consistent to cache the key and set property-wrapped `@property` function and then we can iterate through the key instead. The only thing we need to concern is the speed.

client/ayon_core/tools/launcher/ui/hierarchy_page.py
```diff
@@ -134,9 +138,25 @@ def __init__(self, controller, parent):
...
 
+        # State for deferred "Locate" navigation
+        self._pending_locate_folder_id = None
+        self._pending_locate_task_name = None
+        self._pending_locate_workfile_id = None
+
         # Post init
         projects_combobox.set_listen_to_selection_change(self._is_visible)
 
+        controller.register_event_callback(
```
moonyuet, line 149:
```suggestion
        self._controller.register_event_callback(
```

client/ayon_core/tools/launcher/ui/hierarchy_page.py
```diff
@@ -134,9 +138,25 @@ def __init__(self, controller, parent):
...
+            "selection.project.changed",
+            self._on_selection_project_changed,
+        )
+        controller.register_event_callback(
```
moonyuet, line 153:
```suggestion
        self._controller.register_event_callback(
```

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,496 @@
...
+    def set_favorite(self, record_id: str, favorite: bool) -> None:
+        """Pin an entry to the top of the history, or unpin it."""
+        with self._lock:
+            if self._items is None:
+                return
+            for item in self._items:
+                if item.record_id == record_id:
+                    break
+            else:
+                return
```
moonyuet, line 235: should we check if the `self._worker` refers to the current thread and clear up the `self._worker` here?

moonyuet (approved):
I only have several comments above. But they are all suggestions only and can be ignored for the efficiency of the software.(If they do slow down the load).

LiborBatek:
@BigRoy  one thing which comes to my mind in terms of improving it...

Lessening the clutter by reducing the full path to just `parent folder/task` or even instead of showing full context show just a very last entity e.g. `workfile name` or so. Could be maybe switchable by knob? Or full path on hoover over or something...

But Im really not sure whether its possible to achieve it or not. So take it with a grain of salt.

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,496 @@
...
+        return self._deserialize_items(raw)
+
+    def _save_to_user_data(self, items: list[RecentActionItem]) -> None:
+        try:
+            # Name and data come from the same response, so the history is
+            # always written back to the user it was read from.
+            user = ayon_api.get_user()
+            user_data = user.get("data")
+            user_data = dict(user_data) if isinstance(user_data, dict) else {}
+            user_data[_USER_DATA_KEY] = self._serialize_items(items)
```
iLLiCiTiT, line 328: Update of data will update only the keys passed in (as far as I know).
```suggestion
            user_data = {_USER_DATA_KEY: self._serialize_items(items)}
```

client/ayon_core/tools/launcher/models/recent_actions.py
```diff
@@ -0,0 +1,496 @@
...
+
+    """
+    if isinstance(icon, IconBase):
+        return icon.to_data()
+    if isinstance(icon, dict):
+        return icon
+    return None
+
+
+class RecentActionsModel:
```
iLLiCiTiT, line 56: Why is this not part of actions model?

client/ayon_core/tools/launcher/ui/hierarchy_page.py
iLLiCiTiT: Current state is better. But I would move it before the variables are stored to `self`.

client/ayon_core/tools/launcher/models/selection.py
```diff
@@ -36,6 +36,10 @@ def set_selected_project(self, project_name: Optional[str]) -> None:
             return
 
         self._project_name = project_name
+        self._folder_id = None
```
iLLiCiTiT, line 39: The goal was that the task widget will trigger `set_selected_task` when the folder does change. The same applies for workfile etc.

So the backend does not set something that is not true in frontend. Frontend should tell backend what is selected...

iLLiCiTiT:
I kinda don't like a lot of changes in the backend and frontend. It changes how it was designed and is going against the assumptions that were made in past. Probably will have dedicate more time to go through this PR.

PR #1784, opened 2026-04-13
Propagate product version to attached reviewable

tadeas-hejnic:
## Changelog Description
The new collector is propagating data from the product to which the reviewable is attached. 

It is not solving the issue with multiple products, but it's propagating at least some relevant data as mentioned in the [issue comment](https://github.com/ynput/ayon-core/issues/1754#issuecomment-4165589844).

## Additional info
For example, the burnins are created just once for all the products, so it should be decided from which product should be the data propagated. Also I am not sure which data should be propagated in addition to `version data`. 

## Testing notes:
1. Publish with attach reviewable
2. Check if the version in burnins matches the version of product (not equal to `v001`)

### DIFF
--- a/client/ayon_core/plugins/publish/collect_attach_data.py
+++ b/client/ayon_core/plugins/publish/collect_attach_data.py
@@ -1,0 +1,127 @@
+import copy
+
+import pyblish.api
+
+
+class CollectAttachData(pyblish.api.ContextPlugin):
+    """Collect version metadata from attached reviewables."""
+
+    order = pyblish.api.CollectorOrder + 0.491
+    label = "Collect data from attached products"
+
+    def process(self, context):
+        self.log.debug("Collecting data from attached products")
+        for instance in context:
+            if not instance.data.get("active", True):
+                continue
+
+            attach_to = self._get_attach_targets(instance)
+            if not attach_to:
+                continue
+
+            attach_instance = self._get_source_attach_instance(
+                instance, attach_to
+            )
+            if attach_instance is None:
+                continue
+
+            self._merge_attached_data(attach_instance, instance)
+
+    def _get_attach_targets(self, instance):
+        publish_attributes = instance.data.get("publish_attributes", {})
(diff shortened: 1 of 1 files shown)


LiborBatek (changes requested):
My review even been the very first one gets the version after the `model` which is not attached by default (aka model version got propagated into review product wrongly)

In my case model before publish been `v003` leading to `v004` after publish

`Review` product not been attached to the `model` purposely and should get `v001` but gets also `v004` after the `model` product version...which is not correct.

Image

and in Ayon and products page:

model been correct while review having incorrect version after the model

Image

iLLiCiTiT:
I don't think this is relevant PR, see [comment](https://github.com/ynput/ayon-core/issues/1754#issuecomment-4242502230).

LiborBatek:
@iLLiCiTiT well imagine having 2 products streams `model` with latest ver `v004` and `Review` in version `v002` but when you attach it to the `model` product via `attach review` action it will still have its own version stream not picking that model one aka `v004` or rather `v005` when new version produced with attached review.

I guess its invalid this way it works in Ayon atm.

iLLiCiTiT:
> well imagine having 2 products streams model with latest ver v004 and Review in version v002 but when you attach it to the model product via attach review action it will still have its own version stream not picking that model one aka v004 or rather v005 when new version produced with attached review.

I do understand that. But the feature literaly is: "Take review of instance A and use it for instance B", as is.

iLLiCiTiT:
To support what is requested, we would have to add plugin inbetween extract review and burnins to copy (duplicate) the review files for each instance that should be using it, prepare new representation for the instance with the duplicated file so following plugins are using that as source for review. There might be caviads but they can be resolved when they appear.

That would probably literaly scratch the existing integrate plugin as that would technically fully replace it. But I would move the publish attributes definition to a collect plugin so if we change the logic "how things actually happen" we don't have to handle old values and keep the collect plugin as is untouched.

@tadeas-hejnic

PR #1787, opened 2026-04-16
Extract Thumbnail: Use exr instead of png for create frame from video

moonyuet:
## Changelog Description
This PR is to ensure ffmpeg writing the correct image with alpha so that it can be eventually correctly transcoded into thumbnail. The purpose of this PR is the fact that the oiiotool cannot read png with alpha and we need to resolve it with exr instead.  
## Additional info
n/a

## Testing notes:
1. Create Render (with video, render sequences)
2. Publish

### DIFF
--- a/client/ayon_core/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_core/plugins/publish/extract_thumbnail.py
@@ -630,7 +630,7 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
         base_name = os.path.basename(video_file_path)
         filename = os.path.splitext(base_name)[0]
         output_thumb_file_path = os.path.join(
-            output_dir, "{}.png".format(filename))
+            output_dir, "{}.exr".format(filename))
 
         # Set video input attributes
         max_int = str(2147483647)


BigRoy:
@iLLiCiTiT @antirotor @jakubjezek001 any ffmpeg knowledge in the house on why these must be flipped here?

@moonyuet can you provide an example input file that fails and the relevant ffmpeg call that would work and the one that would fail so someone else can easily test the same in isolation?

moonyuet:
> @iLLiCiTiT @antirotor @jakubjezek001 any ffmpeg knowledge in the house on why these must be flipped here?
> 
> @moonyuet can you provide an example input file that fails and the relevant ffmpeg call that would work and the one that would fail so someone else can easily test the same in isolation?

It isn't failed (or created any bug) and it is creating black thumbnail for reviewables when publishing render product types in the channels other than `rgb`.

BigRoy:
> It isn't failed (or created any bug) and it is creating black thumbnail for reviewables when publishing render product types in the channels other than `rgb`.

Provide the relevant command line call that generates the thumbnail (both working and the black one). It should appear in the logs.

client/ayon_core/plugins/publish/extract_thumbnail.py
```diff
@@ -653,12 +653,15 @@ def _create_frame_from_video(
 
         # Build command args
         cmd_args = []
+        # Ensure the ffpmeg has the full stream
+        # before seeking to the position
+        cmd_args.extend(["-i", video_file_path])
```
iLLiCiTiT:
```suggestion
        cmd_args = [
            "-i", video_file_path,
        ]
```

iLLiCiTiT:
Both do work for me. It looks like the `-ss` before input is "for all streams" and `-ss` after input is "for all streams of the input". In case of thumbnail extraction where is only one input it should not matter?

~~Isn't that black thumbnail caused by having offset that leads after the end of the input file?~~ That would fail to create the output file.

I'm ok with this PR as it does not really change anything. But related to the bug, we need input file and arguments that can reproduce the issues, so we can verify this does fix it.

moonyuet:
> But related to the bug, we need input file and arguments that can reproduce the issues, so we can verify this does fix it.



> Both do work for me. It looks like the `-ss` before input is "for all streams" and `-ss` after input is "for all streams of the input". In case of thumbnail extraction where is only one input it should not matter?
> 
> ~Isn't that black thumbnail caused by having offset that leads after the end of the input file?~ That would fail to create the output file.
> 
> I'm ok with this PR as it does not really change anything. But related to the bug, we need input file and arguments that can reproduce the issues, so we can verify this does fix it.

I am not sure if I am allowed to share the file as it is shared by the client, you can access the [file](https://discord.com/channels/517362899170230292/1479127436712218847/1479420185013260400)with matte channel 
To verify the fixes, you need to do the testing below.
I will also edit the testing steps for the github issue too.

Testing with develop branch of core addon
1. Create Render Instance(e.g. RenderTestAlpha) and exposes `channels` in `ayon+settings://nuke/create/CreateWriteRender/exposed_knobs`
2. change the channel to `all` or `rgba` 
3. Publish
4. Check the publish of the render instance e.g. RenderTestAlpha in the server
5. You would find the black thumbnail created for that render instance

Testing with the branch of core addon
1. Repeat the step 1-4 above
2. You would find the thumbnail created for that render instance

iLLiCiTiT:
Wait, what? It is single file? I guess not, signle file cannot have `-ss`.

We need input mp4/mov file and arguments that are passed to ffmpeg in this plugin. Not full publishing step.

moonyuet:
> We need input mp4/mov file and arguments that are passed to ffmpeg in this plugin. Not full publishing step.

I can't replicate myself with mp4/mov/ single image to be honest,
but with the images the client shared, I can replicate.
I can still manage to get this with some contents below: https://github.com/user-attachments/assets/642f18a4-9742-43b0-b059-4a80b826ca6e
Testing command
`./ffmpeg.exe -ss 5.0 -i D:\test.mp4 -analyzeduration 2147483647 -probesize 2147483647 -y -frames:v 1 D:/output_ffmpeg.jpg`
The command for the code changed:
`./ffmpeg.exe -i D:\test.mp4 -ss 5.0 -analyzeduration 2147483647 -probesize 2147483647 -y -frames:v 1 D:/output_ffmpeg.jpg`

iLLiCiTiT:
The issue is with alpha not with time offset. The thumbnail is empty no matter where the `-ss` is.

moonyuet:
@BigRoy can you take a look at it?

BigRoy (approved):
This avoids the alpha issue for those cases where it's relevant. 

@LiborBatek can you just confirm that publishes (especially those with alphas, incl. e.g. renders from Maya, 3dsmax, etc. get their thumbnails generated and they still look as intended. This should really only affect the thumbnail alpha.)

LiborBatek:
so far works with `maya` and `3dsmax` Im now in the middle of testing it with `blender`... should be finnished soon

so I will update here asap

LiborBatek:
probably not related to this PR but I am experiencing constant crashes while Publisher showing up...

any ideas what might be going on??

Screenshot 2026-04-17 150950

And the crash log:

[Blnd_crash_log.txt](https://github.com/user-attachments/files/26827246/Blnd_crash_log.txt)

iLLiCiTiT:
Not related to this PR at all. Please use different threads to resolve.

moonyuet:
> probably not related to this PR but I am experiencing constant crashes while Publisher showing up...
> 
> any ideas what might be going on??
> 
> Screenshot 2026-04-17 150950
> And the crash log:
> 
> [Blnd_crash_log.txt](https://github.com/user-attachments/files/26827246/Blnd_crash_log.txt)

It is probably and most likely caused by the merge of the PR in blender where the time interval has been changed in Blender https://github.com/ynput/ayon-blender/pull/289 so not related to this PR.

PR #1791, opened 2026-04-17
Create hero version: Convert create hero version loader to action

iLLiCiTiT:
## Changelog Description
Converted Create hero version to loader action instead of loader plugin.

## Additional info
Because the action can work on all of selected versions at once the logic changed a little to take that into account. The action is shown only in browser and not in hosts, not sure if that is correct behavior?

## Testing notes:
1. It is possible to create hero version from existing versions.


kalisp:
I am not sure if it is fault of conversion or if it was there before, but creation of  Hero version from version with multiple representations copies over only single representation.

image

image

iLLiCiTiT:
> I am not sure if it is fault of conversion or if it was there before, but creation of Hero version from version with multiple representations copies over only single representation.

Try it without this PR, but I didn't touch that part (as far as I know).

EDITED:
Might be because you don't have `<_{output}>` in delivery template, or we don't handle it, as the only difference is missing the output specification.

kalisp (approved):
That mentioned issue is in develop so it could be fixed in different PR.

(It has an issue with multiple representations for `image` product type, but not for `model`, but repres that I tested had different extensions and with those it worked as expected.)

kalisp:
https://github.com/ynput/ayon-core/issues/1793

PR #1794, opened 2026-04-21
Oiio parallel execution

robertokker:
## Changelog Description
Previously, the _translate_to_sequence method used a hardcoded, highly restrictive clique pattern ([clique.PATTERNS["frames"]]), which only matched standard dot-padded sequences like filename.1001.exr. If an incoming representation used underscoring formatting (like _1001.exr), the matcher failed and simply grouped all files into the "remainders" bin. This forced the code to default to the single-file fallback block, executing OIIO identically file-by-file in separate subprocesses.

The Fix
I replaced the _translate_to_sequence hack and completely decoupled the processing directly inside the process() loop using a natively robust implementation constraint:

Native Sequence Detection: Now runs clique.assemble(..., assume_padded_when_ambiguous=True) directly on files_to_convert to accurately chunk sequences regardless of numbering separators (dots, underscores, etc.).
Wildcard Preparation: The script dynamically formats the OIIO expected wildcard sequence path directly into collection.format(f"{{head}}#{{tail}}") giving # formatting across sequences.
Optimized Execution Batch:
By creating the standardized parameter array upfront, oiio_color_convert uses parallel_frames=True and receives exactly the single overarching sequence wildcard query.
Using the underlying ayon_core.lib.execute.run_subprocess logic, it spins up OpenImageIO precisely only once per complete sequence context (speeding up transcode speeds significantly without redundant overhead).
Stable Fallback: If standard files, matte-paintings, or completely unrelated single assets hit the tool, we bypass wildcard definitions and loop through a stable single-file fallback.

## Additional info
Although i wanted this to write myself, I'm definitely not keen enough for this to write completely myself.
So although I've been testing it, reviewing it most of the code has been written with the help of Gemini.
I tried to keep it clean so no incomplete slop was created.

But the results are that a shot exported from hiero this part of the process used to take 4:20 and with this implementation it was just 19 seconds
New
image

Old
image


client/ayon_core/hooks/pre_ocio_hook.py
```diff
@@ -76,3 +76,29 @@ def execute(self):
...
 
         self.launch_context.env["OCIO"] = ocio_path
+
+        custom_vars = config_data.get("custom_vars", {})
+        for var_name, var_value in custom_vars.items():
+            if var_value:
+                # Normalize path to fix mixed slashes like `W:\/AY1_Ayon/foo`
+                import os
+                var_value = os.path.normpath(var_value)
+                if self.host_name in ["nuke", "hiero"]:
```
iLLiCiTiT, line 86: If host specific logic is needed then it should be moved to the host addons, or do it all the time.

client/ayon_core/hooks/pre_ocio_hook.py
```diff
@@ -76,3 +76,29 @@ def execute(self):
             f"Setting OCIO environment to config path: {ocio_path}")
 
         self.launch_context.env["OCIO"] = ocio_path
+
+        custom_vars = config_data.get("custom_vars", {})
```
iLLiCiTiT, line 80: How is the custom vars logic related to OIIO processing?

robertokker: ohh nooo.
I made a new branch from another branch.
It shouldnt be includeded.

Let me close this down and push a normal one as it's just be extract_color_transcode.py

So sorry

robertokker:
wrong pull request, i accidently made a branch of a branch.
I'll push the correct one

robertokker:
fixed in this one
https://github.com/ynput/ayon-core/pull/1796

PR #1800, opened 2026-04-22
Extract USD layer: Backup profile has filled values

iLLiCiTiT:
## Changelog Description
If no profile is matched then define profile with all keys that should be available in profile from settings.

## Additional info
It is better to know what actually should be in the profile is none is available. Avoid having `.get(..., <default>)` which might potentially hide a bug in key vs. attribute difference.

## Testing notes:
1. Extract USD layer should show it's attributes, if there is matching profile and if there is no matching profile.


BigRoy (approved):
Nice. @MustafaJafar any chance you can test to confirm all works?

BigRoy (approved):

PR #1802, opened 2026-04-22
Publisher instances sort by folder path and product name and fix shif…

rischpierre:
## Changelog Description
Hello, I noticed a  bug on the Publisher regarding the order of instances and the mutlitple selection in the case of  multishot workflow.  
I did not saw any issue regarding this, so I created this MR directly  
here is the bug ( you can see the selection is messed up)  
Peek 2026-04-15 09-20

and after this fix
Peek 2026-04-15 09-44

I also sorted the instances by shot and then product name so it is easier for the artists.  

## Testing notes:
1. Open a DCC (gaffer in my case) with multiple shot publisher nodes containing different shots 
2. ctrl+click select the instances

### DIFF
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -816,12 +816,10 @@ class InstanceCardView(AbstractInstanceView):
 
             widget_idx += 1
 
-            instances = sorted(
+            for instance in sorted(
                instances_by_group[group_name],
-               key=lambda i: (i.get_folder_path(), getattr(i, "product_name", ""))
-            )
-
-            for instance in instances:
+               key=lambda i: (i.get_folder_path() or "", i.product_name))
+            ):
                 group_by_instance_id[instance.id] = group_name
                 instance_ids_by_group_name[group_name].append(instance.id)
 
--- a/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
+++ b/client/ayon_core/tools/publisher/widgets/card_view_widgets.py
@@ -816,7 +816,11 @@ class InstanceCardView(AbstractInstanceView):
 
             widget_idx += 1
 
-            instances = instances_by_group[group_name]
+            instances = sorted(
+               instances_by_group[group_name],
+               key=lambda i: (i.get_folder_path(), getattr(i, "product_name", ""))
+            )
+
             for instance in instances:
                 group_by_instance_id[instance.id] = group_name
                 instance_ids_by_group_name[group_name].append(instance.id)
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -816,7 +816,11 @@ def refresh(self):
 
             widget_idx += 1
 
-            instances = instances_by_group[group_name]
+            instances = sorted(
+               instances_by_group[group_name],
+               key=lambda i: (i.get_folder_path(), getattr(i, "product_name", ""))
+            )
+
             for instance in instances:
```
iLLiCiTiT:
```suggestion
            for instance in sorted(
               instances_by_group[group_name],
               key=lambda i: (i.get_folder_path() or "", i.product_name))
            ):
```

iLLiCiTiT: Why was the sorting moved here?

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
```diff
@@ -816,8 +816,10 @@ def refresh(self):
 
             widget_idx += 1
 
-            instances = instances_by_group[group_name]
-            for instance in instances:
+            for instance in sorted(
+               instances_by_group[group_name],
+               key=lambda i: (i.get_folder_path() or "", i.product_name))
```
iLLiCiTiT, line 821: Just realized this is `InstanceItem` not `CreatedInstance`.
```suggestion
               key=lambda i: (i.folder_path or "", i.product_name))
```

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
rischpierre: If I remember correctly, initially, there was a sort in a function, but it was not replicated in the function that was calculating the index for the ctrl+click select.
and that function is relying on group_by_instance_id or instance_ids_by_group_name
that is why the sorting needs to be applied on call level above.
Sorry it is  a bit vague in my mind

client/ayon_core/tools/publisher/widgets/card_view_widgets.py
iLLiCiTiT: Still same question, why this was moved here? It makes more sense to sort them in `_update_instance_widgets` if possible.

LiborBatek:
@BigRoy 

I have tested first in our recent addons version...basically it works ok using selections via `CTRL+ click` while it failed using `SHIFT+ click` and started to "randomly selecting" while using different order as described in this PR

However pulling this branch gives me errors while trying to even open `Publisher` window...

```
# Error: SyntaxError: file c:\Work\REPO\ayon-core\client\ayon_core\tools\publisher\widgets\overview_widget.py line 10: unmatched ')'
```

I guess this branch needs updating before testing to remove any issues...

PR #1803, opened 2026-04-22
Settings registry: Enhancements

iLLiCiTiT:
## Changelog Description
AYON registry is using local launcher dir instead of storage dir. JSON and INI settings registry also have `filepath` property.

## Additional info
Use local directory instead of storage directory which can be shared. Also gave option to get the path to the registry if it is filebased registry.

## Testing notes:
1. Validate code changes.


kalisp (approved):
LGTM

PR #1807, opened 2026-04-27
Collect resources: Fill originalDirname

### LINKED ISSUE #1696: YN-0486: Missing originalDirname

During publishing, plugin

```
collect_resources_path
```

will format anatomy template to get the path. If that template has

```
{originalDirname}
```

in it, it will fail because it is using only data stored on

```
anatomyData
```

instance key, but

```
originalDirname
```

is collected directly to instance data. 

During publishing, plugin `collect_resources_path` will format anatomy template to get the path. If that template has `{originalDirname}` in it, it will fail because it is using only data stored on `anatomyData` instance key, but `originalDirname` is collected directly to instance data.

_No response_

### Additional context:

_This issue was automatically created from AYON ticket [YN-0486]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=419178c005dd11f18b96878abdd0e184)

iLLiCiTiT:
## Changelog Description
Fill in `originalDirname` in collect resources.

## Additional info
If publish template for the instance does have original dirname it can fail to collect resources because of missing `originalDirname` in template data.

## Testing notes:
1. Define a template which will have `{orignalDirname}` in directory template.
2. Create an instance that will use the template (settings for that are `ayon+settings://core/tools/publish/template_name_profiles`). I've used `image` product base type in traypublisher for testing.
3. Publish.
4. Collect resources path should not fail.

Resolves https://github.com/ynput/ayon-core/issues/1696

### DIFF
--- a/client/ayon_core/plugins/publish/collect_resources_path.py
+++ b/client/ayon_core/plugins/publish/collect_resources_path.py
@@ -13,7 +13,7 @@ import copy
 
 import pyblish.api
 
-from ayon_core.pipeline.publish import get_publish_template_name
+from ayon_core.pipeline.publish import get_publish_template_name, PublishError
 
 
 class CollectResourcesPath(pyblish.api.InstancePlugin):
@@ -109,13 +109,14 @@ class CollectResourcesPath(pyblish.api.InstancePlugin):
             original_directory = instance.data.get("originalDirname")
             if not original_directory:
                 original_directory = instance.data.get("stagingDir")
-                if not original_directory:
-                    self.log.warning(
-                        "Publish template requires 'originalDirname'"
-                        " but 'originalDirname' is not set on instance"
-                        " and 'stagingDir' is not yet filled."
-                    )
-                    return
+
+            if not original_directory:
+                raise PublishError(
+                    "Publish template requires 'originalDirname'"
+                    " but 'originalDirname' is not set on instance"
+                    " and 'stagingDir' is not yet filled."
+                )
+
             template_data["originalDirname"] = original_directory
 
         publish_folder = os.path.normpath(
--- a/client/ayon_core/plugins/publish/collect_resources_path.py
+++ b/client/ayon_core/plugins/publish/collect_resources_path.py
@@ -106,7 +106,7 @@ class CollectResourcesPath(pyblish.api.InstancePlugin):
             "publish", template_name, "directory")
 
         if "{originalDirname}" in publish_template:
-            original_directory =  instance.data.get("originalDirname")
+            original_directory = instance.data.get("originalDirname")
             if not original_directory:
                 original_directory = instance.data["stagingDir"]
             template_data["originalDirname"] = original_directory
--- a/client/ayon_core/plugins/publish/collect_resources_path.py
+++ b/client/ayon_core/plugins/publish/collect_resources_path.py
@@ -105,6 +105,12 @@ class CollectResourcesPath(pyblish.api.InstancePlugin):
         publish_template = anatomy.get_template_item(
             "publish", template_name, "directory")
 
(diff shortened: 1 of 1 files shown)


moonyuet:
As far as I tested with `{originalDirname}` as template in my side (I dont know much about the scope of this PR) in traypublisher, it is working with the path based within the project directory.
Enclosed the publish reports 

[publish-report-260428-14-44.json](https://github.com/user-attachments/files/27153736/publish-report-260428-14-44.json)

moonyuet (approved):
Approved with the comment stated above

client/ayon_core/plugins/publish/collect_resources_path.py
```diff
@@ -105,6 +105,12 @@ def process(self, instance):
         publish_template = anatomy.get_template_item(
             "publish", template_name, "directory")
 
+        if "{originalDirname}" in publish_template:
+            original_directory = instance.data.get("originalDirname")
+            if not original_directory:
+                original_directory = instance.data["stagingDir"]
```
BigRoy: This would work with tray publisher - but pretty sure apps like e.g. Maya don't have `stagingDir` set until the Extractor has run because that is the one creating it. Not super-problematic, but something to be wary of that this part may not be the holy grail to solving thing? 🤔 (In those cases those are also 'temporary directories' so you wouldn't really want to be using `originalDirname` to begin with though - unless you're using custom staging dirs that are persistent?)

moonyuet: There is some edge case of which we cannot publish with original directory name across the projects.
[publish-report-260428-14-43.json](https://github.com/user-attachments/files/27165289/publish-report-260428-14-43.json)

MilaKudr:
Hope I test it correctly, but had this error during validation before publishing.
```
DEBUG: Looking for matching profile for: host_names: "nuke" | product_base_types: "render" | task_names: "GENERIC" | task_types: "Generic"
DEBUG: "nuke" not found in "host_names": ['substancedesigner']
DEBUG: "nuke" not found in "host_names": ['maya']
DEBUG: "nuke" not found in "host_names": ['traypublisher']
DEBUG: "nuke" not found in "host_names": ['max']
DEBUG: "nuke" not found in "host_names": ['unreal']
DEBUG: Profile selected: {'host_names': [], 'product_base_types': ['review', 'render', 'prerender'], 'task_names': [], 'task_types': [], 'template_name': 'render'}
  File "C:\Users\mila\AppData\Local\Ynput\AYON\dependency_packages\ayon_2602231726_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\Users\mila\AppData\Local\Ynput\AYON\addons\core_1.9.0+dev\ayon_core\plugins\publish\collect_resources_path.py", line 111, in process
    original_directory = instance.data["stagingDir"]
                         ~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'stagingDir'
```

image

client/ayon_core/plugins/publish/collect_resources_path.py
iLLiCiTiT: Added warning log. Not sure if we can do anything about it?

iLLiCiTiT: Should it crash instead?

BigRoy: I think the fact that we're now collecting nothing will also be confusing - because pretty sure this now moves the issue just somewhere. I think this should really be a `PublishError`.

iLLiCiTiT: Changed

PR #1810, opened 2026-04-29
Run plug-ins regardless of hosts: Collect Audio, Extract Review, Extract Burnin and Clean Up

BigRoy:
## Changelog Description

Run plug-ins regardless of hosts: Collect Audio, Extract Review, Extract Burnin and Clean Up.

## Additional info

This will most definitely have side-effects. 

Particularly because suddenly these plug-ins will trigger on hosts that previously were not in these lists.
I'd especially investigate the more esoteric hosts - like webpublisher, batch publisher or certain ones like mocha. 

Note that, except for Clean Up plug-in these changed plug-ins only target "review" family (and Extract Burnin also "burnin" family; although I know no case where we add burnin as family)

If we find one plug-in particularly dangerous to change, we can move it into a separate PR - e.g. the Clean Up one.

## Testing notes:

- Clean Up plug-in does not cause problems in:
  - [x] batchdelivery host
  - [x] traypublisher host
  - [x] (3ds)max host
- Collect Folder Audio plug-in ('review' family only) does not cause problems in:
  - [x] workflow host for 'review' family
- All these plug-ins do not cause problems in:
  - [ ] Mocha
  - [ ] 3dequalizer
  - [ ] ... and more hosts that were missing

### DIFF
--- a/client/ayon_core/plugins/publish/cleanup.py
+++ b/client/ayon_core/plugins/publish/cleanup.py
@@ -20,24 +20,6 @@ class CleanUp(pyblish.api.InstancePlugin):
 
     order = pyblish.api.IntegratorOrder + 10
     label = "Clean Up"
-    hosts = [
-        "aftereffects",
-        "blender",
-        "celaction",
-        "flame",
-        "fusion",
-        "harmony",
-        "hiero",
-        "houdini",
-        "maya",
-        "nuke",
-        "photoshop",
-        "resolve",
-        "tvpaint",
-        "unreal",
-        "webpublisher",
-        "shell"
-    ]
     settings_category = "core"
 
     exclude_families = ["clip"]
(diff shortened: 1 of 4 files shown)


iLLiCiTiT:
I'd really love to approve it, but it has to be tested in hosts I even don't know how to use 😞 .

Will try traypublisher with the cleanup plugin, but I do believe that one will cause issues.

iLLiCiTiT:
So, traypublisher and cleanup, for all cases instances have staging set to temp directory and that is marked for explicit cleanup, except for editorial publishing, that one needs retest.

EDITED:
Editorial is ok, because it is not in temp it is skipped.

moonyuet (approved):
It definitely has side-effect to 3dsmax host, of which I will make the PR for the change in regards of that.

jakubjezek001 (approved):

BigRoy:
I'd say, let's merge @iLLiCiTiT

iLLiCiTiT:
You really wanna see the world burn? Ok.

PR #1812, opened 2026-04-30
UILabelDef widget has browser abilities

iLLiCiTiT:
## Changelog Description
UILabelDef widget allows to select text or click uris.

## Testing notes:
You have to use UILabelDef to be able to test it. This is one example how it can be tested.
1. Start tray and open Browser tool.
2. Run `Delete versions` on a product.
3. Submit the first dialog, but DO NOT confirm the second.
4. Second dialog contains `UILabelDef` with `"Going to delete n versions"` where you should be able to select the text.

### DIFF
--- a/client/ayon_core/tools/attribute_defs/widgets.py
+++ b/client/ayon_core/tools/attribute_defs/widgets.py
@@ -353,6 +353,9 @@ class LabelAttrWidget(_BaseAttrDefWidget):
         label = self.attr_def.label
         if label:
             input_widget.setText(str(label))
+        input_widget.setWordWrap(True)
+        input_widget.setTextInteractionFlags(QtCore.Qt.TextBrowserInteraction)
+        input_widget.setOpenExternalLinks(True)
 
         self._input_widget = input_widget
 


MilaKudr (commented):
Image

Image

BigRoy:
I think the `input_widget.setOpenExternalLinks(True)` still needs to be tested - because I think you need to implement a link click handler in Qt for it to do anything, right?

So check with a label that has a hyperlink?

iLLiCiTiT:
> I think the input_widget.setOpenExternalLinks(True) still needs to be tested - because I think you need to implement a link click handler in Qt for it to do anything, right?

Yes, but that is something Mila is not able to test as that requires to implement it in code.

I can confirm it woks
image

BigRoy:
> > I think the input_widget.setOpenExternalLinks(True) still needs to be tested - because I think you need to implement a link click handler in Qt for it to do anything, right?
> 
> Yes, but that is something Mila is not able to test as that requires to implement it in code.
> 
> I can confirm it woks image

Works for me in Publisher UI too: 
image

But the styling makes it near impossible to read. I wonder why yours is slightly better? 🤔 
Anyway, if we can fix the URL styling that'd be great!

iLLiCiTiT:
> Anyway, if we can fix the URL styling that'd be great!

Tried, failed. It is possible to change url look of `QLabel` only by changing palette on `QApplication`, which might be dangerous.

PR #1813, opened 2026-04-30
only use label if it is equal to the variant_label

vincentullmann:
## Changelog Description
Launcher: `full_label` no longer duplicates the label when it matches the variant label.

## Additional info
This allows admins to "ungroup" application variants by setting the `group_label = variant_label`  thus generating unique groups

## Testing notes:
1. start with this step
2. follow this step


iLLiCiTiT:
Rejected, this affects all actions not just applications.

PR #1814, opened 2026-04-30
Validate Product Name Uniqueness: allow it if instances target different versions

### LINKED ISSUE #1815: YN-0703: multiple versions of single products does not fail validation

Core validator should not reise if multiple version product instances are under the same context. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0703]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=48f70ae0449d11f1a838dd13846dfdf9)

BigRoy:
## Changelog Description

Allow multiple instances targeting the same product but a different (explicit) version number to be published from a single context

## Additional info

Fix https://github.com/ynput/ayon-core/issues/1815

Required for things like CSV ingests where multiple instances may be explicitly targeting different versions.

Note that [ValidateVersion](https://github.com/ynput/ayon-core/blob/32ea97af45bef30951c65e235f541cc2f2827e46/client/ayon_core/plugins/publish/validate_version.py#L11) would still trigger if configured with the profile disallowing not publishing a 'latest' version, etc. but that's a correct separate validation from this case. Currently it'd even still allow it if e.g. both instances, represent newer versions than the current existing one which I think is perfect too.

## Testing notes:

1. Publish multiple instances to same target context+product with different `version` data should be allowed.

### DIFF
--- a/client/ayon_core/plugins/publish/validate_unique_subsets.py
+++ b/client/ayon_core/plugins/publish/validate_unique_subsets.py
@@ -56,28 +56,39 @@ class ValidateProductUniqueness(pyblish.api.ContextPlugin):
                 ).format(instance.name))
                 continue
 
-            instance_per_folder_product[(folder_path, product_name)].append(
-                instance
+            # Note: version may be optionally set, but when set that will be
+            #  the resulting target version. If two instances target the same
+            #  product but a different version we will allow it.
+            # TODO: This may have the caveat where one instance *may* target
+            #  None (next version) and that next version matching exactly an
+            #  explicitly set version but that would really be an edge case.
+            version = instance.data.get("version")
+            key = (
+                folder_path,
+                product_name,
+                version
             )
+            instance_per_folder_product[key].append(instance)
 
         non_unique = []
-        for (folder_path, product_name), instances in (
+        for (folder_path, product_name, version), instances in (
             instance_per_folder_product.items()
         ):
             # A single instance per folder, product is fine
             if len(instances) < 2:
                 continue
 
-            non_unique.append(
-                "{} > {}".format(folder_path, product_name)
-            )
+            label = f"{folder_path} > {product_name}"
+            if version is not None:
+                label += f" (version {version})"
+            non_unique.append(label)
 
         if not non_unique:
             # All is ok
             return
 
         msg = (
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Should we allow to publish multiple versions of the same product during one publishing? If yes, shouldn't this plugin be disabled? We really don't know what version it will be at this moment and if one instance A has version `3` instance B has `None` and there are currently 2 published versions, they technically will try to publish to same version (if instance B is processed by integrate asset first).

We can allow it only if all instnaces have explicit version filled (and the versions are different).

BigRoy:
> Should we allow to publish multiple versions of the same product during one publishing? If yes, shouldn't this plugin be disabled? We really don't know what version it will be at this moment and if one instance A has version `3` instance B has `None` and there are currently 2 published versions, they technically will try to publish to same version (if instance B is processed by integrate asset first).

It's the caveat that I mentioned in the code indeed. But perhaps disallowing mix of 'implicit latest' together with 'explicit versions' is a solid idea.

BigRoy:
> Should we allow to publish multiple versions of the same product during one publishing? If yes, shouldn't this plugin be disabled? We really don't know what version it will be at this moment and if one instance A has version `3` instance B has `None` and there are currently 2 published versions, they technically will try to publish to same version (if instance B is processed by integrate asset first).
> 
> We can allow it only if all instnaces have explicit version filled (and the versions are different).

Took another look at this, but this actually doesn't seem to be true. The `instance.data["version"]` should at this point in time always be an integer, due to it being set here: https://github.com/ynput/ayon-core/blob/18ea9ed9b4f24a774375dc131c906d4e0be3de8f/client/ayon_core/plugins/publish/collect_anatomy_instance_data.py#L380

As such, the case where `version` _may_ be `None` would actually be very special and I'd even say... a potential bug?
As such, I think current code is fine.

iLLiCiTiT:
Looks like you're right, but the code does `instance.data.get("version")` so we should either use `instance.data["version"]` or/and validate if the version is not `None`.

BigRoy:
@iLLiCiTiT could you review again please? :)

iLLiCiTiT (approved):
Looks dangerous, but code looks ok.

BigRoy:
Can we merge this? :)

PR #1818, opened 2026-05-08
Implementation of custom frame range collector

moonyuet:
## Changelog Description
This PR is to implement the custom frame range collector with the custom frame range migrated from deadline addon.

## Additional info
Tested with https://github.com/ynput/ayon-deadline/pull/252

## Testing notes:
1. Create Render
2. Test with both local and farm rendering. 
3. Test with or without custom frame range


client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -0,0 +1,117 @@
...
+                instance_change["changes"]
+            )
+
+            instance = instance_change["instance"]
+            # recalculate only if context changes
+            if (
+                "task" not in instance_change
+                and "folderPath" not in instance_change
+                and not custom_frame_change
+            ):
```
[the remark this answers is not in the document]
moonyuet: `instance_change["changes"]` is to track on the state of the custom frame range toggle changes and the check on the context change is to re-track on the state of the custom frame range.

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -0,0 +1,130 @@
...
+                "Custom frames are not used, "
+                "skipping collection of frame range."
+            )
+            return
+
+        frames = attr_values.get("frames")
+        if not frames:
+            raise KnownPublishError("Please fill `Custom Frames` value")
+
+        instance.data["custom_frames"] = frames
```
moonyuet:
```suggestion
        instance.data["customFrames"] = frames
```

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -0,0 +1,130 @@
...
+            "render", "render.local", "render.farm", "render.frames_farm",
+            "prerender", "prerender.farm", "prerender.frames_farm",
+            "renderlayer", "imagesequence", "image",
+            "vrayscene", "maxrender",
+            "arnold_rop", "mantra_rop",
+            "karma_rop", "vray_rop", "redshift_rop",
+            "renderFarm", "usdrender", "publish.hou",
+            "remote_publish_on_farm",
+            "deadline"
+    ]
```
BigRoy, line 25: Ouch, this is painful. We'll need to define a better way to when this plug-in should be relevant.

@iLLiCiTiT let's discuss when we can find some time.

LiborBatek (approved):
Works for me together with `ayon-3dsmax` and `ayon-deadline` with related branches...

Image

client/ayon_core/plugins/publish/collect_custom_frame_range.py
iLLiCiTiT: Yep you're right. This might the breakpoint for me. We will have to tackle the "families" thing.

iLLiCiTiT:
Currently shows the attribute at places where it should not be visible. Has to be resolved first (connected to https://github.com/ynput/ayon-core/pull/1818#discussion_r3307266441 )

antirotor:
So is there any follow-up issue somewhere? I guess it is bigger topic with families and all, but it shouldn't hold up resolution of this? If it works, lets merge it, but create the issue to fix the families?

iLLiCiTiT:
> If it works, lets merge it, but create the issue to fix the families?

It works where it is needed, and it breaks places where it should not be used. So yes, it does resolve the issue, but creates multiple other issues.

BigRoy:
> So yes, it does resolve the issue, but creates multiple other issues.

Let's describe the new issues it creates - track them by priority and resolve the ones we absolutely must and then get to making tickets for those that can be fixed later please.

moonyuet:
What should we do with this PR?

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -0,0 +1,130 @@
...
+        for instance_change in event["changes"]:
+            custom_frame_change = cls._get_publish_use_custom_frames_value(
+                instance_change["changes"]
+            )
+
+            instance = instance_change["instance"]
+            # recalculate only if context changes
+            if (
+                "task" not in instance_change
+                and "folderPath" not in instance_change
```
BigRoy: Why do we care about task / folderPath changes? 🤔

moonyuet: The comment said it is recalculated only if context changes. 
But I dont think we need the check. 
[e05ce19](https://github.com/ynput/ayon-core/pull/1818/commits/e05ce19d2b054fc44c8e27f764bc6f0e174928d2)

iLLiCiTiT:
For reference, [this PR](https://github.com/ynput/ayon-core/pull/1994) is a must have for this feature to avoid showing this everywhere. With that we can limit the plugin to only one, very specifi family that has to be added to instances which actually does support the feature.

LiborBatek (commented):
I have tested it with all major 3d hosts and it worked well



- [x]  Maya 2027
- [x]  3dsmax 2026
- [x]  Blender 5.2.1 LTS

There was a bit suspicious doubling in publisher of the settings regarding `Use Custom Frames` while in the `Blender` host thought...not sure why but rather noting it here.

Other than that...seems functional in both local/farm rendering...

Image

moonyuet:
> I have tested it with all major 3d hosts and it worked well
> 
> * [x]   Maya 2027
> * [x]   3dsmax 2026
> * [x]   Blender 5.2.1 LTS
> 
> There was a bit suspicious doubling in publisher of the settings regarding `Use Custom Frames` while in the `Blender` host thought...not sure why but rather noting it here.
> 
> Other than that...seems functional in both local/farm rendering...
> 
> Image

You need to test along with https://github.com/ynput/ayon-deadline/pull/252

LiborBatek:
@moonyuet yeah, you are right...while using the correct `ayon-deadline` branch it works and shows just once...which is good!

image


However `Blender` failed to render just those `custom frames` and produce full length sequence when using `local render` however all been working when using farm rendering...

Im providing full publish report while using the `local render` option:

[b3d-local-custom-fr-publish-report-260901-14-27.json.txt](https://github.com/user-attachments/files/31690918/b3d-local-custom-fr-publish-report-260901-14-27.json.txt)

moonyuet:
> However `Blender` failed to render just those `custom frames` and produce full length sequence when using `local render` however all been working when using farm rendering...
> 
> Im providing full publish report while using the `local render` option:
> 
> [b3d-local-custom-fr-publish-report-260901-14-27.json.txt](https://github.com/user-attachments/files/31690918/b3d-local-custom-fr-publish-report-260901-14-27.json.txt)

If we need to make it work for Blender in terms of local rendering, we need to make some edits so that it would collect the expected files by custom frame range.

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -0,0 +1,125 @@
...
+            "render", "render.local", "render.farm", "render.frames_farm",
+            "prerender", "prerender.farm", "prerender.frames_farm",
+            "renderlayer", "imagesequence", "image",
+            "vrayscene", "maxrender",
+            "arnold_rop", "mantra_rop",
+            "karma_rop", "vray_rop", "redshift_rop",
+            "renderFarm", "usdrender", "publish.hou",
+            "remote_publish_on_farm",
+            "deadline"
+    ]
```
iLLiCiTiT, line 25: These should be exactly one family here, the instances that do allow it should add the family to the instances.

Don't know what exactly the family would be, but probably something like `"custom.frame.range"`.

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -0,0 +1,125 @@
...
+        defs.append(
+            TextDef(
+                "frames",
+                label="Custom Frames",
+                default="",
+                tooltip="Explicit frames to be rendered. (1001,1003-1004)(2x)",
+                visible=custom_frames_visible
+            )
+        )
+        return defs
```
iLLiCiTiT:
```suggestion
        use_custom_frames = (
            cls._get_publish_use_custom_frames_value(instance.data) or "none"
        )

        # explicit frames to render - for test renders
        use_custom_frames_enum_values = [
            {"value": "none", "label": "Disabled"},
            {"value": "custom_only", "label": "Custom Frames Only"},
            {"value": "reuse_last_version", "label": "Reuse from Last Version"}
        ]
        custom_frames_visible = cls._is_custom_frames_used(use_custom_frames)
        return [
            EnumDef(
                "use_custom_frames",
                label="Use Custom Frames",
                default=use_custom_frames,
                items=use_custom_frames_enum_values,
            ),
            TextDef(
                "frames",
                label="Custom Frames",
                default="",
                tooltip="Explicit frames to be rendered. (1001,1003-1004)(2x)",
                visible=custom_frames_visible
            ),
        ]
```

client/ayon_core/plugins/publish/collect_custom_frame_range.py
iLLiCiTiT: Made a PR that is proposing how it could work https://github.com/ynput/ayon-core/pull/2040 . I didn't validate it...

PR #1823, opened 2026-05-12
Update speedcopy dependency version to 2.2

antirotor:
## Changelog Description
Bump version of [speedcopy](https://github.com/antirotor/speedcopy)

## Additional info
New version of speedcopy fixes some macos related issues (crashes in integrator).

## Testing notes:
Test asset integration on all platforms (publish)

> [!WARNING]
> You need to build new dependency package

### DIFF
--- a/client/ayon_core/lib/file_transaction.py
+++ b/client/ayon_core/lib/file_transaction.py
@@ -10,6 +10,7 @@ from typing import List, Optional
 from ayon_core.lib import create_hard_link
 
 import speedcopy
+import speedcopy.version
 
 _IS_MACOS = platform.system().lower() == "darwin"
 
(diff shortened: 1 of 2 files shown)


BigRoy (approved):
LGTM - would be good for @jakubjezek001 or @rdelillo confirm the Mac OS side.
@LiborBatek perhaps you can do a windows test run?

iLLiCiTiT:
> LGTM - would be good for @jakubjezek001 or @rdelillo confirm the Mac OS side.

Speedcopy usage is disabled on mac at this moment, you'd have to remove the condition to test it (`ayon_core/lib/file_transaction.py`).

BigRoy:
> Speedcopy usage is disabled on mac at this moment, you'd have to remove the condition to test it

Very good point - so I guess we should be enabling it again then with this PR as well.

iLLiCiTiT:
> Very good point - so I guess we should be enabling it again then with this PR as well.

Not sure if it's good idea. It should bump minor version in that case.

BigRoy:
> Not sure if it's good idea. It should bump minor version in that case.

The hardest bit is that we have little checks to ensure a dependency package _actually matches_ your current addon dependencies. Because then we could let the admin know that dependency package would need updating.

BigRoy:
@iLLiCiTiT how about like this? (That way accidental older dependency packages don't get in the way)

iLLiCiTiT (approved):

PR #1836, opened 2026-05-14
Supports single frame publish for custom frame

moonyuet:
## Changelog Description
This PR is to support single frame publish for custom frame.
Resolve https://github.com/ynput/ayon-core/issues/1835

## Additional info
Make it draft as not sure if it is the best solution.

## Testing notes:
1. Launch your host to publish render with custom frames
2. Publish

PR #1839, opened 2026-05-18
Allow integrator to take repre files with a single file as string instead of list

### LINKED ISSUE #1777: YN-0632: Core: Allow integrator to take repre files with single file as a List

Allow Integrator to take a representation files with a single file in a list to allow integration single frame "sequences" instead of requiring single frames to only be supported as str (edited)Wednesday, April 1, 2026 13:35

<br>

Connected with:

<br>

[YN-0631](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=5b1e39a02e6811f1af7617d58b7cba9a)

<br> 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0632]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=22f860302e6a11f1b898c1825cf2e3a5)

moonyuet:
## Changelog Description
This PR is to implement the `get_collections` function and replace `clique.assemble` with that 
Fix https://github.com/ynput/ayon-core/pull/1836, https://github.com/ynput/ayon-core/issues/1855

## Additional info
The following product is adjusted for the testing:
Texture product from Substance Painter
Review product from Blender
Render product in general
Test this PR with the branch `chore/testing_representation_files_with_list_of_string` from Substance Painter : https://github.com/ynput/ayon-substance-painter/tree/chore/testing_representation_files_with_list_of_string
and the branch `chore/testing_repre_file_with_list_of_files` from Blender: https://github.com/ynput/ayon-blender/tree/chore/testing_repre_file_with_list_of_files

## Testing notes:

- Publishing single frame should work
- Publishing multiple frames should work
- Publishing single frame with integrations that store single frame as `list: str = ["frame.1001.exr"]` should work - would require dependent PRs in other integrations to check with.

With the last case, validate after that the published file has indeed been named with the UDIM and frame number in it instead of turned into a single file without frame/UDIM.

1. Create Render
2. Publish both local/farm rendering


iLLiCiTiT:
I do remember we were discussing this and the result was that we should not allow this and we should fix the source of the issue instead. Or am I wrong? Do you remember @BigRoy ?

BigRoy:
> I do remember we were discussing this and the result was that we should not allow this and we should fix the source of the issue instead. Or am I wrong? Do you remember @BigRoy ?

You are not wrong - **but** it was for the initial hotfix for a particular case which was fixed here: https://github.com/ynput/ayon-core/pull/1836

However, long-term I believe we do want to 'simplify' things if we can. The question was more, that if we provide a single file in a list, whether we should still assume it to be a sequence (or 1 frame?) by the fact of it being a list or should we enforce it to be a frame-less file (as this PR would make it do; but that's also what the `list[str]` -> `str` on `representation.files` would already be doing.) So it wouldn't technically differ currently.

Anyway, we held off from allowing this fix initially due to the worry of edge cases it may somehow affect. But if we can reasonable avoid having to make it a `str` and perhaps even rely on it always being a `list[str]` it'd make the variable `repre["files"]` during publishing much more consistent.

This 'issue' though describes that single frame in `list[str]` should be a single frame sequence - which does mean that the fix done here is incorrect.

Anyway, main question is:

- is there still reason for having `files: str = "singlefile"` and `files: list[str] = ["single_frame.exr"]"?
- can we actually safely detect always the frame number from a single file in the list e.g. `sh001.exr` might refer to the shot number, not the frame number?

iLLiCiTiT:
I agree with the questions, but as you said this PR is not handling that and the question is out of scope of this PR (and the source issue) -> this is not the right place to ask the questions.

antirotor:
But this will break single frame sequences, right? I think we should communicate it somehow if that is the solution (even temporary one)

jakubjezek001:
I am not sure if this is related but there might be also case where single frames with slates will need to be published. It might be really rare use case. And it will in fact become two frame sequence.

iLLiCiTiT:
> I am not sure if this is related but there might be also case where single frames with slates will need to be published. It might be really rare use case. And it will in fact become two frame sequence.

Still, with current publish logic we have, you can't do that. And even if we'd do that then the integrate plugin would get list of files. This PR does not solve anything, only allows something what would cause issues in other plugins.

BigRoy:
I think the plan will need to be to support both:

- files: str -> single file, not in a sequence.
- files: list[str] -> single or more files, part of a sequence.

In the case of the latter - it will need to be an identifiable sequence e.g. `hello.1001.exr` or `hello_1001.exr`.
Whenever it's a single frame files list, we should also handle it as a sequence, preserving `frame` token, etc.

Which would make it so that every plug-in that assumes `list[str]` is not a single file currently (for whatever reason) will need to support the case when it does get a single frame sequence.

Auto-converting `files: list[str] = ["hello.1001.exr"]` to `files: str = "hello.1001.exr"` just because it's a single frame is the wrong approach just because it then starts to assume that maintaining that it initially was a frame sequence can be ignored.

In short, we should work towards supporting single frame sequences, with the integrator (and other plug-ins) allowing to support `files: list[str] = ["hello.1001.exr"]`

iLLiCiTiT:
Yep, but again, this PR does not solve that issue, and is far from solving the issue.

BigRoy:
> Yep, but again, this PR does not solve that issue, and is far from solving the issue.

Yup, just mentioning it here to steer either this PR or make a new one.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -615,6 +615,10 @@ def prepare_representation(
 
         # required representation keys
         files = repre["files"]
+        #  a single frame should be str in representation[files]
+        if isinstance(files, (list, tuple)) and len(files) == 1:
+            files: str = files[0]
+
```
BigRoy: With this PR this should not be true anymore.
```suggestion
```

BigRoy:
Note that to review this correctly, ensure that from an integration a single frame for the sequence is stored as
```
repre = {
    ...
    "files": ["frame.1001.exr"] 
}
```
It must be a `list: str`.
Then after validate that the publish template did preserve the frame number, and didn't turn it into a 'single file' naming template.

I'd also recommend to test with substance painter perhaps, with a single frame `UDIM` scenario.

Also - it's **NOT** required for a representation to use `list[str]`. They are still ok to use `str` for single frame, we're just - like before - not handling that as a sequence but always ignore frame/udims, etc.

moonyuet:
> Note that to review this correctly, ensure that from an integration a single frame for the sequence is stored as
> 
> ```
> repre = {
>     ...
>     "files": ["frame.1001.exr"] 
> }
> ```
> 
> It must be a `list: str`. Then after validate that the publish template did preserve the frame number, and didn't turn it into a 'single file' naming template.
> 
> I'd also recommend to test with substance painter perhaps, with a single frame `UDIM` scenario.

I will copy and put this in the testing notes.

moonyuet:
I tested in SP: It publishes the udim successfully(both sequence and single frame)
[publish-report-260806-17-38.json](https://github.com/user-attachments/files/30781669/publish-report-260806-17-38.json)
[publish-report-260806-17-43.json](https://github.com/user-attachments/files/30781871/publish-report-260806-17-43.json)

BigRoy:
> I tested in SP: It publishes the udim successfully(both sequence and single frame) [publish-report-260806-17-38.json](https://github.com/user-attachments/files/30781669/publish-report-260806-17-38.json) [publish-report-260806-17-43.json](https://github.com/user-attachments/files/30781871/publish-report-260806-17-43.json)

Can you try some that also go through Extract Review, OIIO transcode, etc. (possibly other apps). Glancing over the reports these do look fine. 🤔 

We may want the integrator to start logging "Handling single file" and "Handling sequence" just so we can keep track which route it ended up taking? 🤔 to confirm it does what we suspect.

moonyuet:
@BigRoy I tested with Review product in 3dsmax. Looks good.
[publish-report-260806-19-55.json](https://github.com/user-attachments/files/30786591/publish-report-260806-19-55.json)
I also tested with several scanerios in traypublisher. No error spotted so far.
[publish-report-260806-20-42.json](https://github.com/user-attachments/files/30788611/publish-report-260806-20-42.json)
image

BigRoy:
That took surprisingly little. @iLLiCiTiT thoughts?

@LiborBatek can you go crazy and try to break it? @moonyuet could you provide Libor some draft branches that he can uyse to test integrations that would then set `list[str]` with single frame?

moonyuet:
@LiborBatek test this PR with the branch `chore/testing_representation_files_with_list_of_string` from Substance Painter and the branch `chore/testing_repre_file_with_list_of_files` from Blender,
You need to test single/multiple udims publishing in SP while testing to publish review product from Blender

As @BigRoy and I discussed regarding to the PR, you need to also test:
- farm submission
- something that is frames, not udim (because that's stored differently on the instance)
- something that goes through extract review, burnin and oiio transcodes.

iLLiCiTiT:
I would be more specific about the function name and location, it should be used only for repre files. We did discuss that the patter will be used only if there is single file, we can't guarantee the format.

BigRoy:
> [I would be more specific about the function name and location, it should be used only for repre files. We did discuss that the patter will be used only if there is single file, we can't guarantee the format](https://github.com/ynput/ayon-core/pull/1839#issuecomment-5207683762)

...

> commit: [make sure the get collections only used within the integrate.py and rename the get_collections to _get_collections_for_repre_files](https://github.com/ynput/ayon-core/pull/1839/commits/a450e65b2d25116a3607c89f5e380c9a13f5ff2a)

I don't think this is the intent here @iLLiCiTiT - we are providing an example on how to best get a single collection through assemble from a list of files, where you KNOW pretty well that it's a sequence.

For that we'll need to fix anything that does `clique.assemble` to begin with.

The reason I'd oppose losing the `patterns` completely is this issue: https://github.com/ynput/ayon-core/issues/1855
That's a massive slow-down and issue on long sequences.

I think we should make it clear enough what the function does and what the intent is (docstring and function name) but it should be a library function, and preferred over using clique directly when we can. So we support the same file names etc everywhere without doubt, and if we need to support more we need to change one place.

To avoid issue https://github.com/ynput/ayon-core/issues/1855 I think the best we can do is:

```
collections, remainder = clique.assemble(patterns=[...])
if not collections and len(files) > 1:
    # Allow any file pattern and find the incrementing digit in the filename
    clique.assemble()
```

EVEN in that latter fallback case, I'd argue that we should be passing a pattern if clique allows that where it does NEVER check for patterns in parent folder names. Because currently passing a path like `/path/to/sq010/sh010/hero01/myfile.1001.exr` will be mad slow for 1000 frames because it will also for all frames check (very slowly somehow) whether those are incrementing digits as well, and considers all permutations. We must make it so that it only considers the files (not the folders) for assembling, because that's our use case here.

If we can't solve that, we must then complicate it more by doing clique assemble on the filenames, and then joining them with the parent directory, etc.

(note that `repre["files"]` SHOULD be filenames to begin with, but like I said - this function is intended to be the thing that efficiently gets you the sequence, just because clique by default is shit.

iLLiCiTiT:
That's why I think we should make it representations specific to avoid the issue of handling file paths and potentially alle to handle more sophisticated file objects in future. But ok.

Still the function name should be more specific, at least use `file` in the name (`get_file_collections`).

moonyuet:
> That's why I think we should make it representations specific to avoid the issue of handling file paths and potentially alle to handle more sophisticated file objects in future. But ok.
> 
> Still the function name should be more specific, at least use `file` in the name (`get_file_collections`).

rename the function to `get_file_collections`: [11ae8a0](https://github.com/ynput/ayon-core/pull/1839/commits/11ae8a05ae949a3e3d5d4490d6272350a3df0f6a)

BigRoy:
Thanks - after some brief more info from @iLLiCiTiT I think we should indeed allow the fallback pattern that if no collection is found from the `assemble` call with the `patterns` to call it once more without patterns.

Then I think we're ok. Keeping it private to the Integrator is fine for now if it requires no changes in other areas of the code to make this work just fine. Because solving this more globally will be done later with what @iLLiCiTiT described before where we keep track of the collection (with frame range pattern, etc.) along the way instead of having to recollect in integrator.

moonyuet:
> Thanks - after some brief more info from @iLLiCiTiT I think we should indeed allow the fallback pattern that if no collection is found from the `assemble` call with the `patterns` to call it once more without patterns.
> 
> Then I think we're ok. Keeping it private to the Integrator is fine for now if it requires no changes in other areas of the code to make this work just fine. Because solving this more globally will be done later with what @iLLiCiTiT described before where we keep track of the collection (with frame range pattern, etc.) along the way instead of having to recollect in integrator.

I added the fallback here: [236190b](https://github.com/ynput/ayon-core/pull/1839/commits/236190bfce3a24fad718570c42f5bfb91055de86)

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -1126,3 +1126,37 @@ def _get_attributes_by_type(self, context):
...
+        collections, remainders = clique.assemble(
+            files,
+            minimum_items=minimum_items,
+            assume_padded_when_ambiguous=True,
+            patterns=[pattern]
+        )
+
+        # If custom pattern yields no collections,
+        # retry default clique parsing.
+        if not collections:
```
BigRoy: This should only apply if `len(files) > 1` because we do want the strict case only for single files.


```suggestion
        if not collections and len(files) > 1:
```

BigRoy (commented):
@moonyuet can you still add the log on whether it handles a sequence or single file as described [before in DMs](https://discord.com/channels/@me/1084823103051411476/1534909658924454081). We'll need that to evaluate whether the test run logs actually are doing what we'd expect.

Other than that, code looks ok, but needs the test scenarios prepared because it requires changes in DCC integrations to ensure they actually take the single frame sequence into the "handle as sequence" case where it is `list[str]`. So please link the relevant branches on other integrations to test with in the testing notes, so that @LiborBatek and possibly others can test it.

moonyuet:
> @moonyuet can you still add the log on whether it handles a sequence or single file as described [before in DMs](https://discord.com/channels/@me/1084823103051411476/1534909658924454081). We'll need that to evaluate whether the test run logs actually are doing what we'd expect.
> 
> Other than that, code looks ok, but needs the test scenarios prepared because it requires changes in DCC integrations to ensure they actually take the single frame sequence into the "handle as sequence" case where it is `list[str]`. So please link the relevant branches on other integrations to test with in the testing notes, so that @LiborBatek and possibly others can test it.

Add log message: [a99feb0](https://github.com/ynput/ayon-core/pull/1839/commits/a99feb08ab33df7b62ed6a2e82ba84e9e56664e7)

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -576,7 +576,7 @@ def _validate_repre_files(self, files, is_sequence_representation):
         if not is_sequence_representation:
             return
 
-        src_collections, remainders = clique.assemble(files)
+        src_collections, remainders = self.get_file_collections(files)
         if len(files) < 2 or len(src_collections) != 1 or remainders:
```
BigRoy: This logic here actually still seems to be raising an error if it were `list[str]` with a single item. So I'm pretty sure this PR doesn't work yet, because with this PR we're supporting a sequence representation but with a single file... so pretty sure this PR hasn't been tested yet.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -1126,3 +1126,40 @@ def _get_attributes_by_type(self, context):
...
+        Returns:
+            tuple[list[clique.Collection], list[str]]: tuple of collections
+            and remainders from clique.
+        """
+        pattern = "(?P<index>(?P<padding>0*)\\d+)\\.\\D+\\d?$"
+        minimum_items = 1 if len(files) == 1 else 2
+        if minimum_items > 1:
+            self.log.debug(f"Handling single file: {files}")
+        else:
+            self.log.debug(f"Handling sequence files: {files}")
```
BigRoy: This log is in the wrong place - once we call `get_file_collections` we are already handling a scenario where it MUST be a sequence, even if it has a single frame.

This log should be here: https://github.com/ynput/ayon-core/pull/1839/changes#diff-181b1a0d60066e0e3ab6db7a72209332064bb6254e56da721b1d0f0b84adb909R690 telling us whether we're handling it as a sequence representation or not (and hence whether it was set as `list[str]` vs `str` on the `repre["files"]`.

BigRoy (changes requested):
Pretty sure this isn't working as intended. See comments.
Put the log in place in the right spot and I'd love to see a publish report that says it is a sequence representation that correctly handles a sequence representation which is actually a single frame.

LiborBatek:
I have tested this PR with Substance Painter while trying to produce UDIMs with multiple tiles / single tile...

While I was able to produce publish of texture set with multiple UDIMs (1001 - 1004) I was not able to publish texture set with UDIM containing just single tile (1001)

And getting this error:

```
  File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2606031445_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
    ~~~~~~^^^^^^^
  File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 173, in process
    self.main_process(instance)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 407, in main_process
    self._render_output_definitions(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        instance,
        ^^^^^^^^^
    ...<3 lines>...
        layer_name
        ^^^^^^^^^^
    )
    ^
  File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 482, in _render_output_definitions
    temp_data = self.prepare_temp_data(instance, repre, output_def)
  File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 710, in prepare_temp_data
    input_frames = list(sorted(cols[0].indexes))
                               ~~~~^^^
IndexError: list index out of range
```

Here Im enclosing full publish json report:

[SP-publish-report-260808-18-09.json.txt](https://github.com/user-attachments/files/30858760/SP-publish-report-260808-18-09.json.txt)

moonyuet:
> I have tested this PR with Substance Painter while trying to produce UDIMs with multiple tiles / single tile...
> 
> While I was able to produce publish of texture set with multiple UDIMs (1001 - 1004) I was not able to publish texture set with UDIM containing just single tile (1001)
> 
> And getting this error:
> 
> ```
>   File "C:\Users\lbate\AppData\Local\Ynput\AYON\dependency_packages\ayon_2606031445_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
>     runner(*args)
>     ~~~~~~^^^^^^^
>   File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 173, in process
>     self.main_process(instance)
>     ~~~~~~~~~~~~~~~~~^^^^^^^^^^
>   File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 407, in main_process
>     self._render_output_definitions(
>     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
>         instance,
>         ^^^^^^^^^
>     ...<3 lines>...
>         layer_name
>         ^^^^^^^^^^
>     )
>     ^
>   File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 482, in _render_output_definitions
>     temp_data = self.prepare_temp_data(instance, repre, output_def)
>   File "C:\Work\REPO\ayon-core\client\ayon_core\plugins\publish\extract_review.py", line 710, in prepare_temp_data
>     input_frames = list(sorted(cols[0].indexes))
>                                ~~~~^^^
> IndexError: list index out of range
> ```
> 
> Here Im enclosing full publish json report:
> 
> [SP-publish-report-260808-18-09.json.txt](https://github.com/user-attachments/files/30858760/SP-publish-report-260808-18-09.json.txt)

[4b0af86](https://github.com/ynput/ayon-core/pull/1839/commits/4b0af86f9f655bfe956d085e54c609c0084de8e8) should be fixing the issue.

@BigRoy @iLLiCiTiT where should we store the `get_file_collections` function as it is used in extract review, burnin and integrator?

BigRoy:
> where should we store the `get_file_collections` function as it is used in extract review, burnin and integrator?

Ah, there we go. This is exactly the kind of change I was expecting we'd need.
I'll leave where to put it to @iLLiCiTiT because I know he'll have opinions. :)

iLLiCiTiT:
> Ah, there we go. This is exactly the kind of change I was expecting we'd need.

Ok, my opinion is that the logic is kinda for representations only. I'd say the function should expect only filenames, not filepaths (also changing the `files` to `filenames` in argument), it also does have 2 possible handlings based on how many files are passed in, which is because of representations, I'd put it into `pipeline/publish/lib.py` next to `fill_sequence_gaps_with_previous_version`.

I also do think that we'll add more logic to the function that will be related to representations, if it would be in `lib/*.py` we can't do that.

moonyuet:
> > Ah, there we go. This is exactly the kind of change I was expecting we'd need.
> 
> Ok, my opinion is that the logic is kinda for representations only. I'd say the function should expect only filenames, not filepaths (also changing the `files` to `filenames` in argument), it also does have 2 possible handlings based on how many files are passed in, which is because of representations, I'd put it into `pipeline/publish/lib.py` next to `fill_sequence_gaps_with_previous_version`.
> 

I have moved the function to [3c45ab6](https://github.com/ynput/ayon-core/pull/1839/commits/3c45ab673e65c146e0cb528da3bac75557dd9a0e)

LiborBatek (changes requested):
When tested again in Substance Painter, this time I was able to produce both `textureSet` with multiple `UDIMs` and also when just single `UDIM` used... which is good!

I have also tested rendering in the `Blender` using the custom branch `ayon-blender` for use with this PR as mentioned above.

I have tested 3 different scenarios to be sure and see how it performs:

1) asset render with frame range `1001-1001` as default

2) asset render using only single frame `1005` but having `1001-1005` frame range in the db

3) asset render using full frame range as in db `1001-1005` aka regular image sequence

All three jobs were succesfully render while the first didnt acted as a frame from sequence but acting as single frame, while the third been regular sequence as expected.

Not sure but I guess that the outcome of 1) and 2) scenario been wrong as the result been `single image` and not `sequence` consisting of single frame...

Here are some published files:

01)

Image

02)

Image

03)

Image

@BigRoy  @moonyuet  I can provide any logs for any of these scenarios if needed

moonyuet:
> Image
> @BigRoy @moonyuet I can provide any logs for any of these scenarios if needed

Do we need to ensure there would be single frame in respect to the refactoring? @BigRoy @iLLiCiTiT

BigRoy:
> Do we need to ensure there would be single frame in respect to the refactoring?

If the representation is `files: list[str]` we **must** handle it as a sequence, if it's a single frame. So if that customized `blender` that @LiborBatek mentioned is using `list[str]` even for single frame renders then it should remain `list[str]` and be handled as seuqence as the integrator.

It's up to the DCC integration at that point whether to use `str` (single file) or `list[str]` (sequence of one or more frames).

Where "handle as single file" or "handle as sequence" is [essentially this log difference](https://github.com/ynput/ayon-core/pull/1839/changes#diff-181b1a0d60066e0e3ab6db7a72209332064bb6254e56da721b1d0f0b84adb909R691-R698).

A single frame `list[str]` would still format as a frame number on the formatted anatomy, a `str` would not.

Does that answer it?

moonyuet:
>A single frame list[str] would still format as a frame number on the formatted anatomy, a str would not.

>Does that answer it?

It is actually my bad. It should be more informative on which produce should be tested. I converted the files from the representation data in the review product instead of the render product in Blender. The render product is actually handling what it does, the files from the representation data should be of string type in terms of render product. (I didn't convert the data type to string) But at least we know how it looks when it is with string 

@LiborBatek can you try to publish the review product in Blender and see? The files from the representation for the review has been converted to the list instead of string.

tests/client/ayon_core/lib/test_file_handling.py
```diff
@@ -0,0 +1,55 @@
...
+            minimum_items=2,
+            assume_padded_when_ambiguous=True,
+            patterns=["(?P<index>(?P<padding>0*)\\d+)\\.\\D+\\d?$"],
+        ),
+        call(
+            files,
+            minimum_items=2,
+            assume_padded_when_ambiguous=True,
+        ),
+    ]
```
BigRoy: I don't understand this test at all - overly convoluted to 'hack' some odd test?

Shouldn't we just be testing simple cases like:

```
sh010.exr (should be disallowed)
sh010.exr + sh011.exr (should be allowed)
frame1001.exr + frame1002.exr
frame_1001.exr + frame_1002.exr
frame.1001.exr + frame.1002.exr
1000.exr + 1002.exr (frames only)
1.exr (single frame; digit only)
1.exr + 10.exr (no padding)
etc.
```

antirotor: It patches the inside to test directly the condition *if custom pattern yields no collection, retry default clique parsing*. In that sense it is testing exactly that. But I agree that we should aslo add more specific tests.

antirotor: btw it is now failing on:

> sh010.exr (should be disallowed)

*(see the updated tests)*

BigRoy: > It patches the inside to test directly the condition if custom pattern yields no collection, retry default clique parsing. In that sense it is testing exactly that. But I agree that we should aslo add more specific tests.

I don't think we care much about how the function works internally, the tests should test what we expect out of it. The tests shouldn't need to enforce that it uses `clique` for example. What does testing this actually help with? I'm a bit unsure why we need to test this particular case so explicitly? I'm probably missing something 🤔

tests/client/ayon_core/lib/test_file_handling.py
```diff
@@ -53,3 +54,81 @@ def test_get_file_collections_retries_default_pattern() -> None:
...
+            [[1]],
+            [],
+            id="single-digit-frame",
+        ),
+        pytest.param(
+            ["1.exr", "10.exr"],
+            [[1, 10]],
+            [],
+            id="unpadded-frames-form-a-sequence",
+        ),
```
BigRoy, line 79: I think this lacks tests for the single frame case for `frame.1001.exr` and `frame_1001.exr` which should both be allowed.

tests/client/ayon_core/lib/test_file_handling.py
antirotor: I can ofc remove that test if it offends but if we state in implementation comment some expected behaviour, we should IMO test if it is so.

LiborBatek (commented):
I was able to publish successfully the `Review` of just one frame `1001-1001` not sure about the outcome tho, but Im enclosing the publish report for your inspection.

Seems like the resulting files are just stills and not mp4 obviously...

Image

[review-publish-report-260812-17-02.json.txt](https://github.com/user-attachments/files/30987128/review-publish-report-260812-17-02.json.txt)

BigRoy:
> review-publish-report-260812-17-02.json.txt

Unfortunately this is still going through the single file:

```
DEBUG: Handling single file representation thumbnail: pump_reviewMain.1001.jpg
DEBUG: Handling single file representation png_png: pump_reviewMain_png.png
```

So something is along the way still converting to `str` instead of `list[str]` @moonyuet

moonyuet:
> > review-publish-report-260812-17-02.json.txt
> 
> Unfortunately this is still going through the single file:
> 
> ```
> DEBUG: Handling single file representation thumbnail: pump_reviewMain.1001.jpg
> DEBUG: Handling single file representation png_png: pump_reviewMain_png.png
> ```
> 
> So something is along the way still converting to `str` instead of `list[str]` @moonyuet

I updated the commit for using `list[str]` in https://github.com/ynput/ayon-blender/tree/chore/testing_repre_file_with_list_of_files
Updated [26a5af6](https://github.com/ynput/ayon-core/pull/1839/commits/26a5af6b9c77dac07088021e109dc9fde79497c9) in this PR for using list[str] in terms of new representation of single frame sequence for extract review.

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1241,7 +1238,7 @@ def input_output_paths(self, new_repre, output_def, temp_data: TempData):
         output_ext_is_image = bool(output_ext in self.image_exts)
         output_is_sequence = bool(
             output_ext_is_image
-            and "sequence" in output_def["tags"]
+            and temp_data.output_frame_end >= temp_data.output_frame_start
```
BigRoy: Wouldn't this also enforce marking single frames as `list[str]` ALWAYS.
Doesn't this cause the reverse problem where something that was `str` now becomes `list[str]` @moonyuet ?

moonyuet: Yes it is enforcing as always for now. I am still figuring out how to get it right from the old_repre["files"]. https://github.com/ynput/ayon-core/pull/1839#issuecomment-5271184051
With the wip commit, you would get the list of string for representation in review
```
DEBUG: Handling sequence representation thumbnail: ['lib_reviewSusanna.1001.jpg']
DEBUG: Template filled: D:/sh_boss_proj/interior_lenny//lib/publish/review/reviewSusanna/v012/il_lib_reviewSusanna_v012.1001.jpg
DEBUG: Handling sequence representation png_png: ['lib_reviewSusanna_png.1001.png']
DEBUG: Template filled: D:/sh_boss_proj/interior_lenny//lib/publish/review/reviewSusanna/v012/il_lib_reviewSusanna_v012_png.1001.png
DEBUG: Executing [POST] http://localhost:5000/api/projects/interior_lenny/operations/background
DEBUG: http
```
For the mp4 burnin, you would get the string type for representation
```
DEBUG: Creating new version ...
DEBUG: Prepared version: v013
DEBUG: http://localhost:5000 "POST /graphql HTTP/1.1" 200 112
DEBUG: Response <RestApiResponse [200]>
DEBUG: Handling sequence representation thumbnail: ['lib_reviewSusanna.1001.jpg']
DEBUG: Template filled: D:/sh_boss_proj/interior_lenny//lib/publish/review/reviewSusanna/v013/il_lib_reviewSusanna_v013.1001.jpg
DEBUG: Handling single file representation h264_png: lib_reviewSusanna_h264burnin.mp4
DEBUG: Executing [POST] http://localhost:5000/api/projects/interior_lenny/operations/background
DEBUG: http://localhost:5000 "POST /api/projects/interior_lenny/operations/background HTTP/1.1" 200 90
```

BigRoy:
```
is_sequence = isinstance(old_repre["files"], (list, tuple))
```
Similar to how integrator would do it?

BigRoy: Also, not sure what the `sequence` tags were for in output definitions, but also doesn't like something we can just suddenly kill. There must be reasons for having it. @jakubjezek001 @iLLiCiTiT ?

BigRoy: But since we are converting types (e.g. sequence to single mp4 or vice versa) we can't just copy from old repre. But if both are frame range based (e.g. both are image sequences), the input has `list[str]` I think the output can remain `list[str]` but if image, has one file and previously was `str` we'd likely want to preserve `str` on the output too.

moonyuet: > But since we are converting types (e.g. sequence to single mp4 or vice versa) we can't just copy from old repre. But if both are frame range based (e.g. both are image sequences), the input has `list[str]` I think the output can remain `list[str]` but if image, has one file and previously was `str` we'd likely want to preserve `str` on the output too.

commit: [64e8059](https://github.com/ynput/ayon-core/pull/1839/commits/64e8059f11e2502a4f691fffb3fa83b34eaae822)

BigRoy: Logic is wrong currently.

This wouldn't work if the source file was a video file (hence `representation["files"] = "singlefile.mp4"`) but the output is rendering 20 frames from it here. Because it would not pass the `self.input_is_sequence(repre)` check.

I'd actually turn this around:
```py
if (
  # If this is a video file
  not output_ext_is_image
  # Or if it's a single frame only and the
  # source was also just a `str` (single file)
  # then handle it as single frame
  or (
      is_single_frame and not self.input_is_sequence(repre)
  )
):
    handle_single_file
else:
    handle_sequence
```

Essentially:
- If the output is a video file (then it must be a single frame and hence we always deal with it as single file)
- If the output an image, but it's a single frame and the input was also not marked as sequence, then handle it as single file.

iLLiCiTiT: So, fisrt of all, I don't understand why would you have to add `"sequence"` tag to render image sequence, I think we can ged rid of that?

With that in mind:
If output is sequence or not is not related to if input is sequence, you can have input video and you might want to create image sequence out of it, so the `self.input_is_sequence(repre)` should not be there.

Possible combinations
```
|- input is video (str)
    | - output is video -> str
    | - output is image sequence -> list[str]
|- input is image (str)
    | - output is video -> str
    | - output is image -> str
|- input is image sequence (list[str])
    | - output is video -> str
    | - output is image -> list[str]
```

BigRoy: Yup, that makes sense. Thanks!
Which essentially boils down to:

```
if output is video:
    return output is video -> str
else:
    if input is image (str):
        return output is image -> str
    else:
        return output is image -> list[str]
```

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -722,8 +721,11 @@ def prepare_temp_data(self, instance, repre, output_def) -> TempData:
             ext = os.path.splitext(repre["files"][0])[1].replace(".", "")
             if ext.lower() in self.alpha_exts:
                 input_allow_bg = True
+
+            input_is_single_image = ext.lower() in self.video_exts
```
BigRoy: This is quite a confusing variable. If the extension is a video, then it's a single image? 🗡️ Maybe `single_file`? 🤔 

Anyway, I can't find @iLLiCiTiT 's logic from [here](https://github.com/ynput/ayon-core/pull/1839#discussion_r3782432293) in this yet.

moonyuet: Implemented [e4d516d](https://github.com/ynput/ayon-core/pull/1839/commits/e4d516d68da5911d64d1d3620b9171cdd29095e4)

LiborBatek (commented):
Now when producing `Review` with just single frame `1001` it seems ok to me as it still reads as single frame of the sequence when inspecting the published version.

Image

Im also enclosing the full publish report so you can inspect whether its been correct or not yet:

[Blender-Review-publish-report-260824-15-49.json.txt](https://github.com/user-attachments/files/31378003/Blender-Review-publish-report-260824-15-49.json.txt)

moonyuet:
> Now when producing `Review` with just single frame `1001` it seems ok to me as it still reads as single frame of the sequence when inspecting the published version.
> 
> Image
> Im also enclosing the full publish report so you can inspect whether its been correct or not yet:
> 
> [Blender-Review-publish-report-260824-15-49.json.txt](https://github.com/user-attachments/files/31378003/Blender-Review-publish-report-260824-15-49.json.txt)

@BigRoy the case above is handling the single frame range in sequence representation now as you can see the enclosed publish report.

LiborBatek (commented):
I have tested it with both single frame `1001-1001` and sequence `1001-1005`

In both case I have succesfully published and got published products. 

Im enclosing the resulting published files for your consideration + respective publish json report for each case (single / sequence of frames).

I had used the `ayon-blender` branch mentioned at the testing steps together with this `ayon-core` branch..

[B3d-single-fr-publish-report-260910-16-04.json.txt](https://github.com/user-attachments/files/32062652/B3d-single-fr-publish-report-260910-16-04.json.txt)

and resulting files in the publish folder:

Image

while when submitting as sequence (1001-1005):

[b3d-seq-publish-report-260910-16-11.json.txt](https://github.com/user-attachments/files/32062681/b3d-seq-publish-report-260910-16-11.json.txt)

and the resulting files:

Image


I will just comment as Im not super sure what we are looking exactly for...but at first glance all seems working and no any errors experienced.

BigRoy (approved):
I was still getting this from Blender with review and single frame on branch `chore/testing_repre_file_with_list_of_files`:

```
2026/09/14 11:24:05 DEBUG: Handling sequence representation thumbnail: ['blender_reviewMain.1001.jpg']
2026/09/14 11:24:05 DEBUG: Template filled: C:/projects/test/apps/blender/publish/review/reviewMain/v008/tes_blender_reviewMain_v008.1001.jpg
2026/09/14 11:24:05 DEBUG: Handling sequence representation png_png: ['blender_reviewMain_png.1001.png']
2026/09/14 11:24:05 DEBUG: Template filled: C:/projects/test/apps/blender/publish/review/reviewMain/v008/tes_blender_reviewMain_v008_png.1001.png
```
Meaning, that it's handling these as sequence instead of single frame - which would still be wrong.

Anyway, if I fixed that in the blender branch on my end - then it does work as this PR intends it to.
Then both the review + render (on farm) work as intended.

BigRoy:
@iLLiCiTiT could you review please?

BigRoy:
@moonyuet your added test is failing. Can you fix please?

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -1414,6 +1414,40 @@ def get_trait_representations(
...
+        tuple[list[clique.Collection], list[str]]: tuple of collections
+        and remainders from clique.
+    """
+    pattern = "(?P<index>(?P<padding>0*)\\d+)\\.\\D+\\d?$"
+    minimum_items = 1 if len(files) == 1 else 2
+    collections, remainders = clique.assemble(
+        files,
+        minimum_items=minimum_items,
+        assume_padded_when_ambiguous=True,
+        patterns=[pattern]
```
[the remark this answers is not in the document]
BigRoy: @moonyuet can you see how viable it is to patch this one in this PR too please. :)

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -688,6 +688,14 @@ def prepare_representation(
             template_data["originalDirname"] = without_root
 
         is_sequence_representation = isinstance(files, (list, tuple))
+        if is_sequence_representation:
+            self.log.debug(
+                f"Handling sequence representation {repre['name']}: {files}"
+            )
```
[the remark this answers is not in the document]
BigRoy: I think this is a valid point - no need to clutter the logs here.

client/ayon_core/pipeline/publish/lib.py
```diff
@@ -1414,6 +1414,40 @@ def get_trait_representations(
...
+
+    Args:
+        files (list[str]): list of sequence file paths
+
+    Returns:
+        tuple[list[clique.Collection], list[str]]: tuple of collections
+        and remainders from clique.
+    """
+    pattern = "(?P<index>(?P<padding>0*)\\d+)\\.\\D+\\d?$"
+    minimum_items = 1 if len(files) == 1 else 2
```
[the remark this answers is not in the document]
BigRoy: valid point actually here @moonyuet - the single file case should still not allow `sh010.exr` to be a "collection", even if it's the only file, right?

BigRoy:
Revoking approval until the copilot issues are resolved, they seem accurate.

tests/client/ayon_core/lib/test_file_handling.py
```diff
@@ -0,0 +1,116 @@
+"""Tests for publish library helpers."""
```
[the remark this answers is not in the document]
moonyuet: [1e07201](https://github.com/ynput/ayon-core/pull/1839/commits/1e07201a4220641e14e4677fa9d41c27f9b10931)

client/ayon_core/pipeline/publish/lib.py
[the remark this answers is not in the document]
moonyuet: [68d0a12](https://github.com/ynput/ayon-core/pull/1839/commits/68d0a122055fe4236d4f6cad29a314e8ce179676)

client/ayon_core/pipeline/publish/lib.py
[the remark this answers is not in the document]
moonyuet: [68d0a12](https://github.com/ynput/ayon-core/pull/1839/commits/68d0a122055fe4236d4f6cad29a314e8ce179676)

antirotor (approved):
Ran a quick test in blender, worked well.

PR #1842, opened 2026-05-19
Added Copy file path action

### LINKED ISSUE #2036: YN-0725: Copy Representation File Path via context menu

**Summary**

Add a frontend-only way to copy a representation file path from the UI via right-click context menu.

**Problem**

The current implementation was started as a backend/action-driven flow, but it was concluded that this can be handled on the frontend instead. A dialog-based approach was explored earlier, but the preferred UX is a context menu on the representation itself.

**Proposed UX**

When a user right-clicks a representation, show a **Copy path** submenu with:

-   **Copy Windows path**
    
-   **Copy Mac path**
    
-   **Copy Linux path**

**Technical implementation: constructing the representation path**

-   Read the representation’s **first file entry** and take its **rootless path**.
    
-   Determine which **root token** the representation/template uses (for example, `work` or another configured root key).
    
-   From the representation/anatomy data, look up the OS-specific root values for that same root token:
    
    -   Windows root
        
    -   Linux root
        
    -   Mac root
        
-   Construct the final path by prepending/replacing with the selected OS root value against the rootless path.
    
-   Expose three copy actions in the context menu:
    
    -   Copy Windows path
        
    -   Copy Linux path
        
    -   Copy Mac path
        

**Notes**

-   Do not use the published full path directly, because that may reflect only the publish platform path; use the **rootless** form as the base.
    
-   Do not include site overrides in this implementation; only generic OS variants are needed.
    
-   This should be doable entirely on the frontend.

### Additional context:

_This issue was automatically created from AYON ticket [YN-0725]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=d88d450fa6e740d7ac85aff3db9267af)

kalisp:
## Changelog Description
Adds Action on Version (`Products` tab) to copy file path of chosen representation and platform to clipboard.

Used for sharing paths, mimics similar plugin in Loader.

## Testing notes:
1. select product in `Products` tab
2. click on action and select platform and representation
Screenshot 2026-05-19 142142

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -1,11 +1,16 @@
-from typing import Any
+from typing import Any, Dict
 
+from api.resolve import StringTemplate
 from ayon_server.addons import BaseServerAddon
 from ayon_server.actions import (
     ActionExecutor,
     ExecuteResponseModel,
     SimpleActionManifest,
 )
+from ayon_server.forms.simple_form import SimpleForm, FormSelectOption
+from ayon_server.helpers.roots import get_roots_for_projects
+from ayon_server.lib.postgres import Postgres
+
 try:
     from ayon_server.logging import logger
 except ImportError:
@@ -60,6 +65,20 @@ class CoreAddon(BaseServerAddon):
                     allow_multiselection=False,
                 )
             )
+            output.append(
+                SimpleActionManifest(
+                    identifier="core.copyfilepath",
+                    label="Copy File Path",
+                    icon={
+                        "type": "material-symbols",
+                        "name": "content_copy",
+                    },
+                    order=100,
+                    entity_type="version",
+                    entity_subtypes=None,
+                    allow_multiselection=False,
+                )
+            )
 
         return output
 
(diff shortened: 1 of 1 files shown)


BigRoy (commented):
@mkolar @Innders this may need actually some design decision.

So, this does work and I quite like having access to it. But I'm not too fond of needing to go through that form after triggering that action on a version.

Image

I'd rather have quicker access to it.
And also, have it support copying multiple paths at once for multiple selected versions perhaps? So you get equivalent to:
```
/path/to/a
/path/to/b
etc.
```

I'm thinking - how feasible would it be to show a dialog that would be equivalent to:

Image

But with essentially a field that is for:
- Windows
- Mac
- Linux

---

Or an alternative would be to put it into the "Share" dialog completely (but I'm worried it's just overcomplicating that dialog)

Or another option is putting it on the representations themselves under context menu:

Image

---

The only reason we went for action originally was that it could go through launcher to resolve your platform's site, etc. but as @kalisp did point out - you most likely want to share the path for someone else, not necessarily your own exact filepath?

Also, having it as action might be a way to get also disallow/disable the functionality - for if you actually don't want to be giving quick and easy access to the publish filepaths?

server/__init__.py
```diff
@@ -93,9 +112,96 @@ async def execute_action(
...
+            value=representation_options[0]["value"],
+        )
+
+        return await executor.get_form_response("Select representation", form)
+
+    async def copy_file_path_action(
+        self,
+        executor: ActionExecutor,
+    ) -> ExecuteResponseModel:
+        form_data: Dict[str, Any] = executor.context.form_data or {}
```
iLLiCiTiT, line 184:
```suggestion
        form_data: dict[str, Any] = executor.context.form_data or {}
```

server/__init__.py
```diff
@@ -1,11 +1,16 @@
-from typing import Any
+from typing import Any, Dict
```
iLLiCiTiT, line 1:
```suggestion
from typing import Any
```

server/__init__.py
```diff
@@ -1,11 +1,16 @@
...
 
+from api.resolve import StringTemplate
```
iLLiCiTiT, line 3: I don't think there is `StringTemplate` on server.
```suggestion
```

server/__init__.py
```diff
@@ -93,9 +112,96 @@ async def execute_action(
...
+                site_id=None,
+                projects=[project_name],
+                platform=form_data["platform"],
+            )
+
+            context = {"root": roots[project_name]}
+            first_rootless_path = form_data["first_rootless_path"]
+            resolved_path = StringTemplate.format_template(
+                first_rootless_path, context
+            )
```
iLLiCiTiT, line 199:
```suggestion
            first_rootless_path = form_data["first_rootless_path"]
            resolved_path = first_rootless_path.format(
                root=roots[project_name],
            )
```

server/__init__.py
```diff
@@ -93,9 +112,96 @@ async def execute_action(
...
+    ) -> ExecuteResponseModel:
+        """Get initial form for Copy File Path action."""
+        form = SimpleForm()
+
+        platform_options: list[FormSelectOption] = []
+        platforms = {"windows": "Windows", "linux": "Linux", "darwin": "OSX"}
+        for value, label in platforms.items():
+            platform_options.append(
+                FormSelectOption(value=value, label=label)
+            )
```
iLLiCiTiT, line 137:
```suggestion
        platform_options: list[FormSelectOption] = [
            FormSelectOption(value="windows", label="Windows"),
            FormSelectOption(value="linux", label="Linux"),
            FormSelectOption(value="darwin", label="macOs"),
        ]
```

server/__init__.py
```diff
@@ -93,9 +112,96 @@ async def execute_action(
...
+            label="Select OS",
+            options=platform_options,
+            value=platform_options[0]["value"],
+        )
+
+        version_id = (
+            executor.context.entity_ids[0]
+            if executor.context.entity_ids
+            else None
+        )
```
iLLiCiTiT, line 150:
```suggestion
        version_id = executor.context.entity_ids[0]
```

server/__init__.py
BigRoy: https://github.com/ynput/ayon-backend/blob/d42f863b02d3b143a880bd7743a7e464f92c25b6/api/resolve/templating.py#L459

Sure is.

server/__init__.py
BigRoy: `StringTemplate` does exist on server.

https://github.com/ynput/ayon-backend/blob/d42f863b02d3b143a880bd7743a7e464f92c25b6/api/resolve/templating.py#L459

Innders:
It might be possible for the share dialog to call the action to receive the correct path

kalisp:
> It might be possible for the share dialog to call the action to receive the correct path

What would that entail? Wrapping current logic into a endpoint for FE to call it?

PR #1845, opened 2026-05-19
Browser right click group error with group by status

### LINKED ISSUE #1844: Browser: right-click group error with group by status

### Current Behavior:

right-clicking a group creates an error when group by status is used.

### Expected Behavior:

nothing should happen.

### Version

1.9.4

MacOS

### Steps To Reproduce:

1. open the new browser
2. group by status
3. right click a header
4. boom !

### Relevant log output:

```shell

```

### Additional context:

_No response_

philippe-ynput:
## Changelog Description
Prevent a contextual menu error on items without a valid version id. This includes groups like Status, that don't carry a featured version id. By product grouping works as usual.

## Additional info
None.

## Testing notes:
1. `tools/make.sh run --use-dev browser`
2. in table view, group by anything and right-click the group.
3. there shouldn't be any error and no contextual menu.

closes #1844

### DIFF
--- a/client/ayon_core/tools/loader/ui/reviews_widget.py
+++ b/client/ayon_core/tools/loader/ui/reviews_widget.py
@@ -160,14 +160,14 @@ class ReviewsWidget(AYContainer):
             row_dict = proxy_idx.data(QtCore.Qt.ItemDataRole.UserRole) or {}
             if row_dict.get("entityType", "") == "Folder":
                 continue
-            version_id = row_dict.get("_version_id", row_dict.get("id", ""))
-            if version_id:
+            version_id = row_dict.get("_version_id") or row_dict.get("id", "")
+            if version_id and not version_id.startswith("grp:"):
                 version_ids.add(version_id)
 
         global_point = self._table.active_view.viewport().mapToGlobal(pos)
 
         if not version_ids or not project_name:
-            log.warning("No version ids or project name")
+            log.debug("No version ids or project name")
             return
 
         action_items = self._controller.get_action_items(


BigRoy (commented):
This PR seems to contain a much bigger diff than the description of the PR describes of what it should solve.

Regarding testing what the PR does, the testing notes seem to work fine:

✅ No errors, I do see log: `- { ayon_core.tools.loader.ui.reviews_widget }: [  No version ids or project name  ]` so seems to work.

---

Note that:

When multi-selecting versions by default it doesn't seem to sort the entries (even though the header does hint it's sorted by name column?)

Image

Clicking the header to change sorting, does affect the sorting - but then deselecting and reselecting makes it show the list unsorted again.

And we may need to keep an eye on refresh/drawing performance:

Multiselecting the versions in the old loader is quite a lot faster than in this browser redesign. It doesn't seem to be the querying that's slow or anything, but just e.g. drag-selecting across the UI feels entirely smooth in the old loader but in this one has a clear 'lag' (albeit relatively small; seems to shift between 0.1-0.3s?) Just pointing out that it does _feel_ slower and I can't get that slow feeling in the old loader - so we may need to remain wary on how this feels.

client/ayon_core/tests/tools/loader/ui/test_review_controller.py
```diff
@@ -22,9 +22,9 @@ def _make_request(
 def test_fetch_product_group_headers_fetches_all_pages_and_deduplicates(
     monkeypatch,
 ):
-    controller = ReviewController()
+    controller = ReviewController(None)
```
BigRoy: Usage seems wrong here - the type hint to `ReviewController` argument defines that it must take a `LoaderController` but you're passing `None` everywhere.

I'm also a bit confused why you made it a required argument if you're passing `None` to it in so many places?

client/ayon_core/tools/loader/ui/review_controller.py
```diff
@@ -663,7 +660,11 @@ def get_action_items(
         Returns:
             List of :class:`ActionItem` objects.
         """
-        if self._loader_controller is None:
+        if not isinstance(self._loader_controller, LoaderController):
+            self.log.warning(
+                "get_action_items called but no loader controller is "
+                "available, returning empty list."
+            )
```
BigRoy: We're passing `None` often to this controller looking at the code changes - are we expecting some spam of warnings in some cases? The input argument type to `__init__` says it requires a `LoaderController` which makes it seem like this shouldn't ever be optional, and hence must always be of that instance type?

client/ayon_core/tools/loader/ui/review_inspector.py
```diff
@@ -312,48 +341,209 @@ def _on_close(self) -> None:
...
+                # If this is a group row, include all its children.
+                model = midx.model()
+                for i in range(model.rowCount(midx)):
+                    child_idx = model.index(i, 0, midx)
+                    child_user_data = (
+                        child_idx.data(QtCore.Qt.ItemDataRole.UserRole) or {}
+                    )
+                    selected_ids.add(child_user_data.get("id", ""))
+            else:
+                selected_ids.add(user_data.get("id", ""))
```
BigRoy, line 406: Shouldn't all these entries have an `id` - if they wouldn't have that would indicate much larger issues. Right?
As such, I'm not sure about needing to rely on `.get` calls here? It will just obfuscate actual issues with the data?

client/ayon_core/tools/loader/ui/review_inspector.py
```diff
@@ -312,48 +341,209 @@ def _on_close(self) -> None:
...
+            form_values={},
+        )
+
+    def _set_tree_mode(self, enabled: bool) -> None:
+        """Enable or disable tree decoration on the view.
+
+        Args:
+            enabled: ``True`` to show collapse/expand indicators.
+        """
+        self._table._apply_tree_mode(enabled)
```
BigRoy, line 439: Technically we shouldn't be touching private methods of other objects - right? Does this indicate a design issue or is this fine?

philippe-ynput: later: https://github.com/ynput/ayon-ui-qt/issues/37

PR #1847, opened 2026-05-20
Define version status for publishing

### LINKED ISSUE #1230: AY-7638_Version status on publishing
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

We don't have option to define default status for an entity type, but we could be able to define status to set on version creation during integration.

It could be defined in setting with profiles, it should NOT live in settings of integration plugin (IntegrateAsset), so it can be reused if needed elsewhere.

### How would you imagine the implementation of the enhancemenent?
_No response_

_No response_

### Additional context:
link to discussion on [Discord](None)

_This issue was automatically created from Clickup ticket [AY-7638](https://app.clickup.com/t/86c2qr1cq)_

moonyuet:
## Changelog Description
This PR is to add integrator for setting version status during publish
Resolve https://github.com/ynput/ayon-core/issues/1230

## Additional info
Marked it as draft due to the ongoing discussion.

## Testing notes:
1. add profile settings in `ayon+settings://core/publish/IntegrateStatus/status_profiles`
2. Open Publisher UI
3. You can find set status options depending on how you set the profile setting
4. Publish

### DIFF
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -96,18 +96,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         if status_profile:
             artist_can_change = status_profile["artist_can_change"]
             default_status = status_profile["default_status"]
-            if not artist_can_change:
-                cls._set_instance_state(
-                    instance, status_state_attr, f"status|{default_status}"
-                )
-                cls.log.debug(
-                    "Artist cannot change status based on profile settings."
-                )
-                return output
 
-            default_status = status_profile["default_status"]
-
-        cls._set_instance_state(instance, status_state_attr, "use_status")
         if default_status not in statuses:
             cls.log.warning(
                 f"Default status '{default_status}' is not available"
@@ -115,6 +104,18 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
                 f"Using '{statuses[0]}' instead."
             )
             default_status = statuses[0]
+
+        if not artist_can_change:
+            cls._set_instance_state(
+                instance, status_state_attr, f"status|{default_status}"
+            )
+            cls.log.debug(
+                "Artist cannot change status based on profile settings."
+            )
+            return output
+
+        cls._set_instance_state(instance, status_state_attr, "use_status")
+
         output.append(EnumDef(
            "status",
             label="Version status",
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -10,13 +10,14 @@ if TYPE_CHECKING:
     from ayon_core.created_instance import CreatedInstance
 
 
-class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
     """Allow user to set status for the published version
     based on profiles defined in settings."""
 
-    order = pyblish.api.IntegratorOrder - 0.01
-    label = "Integrate Status"
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Collect Status"
 
+    enabled = False
     status_profiles: list[dict] = []
 
     def process(self, instance):
@@ -27,7 +28,7 @@ class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             # already set so we won't override it
             return
         attr_values = self.get_attr_values_from_data(instance.data)
-        status = attr_values.get("status")
+        status = attr_values.get("status", "")
         if status:
             instance.data["status"] = status
 
@@ -69,12 +70,18 @@ class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             logger=cls.log
         )
         default_status = None
-        showed_in_publisher_ui = True
+        artist_can_change = True
         if status_profile:
-            showed_in_publisher_ui = status_profile.get(
-                "show_in_publisher_ui", True
+            artist_can_change = status_profile.get(
+                "artist_can_change", True
             )
-            default_status = status_profile["default_status"]
+            if not artist_can_change:
+                cls.log.debug(
+                    "Artist cannot change status based on profile settings."
+                )
+                return []
+
+            default_status = status_profile.get("default_status", "")
 
         if default_status not in statuses:
             cls.log.warning(
@@ -88,7 +95,6 @@ class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
                 "status",
                 label="Version status",
                 items=statuses,
-                visible=showed_in_publisher_ui,
                 default=default_status,
             )
         ]
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -61,7 +61,8 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         ]
         if not statuses:
             cls.log.warning("No version statuses found in current project.")
-            return []
+            self._set_instance_state(instance, status_state_attr, "dont_use")
+            return output 
         folder_path = instance.get("folderPath")
         folder_entity = create_context.get_folder_entity(folder_path)
         task_entity = None
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -13,8 +13,10 @@ if TYPE_CHECKING:
 
 
 class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
-    """Allow user to set status for the published version
-    based on profiles defined in settings."""
+    """Allow to define version status for published version.
+    
+    It is possible to change setttings so user can set status in publisher UI.
+    """
 
     order = pyblish.api.CollectorOrder + 0.499
     label = "Collect Version Status"
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -136,4 +136,3 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             return
 
         plugin_attributes["status_state"] = state
-        cls._set_instance_state(instance, status_state_attr, "dont_use")
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -110,7 +110,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
                 )
                 return []
 
-            default_status = status_profile.get("default_status", "")
+            default_status = status_profile["default_status"]
 
         self._set_instance_state(instance, status_state_attr, "use_status")
         if default_status not in statuses:
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -103,6 +103,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
 
             default_status = status_profile.get("default_status", "")
 
+        self._set_instance_state(instance, status_state_attr, "use_status")
         if default_status not in statuses:
             cls.log.warning(
                 f"Default status '{default_status}' is not available"
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -14,7 +14,6 @@ if TYPE_CHECKING:
 
 class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
     """Allow to define version status for published version.
-    
     It is possible to change setttings so user can set status in publisher UI.
     """
 
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -37,7 +37,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         if status_state == "use_status":
             status = attr_values.get("status", "")
         elif status_state.startswith("status|"):
-            status = status_state.split("|", 1)[1]
+            status = status_state.removeprefix("status|")
         else:
             return
         if status:
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -36,7 +36,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
 
         if status_state == "use_status":
             status = attr_values.get("status", "")
-        elif status_state.startswith("use|"):
+        elif status_state.startswith("status|"):
             status = status_state.split("|", 1)[1]
         else:
             return
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -91,7 +91,11 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             artist_can_change = status_profile.get(
                 "artist_can_change", True
             )
+            default_status = status_profile.get("default_status", "")
             if not artist_can_change:
+                self._set_instance_state(
+                    instance, status_state_attr, f"status|{default_status}"
+                )
                 cls.log.debug(
                     "Artist cannot change status based on profile settings."
                 )
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -17,7 +17,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
     based on profiles defined in settings."""
 
     order = pyblish.api.CollectorOrder + 0.499
-    label = "Collect Status"
+    label = "Collect Version Status"
 
     enabled = False
     status_profiles: list[dict] = []
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -98,11 +98,12 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             default_status = status_profile["default_status"]
 
         if default_status not in statuses:
-            cls.log.warning(
-                f"Default status '{default_status}' is not available"
-                f" on project: {project_entity['name']}"
-                f"Using '{statuses[0]}' instead."
-            )
+            if status_profile:
+                cls.log.warning(
+                    f"Default status '{default_status}' is not available"
+                    f" on project: {project_entity['name']}"
+                    f"Using '{statuses[0]}' instead."
+                )
             default_status = statuses[0]
 
         if not artist_can_change:
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 import pyblish.api
-from ayon_core.lib import EnumDef, filter_profiles
+from ayon_core.lib import EnumDef, TextDef, filter_profiles
 from ayon_core.pipeline.publish import AYONPyblishPluginMixin
 
 if TYPE_CHECKING:
@@ -31,7 +31,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         status_state = attr_values.get("status_state")
         if status_state == "dont_use":
             return
-            
+
         if status_state == "use_status":
             status = attr_values.get("status", "")
         elif status_state.startswith("use|"):
@@ -50,18 +50,9 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         )
         output = [status_state_attr]
         if not cls.status_profiles:
-             self._set_instance_state(instace, status_state_attr, "dont_use")
-             return output
-
-    def _set_instance_state(self, instance, attr, state):
-        attr.default = state
-        plugin_attributes = instance.publish_attributes.get(cls.__name__)
-        if plugin_attributes is None:
-            return
-
-        plugin_attributes["status_state"] = state
-            self._set_instance_state(instance, status_state_attr, "dont_use")
+            cls._set_instance_state(instance, status_state_attr, "dont_use")
             return output
+
         project_entity = create_context.get_current_project_entity()
         statuses = [
             status["name"]
@@ -70,8 +61,9 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         ]
         if not statuses:
             cls.log.warning("No version statuses found in current project.")
-            self._set_instance_state(instance, status_state_attr, "dont_use")
-            return output 
+            cls._set_instance_state(instance, status_state_attr, "dont_use")
+            return output
+
         folder_path = instance.get("folderPath")
         folder_entity = create_context.get_folder_entity(folder_path)
         task_entity = None
@@ -102,17 +94,17 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             )
             default_status = status_profile.get("default_status", "")
             if not artist_can_change:
-                self._set_instance_state(
+                cls._set_instance_state(
                     instance, status_state_attr, f"status|{default_status}"
                 )
                 cls.log.debug(
                     "Artist cannot change status based on profile settings."
                 )
-                return []
+                return output
 
             default_status = status_profile["default_status"]
 
-        self._set_instance_state(instance, status_state_attr, "use_status")
+        cls._set_instance_state(instance, status_state_attr, "use_status")
         if default_status not in statuses:
             cls.log.warning(
                 f"Default status '{default_status}' is not available"
@@ -127,3 +119,18 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             default=default_status,
         ))
         return output
+
+    @classmethod
+    def _set_instance_state(
+        cls,
+        instance: "CreatedInstance",
+        status_state_attr: "TextDef",
+        state: str
+    ) -> None:
+        status_state_attr.default = state
+        plugin_attributes = instance.publish_attributes.get(cls.__name__)
+        if plugin_attributes is None:
+            return
+
+        plugin_attributes["status_state"] = state
+        cls._set_instance_state(instance, status_state_attr, "dont_use")
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -69,19 +69,22 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         folder_path = instance.get("folderPath")
         folder_entity = create_context.get_folder_entity(folder_path)
         task_entity = None
+        task_name = None
+        task_type = None
         if folder_entity:
             task_name = instance.get("task")
             task_entity = create_context.get_task_entity(
                 folder_path, task_name
             )
+            if task_entity:
+                task_type = task_entity["taskType"]
+
         filter_data = {
             "host_names": create_context.host_name,
-            "task_types": task_entity["taskType"],
-            "task_names": instance.get("task"),
+            "task_types": task_type,
+            "task_names": task_name,
             "product_base_types": instance.product_base_type,
         }
-        if task_entity:
-            filter_data["task_types"] = task_entity["taskType"]
 
         status_profile = filter_profiles(
             cls.status_profiles,
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -94,10 +94,8 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         default_status = None
         artist_can_change = True
         if status_profile:
-            artist_can_change = status_profile.get(
-                "artist_can_change", True
-            )
-            default_status = status_profile.get("default_status", "")
+            artist_can_change = status_profile["artist_can_change"]
+            default_status = status_profile["default_status"]
             if not artist_can_change:
                 cls._set_instance_state(
                     instance, status_state_attr, f"status|{default_status}"
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -111,11 +111,10 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
                 f"Using '{statuses[0]}' instead."
             )
             default_status = statuses[0]
-        return [
-            EnumDef(
-                "status",
-                label="Version status",
-                items=statuses,
-                default=default_status,
-            )
-        ]
+        output.append(EnumDef(
+           "status",
+            label="Version status",
+            items=statuses,
+            default=default_status,
+        ))
+        return output
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -28,7 +28,16 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             # already set so we won't override it
             return
         attr_values = self.get_attr_values_from_data(instance.data)
-        status = attr_values.get("status", "")
+        status_state = attr_values.get("status_state")
+        if status_state == "dont_use":
+            return
+            
+        if status_state == "use_status":
+            status = attr_values.get("status", "")
+        elif status_state.startswith("use|"):
+            status = status_state.split("|", 1)[1]
+        else:
+            return
         if status:
             instance.data["status"] = status
 
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -37,7 +37,8 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         cls, create_context: "CreateContext", instance: "CreatedInstance"
     ):
         if not cls.status_profiles:
-            return []
+            self._set_instance_state(instance, status_state_attr, "dont_use")
+            return output
         project_entity = create_context.get_current_project_entity()
         statuses = [
             status["name"]
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -31,7 +31,7 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
             return
         attr_values = self.get_attr_values_from_data(instance.data)
         status_state = attr_values.get("status_state")
-        if status_state == "dont_use":
+        if not status_state or status_state == "dont_use":
             return
 
         if status_state == "use_status":
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -6,8 +6,10 @@ from ayon_core.lib import EnumDef, TextDef, filter_profiles
 from ayon_core.pipeline.publish import AYONPyblishPluginMixin
 
 if TYPE_CHECKING:
-    from ayon_core.create_context import CreateContext
-    from ayon_core.created_instance import CreatedInstance
+    from ayon_core.pipeline.create import (
+        CreateContext,
+        CreatedInstance,
+    )
 
 
 class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
--- a/client/ayon_core/plugins/publish/collect_status.py
+++ b/client/ayon_core/plugins/publish/collect_status.py
@@ -36,7 +36,21 @@ class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
     def get_attr_defs_for_instance(
         cls, create_context: "CreateContext", instance: "CreatedInstance"
     ):
+        status_state_attr = TextDef(
+            "status_state", visible=False, default="dont_use"
+        )
+        output = [status_state_attr]
         if not cls.status_profiles:
+             self._set_instance_state(instace, status_state_attr, "dont_use")
+             return output
+
+    def _set_instance_state(self, instance, attr, state):
+        attr.default = state
+        plugin_attributes = instance.publish_attributes.get(cls.__name__)
+        if plugin_attributes is None:
+            return
+
+        plugin_attributes["status_state"] = state
             self._set_instance_state(instance, status_state_attr, "dont_use")
             return output
         project_entity = create_context.get_current_project_entity()
--- a/client/ayon_core/plugins/publish/integrate_status.py
+++ b/client/ayon_core/plugins/publish/integrate_status.py
@@ -43,7 +43,6 @@ class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
         if not statuses:
             cls.log.warning("No version statuses found in current project.")
             return []
-        default_status = None
         folder_path = instance.get("folderPath")
         folder_entity = create_context.get_folder_entity(folder_path)
         task_entity = None
--- a/client/ayon_core/plugins/publish/integrate_status.py
+++ b/client/ayon_core/plugins/publish/integrate_status.py
@@ -2,19 +2,23 @@ from typing import List
 
 import pyblish.api
 from ayon_core.lib import EnumDef, filter_profiles
-from ayon_core.pipeline.publish import AYONPyblishPluginMixin
+from ayon_core.pipeline.publish import OptionalPyblishPluginMixin
 
 
-class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+class IntegrateStatus(pyblish.api.InstancePlugin, OptionalPyblishPluginMixin):
     """Allow user to set status for the published version
     based on profiles defined in settings."""
 
     order = pyblish.api.IntegratorOrder - 0.01
     label = "Integrate Status"
 
+    optional = True
     status_profiles: List[dict] = []
 
     def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+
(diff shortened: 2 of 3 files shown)


client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,69 @@
...
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        status = attr_values.get("status")
+        instance.data["status"] = status
+
+    @classmethod
+    def get_attribute_defs(cls):
+        if not cls.status_profiles:
+            return []
+        statuses = get_project_status()
```
iLLiCiTiT:
```suggestion
    @classmethod
    def get_attr_defs_for_instance(
        cls, create_context: "CreateContext", instance: "CreatedInstance"
    ):
        if not cls.status_profiles:
            return []
        project_entity = context.get_current_project_entity()
        statuses = [
            status["name"]
            for status in project_entity["statuses"]
        ]
        if not statuses:
            return []

```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,69 @@
...
+from ayon_core.pipeline.publish import AYONPyblishPluginMixin
+from ayon_core.pipeline import get_current_project_name
+
+
+def get_project_status() -> List[str]:
+    """Return available status names in current project"""
+    project_name = get_current_project_name()
+    project_entity = ayon_api.get_project(project_name)
+    statuses = project_entity["statuses"]
+    return [status["name"] for status in statuses]
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,69 @@
...
+            filter_data,
+            logger=self.log
+        )
+        if status_profile is None:
+            self.log.debug("No matching status profile found.")
+            return
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        status = attr_values.get("status")
+        instance.data["status"] = status
```
iLLiCiTiT:
```suggestion
        instance.data["versionStatus"] = status
```
@BigRoy didn't we discuss to add it to `"versionData"`?

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,69 @@
...
+        version_data = instance.data.setdefault("versionData", {})
+        if "status" in version_data:
+            # already set so we won't override it
+            return
+        folder_entity = instance.data["folderEntity"]
+        task_entity = instance.data["taskEntity"]
+        filter_data = {
+            "host_names": instance.context.data["hostName"],
+            "task_types": task_entity["taskType"],
+            "task_names": task_entity["name"],
```
iLLiCiTiT:
```suggestion
        task_entity = instance.data["taskEntity"]
        task_name = task_type = None
        if task_entity:
            task_name = task_entity["name"]
            task_type = task_entity["taskType"]

        filter_data = {
            "host_names": instance.context.data["hostName"],
            "task_types": task_type,
            "task_names": task_name ,
```

client/ayon_core/plugins/publish/integrate_status.py
moonyuet: I actually add the version data on top. 
it should be `version_data["status"] = status`

iLLiCiTiT:
I have a question, does it make sense to have a selection of statuses in publisher? Shouldn't the studio decide what the status will be in the profiles?

Considering they might want also decide based on other criteria in future, for example use different status when sending something to farm, it would make more sense. I would not allow to let artists choose any status, maybe a predefined set of statuses, but I do believe artists don't care.

moonyuet:
> I have a question, does it make sense to have a selection of statuses in publisher? Shouldn't the studio decide what the status will be in the profiles?
> 
> Considering they might want also decide based on other criteria in future, for example use different status when sending something to farm, it would make more sense. I would not allow to let artists choose any status, maybe a predefined set of statuses, but I do believe artists don't care.

I actually agree with this. This would prevent artists to accidentally set the status with the integrators.
But I supposed that we need to expose the status enum from the backend so that we can use it as the setting in the profiles.

server/settings/publish_plugins.py
```diff
@@ -1298,6 +1327,10 @@ class PublishPuginsModel(BaseSettingsModel):
         default_factory=PreIntegrateThumbnailsModel,
         title="Override Integrate Thumbnail Representations"
     )
+    IntegrateStatus: IntegrateStatusModel = SettingsField(
```
iLLiCiTiT: Another question, does this setting have to be "for a plugin", shouldn't we make it as dedicated settings for statuses? @BigRoy opinions?

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,71 @@
...
+    def get_attr_defs_for_instance(
+        cls, create_context: "CreateContext", instance: "CreatedInstance"
+    ):
+        if not cls.status_profiles:
+            return []
+        project_entity = cls.create_context.get_current_project_entity()
+        statuses = [
+            status["name"]
+            for status in project_entity["statuses"]
+        ]
```
BigRoy, line 64: These should be filtered further to statuses that are scoped to `version` entity types.

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,71 @@
...
+        task_entity = instance.data["taskEntity"]
+        task_name = task_type = None
+        if task_entity:
+            task_name = task_entity["name"]
+            task_type = task_entity["taskType"]
+
+        filter_data = {
+            "host_names": instance.context.data["hostName"],
+            "task_types": task_type,
+            "task_names": task_name ,
```
BigRoy:
```suggestion
            "task_names": task_name,
```

server/settings/publish_plugins.py
```diff
@@ -1111,6 +1111,28 @@ class PreIntegrateThumbnailsModel(BaseSettingsModel):
     )
 
 
+class IntegrateStatusProfile(BaseSettingsModel):
+    _layout = "expanded"
```
BigRoy, line 1263: Can we add a toggle to enable/disable this plug-in (when disabled the attribute also should not show of course).

Then, I'd prefer also for the attribute to be in the UI to be optional - so I feel like we should have a `allow_override_in_publisher` kind of attribute as well to define whether the artist can set it to something else too and allow disabling that.

server/settings/publish_plugins.py
BigRoy: Hmm - not sure. It may make little sense as being a plug-in setting in the plug-ins list - but it's the same with all the others.

Where would you put it instead?
- It allows to pick a default status for version per context
- It allows the user to override it in the publisher UI (and maybe even the statuses that are allowed to be picked by the user should be customizable too?)

server/settings/publish_plugins.py
```diff
@@ -1111,6 +1111,28 @@ class PreIntegrateThumbnailsModel(BaseSettingsModel):
...
+    task_types: list[str] = SettingsField(
+        default_factory=list,
+        title="Task types",
+        enum_resolver=task_types_enum
+    )
+    task_names: list[str] = SettingsField(
+        default_factory=list,
+        title="Task names",
+    )
+    status: bool = SettingsField(title="Status")
```
BigRoy: I was expecting this to be `str` and to be the 'default status' value for that particular context. Also, can we make this a status enum (for versions?)?

moonyuet: As far as I explore, we dont have the status enum exposed in backend: https://github.com/ynput/ayon-backend/blob/develop/ayon_server/settings/enum.py

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -55,10 +59,10 @@ def get_attr_defs_for_instance(
     ):
         if not cls.status_profiles:
             return []
-        project_entity = cls.create_context.get_current_project_entity()
+        version_entity = cls.create_context.get_current_version_entity()
```
BigRoy: Pretty sure that `get_current_version_entity` is not an existing method 🤔 

Statuses are still on the project, you just need to compare whether `version` is in the status scope for each entry?

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,75 @@
...
+    def get_attr_defs_for_instance(
+        cls, create_context: "CreateContext", instance: "CreatedInstance"
+    ):
+        if not cls.status_profiles:
+            return []
+        version_entity = cls.create_context.get_current_version_entity()
+        statuses = [
+            status["name"]
+            for status in version_entity["statuses"]
+        ]
```
BigRoy: I think it would be something like this:
```suggestion
        project_entity = cls.create_context.get_current_project_entity()
        statuses = [
            status["name"]
            for status in project_entity ["statuses"]
            if "version" in status["scope"]
        ]
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,75 @@
...
+        if not cls.status_profiles:
+            return []
+        version_entity = cls.create_context.get_current_version_entity()
+        statuses = [
+            status["name"]
+            for status in version_entity["statuses"]
+        ]
+
+        return [
+            EnumDef(
```
BigRoy: I think this should maybe somehow also be set in `default` value through the profiles if we're allowing profiles - otherwise I wouldn't know how to pick either the one from the UI (attr-values) versus the status as set in the profile? 🤔 

I originally thought that the profile would essentially state what the default status would be for that publish context. @iLLiCiTiT what's your thought there? And then we may add product base type and product name to the profile as well (and host!) so you could say that e.g. all reviews would go to "Pending Review" by default, but all other products default to a status you may have named "Data".

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+        task_name = task_type = None
+        if task_entity:
+            task_name = task_entity["name"]
+            task_type = task_entity["taskType"]
+
+        filter_data = {
+            "host_names": instance.context.data["hostName"],
+            "task_types": task_type,
+            "task_names": task_name,
+            "folder_paths": folder_entity["path"]
```
iLLiCiTiT:
```suggestion
            "product_base_types": instance.data["productBaseType"],
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if not self.status_profiles:
+            self.log.debug("No status profiles defined in settings.")
+            return
+
+        version_data = instance.data.setdefault("versionData", {})
+        if "status" in version_data:
+            # already set so we won't override it
+            return
+        folder_entity = instance.data["folderEntity"]
```
iLLiCiTiT:
```suggestion

```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,76 @@
...
+                "status",
+                label="Set status",
+                items=statuses,
+                default=statuses[0]
+            )
+        ]
```
iLLiCiTiT:
```suggestion
            default_status = None
            folder_path = instance.get("folderPath")
            folder_entity = context.get_folder_entity(folder_path)
            task_entity = None
            if folder_entity:
                task_name = instance.get("task")
                task_entity = context.get_task_entity(folder_path, task_name)

            if task_entity:
                filter_data = {
                    "host_names": context.host_name,
                    "task_types": task_entity["taskType"],
                    "task_names": task_entity["name"],
                    "product_base_types": instance.product_base_type,
                }
                status_profile = filter_profiles(
                    self.status_profiles,
                    filter_data,
                    logger=self.log
                )
                if status_profile:
                    default_status = status_profile["default_status"]

        if default not in statuses:
            default = statuses[0]

        return [
            EnumDef(
                "status",
                label="Version status",
                items=statuses,
                default=default_status,
            )
        ]
```

server/settings/publish_plugins.py
```diff
@@ -1111,6 +1111,31 @@ class PreIntegrateThumbnailsModel(BaseSettingsModel):
...
```
iLLiCiTiT:
```suggestion
    default_status: bool = SettingsField(
        title="Default status",
        enum_resolver=StatusesEnumResolver,
    )
```
Requires import `from ayon_server.enum.resolvers import StatusesEnumResolver`

client/ayon_core/plugins/publish/collect_status.py
iLLiCiTiT: This is suggestion to be able to define default status using settings.

There are more options how to achieve that. We can also add one more item to statuses saying: "Use default" which means that the one from settings used.

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+
+        if not self.status_profiles:
+            self.log.debug("No status profiles defined in settings.")
+            return
+
+        version_data = instance.data.setdefault("versionData", {})
```
iLLiCiTiT: I just found out that we already do handle version status in integrate.py. We use `instance.data.get("status")`, this does use `["versionData"]["status"]`. If we keep it under `versionData` we also have to handle it in integrate.py. With this change it might stop working for someone who already has logic for statuses.

BigRoy: In this case - let's keep it at `instance.data["status"]` then. It's a fine key for it - it's more explicit than `versionData` which may also go to `data`, etc.

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+            filter_data,
+            logger=self.log
+        )
+        if status_profile is None:
+            self.log.debug("No matching status profile found.")
+            return
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        status = attr_values.get("status")
+        version_data["status"] = status
```
iLLiCiTiT:
```suggestion
        if status:
            instance.data["status"] = status
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+    status_profiles: List[dict] = []
+
+    def process(self, instance):
+        if not self.is_active(instance.data):
+            return
+
+        if not self.status_profiles:
+            self.log.debug("No status profiles defined in settings.")
+            return
+
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+            "folder_paths": folder_entity["path"]
+        }
+        status_profile = filter_profiles(
+            self.status_profiles,
+            filter_data,
+            logger=self.log
+        )
+        if status_profile is None:
+            self.log.debug("No matching status profile found.")
+            return
```
iLLiCiTiT: Based on https://github.com/ynput/ayon-core/pull/1847/changes#r3279783880 no need to check profiles here.
```suggestion
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+
+        if not self.status_profiles:
+            self.log.debug("No status profiles defined in settings.")
+            return
+
+        version_data = instance.data.setdefault("versionData", {})
+        if "status" in version_data:
+            # already set so we won't override it
+            return
```
iLLiCiTiT:
```suggestion
        if instance.data.get("status"):
            # already set so we won't override it
            return
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
+from typing import List
+
+import pyblish.api
+from ayon_core.lib import EnumDef, filter_profiles
+from ayon_core.pipeline.publish import OptionalPyblishPluginMixin
```
iLLiCiTiT: It doesn't make sense to use `OptionalPyblishPluginMixin` if you override `get_attr_defs_for_instance`. In that case it is either enabled or not.
```suggestion
from ayon_core.pipeline.publish import AYONPyblishPluginMixin
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class IntegrateStatus(pyblish.api.InstancePlugin, OptionalPyblishPluginMixin):
```
iLLiCiTiT:
```suggestion
class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,76 @@
...
+    """Allow user to set status for the published version
+    based on profiles defined in settings."""
+
+    order = pyblish.api.IntegratorOrder - 0.01
+    label = "Integrate Status"
+
+    optional = True
```
iLLiCiTiT:
```suggestion
```

server/settings/publish_plugins.py
```diff
@@ -1111,6 +1111,31 @@ class PreIntegrateThumbnailsModel(BaseSettingsModel):
     )
 
 
+class IntegrateStatusProfile(BaseSettingsModel):
+    _layout = "expanded"
+    enabled: bool = SettingsField(True, title="Enabled")
+    active: bool = SettingsField(True, title="Active")
+    optional: bool = SettingsField(False, title="Optional")
```
iLLiCiTiT:
```suggestion
```

server/settings/publish_plugins.py
moonyuet: I got this error from docker when I use this as enum_resolver, but fixed it with the lambda.
```
ERROR   logging                    | Unhandled exception: TypeError: BaseEnumResolver.__init__() got multiple values for argument 'self'

    traceback:

      TypeError: BaseEnumResolver.__init__() got multiple values for argument 'self'


      ayon_server/api/context.py:77

      response = await call_next(request)


      api/addons/studio_settings.py:50

      await postprocess_settings_schema(schema, model, context=context)


      ayon_server/settings/postprocess.py:252

      await postprocess_settings_schema(


      ayon_server/settings/postprocess.py:141

      enum_values, enum_labels = await process_enum(
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,75 @@
+from typing import List, TYPE_CHECKING
```
BigRoy:
```suggestion
from __future__ import annotations
from typing import TYPE_CHECKING
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+    """Allow user to set status for the published version
+    based on profiles defined in settings."""
+
+    order = pyblish.api.IntegratorOrder - 0.01
+    label = "Integrate Status"
+
+    status_profiles: List[dict] = []
```
BigRoy:
```suggestion
    status_profiles: list[dict] = []
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,75 @@
...
+            status_profile = filter_profiles(
+                cls.status_profiles,
+                filter_data,
+                logger=cls.log
+            )
+            if status_profile:
+                default_status = status_profile["default_status"]
+
+        if default_status not in statuses:
+            default_status = statuses[0]
```
BigRoy, line 108: Can we log a warning here?

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,75 @@
...
+                "task_names": task_entity["name"],
+                "product_base_types": instance.product_base_type,
+            }
+            status_profile = filter_profiles(
+                cls.status_profiles,
+                filter_data,
+                logger=cls.log
+            )
+            if status_profile:
+                default_status = status_profile["default_status"]
```
BigRoy: I think we should also allow to filter profile without task entity.

```suggestion
        filter_data = {
            "host_names": cls.create_context.host_name,
            "task_types": task_entity["taskType"],
            "task_names": instance.get("task"),
            "product_base_types": instance.product_base_type,
        }
        if task_entity:
            filter_data["task_types"] = task_entity["taskType"]
        
        status_profile = filter_profiles(
            cls.status_profiles,
            filter_data,
            logger=cls.log
        )
        default_status = None
        if status_profile:
            default_status = status_profile["default_status"]
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,84 @@
...
+        cls, create_context: "CreateContext", instance: "CreatedInstance"
+    ):
+        if not cls.status_profiles:
+            return []
+        project_entity = create_context.get_current_project_entity()
+        statuses = [
+            status["name"]
+            for status in project_entity["statuses"]
+            if "version" in status["scope"]
+        ]
```
BigRoy, line 64: The `EnumDef` when not in multiselect mode _must_ have an item. Let's avoid issues further down if for whatever reason someone does the odd thing of scoping no statuses whatsoever to "version"
```suggestion
        ]
        if not statuses:
            cls.log.warning("No version statuses found in current project.")
            return []
```

server/settings/publish_plugins.py
BigRoy: @moonyuet we're now lacking a way to disable the set status to show in UI or not. I think we can just add a global "enabled" checkbox for the plug-in (disabled by default?)

That would leave us then in a place where there's no way you can set per profile the default status but disallow the user to override it in the UI - which I think we should allow to do as well.. the only annoying bit is, as soon as "exposed in UI" is disabled we'll need to duplicate the profile filtering to also apply in `process` because then we'll need to get to the 'status' from the running instance as well. 🤔 Complicating the code - but I don't see any other way.

- [x] Global enabled checkbox (to disable the plug-in completely) (default: False)
- [ ] ~~Per-profile "Enabled" checkbox (this way you can also very selectively "exclude" status setting behavior for certain profiles.) (default: True)~~ - as discussed in DM, let's skip this.
- [x] Per-profile add a "Show in publisher UI" checkbox with description to "Allow artist to override the version status." (default: True)

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,92 @@
...
+                f"available statuses: {statuses}"
+                f"Use {statuses[0]} instead."
+            )
+            default_status = statuses[0]
+        return [
+            EnumDef(
+                "status",
+                label="Version status",
+                items=statuses,
+                visible=showed_in_publisher_ui,
```
BigRoy: There is one important downside to this - and that is that if it's made invisible, but was visible prior with e.g. a different status - the "default" value would not be enforced but it'll remain the overridden value. Which is a bit annoying - because if not shown in UI what we _actually_ want is to enforce the default status regardless.

Which is the annoying bit I [explained here](https://github.com/ynput/ayon-core/pull/1847/changes#r3287458993) to avoid that. We'll need to essentially not rely on the EnumDef at that point and get to the correct 'status' for the instance within the `process` method. @iLLiCiTiT or do you have other tricks up your sleeve?

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,92 @@
...
+        project_entity = create_context.get_current_project_entity()
+        statuses = [
+            status["name"]
+            for status in project_entity["statuses"]
+            if "version" in status["scope"]
+        ]
+        if not statuses:
+            cls.log.warning("No version statuses found in current project.")
+            return []
+        default_status = None
```
iLLiCiTiT: Doubled variable
```suggestion

```

client/ayon_core/plugins/publish/integrate_status.py
iLLiCiTiT: Few ideas

### We can set the value on the instance if not visible.
This way we make sure the value is set.
```python
if not showed_in_publisher_ui:
    plugin_values = instance.publish_attributes.get(self.__class__)
    if plugin_values is not None:
        plugin_values["status"] = default_status
```
We should do the same in case there are no profiles, but set it to empty string `""` instead of `default_status`.

### Add information about the status using another attribute
Always return `TextDef` with `visible=False` containing information if the status value should be considered and how: `ignore`, `use_default`, `use_value`.

More complicated, but is very explicit, and the status set on instances is not lost.

BigRoy: Either way we'll need to pass on to the instance which value to use - technically the `TextDef` would still be stored in data on the instance, and hence can 'persist' the wrong value meaning we'll always need to still push the right value - which is what the first is doing.

In short, I think we should just do the former, if not shown in UI - then force push the value to the instance. (if that indeed works.) If then Enum is visible again, I'd expect it indeed to use the last value.

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,91 @@
...
+import pyblish.api
+from ayon_core.lib import EnumDef, filter_profiles
+from ayon_core.pipeline.publish import AYONPyblishPluginMixin
+
+if TYPE_CHECKING:
+    from ayon_core.create_context import CreateContext
+    from ayon_core.created_instance import CreatedInstance
+
+
+class IntegrateStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
```
iLLiCiTiT: BTW can this be collector? Considering extractors can remove the source instance it might be better. BTW this is one of reasons why I don't think the settings should live under class name of the plugin.

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,91 @@
...
+    """Allow user to set status for the published version
+    based on profiles defined in settings."""
+
+    order = pyblish.api.IntegratorOrder - 0.01
+    label = "Integrate Status"
+
+    status_profiles: list[dict] = []
+
+    def process(self, instance):
+        if instance.data.get("status"):
```
iLLiCiTiT, line 30: This solves case when the profiles are empty.
```suggestion
        if not self.status_profiles:
            return

        if instance.data.get("status"):
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,91 @@
...
+        if status_profile:
+            showed_in_publisher_ui = status_profile.get(
+                "show_in_publisher_ui", True
+            )
+            default_status = status_profile["default_status"]
+
+        if default_status not in statuses:
+            cls.log.warning(
+                f"Default status '{default_status}' is not in the list of "
+                f"available statuses: {statuses}"
```
iLLiCiTiT:
```suggestion
                f"Default status '{default_status}' is not available"
                f" on project: {project_entity['name']}"
```

client/ayon_core/plugins/publish/integrate_status.py
```diff
@@ -0,0 +1,91 @@
...
+                f"Use {statuses[0]} instead."
```
iLLiCiTiT:
```suggestion
                f"Using '{statuses[0]}' instead."
```

server/settings/publish_plugins.py
```diff
@@ -1111,6 +1128,39 @@ class PreIntegrateThumbnailsModel(BaseSettingsModel):
...
+    task_types: list[str] = SettingsField(
+        default_factory=list,
+        title="Task types",
+        enum_resolver=task_types_enum
+    )
+    task_names: list[str] = SettingsField(
+        default_factory=list,
+        title="Task names",
+    )
+    showed_in_publisher_ui: bool = SettingsField(
```
iLLiCiTiT: I don't think this is right terminology. It is more like `artist_can_change`, it is not about "showing" in UI but giving artist option to choose different status from the default one.

I would also move this option below `default_status`. Reading UI: Set filters -> Choose default status -> Give artist permission to choose different status.

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+        default_status = None
+        artist_can_change = True
+        if status_profile:
+            artist_can_change = status_profile.get(
+                "artist_can_change", True
+            )
+            if not artist_can_change:
+                cls.log.debug(
+                    "Artist cannot change status based on profile settings."
+                )
```
BigRoy: I think we're still lacking setting the 'default value' in this scenario? When not expose to artist but there is a profile that matches we should still be applying the default value?

server/settings/publish_plugins.py
BigRoy: @moonyuet I think we're still lacking the global enabled checkbox here?

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+        status = attr_values.get("status", "")
+        if status:
+            instance.data["status"] = status
+
+    @classmethod
+    def get_attr_defs_for_instance(
+        cls, create_context: "CreateContext", instance: "CreatedInstance"
+    ):
+        if not cls.status_profiles:
+            return []
```
iLLiCiTiT:
```suggestion
            self._set_instance_state(instance, status_state_attr, "dont_use")
            return output
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+        attr_values = self.get_attr_values_from_data(instance.data)
+        status = attr_values.get("status", "")
+        if status:
+            instance.data["status"] = status
+
+    @classmethod
+    def get_attr_defs_for_instance(
+        cls, create_context: "CreateContext", instance: "CreatedInstance"
+    ):
+        if not cls.status_profiles:
```
iLLiCiTiT, line 55:
```suggestion
        status_state_attr = TextDef(
            "status_state", visible=False, default="dont_use"
        )
        output = [status_state_attr]
        if not cls.status_profiles:
             self._set_instance_state(instace, status_state_attr, "dont_use")
             return output

    def _set_instance_state(self, instance, attr, state):
        attr.default = state
        plugin_attributes = instance.publish_attributes.get(cls.__name__)
        if plugin_attributes is None:
            return

        plugin_attributes["status_state"] = state
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+            return []
+        project_entity = create_context.get_current_project_entity()
+        statuses = [
+            status["name"]
+            for status in project_entity["statuses"]
+            if "version" in status["scope"]
+        ]
+        if not statuses:
+            cls.log.warning("No version statuses found in current project.")
+            return []
```
iLLiCiTiT:
```suggestion
            self._set_instance_state(instance, status_state_attr, "dont_use")
            return output 
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+            logger=cls.log
+        )
+        default_status = None
+        artist_can_change = True
+        if status_profile:
+            artist_can_change = status_profile.get(
+                "artist_can_change", True
+            )
+            if not artist_can_change:
+                cls.log.debug(
```
iLLiCiTiT:
```suggestion
            default_status = status_profile.get("default_status", "")
            if not artist_can_change:
                self._set_instance_state(
                    instance, status_state_attr, f"status|{default_status}"
                )
                cls.log.debug(
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+                    "Artist cannot change status based on profile settings."
+                )
+                return []
+
+            default_status = status_profile.get("default_status", "")
+
```
iLLiCiTiT, line 100:
```suggestion

        self._set_instance_state(instance, status_state_attr, "use_status")
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+            )
+            default_status = statuses[0]
+        return [
+            EnumDef(
+                "status",
+                label="Version status",
+                items=statuses,
+                default=default_status,
+            )
+        ]
```
iLLiCiTiT:
```suggestion
        output.append(EnumDef(
           "status",
            label="Version status",
            items=statuses,
            default=default_status,
        ))
        return output
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def process(self, instance):
+        if not self.status_profiles:
+            return
+
+        if instance.data.get("status"):
+            # already set so we won't override it
+            return
+        attr_values = self.get_attr_values_from_data(instance.data)
+        status = attr_values.get("status", "")
```
iLLiCiTiT:
```suggestion
        status_state = attr_values.get("status_state")
        if status_state == "dont_use":
            return
            
        if status_state == "use_status":
            status = attr_values.get("status", "")
        elif status_state.startswith("use|"):
            status = status_state.split("|", 1)[1]
        else:
            return
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,100 @@
...
+            artist_can_change = status_profile.get(
+                "artist_can_change", True
+            )
+            if not artist_can_change:
+                cls.log.debug(
+                    "Artist cannot change status based on profile settings."
+                )
+                return []
+
+            default_status = status_profile.get("default_status", "")
```
iLLiCiTiT:
```suggestion
            default_status = status_profile["default_status"]
```

server/settings/publish_plugins.py
iLLiCiTiT: > @moonyuet I think we're still lacking the global enabled checkbox here?

That is on `CollectStatusModel`, or you mean to "disable" the status per profile?

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,136 @@
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+import pyblish.api
+from ayon_core.lib import EnumDef, TextDef, filter_profiles
+from ayon_core.pipeline.publish import AYONPyblishPluginMixin
+
+if TYPE_CHECKING:
+    from ayon_core.create_context import CreateContext
+    from ayon_core.created_instance import CreatedInstance
```
iLLiCiTiT:
```suggestion
    from ayon_core.pipeline.create import (
        CreateContext,
        CreatedInstance,
    )
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,136 @@
...
+                folder_path, task_name
+            )
+        filter_data = {
+            "host_names": create_context.host_name,
+            "task_types": task_entity["taskType"],
+            "task_names": instance.get("task"),
+            "product_base_types": instance.product_base_type,
+        }
+        if task_entity:
+            filter_data["task_types"] = task_entity["taskType"]
```
iLLiCiTiT:
```suggestion
        task_name = None
        task_type = None
        if folder_entity:
            task_name = instance.get("task")
            task_entity = create_context.get_task_entity(
                folder_path, task_name
            )
            if task_entity:
                task_type = task_entity["taskType"]

        filter_data = {
            "host_names": create_context.host_name,
            "task_types": task_type,
            "task_names": task_name,
            "product_base_types": instance.product_base_type,
        }
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,136 @@
...
+            filter_data,
+            logger=cls.log
+        )
+        default_status = None
+        artist_can_change = True
+        if status_profile:
+            artist_can_change = status_profile.get(
+                "artist_can_change", True
+            )
+            default_status = status_profile.get("default_status", "")
```
iLLiCiTiT:
```suggestion
            artist_can_change = status_profile["artist_can_change"]
            default_status = status_profile["default_status"]
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,136 @@
...
+    def process(self, instance):
+        if not self.status_profiles:
+            return
+
+        if instance.data.get("status"):
+            # already set so we won't override it
+            return
+        attr_values = self.get_attr_values_from_data(instance.data)
+        status_state = attr_values.get("status_state")
+        if status_state == "dont_use":
```
iLLiCiTiT:
```suggestion
        if not status_state or status_state == "dont_use":
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,136 @@
...
+            return
+
+        if status_state == "use_status":
+            status = attr_values.get("status", "")
+        elif status_state.startswith("use|"):
```
iLLiCiTiT:
```suggestion
        elif status_state.startswith("status|"):
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,136 @@
...
+            status = status_state.split("|", 1)[1]
```
iLLiCiTiT:
```suggestion
            status = status_state.removeprefix("status|")
```

iLLiCiTiT (approved):

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,138 @@
...
+            default_status = status_profile["default_status"]
+
+        cls._set_instance_state(instance, status_state_attr, "use_status")
+        if default_status not in statuses:
+            cls.log.warning(
+                f"Default status '{default_status}' is not available"
+                f" on project: {project_entity['name']}"
+                f"Using '{statuses[0]}' instead."
+            )
+            default_status = statuses[0]
```
BigRoy, line 108: Should this fallback be set also before the above `_set_instance_state` call with the `status|{default_status}` value just so we never end up submitting with an invalid status if not exposed to artist as override @iLLiCiTiT ?

BigRoy (approved):
Code looks good to me - it's a lot more complex than I expected it needed to be. I feel like the whole `_set_instance_state` juggling is not very nice, but I guess we need something to pass on the value. 🤔 

Just the invalid default status fallback to one of the existing statuses I think may need to be earlier, see comment.

client/ayon_core/plugins/publish/collect_status.py
iLLiCiTiT, line 108:
```suggestion
        if status_profile:
            artist_can_change = status_profile["artist_can_change"]
            default_status = status_profile["default_status"]

        if default_status not in statuses:
            cls.log.warning(
                f"Default status '{default_status}' is not available"
                f" on project: {project_entity['name']}"
                f"Using '{statuses[0]}' instead."
            )
            default_status = statuses[0]

        if not artist_can_change:
            cls._set_instance_state(
                instance, status_state_attr, f"status|{default_status}"
            )
            cls.log.debug(
                "Artist cannot change status based on profile settings."
            )
            return output

        cls._set_instance_state(instance, status_state_attr, "use_status")

```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,139 @@
...
+        if status_profile:
+            artist_can_change = status_profile["artist_can_change"]
+            default_status = status_profile["default_status"]
+
+        if default_status not in statuses:
+            cls.log.warning(
+                f"Default status '{default_status}' is not available"
+                f" on project: {project_entity['name']}"
+                f"Using '{statuses[0]}' instead."
+            )
```
iLLiCiTiT: Sorry, one more fix needed.
```suggestion
            if status_profile:
                cls.log.warning(
                    f"Default status '{default_status}' is not available"
                    f" on project: {project_entity['name']}"
                    f"Using '{statuses[0]}' instead."
                )
```

LiborBatek (approved):
Works well and also with filtering set in the profile...

Image

And when exposed to artist, again good!

Image

I have tested in different host apps and with different products and set statuses and all been working nicely!


Note: One small thing which bugs me is lack of all statuses in the settings in frontend - should list all status types imho (same as when in the Publisher UI which is fine)

Image

LiborBatek:
btw its an awesome feature tbh!!! loving it...

iLLiCiTiT:
> Note: One small thing which bugs me is lack of all statuses in the settings in frontend - should list all status types imho (same as when in the Publisher UI which is fine)

Studio settings are showing statuses based on default project anatomy. You have to go to project settings to see what you see in publisher.

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,140 @@
...
+        CreatedInstance,
+    )
+
+
+class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+    """Allow user to set status for the published version
+    based on profiles defined in settings."""
+
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Collect Status"
```
iLLiCiTiT:
```suggestion
    label = "Collect Version Status"
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,140 @@
...
+if TYPE_CHECKING:
+    from ayon_core.pipeline.create import (
+        CreateContext,
+        CreatedInstance,
+    )
+
+
+class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+    """Allow user to set status for the published version
+    based on profiles defined in settings."""
```
iLLiCiTiT:
```suggestion
    """Allow to define version status for published version.
    
    It is possible to change setttings so user can set status in publisher UI.
    """
```

client/ayon_core/plugins/publish/collect_status.py
```diff
@@ -0,0 +1,142 @@
...
+if TYPE_CHECKING:
+    from ayon_core.pipeline.create import (
+        CreateContext,
+        CreatedInstance,
+    )
+
+
+class CollectStatus(pyblish.api.InstancePlugin, AYONPyblishPluginMixin):
+    """Allow to define version status for published version.
+
```
iLLiCiTiT:
```suggestion

```

sebastianbrandhuber:
@iLLiCiTiT @LiborBatek 
Since you were talking about this earlier here:

How is it possible to change the default status in the studio settings (not as a project override),
and get all the statuses from my studio anatomy settings?

image

Because this list does not reflect what we have set up in our studio anatomy template.
Only if I go to the project settings for the addon, I am able to get the full list:
(And I do not want to set this every time we create a project...)

image

Any help is appreciated,
thanks!

iLLiCiTiT:
Made PR https://github.com/ynput/ayon-core/pull/1942

PR #1850, opened 2026-05-20
Set version tags on publish

### LINKED ISSUE #1673: YN-0458: Set version tags on publish

We want to set tags for a version, directly on publish.

  

Possibly from other plug-ins, e.g. through setting some instance data but an alternative is exposing as attribute in the UI to potentially pick from existing tags. 

We need to provide some way to set the tags for the instances.

```
instance.data["versionData"]["tags"] = ["tag1", "tag2"]
# OR
instance.data["tags"] = ["tag1", "tag2"]
```

Since we have tags on all entity types I feel like explicitly putting it in version data may feel the most explicit.

With that, we could e.g. add this to the [Integrator:](https://github.com/ynput/ayon-core/blob/10d3bae2edf8d91af1aebf123625cd25e86449a1/client/ayon_core/plugins/publish/integrate.py#L464) (and also copy that in the Integrate Hero Version integrator)

```
# Special behavior for tags - they are neither `data` nor `attributes`
tags: Optional[list[str]] = all_version_data.pop("tags", None)
if tags is not None:
    # If not an iterable of strings, then log warning and set to None
    if not isinstance(tags, (list, tuple, set)) or not all(
        isinstance(tag, str) for tag in tags
    ):
    self.log.warning(
        "Version tags must be an iterable of strings. "
        "Got: {}".format(repr(tags))
    )
    tags = None
```

And pass

```
tags=tags
```

to the

```
new_version_entity
```

call.

  

Then all you'd need to make sure is to have e.g. something doing:

```
version_data: dict[str, Any] = instance.data.setdefault("versionData", {})
version_data["tags"] = ["Guardian Approved"]
```

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1463602089304260743/1466223372290625689)

_This issue was automatically created from AYON ticket [YN-0458]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=890fba20fd4711f0a31db1f99397efdd)

MustafaJafar:
## Changelog Description
- Add Tags Enum to integrator, it fetches the tags list from the project anatomy.
- Add tags to version entity on integrator and hero integrator

image

## Testing notes:
1. Set tags in publisher
2. Publish


BigRoy:
Can we move the version tags setting of values into a separate plug-in, that can also be disabled, similar to @moonyuet 's work here: https://github.com/ynput/ayon-core/pull/1847

BigRoy (changes requested):
As [stated above](https://github.com/ynput/ayon-core/pull/1850#issuecomment-4502146376):
- make sure it's a separate plug-in

But also:
- make sure it can be disabled in settings
- have it disabled by default so the attribute does not show in publisher UI.
- if tags are already in versionData when it reaches this plug-in with the attribute def, then append the selected tags

The only change in integrator should be that it'd be able to take this data from `instance.data` and apply it to the version.

Later we may want to add profiles for defaults, but I don't think we should now. Because it'll open a can of worms of what may need to be adjustable (like which tags you might be allowed to select from per context and whatnot)

MustafaJafar:
> The only change in integrator should be that it'd be able to take this data from `instance.data` and apply it to the version.

I'm currently depending on `instance.data["versionData"]["tags"]` 
I'm now checking relying on `instance.data["tags"]`.

edit: I checked `tags` in instance data, I only found we add them below representations. so dedicating the instance data tags for versions seems fine.

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -250,6 +250,7 @@ def integrate_instance(
             data=copy.deepcopy(src_version_entity["data"]),
             attribs=copy.deepcopy(src_version_entity["attrib"]),
             entity_id=entity_id,
+            tags=instance.data.get("tags", None),
```
MustafaJafar: I didn't add the same check from integrate.py as it has already been done once.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,64 @@
...
+    def process(self, instance):
+
+        tags = instance.data.setdefault("tags", [])
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        version_tags = attr_values.get("version_tags", [])
+
+        if (
+            isinstance(tags, (list, tuple, set))
+            and all(isinstance(tag, str) for tag in tags)
```
BigRoy: If we really think something in this iterable may not be a `str` I think we should iterate over and warn on those that aren't. Now it'd just silently start skipping them which may seem confusing. I'd rather have it error then so I know about it instead of ignoring them completely. A warning is a good middle ground.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,64 @@
...
+        ):
+            tags.extend([t for t in version_tags if t not in tags])
+
+        self.log.debug(f"Collected Tags: {tags}")
+
+    @classmethod
+    def get_attr_defs_for_instance(cls, create_context, instance):
```
BigRoy, line 44: Please confirm the attributes also don't show if the plug-in is disabled. I forgot whether `get_attr_defs_for_instance` gets called or not if the class was not enabled. If it does, we'll need to return `[]` if `not cls.enabled`.

client/ayon_core/plugins/publish/integrate_hero_version.py
BigRoy: There is the caveat that `instance.data.tags` would still hold the wrong data (because we don't actually filter it and store the filtered back into `instance.data.tags`

client/ayon_core/plugins/publish/collect_version_tags.py
MustafaJafar: > Please confirm the attributes also don't show if the plug-in is disabled.

they disappear when plugin is disabled.

client/ayon_core/plugins/publish/collect_version_tags.py
MustafaJafar: Tbh, I've no hard opinion about what we really think. 
I went through code roughly in core addon and other addons as well, I didn't find any usage for 
`instance.data["tags"]` . some times we have `"tags"` key but only on representation data.
Also, I'm not sure why/how it won't be a list or tuple or set either.

I merely added this check because I copied it from the issue description.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,64 @@
...
+
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "CollectVersionTags"
+    settings_category = "core"
+
+    enabled = False
+
+    def process(self, instance):
+
+        tags = instance.data.setdefault("tags", [])
```
BigRoy: Have we considered making this a `set` actually?

BigRoy (approved):
Seems to work as intended.

client/ayon_core/plugins/publish/collect_version_tags.py
BigRoy:
```suggestion
        tags: set[str] = set(instance.data.get("tags", set()))
```

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,64 @@
...
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        version_tags = attr_values.get("version_tags", [])
+
+        if (
+            isinstance(tags, (list, tuple, set))
+            and all(isinstance(tag, str) for tag in tags)
+        ):
+            tags.extend([t for t in version_tags if t not in tags])
```
BigRoy:
```suggestion
        tags.update(version_tags)
        instance.data["tags"] = tags
```

With the set conversion logic above we will need to apply it back, because we're making a copy of the `set` (or iterable because we're converting it.) Hence the `setdefault` logic wouldn't make much sense there.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,64 @@
...
+    settings_category = "core"
+
+    enabled = False
+
+    def process(self, instance):
+
+        tags = instance.data.setdefault("tags", [])
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        version_tags = attr_values.get("version_tags", [])
```
BigRoy, line 27:
```suggestion
        version_tags: list[str] = attr_values.get("version_tags", [])
```

BigRoy (commented):
Just some convert to `set` nodes.

client/ayon_core/plugins/publish/collect_version_tags.py
MustafaJafar: Well, I just wish to have data structures for instances. 
And, this will be where we can say this is a set. 

But, let's start dealing with it as set.

MustafaJafar: It seems the api doesn't expect a set.
```python
File "E:\Ynput\ayon-launcher\.venv\Lib\site-packages\ayon_api\server_api.py", line 2529, in _prepare_operations_body
    raise ValueError("Couldn't json parse body: {}".format(
ValueError: Couldn't json parse body: {
    "id": "486a7084-e856-4a76-8742-458e5917acca",
    "type": "create",
    "entityType": "version",
    "entityId": "189def0555ab11f188adeb9ed0a5648a",
    "data": {
        "id": "189def0555ab11f188adeb9ed0a5648a",
        "version": 10,
        "productId": "70b38e06537311f1805577bc0084b141",
        "attrib": {
            "families": [
                "model",
                "usd",
                "usdrop"
            ],
            "source": "{root[work]}/Trash_Can/assets/props/crush_soda_can/work/modeling/trshcn_crush_soda_can_modeling_v001.hip",
            "comment": "",
            "machine": "DESKTOP-JHL3D9H",
            "fps": 24.0,
            "frameStart": 1001,
            "frameEnd": 1001,
            "handleEnd": 0,
            "handleStart": 0
        },
        "data": {
            "time": "20260522T085434Z",
            "author": "admin",
            "ayon_app_name": "houdini/21.0.700",
            "host_name": "houdini"
        },
        "taskId": "6eef052251d811f1af2fe3d02c3fdec3",
        "tags": "< Failed value <class 'set'> > {'for reel', 'important'}"
    }
}
```

MustafaJafar:
@BigRoy Hey, I've converted it to be a set and adjusted the integrator plugin a bit to work with tags as a set.
It works on my side. feel free to merge it if there are no further comments.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,61 @@
...
+
+        tags: set[str] = set(instance.data.get("tags", set()))
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        version_tags: list[str] = attr_values.get("version_tags", [])
+
+        tags.update(version_tags)
+        instance.data["tags"] = tags
+
+        self.log.debug(f"Collected Tags: {tags}")
```
BigRoy:
```suggestion
        attr_values = self.get_attr_values_from_data(instance.data)
        version_tags: list[str] = attr_values.get("version_tags", [])
        if not version_tags:
            return

		self.log.debug(f"Adding tags: {version_tags}")
        tags: set[str] = set(instance.data.get("tags", set()))
        tags.update(version_tags)
        instance.data["tags"] = tags
        self.log.debug(f"Collected tags: {tags}")
```

Should we skip doing anything to `tags` if no version tags have been selected at all?

BigRoy (approved):
LGTM - just a cosmetic note. @iLLiCiTiT can you glance over it once more (incl. the cosmetic note) and merge if you're happy.

client/ayon_core/plugins/publish/collect_version_tags.py
MustafaJafar: makes sense 
https://github.com/ynput/ayon-core/pull/1850/commits/da27016cd3c3e6277d3f51ca334633465fa37473

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,63 @@
...
+        project_entity = create_context.get_current_project_entity()
+
+        for tag in project_entity["tags"]:
+
+            items.append(
+                {
+                    "label": tag["name"],
+                    "value": tag["name"],
+                }
+            )
```
iLLiCiTiT: No need to created dict items if value and label are the same.
```suggestion
        project_entity = create_context.get_current_project_entity()
        items = list(project_entity["tags"])
        if not items:
            return []

```

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,63 @@
...
+            return []
+
+        project_entity = create_context.get_current_project_entity()
+        items = [
+            {
+                "label": tag["name"],
+                "value": tag["name"],
+            }
+            for tag in project_entity["tags"]
+        ]
```
iLLiCiTiT:
```suggestion
        items = [tag["name"] for tag in project_entity["tags"]]
```

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,63 @@
...
+        attr_values = self.get_attr_values_from_data(instance.data)
+        version_tags: list[str] = attr_values.get("version_tags", [])
+
+        if not version_tags:
+            return
+
+        self.log.debug(f"Adding tags: {version_tags}")
+        tags: set[str] = set(instance.data.get("tags", set()))
+        tags.update(version_tags)
+        instance.data["tags"] = tags
```
iLLiCiTiT: I'm really not keen on using `tags` key, it should be `versionTags` at least.
Also if we do allow it to be `list` or `tuple` then we have to validate and convert is here if it is already filled then make sure it is a `set`.
```suggestion
        tags = instance.data.get("versionTags")
        if tags is None:
            tags = set()
        elif not isinstance(tags, set):
            tags = set(tags)
            
        tags.update(version_tags)
        instance.data["versionTags"] = tags
```

client/ayon_core/plugins/publish/collect_version_tags.py
MustafaJafar: This is how tags from project entity looks like.
it doesn't have a label or value.
will this still be okay with the EnumDef
```
[
    {'name': 'important', 'color': '#ff2450'}, 
    {'name': 'for reel', 'color': '#5be1c6'}
]
```

client/ayon_core/plugins/publish/collect_version_tags.py
MustafaJafar: > if we do allow it to be `list` or `tuple`

I don't think we force it to be a specific type as there are no structures/data models for instance data as far as I know.

I checked core addon and some dcc addons, we don't have `"tags"` or `"versionTags"` on `instance.data` level.

BigRoy: It's allowed to be other types currently - but in this case `set` I think is the best type for it so we'll want to convert it.
I do agree with `versionTags` over `tags` actually, so let's do that yes.  (of course update the integrator code for that too)

So fine with this code change.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,60 @@
...
+
+        self.log.debug(f"Adding tags: {version_tags}")
+        tags = instance.data.get("versionTags")
+        if tags is None:
+            tags = set()
+        elif not isinstance(tags, set):
+            tags = set(tags)
+
+        tags.update(version_tags)
+        self.log.debug(f"Collected version tags: {tags}")
```
iLLiCiTiT, line 41:
```suggestion
        instance.data["versionTags"] = tags
        self.log.debug(f"Collected version tags: {tags}")
```

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @classmethod
+    def get_attr_defs_for_instance(cls, create_context, instance):
+        if not cls.instance_matches_plugin_families(instance):
+            return []
+
+        project_entity = create_context.get_current_project_entity()
```
iLLiCiTiT: There are no families filters on the plugin
```suggestion
        project_entity = create_context.get_current_project_entity()
```

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -243,13 +243,20 @@ def integrate_instance(
         if old_version:
             entity_id = old_version["id"]
 
+        tags = instance.data.get("version_tags", None)
```
iLLiCiTiT:
```suggestion
        tags = instance.data.get("version_tags")
```

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,60 @@
...
+        items = [tag["name"] for tag in project_entity["tags"]]
+        if not items:
+            return []
```
iLLiCiTiT, line 51: I think the conversation was here already, or it was different PR?

In case `items` are empty you have to pop the attribute value from the instance if it was set. If the instance already has set tags, and now items would be empty, the old value would stay on the instance.

So either pop the value, or store the information if tags were set on the instance itself, e.g. as another hidden attribute that is always returned from this method.

I think it is safe to pop it.
```suggestion
            plugin_attributes = instance.publish_attributes.get(cls.__name__)
            if plugin_attributes is not None:
                plugin_attributes.pop("version_tags", None)
            return []
```

MustafaJafar:
@iLLiCiTiT Thank you. let me know if there are further actions.


@BigRoy it's worth mentioning that the enum disappears if the project has no tags. is this the desired UX?

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +476,23 @@ def prepare_version(
         host_name = instance.context.data["hostName"]
         version_data["host_name"] = host_name
 
+        tags = instance.data.get("versionTags", None)
+        if tags is not None:
+            # Check if tags is an iteratable.
+            if not isinstance(tags, (list, tuple, set)):
+                raise PublishError(
+                    "Tags must be an itertable."
```
BigRoy:
```suggestion
                    "Tags must be an iterable."
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +476,23 @@ def prepare_version(
...
+                    f"Instead got type {type(tags)}"
```
BigRoy:
```suggestion
                    f" Instead got type {type(tags)}"
```

BigRoy (approved):
LGTM - didn't test again though.
Just one cosmetic change to the logged message.

client/ayon_core/plugins/publish/collect_version_tags.py
```diff
@@ -0,0 +1,61 @@
...
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Collect Version Tags"
+    settings_category = "core"
+
+    enabled = False
+
+    def process(self, instance):
+
+        attr_values = self.get_attr_values_from_data(instance.data)
+        version_tags: list[str] = attr_values.get("version_tags", [])
```
iLLiCiTiT: `list[str]` might break in older pythons and there is no need for default value.
```suggestion
        version_tags = attr_values.get("version_tags")
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +476,23 @@ def prepare_version(
         host_name = instance.context.data["hostName"]
         version_data["host_name"] = host_name
 
+        tags = instance.data.get("versionTags", None)
+        if tags is not None:
+            # Check if tags is an iteratable.
+            if not isinstance(tags, (list, tuple, set)):
+                raise PublishError(
```
iLLiCiTiT, line 505: This error is not for artists, I would say this should use `KnownPublishError`.
```suggestion
                raise KnownPublishError(
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +476,23 @@ def prepare_version(
...
+                    "Tags must be an itertable."
+                    f"Instead got type {type(tags)}"
+                )
+            # Error if not all are string.
+            if not all(isinstance(tag, str) for tag in tags):
+                raise PublishError(
```
iLLiCiTiT, line 511:
```suggestion
                raise KnownPublishError(
```

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -243,13 +243,20 @@ def integrate_instance(
         if old_version:
             entity_id = old_version["id"]
 
+        tags = instance.data.get("version_tags")
```
iLLiCiTiT:
```suggestion
        tags = instance.data.get("versionTags")
```

BigRoy:
@MustafaJafar would be great if you can finish above comments 🙇‍♂️

MustafaJafar:
@BigRoy I pushed changes and gave it another test run on my end and it worked.

client/ayon_core/plugins/publish/integrate.py
BigRoy: Isn't this recommendation wrong @iLLiCiTiT 

seeing that `KnownPublishError` is 'deprecated', see: https://github.com/ynput/ayon-core/blob/70b22b1c0e0a9d72240fcca1b583ddb25636a9ab/client/ayon_core/pipeline/publish/publish_plugins.py#L36-L44

BigRoy: @iLLiCiTiT just checking?

iLLiCiTiT: Oh, I think that deprecation should be removed. Sometimes it makes sense to raise an error that doesn't have to have "user-friendly" message and title.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +475,23 @@ def prepare_version(
...
+            if not isinstance(tags, (list, tuple, set)):
+                raise KnownPublishError(
+                    "Tags must be an iterable."
+                    f" Instead got type {type(tags)}"
+                )
+            # Error if not all are string.
+            if not all(isinstance(tag, str) for tag in tags):
+                raise KnownPublishError(
+                    "Version tags must be an iterable of strings. "
+                    "Got: {}".format(repr(tags))
```
iLLiCiTiT: `repr` does not output the type, the type can implement `__repr__` which can be anything
```suggestion
                    f"Got: {type(tags)}"
```

client/ayon_core/plugins/publish/integrate.py
BigRoy: Can you give some example on how to use that one differently from PublishError?

BigRoy: @iLLiCiTiT your repo, your call at this point - make the change to your liking and merge it if the rest is ok for you. Let's discuss PublishError vs KnownPublishError sometime soon.

iLLiCiTiT: `PublishError` does show the error message to user, or no? `KnowPublishError` shows something like "something went wrong".

iLLiCiTiT: Nope, I'm wrong, it just shows the message of the expcetion. In that case we don't have a way how to show generic user error message with `PublishError`.

BigRoy: I'd rather have more errors that are more descriptive, than less. So we should really be recommending `KnownPublishError` if that's the only one actually showing anything remotely useful to the user. I think always if we raise a dedicated error, even if obscure. _Anything_ is better than "Not your fault" message - it's one of the bigger complaints I've been hearing over time. The not your fault message telling you nothing.

Anyway, let's make sure it's the one that is more descriptive to the user - and if that is `KnownPublishError` then use that. And also make sure it's not marked deprecated anymore.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +475,23 @@ def prepare_version(
         host_name = instance.context.data["hostName"]
         version_data["host_name"] = host_name
 
+        tags = instance.data.get("versionTags", None)
+        if tags is not None:
+            # Check if tags is an iteratable.
+            if not isinstance(tags, (list, tuple, set)):
+                raise KnownPublishError(
```
iLLiCiTiT:
```suggestion
                raise PublishError(
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -475,6 +475,23 @@ def prepare_version(
...
+                    "Tags must be an iterable."
+                    f" Instead got type {type(tags)}"
+                )
+            # Error if not all are string.
+            if not all(isinstance(tag, str) for tag in tags):
+                raise KnownPublishError(
```
iLLiCiTiT:
```suggestion
                raise PublishError(
```

client/ayon_core/plugins/publish/integrate.py
iLLiCiTiT: `KnownPublishError` does what `PublishError` does by default.

The thing is that sometimes you want to write technical information for a dev to the message which probably should not be shown to an artist.

BigRoy: > The thing is that sometimes you want to write technical information for a dev to the message which probably should not be shown to an artist.

I see - so both show the message, but one (which one?) allows to attach _additional messages_ that are shown elsewhere? if so, where?

iLLiCiTiT: > I see - so both show the message, but one (which one?) allows to attach additional messages that are shown elsewhere? if so, where?

No, I thought that `KnownPublishError` does not show the message to artist -> so you could write more technical information there, but I was wrong.

The point is `PublishError` should be used (because it does what `KnownPublishError` does) and `KnownPublishErro` is really deprecated. To add more information to the error we would have to modify pyblish, or publish report viewer.

client/ayon_core/plugins/publish/integrate_hero_version.py
```diff
@@ -243,13 +243,20 @@ def integrate_instance(
         if old_version:
             entity_id = old_version["id"]
 
+        tags = instance.data.get("versionTags")
+        if tags is not None:
+            # Tags contents are checked at eariler step "IntegrateAsset"
```
BigRoy:
```suggestion
            # Tags contents are checked at earlier step "IntegrateAsset"
```

BigRoy:
@MustafaJafar can you resolve conflicts?

@iLLiCiTiT then ready to merge?

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -498,6 +499,23 @@ def prepare_version(
         host_name = instance.context.data["hostName"]
         version_data["host_name"] = host_name
 
+        tags = instance.data.get("versionTags", None)
```
iLLiCiTiT:
```suggestion
        tags = instance.data.get("versionTags")
```

iLLiCiTiT (approved):
Code looks ok. Tags are added to version,

PR #1852, opened 2026-05-21
Push to project: Re-upload reviewables

### LINKED ISSUE #1704: YN-0499: Library include reviewable during copy

We need to include reviewable when using Library copy function. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1347586551923867648/1472937696606556263)

_This issue was automatically created from AYON ticket [YN-0499]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=7ad546100b5111f1ab620584a12429a3)

iLLiCiTiT:
## Changelog Description
Use reviewables from source version in the new version.

## Additional info
It is necessary to download the file and upload it back for the new version.

## Testing notes:
1. Open browser.
2. Select a version with reviewables and run `Push to project` action on it.
3. Fill form for the push.
4. Hit Publish.
5. New version in AYON should have the reviewable.

Resolves https://github.com/ynput/ayon-core/issues/1704

### DIFF
--- a/client/ayon_core/tools/push_to_project/models/integrate.py
+++ b/client/ayon_core/tools/push_to_project/models/integrate.py
@@ -2,6 +2,7 @@ import os
 import re
 import copy
 import itertools
+import shutil
 import sys
 import tempfile
 import traceback
@@ -442,6 +443,7 @@ class ProjectPushItemProcess:
         self._src_folder_entity = None
         self._src_product_entity = None
         self._src_version_entity = None
+        self._src_reviewable_files = None
         self._src_repre_items = None
 
         self._project_entity = None
@@ -494,6 +496,8 @@ class ProjectPushItemProcess:
             self._make_sure_version_exists()
             self._log_info("Prerequirements were prepared")
             self._integrate_representations()
+            self._log_info("Representations created")
+            self._reupload_reviewables()
             self._log_info("Integration finished")
 
         except PushToProjectError as exc:
@@ -604,7 +608,7 @@ class ProjectPushItemProcess:
             ))
             raise PushToProjectError(self._status.fail_reason)
 
-        anatomy = Anatomy(src_project_name)
+        anatomy = Anatomy(src_project_name, project_entity=project_entity)
 
         repre_entities = ayon_api.get_representations(
             src_project_name,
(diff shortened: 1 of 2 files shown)


client/ayon_core/tools/push_to_project/models/integrate.py
```diff
@@ -1418,6 +1436,45 @@ def _create_src_version_description(
...
+                    )
+
+                ext = os.path.splitext(dst_path)[1].lower()
+                counter = names_by_ext.setdefault(ext, 0)
+                counter_str = ""
+                if counter > 0:
+                    counter_str = f"_{counter + 1}"
+                names_by_ext[ext] += 1
+
+                filename = f"{self._product_name}{counter_str}"
```
BigRoy: Should we try and match the original source filename from the original upload?

BigRoy (approved):
Looks good, works for me - reviewables came through as well and can be played on web.

I did wonder whether we should keep the filename consistent though?

Original:

Image

Pushed:

Image

From a production perspective I'd prefer the filename to match the original I pulled it from.

client/ayon_core/tools/push_to_project/models/integrate.py
iLLiCiTiT: By default we use filename that is calculated based on publish template. The default publish template is using project name/code in the filename, is that good idea?

BigRoy: I believe the filename for the reviewables is whatever you uploaded it with - yes from desktop tools it'd have the "original names" from that product, but you may have uploaded them manually as well from some other source.

But maybe preserving that name could still be more confusing than being worth to track what file it was originally.

@mkolar what do you reckon here from your perspective?

64qam (approved):

mkolar:
We should definitely keep the name of the original file if that's how it was uploaded in the first place

client/ayon_core/tools/push_to_project/models/integrate.py
BigRoy: As stated [here](https://github.com/ynput/ayon-core/pull/1852#issuecomment-4517340686) let's preserve the original filename please.

BigRoy (changes requested):
Change:
- [x] It should preserve original filename of the reviewables on push.

client/ayon_core/tools/push_to_project/models/integrate.py
```diff
@@ -1418,6 +1436,45 @@ def _create_src_version_description(
...
+                    )
+
+                ext = os.path.splitext(dst_path)[1].lower()
+                counter = names_by_ext.setdefault(ext, 0)
+                counter_str = ""
+                if counter > 0:
+                    counter_str = f"_{counter + 1}"
+                names_by_ext[ext] += 1
+
+                filename = f"{self._product_name}{counter_str}{ext}"
```
BigRoy: Filename should be inheriting from the source.
For me it doesn't currently.

I think you forgot to push @iLLiCiTiT ?

BigRoy (changes requested):

BigRoy (approved):
Works.

PR #1853, opened 2026-05-21
Integrate: Representation rootless paths

iLLiCiTiT:
## Changelog Description
Prefer root from publish template for representation files information.

## Additional info
In case someone has 2 roots that are the same or one is nested under other one the `find_root_template_from_path` might not return the root which was actually used for publishing. In my case I had `publish` and `delivery` roots set to similar paths, the representation template attribute had `{root[publish]}` but rootless paths on representation files info have `{root[delivery]}`.

Renamed `RootItem` in antomy code to `AnatomyRoot` as it makes more sense. Kept `RootItem` for backwards compatibility but it should not be used anywhere, yet.

Bug discovered during https://github.com/ynput/ayon-core/pull/1852 . I was not able to push to other project because of that.

## Testing notes:
1. Define 2 roots with the same path on a project.
2. Use the second (alphabetically) in publish template.
3. Publish a representation.
4. Check the representation `attrib.template` and `files.0.path` if the root used in the template is the same.

### DIFF
--- a/client/ayon_core/pipeline/anatomy/roots.py
+++ b/client/ayon_core/pipeline/anatomy/roots.py
@@ -8,7 +8,7 @@ from ayon_core.lib.path_templates import FormatObject
 from .exceptions import RootMissingEnv
 
 
-class RootItem(FormatObject):
+class AnatomyRoot(FormatObject):
     """Represents one item or roots.
 
     Holds raw data of root item specification. Raw data contain value
@@ -243,6 +243,11 @@ class RootItem(FormatObject):
         return (result, output)
 
 
+# TODO remove
+# Backwards compatibility added 2026/05/21
+RootItem = AnatomyRoot  # DEPRECATED
+
+
 class AnatomyRoots:
     """Object which should be used for formatting "root" key in templates.
 
@@ -296,7 +301,7 @@ class AnatomyRoots:
             src_platform (Optional[str]): Specify source platform. This is
                 recommended to not use and keep unset until you really want
                 to use specific platform.
-            roots (Optional[Union[dict, RootItem])): It is possible to remap
+            roots (Optional[Union[dict, AnatomyRoot])): It is possible to remap
                 path with different roots then instance where method was
                 called has.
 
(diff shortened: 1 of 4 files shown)


client/ayon_core/pipeline/anatomy/roots.py
```diff
@@ -243,6 +243,11 @@ def find_root_template_from_path(self, path):
         return (result, output)
 
 
+# TODO remove
+# Backwards compatibility added 2026/05/21
+RootItem = AnatomyRoot  # DEPRECATED
```
BigRoy: Can we add deprecation warning?

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -212,6 +221,19 @@ def register(self, instance, file_transactions, filtered_repres):
             )
 
         template_name = self.get_template_name(instance)
+        self.log.debug(f"Anatomy template name: {template_name}")
+        anatomy = instance.context.data["anatomy"]
+        publish_template = anatomy.get_template_item("publish", template_name)
+
+        # Prepare prefered root to use for representation files
```
BigRoy:
```suggestion
        # Prepare preferred root to use for representation files
```

BigRoy (approved):
Although I see how this could work I do feel like the code is overly convoluted - but seems necessary.

I wonder if similar issue then exists in Hero version integrator?

client/ayon_core/pipeline/anatomy/roots.py
iLLiCiTiT: We would have to add full class inheritance. And that would still not cover the only possible usage (which would be typehinting).

antirotor: but `RootItem` could inherit from `AnatomyRoot`, and have deprecation warning both in the class docstring and the constructor? That would solve it, no?

antirotor (approved):
Not sure about the Hero version integrator, but the trait one could be affected too. Can be fixed later.

https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/integrate_traits.py

PR #1861, opened 2026-05-26
Workfile templates: Save workfile entity when template is used

### LINKED ISSUE #1817: Setting option to save first workfile created from workfile_builder

In pre_copy_template_workfile for photoshop, blender, etc, the template is created and copied to the last_workfile_path but is never saved to DB. This means when user does a "save as" in photoshop eg, it saves a v002, but while a v001 exists in the workfile folder location, it is never saved, so isn't visable in launcher.  

This new template v001 file will always be a copy of the template, but creates a mismatch if it exists in workfiles but doesn't show in launcher, eg.

Save_created_workfile() function in ayon-core/hooks/pre_copy_template_workfile.py

save_created_workfile bool in photoshop, blender etc workfile_builder settings.

_No response_

### Additional context:

_No response_

iLLiCiTiT:
## Changelog Description
Save workfile entity when template is being used.

## Additional info
Because we do store only last workfile path, which can actually not hold last workfile ATM, we don't have access to rootless path and version available. We should change what we actually do store to launch context and store also additional information that are available when last workfile is being prepared so we don't have to parse it out recursively which can lead to use wrong rootless path (wrong root in the path) or wront version.

## Testing notes:
1. Make sure you don't have any workfiles yet available for a task.
2. Make sure to have set up settings to use template for first version.
3. Start a host on the task.
4. First workfile should be copied using the template and workfile entity should be created (can be validated in launcher tool).

Resolves https://github.com/ynput/ayon-core/issues/1817

### DIFF
--- a/client/ayon_core/hooks/pre_copy_template_workfile.py
+++ b/client/ayon_core/hooks/pre_copy_template_workfile.py
@@ -1,11 +1,19 @@
 import os
 import shutil
+
 from ayon_core.settings import get_project_settings
-from ayon_applications import PreLaunchHook, LaunchTypes
+
+from ayon_core.pipeline.template_data import get_template_data
 from ayon_core.pipeline.workfile import (
     get_custom_workfile_template,
-    get_custom_workfile_template_by_string_context
+    get_custom_workfile_template_by_string_context,
+    save_workfile_info,
+    find_workfile_rootless_path,
+    get_last_workfile_with_version_from_paths,
+    get_workfile_template_key,
 )
+
+from ayon_applications import PreLaunchHook, LaunchTypes
 
 
 class CopyTemplateWorkfile(PreLaunchHook):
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Makes sense, but also feels like this should be possible with fewer lines of code with some higher level API but I guess those are all focused on running within a DCC when host is available already, which is not the case here.

Didn't test it though.

iLLiCiTiT:
> but also feels like this should be possible with fewer lines of code with some higher level API

The biggest issue is that the last workfile path is just string path. No information about template, so there is no way how to get rootless path 100% accurate. Using `find_rootless_path` uses "first matching" which does not have to be the one which was actually used (that causes issues for rootless paths on representations already). So the ideal way is to store more information about last workfile to context. That has few pre-requirements affecting all host addons.

PR #1860, opened 2026-05-26
Add `editorial_pkg` product type to thumbnail extractor

jakubjezek001:
## Changelog Description
`editorial_pkg` was missing from the supported `families` in the `ExtractThumbnail` plugin, causing editorial package products to never generate thumbnails.

## Additional info
Single-line change in `client/ayon_core/plugins/publish/extract_thumbnail.py` — `"editorial_pkg"` appended to the `families` list of the `ExtractThumbnail` pyblish plugin.

## Testing notes:
1. Publish an editorial package product (e.g. from a host that produces `editorial_pkg` instances).
2. Confirm a thumbnail is extracted and visible in the target tracker (e.g. ftrack).

## Dependency
Close #1859
[YN-0626](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=task&id=da676290325a11f1b5cd3f72ee2f91ca)

PR #1863, opened 2026-06-01
Remove `hosts` and `families` filtering from `ExtractThumbnail`

BigRoy:
## Changelog Description

Remove `hosts` and `families` filtering from `ExtractThumbnail`

## Additional info

Follow up to: https://github.com/ynput/ayon-core/pull/1860#discussion_r3332652060

This will need some follow up changes. The class name for example is not unique. Same class name exists in:
```
client/ayon_core/plugins/publish/extract_thumbnail.py
client/ayon_maya/plugins/publish/extract_thumbnail.py
client/ayon_hiero/plugins/publish/extract_thumbnail.py
client/ayon_blender/plugins/publish/extract_thumbnail.py
client/ayon_max/plugins/publish/extract_thumbnail.py
```

This will also need thorough testing whether it does not conflict with logic in certain DCCs like e.g. the ones mentioned here that also roll their own Extract Thumbnail plug-in.

## Testing notes:

1. Extract Thumbnail should work
2. Extract Thumbnail should not cause issues in any of the DCCs/integrations.

### DIFF
--- a/client/ayon_core/plugins/publish/extract_thumbnail.py
+++ b/client/ayon_core/plugins/publish/extract_thumbnail.py
@@ -93,24 +93,7 @@ class ExtractThumbnail(pyblish.api.InstancePlugin):
 
     label = "Extract Thumbnail"
     order = pyblish.api.ExtractorOrder + 0.49
-    families = [
-        "imagesequence", "render", "render2d", "prerender",
-        "source", "clip", "take", "online", "image"
-    ]
-    hosts = [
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
NOTE: We should also find out why the representations are being removed here if they have `"delete"` tag. That looks as potential rist.

I think the other DCCs are using in-DCC logic for thumbnail extraction. E.g. maya viewport.

PR #1868, opened 2026-06-01
Push to project: Match task types case-insensitively

vincentullmann:
# DRAFT PR:
not 100% sure if this is the correct approach. Will raise it in discord and update the PR later based on the outcome

## Changelog Description
lookup task types case-insensitive

## Additional info
- moved the `task_type in self._project_entity["taskTypes"]` loop into a `_get_dst_task_type` helper method
- updated `_fill_or_create_destination_folder` and `_make_sure_task_exists` to use the helper
- updated `_get_dst_folder_type` to the same shape as `_get_dst_task_type`

## Testing notes:
1. start with this step
2. follow this step

### DIFF
--- a/client/ayon_core/tools/push_to_project/models/integrate.py
+++ b/client/ayon_core/tools/push_to_project/models/integrate.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import os
 import re
 import copy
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/push_to_project/models/integrate.py
```diff
@@ -880,14 +903,8 @@ def _fill_or_create_destination_task(self):
...
-        task_type_name = task_info["taskType"]
-        task_types_by_name = {
-            task_type["name"]: task_type
-            for task_type in self._project_entity["taskTypes"]
-        }
-        task_type_info = copy.deepcopy(
-            task_types_by_name.get(task_type_name, {})
-        )
+        task_type_info = self._get_dst_task_type(task_info["taskType"])
+        task_type_info = copy.deepcopy(task_type_info)
```
iLLiCiTiT, line 903:
```suggestion
        task_type_info = copy.deepcopy(
            self._get_dst_task_type(task_info["taskType"])
        )
```

iLLiCiTiT:
I think it is not necessary, the one part of code is preparing existing task by name and using it's task type, if it is there, the task type is 100% on the project.

Didn't go deep into the code, but looks like it.

vincentullmann:
closing in favor of #1869

PR #1869, opened 2026-06-01
Push To Project: Use task type capitalisation matched in destination project

vincentullmann:
## Changelog Description
changes the task_type lookup to use the capitalisation found on the destination project

## Additional info
Fixes an issue where the same task type uses different capitalisation across projects.

task type matching is already case-insensitive in most places, but the created/returned `taskType` did still use the capitalisation from the source project.
With this change, the task type returned will use the capitalisation found in the destination project.

## Testing notes:
1. publish an asset to project A using a task_type with 1 capitalisation (eg.: ``"Animation"`)
2. update anatomy on project B to have a different capitalisation for the same task type (eg.: `"ANIMation"`)
3. push the product from project A to B


BigRoy (approved):
LGTM

iLLiCiTiT (approved):

PR #1871, opened 2026-06-03
Integrate published versions into AYON entity lists

jakubjezek001:
## Changelog Description
Two new publish plugins enable automatic population of published versions into AYON entity lists. A collector resolves list names from anatomy templates and profile-based settings; an integrator creates or appends to those lists on the server — including on-demand folder hierarchy creation and `review-session` typing. 

## Additional info
- `IntegrateVersionToList` (`IntegratorOrder + 0.49`) groups version IDs by list name, then creates new lists or appends to existing ones. List folder hierarchies are created on demand via direct `ayon_api` calls (temporary workaround until `ayon_api` exposes these endpoints natively — marked with `TODO` comments).
- Server-side settings expose a `CollectVersionToList` profile model with per-host/task/product filtering and per-list configuration (label template, list folders, review list flag).

## Testing notes:
1. Configure a `CollectVersionToList` profile in AYON settings (Studio → Publish → Collect Version to List) targeting a host and product type.
2. Publish an asset matching that profile.
3. Verify a new entity list appears in the AYON server project with the expected name and the published version as a member.
4. Publish a second version with the same list name — confirm it appends to the existing list rather than creating a duplicate.
5. Test with `list_folders` configured — confirm the folder hierarchy is created on the server.
6. Test with `list_type:review-session` — confirm the list type is set to `review-session`.

## Dependency
Close ynput/ayon-traypublisher#148
[YN-0770](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=b255809cc9a64de79c4617fc95d89256)

### DIFF
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -7,8 +7,16 @@ ListType = Literal["generic", "review-session"]
 
 
 @dataclass
+class ListConfigFolder:
+    label: str
+    scope: list[str] = field(default_factory=list)
+    color: str | None = None
+    icon: str | None = None
+
+
+@dataclass
 class ListConfig:
     """Define a list."""
     name: str
     list_type: ListType = "generic"
-    list_folders: list[str] = field(default_factory=list)
+    list_folders: list[ListConfigFolder] = field(default_factory=list)
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -1,0 +1,13 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Literal
+
+ListType = Literal["generic", "review-session"]
+
+@dataclass
+class ListConfig:
+    """Define a list."""
+    name: str
+    parent_folders: list[str] | None = None
+    list_type: ListType = "generic"
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -10,5 +10,5 @@ ListType = Literal["generic", "review-session"]
 class ListConfig:
     """Define a list."""
     name: str
-    parent_folders: list[str] = field(default_factory=list)
     list_type: ListType = "generic"
+    list_folders: list[str] = field(default_factory=list)
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -9,7 +9,8 @@ ListType = Literal["generic", "review-session"]
 @dataclass
 class ListConfigFolder:
     label: str
-    scope: list[str] = field(default_factory=list)
+    # Empty list means the folder is visible for all list types
+    scope: list[ListType] = field(default_factory=list)
     color: str | None = None
     icon: str | None = None
 
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Literal
 
 ListType = Literal["generic", "review-session"]
@@ -10,5 +10,5 @@ ListType = Literal["generic", "review-session"]
 class ListConfig:
     """Define a list."""
     name: str
-    parent_folders: list[str] | None = None
+    parent_folders: list[str] = field(default_factory=list)
     list_type: ListType = "generic"
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -18,6 +18,6 @@ class ListConfigFolder:
 @dataclass
 class ListConfig:
     """Define a list."""
-    name: str
+    label: str
     list_type: ListType = "generic"
     list_folders: list[ListConfigFolder] = field(default_factory=list)
--- a/client/ayon_core/pipeline/structures.py
+++ b/client/ayon_core/pipeline/structures.py
@@ -5,6 +5,7 @@ from typing import Literal
 
 ListType = Literal["generic", "review-session"]
 
+
 @dataclass
 class ListConfig:
     """Define a list."""
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -42,9 +42,8 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
         if profile:
             processing_items.append(profile)
 
-        anatomy_data: dict[str, Any] = instance.data["anatomyData"]
         anatomy: Anatomy = instance.context.data["anatomy"]
-        template_data = deepcopy(anatomy_data)
+        template_data = deepcopy(instance.data["anatomyData"])
         template_data.update({
             "root": anatomy.roots,
             "platform": platform.system().lower(),
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -29,31 +29,43 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
     profiles = []
 
     def process(self, instance):
+        version_lists_templates = instance.data.get(
+            "versionListsTemplates", [])
         profile = self._get_config_from_profile(instance)
-        if not profile:
+        if not profile and not version_lists_templates:
             self.log.debug(f"No profile found for instance {instance}")
             return
 
+        processing_items = []
+        if version_lists_templates:
+            processing_items += version_lists_templates
+        if profile:
+            processing_items.append(profile)
+
         anatomy_data: dict[str, Any] = instance.data["anatomyData"]
         anatomy: Anatomy = instance.context.data["anatomy"]
-        name_template = profile["name_template"]
         template_data = deepcopy(anatomy_data)
         template_data.update({
             "root": anatomy.roots,
             "platform": platform.system().lower(),
         })
-
-        list_name = StringTemplate.format_strict_template(
-            name_template, template_data)
         version_lists: list[ListConfig] = instance.data.setdefault(
             "versionLists", [])
-        version_lists.append(
-            ListConfig(
-                name=list_name,
-                parent_folders=profile.get("parent_folders", None),
-                is_review_list=profile.get("is_review_list", False),
+        for item in processing_items:
+            name_template = item["name_template"]
+            data = item.get("data", {})
+            if data:
+                template_data.update(data)
+            list_name = StringTemplate.format_strict_template(
+                name_template, template_data)
+            version_lists.append(
+                ListConfig(
+                    name=list_name,
+                    parent_folders=item.get("parent_folders", None),
+                    is_review_list=item.get("is_review_list", False),
+                )
             )
-        )
+        self.log.debug(f"Collected version lists: {version_lists}")
 
     def _get_config_from_profile(
         self,
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -28,14 +28,24 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
             self.log.debug(f"No profile found for instance {instance}")
             return
 
+        list_folders = []
+        for item in profile["list_folders"]:
+            scope = []
+            if item["scope_def"] == "list_type":
+                scope.append(profile["list_type"])
+
+            list_folders.append(
+                ListConfigFolder(
+                    label=item["label"],
+                    scope=scope,
+                )
+            )
+
         version_lists.append(
             ListConfig(
                 name=profile["list_name"],
                 list_type=profile["list_type"],
-                list_folders=[
-                    ListConfigFolder(label=name)
-                    for name in profile["parent_folders"]
-                ],
+                list_folders=list_folders,
             )
         )
         self.log.debug(f"Collected version lists: {version_lists}")
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -25,10 +25,10 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
             self.log.debug(f"Version lists already collected: {version_lists}")
             return
 
-        name_template = profile["name_template"]
+        name = profile["name"]
         version_lists.append(
             ListConfig(
-                name=name_template,
+                name=name,
                 parent_folders=profile.get("parent_folders", None),
                 list_type=profile["list_type"],
             )
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -43,7 +43,7 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
 
         version_lists.append(
             ListConfig(
-                name=profile["list_name"],
+                label=profile["list_name"],
                 list_type=profile["list_type"],
                 list_folders=list_folders,
             )
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -45,7 +45,6 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
 
         list_name = StringTemplate.format_strict_template(
             name_template, template_data)
-
         version_lists: list[ListConfig] = instance.data.setdefault(
             "versionLists", [])
         version_lists.append(
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -10,6 +10,7 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
 
     order = pyblish.api.CollectorOrder + 0.499
     label = "Collect Version to List"
+    settings_category = "core"
 
     profiles = []
 
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -30,7 +30,7 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
         version_lists.append(
             ListConfig(
                 name=name,
-                parent_folders=profile.get("parent_folders", None),
+                parent_folders=profile["parent_folders"],
                 list_type=profile["list_type"],
             )
         )
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -25,7 +25,7 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
             self.log.debug(f"Version lists already collected: {version_lists}")
             return
 
-        name = profile["name"]
+        name = profile["list_name"]
         version_lists.append(
             ListConfig(
                 name=name,
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -15,15 +15,17 @@ class CollectVersionToList(pyblish.api.InstancePlugin):
     profiles = []
 
     def process(self, instance):
+        if "versionLists" in instance.data:
+            version_lists = instance.data["versionLists"]
+            self.log.debug(f"Version lists already collected: {version_lists}")
+            return
+
+        version_lists: list[ListConfig] = []
+        instance.data["versionLists"] = version_lists
+
         profile = self._get_profile_for_instance(instance)
         if not profile:
             self.log.debug(f"No profile found for instance {instance}")
-            return
-        version_lists: list[ListConfig] = instance.data.setdefault(
-            "versionLists", [])
-
-        if version_lists:
-            self.log.debug(f"Version lists already collected: {version_lists}")
             return
 
         name = profile["list_name"]
--- a/client/ayon_core/plugins/publish/collect_version_to_list.py
+++ b/client/ayon_core/plugins/publish/collect_version_to_list.py
@@ -1,0 +1,96 @@
+from __future__ import annotations
+
+import platform
+from copy import deepcopy
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any
+
+import pyblish.api
+from ayon_core.lib import StringTemplate, filter_profiles
+
+if TYPE_CHECKING:
+    from ayon_core.pipeline import Anatomy
+
+
+@dataclass
+class ListConfig:
+    """Define a list."""
+    name: str
+    parent_folders: list[str] | None = None
+    is_review_list: bool = False
+
+
+class CollectVersionToList(pyblish.api.InstancePlugin):
+    """Collect Version to List."""
+
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Collect Version to List"
+
+    profiles = []
+
+    def process(self, instance):
+        profile = self._get_config_from_profile(instance)
+        if not profile:
+            self.log.debug(f"No profile found for instance {instance}")
+            return
+
+        anatomy_data: dict[str, Any] = instance.data["anatomyData"]
+        anatomy: Anatomy = instance.context.data["anatomy"]
+        name_template = profile["name_template"]
+        template_data = deepcopy(anatomy_data)
+        template_data.update({
+            "root": anatomy.roots,
+            "platform": platform.system().lower(),
+        })
+
+        list_name = StringTemplate.format_strict_template(
+            name_template, template_data)
+
+        version_lists: list[ListConfig] = instance.data.setdefault(
+            "versionLists", [])
+        version_lists.append(
+            ListConfig(
+                name=list_name,
+                parent_folders=profile.get("parent_folders", None),
+                is_review_list=profile.get("is_review_list", False),
+            )
+        )
+
+    def _get_config_from_profile(
+        self,
+        instance: pyblish.api.Instance
+    ) -> dict | None:
+        """Returns profile for the given instance."""
+        host_name = instance.context.data["hostName"]
+        product_base_type = instance.data.get("productBaseType")
+        if not product_base_type:
+            product_base_type = instance.data["productType"]
+        product_name = instance.data["productName"]
+        task_data = instance.data["anatomyData"].get("task", {})
+        task_name = task_data.get("name")
+        task_type = task_data.get("type")
+        filtering_criteria = {
+            "host_names": host_name,
+            "product_base_types": product_base_type,
+            "product_names": product_name,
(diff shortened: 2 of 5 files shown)


client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,107 @@
...
+if TYPE_CHECKING:
+    from ayon_core.pipeline import Anatomy
+
+
+@dataclass
+class ListConfig:
+    """Define a list."""
+    name: str
+    parent_folders: list[str] | None = None
+    is_review_list: bool = False
```
BigRoy: Just for the fact it may be any other list config - and also to clarify what it means if it's not a review list should we make this `list_type: ListType = "generic"` where `ListType` is `Literal["generic", "review-session"]`? @iLLiCiTiT what do you think?

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,107 @@
...
+            version_lists.append(
+                ListConfig(
+                    name=list_name,
+                    parent_folders=item.get("parent_folders", None),
+                    is_review_list=item.get("is_review_list", False),
+                )
+            )
+        self.log.debug(f"Collected version lists: {version_lists}")
+
+    def _get_config_from_profile(
```
BigRoy: Name is a bit odd, it doesn't get anything from a profile.. but it filters all the profiles. So this would either be `_get_profile_for_instance` or `_get_profile_from_profiles`?

```suggestion
    def _get_profile_for_instance(
```
This may make most sense?

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,107 @@
...
+            name_template = item["name_template"]
+            data = item.get("data", {})
+            if data:
+                template_data.update(data)
+            list_name = StringTemplate.format_strict_template(
+                name_template, template_data)
+            version_lists.append(
+                ListConfig(
+                    name=list_name,
+                    parent_folders=item.get("parent_folders", None),
```
BigRoy: I'd say that each entry of the parent folders should also be allowed to use formatting, so I think we should also trigger the strict formatting there too?

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,107 @@
...
+        })
+        version_lists: list[ListConfig] = instance.data.setdefault(
+            "versionLists", [])
+        for item in processing_items:
+            name_template = item["name_template"]
+            data = item.get("data", {})
+            if data:
+                template_data.update(data)
+            list_name = StringTemplate.format_strict_template(
+                name_template, template_data)
```
BigRoy: I would've personally done this formatting in the integrator itself - and the ListConfig would be the non-formatted one. That way, the `versionListsTemplates` key itself would also become redundant - because it'd just mean that other Collectors would just add to `versionLists` and not use some sort of `versionListTemplates` key.

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,252 @@
...
+
+    @staticmethod
+    def _existing_list(
+        project_name: str,
+        list_label: str,
+        list_type: str | None = None,
+    ) -> dict | None:
+        all_lists = ayon_api.get_entity_lists(
+            project_name=project_name,
+        )
```
BigRoy: We should query this once, and then have the logic just take the `existing_lists` key to avoid doing many queries.

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def _create_entity_list(
+        self,
+        project_name: str,
+        entity_type: str,
+        label: str,
+        items: list[dict],
+        list_type: str | None = None,
+        entity_list_folder_id: str | None = None,
+    ) -> dict:
```
BigRoy, line 373: Can we add a docstring?

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,252 @@
...
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+import ayon_api
+import pyblish.api
+from ayon_api.utils import create_entity_id
+
+if TYPE_CHECKING:
+    from ayon_core.plugins.publish.collect_version_to_list import ListConfig
```
BigRoy: I have a feeling we should instead put this `ListConfig` somewhere else so we avoid the need to import from the other file, even if it's only for type checking. @iLLiCiTiT any good place?

client/ayon_core/plugins/publish/collect_version_to_list.py
jakubjezek001: There are only two types of list entity - either `generic` or `review-sesison` and generic it is by default if nothing is set.

client/ayon_core/plugins/publish/collect_version_to_list.py
jakubjezek001: can we actually pass any object from creators to publishing context? I was assuming it is not possible and so I am passing it as dictionary.

client/ayon_core/plugins/publish/integrate_versions_to_list.py
jakubjezek001: would that work to pass it via context.data?

client/ayon_core/plugins/publish/collect_version_to_list.py
BigRoy: > There are only two types of list entity

Currently yes - the bool is not extendable, but if another list type comes along... then the `str` Literal way allows us to expand in a backwards compatible manner. Also, I feel it's more descriptive than the bool too.

client/ayon_core/plugins/publish/collect_version_to_list.py
BigRoy: Dataclasses - I think so - but I'd personally say it likely comes from a Collector, not directly from a Creator though. The creator in your case may need to set the filename but the collcetor has fine access to that metadata to define the list entry?

client/ayon_core/plugins/publish/integrate_versions_to_list.py
BigRoy: We could, but I don't see the need - we're on a context plugin here and we don't need to pass it to other plug-ins. We should just do `entity_lists = get_entity_lists` and then the `_existing_lists()` call just takes that `entity_lists` as argument instead of querying.

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,107 @@
...
+
+        processing_items = []
+        if version_lists_templates:
+            processing_items += version_lists_templates
+        if profile:
+            processing_items.append(profile)
+
+        anatomy_data: dict[str, Any] = instance.data["anatomyData"]
+        anatomy: Anatomy = instance.context.data["anatomy"]
+        template_data = deepcopy(anatomy_data)
```
iLLiCiTiT:
```suggestion
        anatomy: Anatomy = instance.context.data["anatomy"]
        template_data = deepcopy(instance.data["anatomyData"])
```

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,107 @@
...
+class CollectVersionToList(pyblish.api.InstancePlugin):
+    """Collect Version to List."""
+
+    order = pyblish.api.CollectorOrder + 0.499
+    label = "Collect Version to List"
+
+    profiles = []
+
+    def process(self, instance):
+        version_lists_templates = instance.data.get(
```
iLLiCiTiT: What is `"versionListsTemplates"`?

I thought it would be that `"versionListsTemplates"` has to be collected and integrate plugin will integrate it. But this plugin does require `"versionListsTemplates"`, it should be the other way (I guess) if `"versionListsTemplates"` is set, then skip "general" plugin.

The general collector we were discussing would set the `"versionListsTemplates"` on instance, but would not be based on it.

client/ayon_core/plugins/publish/integrate_versions_to_list.py
iLLiCiTiT: 100% not in plugin, must be in core's api. e.g. `ayon_core.pipeline.publish.structures` or something like that, we might add more similar things over time.

client/ayon_core/plugins/publish/collect_version_to_list.py
jakubjezek001: the key `versionListsTemplates` is set at CSV creator because we need to have exception for csv presets.

client/ayon_core/plugins/publish/collect_version_to_list.py
iLLiCiTiT: So add this
```python
ListType = Literal["generic", "review-session"]
```
And change the typehint to:
```suggestion
    list_type: ListType = "generic"
```

server/settings/publish_plugins.py
```diff
@@ -201,6 +201,81 @@ def validate_unique_outputs(cls, value):
...
+
+
+class CollectVersionToListModel(BaseSettingsModel):
+    enabled: bool = SettingsField(True, title="Enabled")
+    profiles: list[CollectVersionToListProfileModel] = SettingsField(
+        default_factory=list,
+        title="Profiles",
+        description=(
+            ""
+        )
```
BigRoy:
```suggestion
```

server/settings/publish_plugins.py
```diff
@@ -201,6 +201,81 @@ def validate_unique_outputs(cls, value):
...
+    )
+    product_names: list[str] = SettingsField(
+        default_factory=list,
+        title="Product names",
+        description="The product names to match this profile to.",
+    )
+    name_template: str = SettingsField(
+        "",
+        title="Name template",
```
BigRoy: Name should be fine, right?
```suggestion
    name: str = SettingsField(
        "",
        title="Name",
```

server/settings/publish_plugins.py
```diff
@@ -201,6 +201,81 @@ def validate_unique_outputs(cls, value):
...
+        default_factory=list,
+        title="Parent folders",
+        description="Folder hierarchy formed from top to bottom.",
+    )
+    list_type: str = SettingsField(
+        "generic",
+        title="List type",
+        description="Define what type of list this profile represents.",
+        enum_resolver=list_type_enum,
+
```
BigRoy:
```suggestion
```

server/settings/publish_plugins.py
```diff
@@ -1429,6 +1508,10 @@ class PublishPuginsModel(BaseSettingsModel):
             },
         ]
     },
+    "CollectVersionToList": {
+        "enabled": True,
```
BigRoy: Since it doesn't come with profiles by default, I wonder if we should disable it by default.
```suggestion
        "enabled": False,
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,288 @@
...
+                )
+
+    def _get_or_create_parent_folder(
+        self,
+        project_name: str,
+        parent_folders: list[str] | None,
+    ) -> str | None:
+        # TODO: Should we apply 'data.scopes' to the folder to make the folder
+        #  list only under e.g. generic list or review-sessions lists instead
+        #  of scoped to both.
```
BigRoy: I think we should solve this so that it defaults scoped to the list type you're now adding? 🤔

BigRoy (commented):
@jakubjezek001 I took the liberty to push a code change with https://github.com/ynput/ayon-core/pull/1871/commits/10821af24769d3a3be8797b13bed335de2cb8396 since it was easier than making massive comments across github.

But it was mostly about readability, taking out the API helpers from the class, staying close to the ListConfig dataclass as long as we can (even in the integrator with the template formatted values) and considering the cases where the list of that name may exist but may not be of valid type. I also added a **todo** about _scopes_ of the parent folders that we generate which I think we should perhaps look into? I haven't checked what happens either if any of the parent folders does exist, but it's currently scoped to the OTHER type than what you're now adding a list into, which I'm sure will complicate things further. 🗡️ 

Anyway, feel free to revert or comment on the code changes - see it more as 'comment' then me enforcing it please.

jakubjezek001:
I needed to rename settings profile field from `name` back to `list_name` because it was creating really bad loocking results at settings layout. I was trying all three layouts and none of them looked good.

client/ayon_core/plugins/publish/integrate_versions_to_list.py
jakubjezek001: I am just afraid that this might fall under a folder name uniqueness. I would rather keep it as it is now and perhaps implement it in future with more robust checking.

BigRoy: > I am just afraid that this might fall under a folder name uniqueness. I would rather keep it as it is now and perhaps implement it in future with more robust checking.

It is, yes - the folders are unique by name (shared between the two). However if you by default create lists on frontend - the default is that it is not shared in scope, but only shows in one of them which I also think is the more reasonable default (and it aligns more with what our client is actually doing too; although in their case the group already exists so it may be less of a problem since it'd never need to create a list for them.)

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def process(self, context):
+        list_config_by_list_name: dict[str, ListConfig] = {}
+        version_ids_by_list_name: dict[str, list[str]] = defaultdict(list)
+        for instance in context:
+            version_lists: list[ListConfig] = instance.data.get(
+                "versionLists", [])
+            version_entity = instance.data.get("versionEntity")
+            if not version_entity or not version_lists:
+                continue
```
BigRoy: @jakubjezek001 @iLLiCiTiT if the ListConfig is of `review-session` - should we confirm here also whether the version entity actually has any reviewables to begin with? If not, skip it with a debug log?

BigRoy (commented):
This is really cool actually - this works nicely as intended, however I feel like we need to find a way to allow the inclusion of versions to be filtered to only if they have a reviewable.

E.g. when using a `review-session` list type there is a very high chance you actually only want to be added items into that list that happen to have any reviewables attached. So I feel like we need some way to say "Only include if it has reviewables." like e.g. the profiles filter having a "Has reviewable" filter? (Although the issue there is that at `CollectorOrder + 0.499` it doesn't have the version entity nor reviewable yet.)

image

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,75 @@
...
+    profiles = []
+
+    def process(self, instance):
+        profile = self._get_profile_for_instance(instance)
+        if not profile:
+            self.log.debug(f"No profile found for instance {instance}")
+            return
+        version_lists: list[ListConfig] = instance.data.setdefault(
+            "versionLists", [])
+
```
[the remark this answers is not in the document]
BigRoy: PR description should be updated.

client/ayon_core/plugins/publish/integrate_versions_to_list.py
jakubjezek001: Yes it makes perfect sense. Will add it.

BigRoy:
Can you also make the linter pass? :)

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,76 @@
...
+        if version_lists:
+            self.log.debug(f"Version lists already collected: {version_lists}")
+            return
```
iLLiCiTiT: We have to check if the key is set.
```suggestion
        if "versionLists" in instance.data:
            version_listst = instance.data["versionLists"]
            self.log.debug(f"Version lists already collected: {version_lists}")
            return
        
        version_lists: list[ListConfig] = []
        instance.data["versionLists"] = version_lists
```

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,76 @@
...
+
+        name = profile["list_name"]
+        version_lists.append(
+            ListConfig(
+                name=name,
+                parent_folders=profile.get("parent_folders", None),
```
iLLiCiTiT:
```suggestion
                parent_folders=profile["parent_folders"],
```

client/ayon_core/plugins/publish/collect_version_to_list.py
iLLiCiTiT: BTW this check should happen before profiles...

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,318 @@
...
+    label = "Integrate Versions to List"
+    order = pyblish.api.IntegratorOrder + 0.49
+    settings_category = "core"
+
+    def process(self, context):
+        list_config_by_list_name: dict[str, ListConfig] = {}
+        version_ids_by_list_name: dict[str, list[str]] = defaultdict(list)
+        for instance in context:
+            version_lists: list[ListConfig] = instance.data.get(
+                "versionLists", [])
```
iLLiCiTiT:
```suggestion
                "versionLists")
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,318 @@
...
+            version_entity = instance.data.get("versionEntity")
+            if not version_entity or not version_lists:
+                continue
+            for list_config in version_lists:
+                # Construct the list config with the formatted name and parent
+                # folder names
+                anatomy: Anatomy = instance.context.data["anatomy"]
+                template_keys = deepcopy(instance.data["anatomyData"])
```
iLLiCiTiT: Please rename `template_keys` to `template_data`.
```suggestion
                template_data = deepcopy(instance.data["anatomyData"])
```

client/ayon_core/pipeline/structures.py
```diff
@@ -0,0 +1,14 @@
...
+from typing import Literal
+
+ListType = Literal["generic", "review-session"]
+
+
+@dataclass
+class ListConfig:
+    """Define a list."""
+    name: str
+    parent_folders: list[str] | None = None
```
iLLiCiTiT:
```suggestion
    parent_folders: list[str]= field(default_factory=list)
```

client/ayon_core/pipeline/structures.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
```
iLLiCiTiT:
```suggestion
from dataclasses import dataclass, field
```

client/ayon_core/pipeline/structures.py
iLLiCiTiT: Allowing `parent_folders` as `None` is confusing and requires a lot of additional checks

client/ayon_core/plugins/publish/integrate_versions_to_list.py
BigRoy: This is technically incorrect with the type hint.

client/ayon_core/plugins/publish/collect_version_to_list.py
BigRoy: I'm actually rather confused by this check - even if there are already `versionLists` - we still want to append those from the profiles right? You allow other plug-ins to define `versionLists` but it's not a replacement for what this plug-in collect but 'additional' to it?

jakubjezek001: Apparently not, I was discussing this with @iLLiCiTiT and was told to do it as first found first served.

jakubjezek001: the original implementation was allowing multiple lists in the key which is originally designed to be list so it was actually making sense to be able to do it. Argument was mainly, that no-one will ever setup two profiles on both and it would be redundant. I quite agree that no-one will ever configure it that way, but since we are on profile matching, there is no way to configure multiple lists adding for single version and that is perhaps something which could be in future required. Also productions might need to create two lists, one generic and another review-session. The `versionLists` key as list was actually designed for that case.

client/ayon_core/plugins/publish/collect_version_to_list.py
```diff
@@ -0,0 +1,78 @@
...
+
+        profile = self._get_profile_for_instance(instance)
+        if not profile:
+            self.log.debug(f"No profile found for instance {instance}")
+            return
+
+        name = profile["list_name"]
+        version_lists.append(
+            ListConfig(
+                name=name,
```
iLLiCiTiT:
```suggestion
        version_lists.append(
            ListConfig(
                name=profile["list_name"],
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+                        self.log.error(
+                            "Conflicting list_type for entity list label "
+                            f"'{list_name}': '{existing_config.list_type}' vs "
+                            f"'{candidate_config.list_type}'. Skipping."
+                        )
+                        continue
+                    if (
+                        existing_config.parent_folders or None
+                    ) != (
+                        candidate_config.parent_folders or None
```
iLLiCiTiT:
```suggestion
                        existing_config.parent_folders
                        != candidate_config.parent_folders
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+                            "Multiple parent_folders configured for "
+                            f"entity list '{list_name}'. Using the "
+                            f"first: {existing_config.parent_folders}"
+                        )
+                else:
+                    list_config_by_list_name[list_name] = candidate_config
+                version_ids_by_list_name[list_name].append(
+                    version_entity["id"]
+                )
+
```
iLLiCiTiT, line 186:
```suggestion

        if not list_config_by_list_name:
            return

```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+        # Get all list entities for the project from server
+        project_name: str = context.data["projectName"]
+        existing_list_entities_by_label: dict[str, dict] = {
+            entity_list["label"]: entity_list
+            for entity_list in ayon_api.get_entity_lists(
+                project_name=project_name
+            )
+        }
+        for list_name, list_config in list_config_by_list_name.items():
```
iLLiCiTiT:
```suggestion
        existing_list_folders = get_entity_list_folders(
            project_name=project_name
        )
        for list_name, list_config in list_config_by_list_name.items():
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+                    list_entity=existing_list,
+                    version_ids=version_ids,
+                )
+
+            # Else create the new list
+            else:
+                # make sure parent folder exists
+                entity_list_folder_id = self._get_or_create_parent_folder(
+                    project_name=project_name,
+                    parent_folders=list_config.parent_folders,
```
iLLiCiTiT:
```suggestion
                    parent_folders=list_config.parent_folders,
                    existing_list_folders=existing_list_folders,
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+                    items=[
+                        {"entityId": version_id}
+                        for version_id in version_ids
+                    ],
+                )
+
+    def _get_or_create_parent_folder(
+        self,
+        project_name: str,
+        parent_folders: list[str] | None,
```
iLLiCiTiT:
```suggestion
        parent_folders: list[str],
        existing_list_folders: list[dict[str, Any]],
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+    ) -> str | None:
+        # TODO: Should we apply 'data.scopes' to the folder to make the folder
+        #  list only under e.g. generic list or review-sessions lists instead
+        #  of scoped to both.
+        if not parent_folders:
+            return None
+
+        existing_list_folders = get_entity_list_folders(
+            project_name=project_name
+        )
```
iLLiCiTiT:
```suggestion
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+            )
+
+    def _create_entity_list(
+        self,
+        project_name: str,
+        entity_type: str,
+        label: str,
+        items: list[dict],
+        list_type: str | None = None,
+        entity_list_folder_id: str | None = None,
```
iLLiCiTiT:
```suggestion
        list_type: str | None,
        entity_list_folder_id: str | None,
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,327 @@
...
+                    continue
+                parent_folder_id = list_folder["id"]
+                break
+            else:
+                # if folder does not exist, create it under the parent folder
+                parent_folder_id = create_entity_list_folder(
+                    project_name=project_name,
+                    folder_name=folder_name,
+                    parent_folder_id=parent_folder_id,
+                )
```
iLLiCiTiT:
```suggestion
                new_folder_id = create_entity_list_folder(
                    project_name=project_name,
                    folder_name=folder_name,
                    parent_folder_id=parent_folder_id,
                )
                # Add new list folder to existing ones
                existing_list_folders.append({
                    "id": new_folder_id,
                    "parentId": parent_folder_id,
                    "label": folder_name,
                })
                parent_folder_id = new_folder_id
```

client/ayon_core/plugins/publish/integrate_versions_to_list.py
```diff
@@ -0,0 +1,395 @@
...
+                ):
+                    # If folder exists, check if scope needs to be updated
+                    #   with the list type.
+                    # NOTE If scope is None or empty list it is scoped for
+                    #   everything. So the list type is added to scope only
+                    #   if scope exists and does not contain the list type.
+                    scope = existing_list_folder["data"].get("scope")
+                    if scope and list_config.list_type not in scope:
+                        scope.append(list_config.list_type)
+                        # --- Temporary bypass ---
```
iLLiCiTiT, line 313: NOTE looks like there is a bug in frontend (not verified yet). If list folder has scope `["generic", "review-session"]` it is not visible in `Lists` tab.

iLLiCiTiT (approved):

PR #1882, opened 2026-06-08
fix: expand user home in anatomy roots

Viodmian:
## Changelog Description
Expand `~` in anatomy root definitions so paths like `~/foo/bar` resolve to the current user's home directory before root values are cleaned and exposed through helpers.

Fixes #1874.

## Additional info
This keeps the existing environment-variable formatting behavior for current-platform roots and applies home expansion to configured root values before they are stored in `raw_data` / `cleaned_data`.

## Testing notes:
1. `python -m pytest tests\client\ayon_core\pipeline\anatomy\test_roots.py -q`
2. `git diff --check`
3. `python -m py_compile client\ayon_core\pipeline\anatomy\roots.py tests\client\ayon_core\pipeline\anatomy\test_roots.py`

PR #1884, opened 2026-06-08
Anatomy: expand user and vars in root paths

### LINKED ISSUE #1874: YN-0773: make sure variables are expanded in roots

Make sure we expand variables in roots. So the root definition like this

&nbsp;

    
    {
    "work": {
        "windows": "~/foo/bar"
    }
    }
    

&nbsp;

will be properly expanded to for example `/home/user/foo/bar` 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1504075221638844506)

_This issue was automatically created from AYON ticket [YN-0773]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=d30cad905da011f1ad0d75e1c92a3963)

antirotor:
## Changelog Description
Expand variables and user `~` in root paths.

Roots like `~/foo` or `$HOME/foo` or `{HOME}/foo` should all resolve correctly. Note that this feature needs to be used with care as dynamic roots can break a lot of things if not used properly.

## Additional info
This is alternative to #1882

## Testing notes:
1) Publishing to roots with `~` should work as expected
2) Loading should work too

### DIFF
--- a/client/ayon_core/pipeline/anatomy/roots.py
+++ b/client/ayon_core/pipeline/anatomy/roots.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import os
 import platform
 import numbers
@@ -17,13 +18,15 @@ class AnatomyRoot(FormatObject):
     is used for formatting of template.
 
     Args:
-        parent (AnatomyRoots): Parent object.
+        parent (AnatomyRoots | None): Parent object.
         root_raw_data (dict): Dictionary containing root values by platform
             names. ["windows", "linux" and "darwin"]
         name (str): Root name which is representing. Used with
             multi root setup otherwise None value is expected.
     """
-    def __init__(self, parent, root_raw_data, name):
+    def __init__(
+            self, parent: AnatomyRoots | None,
+            root_raw_data: dict, name: str):
         super().__init__()
         self._log = None
         lowered_platform_keys = {
@@ -45,12 +48,13 @@ class AnatomyRoot(FormatObject):
             self.value = (
                 os.path.expandvars(
                     os.path.expanduser(
-                        lowered_platform_keys[current_platform].format_map(
-                            os.environ
-                        )
-                    )
-                )
-            )
+                        lowered_platform_keys[current_platform])))
+        except AttributeError as e:
+            msg = f"Missing root definition for platform {current_platform}."
+            raise RootMissingEnv(msg) from e
+
+        try:
+            self.value = self.value.format_map(os.environ)
         except KeyError as e:
             result = StringTemplate(self.value).format(os.environ.copy())
             is_are = "is" if len(result.missing_keys) == 1 else "are"
@@ -208,7 +212,7 @@ class AnatomyRoot(FormatObject):
         ``os.path.expanduser``) so that the following mismatches are handled:
 
         - ``path`` is already expanded but the stored root still contains
-          ``~`` or ``%VAR%`` / ``$VAR`` tokens (or vice-versa).
+          ``~` / ``$VAR`` tokens (or vice-versa).
 
         Args:
             path (str): Path where root value should be found.
--- a/client/ayon_core/pipeline/anatomy/roots.py
+++ b/client/ayon_core/pipeline/anatomy/roots.py
@@ -41,17 +41,23 @@ class RootItem(FormatObject):
         #   as production safe. Some features may not work as expected, for
         #   example USD resolver or site sync.
         try:
-            self.value = lowered_platform_keys[current_platform].format_map(
-                os.environ
+            self.value = (
+                os.path.expandvars(
+                    os.path.expanduser(
+                        lowered_platform_keys[current_platform].format_map(
+                            os.environ
+                        )
+                    )
+                )
             )
-        except KeyError:
+        except KeyError as e:
             result = StringTemplate(self.value).format(os.environ.copy())
             is_are = "is" if len(result.missing_keys) == 1 else "are"
             missing_keys = ", ".join(result.missing_keys)
             raise RootMissingEnv(
                 f"Root \"{name}\" requires environment variable/s"
                 f" {missing_keys} which {is_are} not available."
-            )
+            ) from e
 
         self.clean_value = self._clean_root(self.value)
 
@@ -196,6 +202,13 @@ class RootItem(FormatObject):
 
         All platform values are checked for this replacement.
 
+        Both the input ``path`` and the stored root values are tried in their
+        original *and* expanded forms (via ``os.path.expandvars`` /
+        ``os.path.expanduser``) so that the following mismatches are handled:
+
+        - ``path`` is already expanded but the stored root still contains
+          ``~`` or ``%VAR%`` / ``$VAR`` tokens (or vice-versa).
+
         Args:
             path (str): Path where root value should be found.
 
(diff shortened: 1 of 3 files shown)


iLLiCiTiT:
What is the use-case of having `~` or env in anatomy roots? Allowing it might block us on implementing server side handling of paths.

antirotor:
> What is the use-case of having `~` or env in anatomy roots? Allowing it might block us on implementing server side handling of paths.

The use case is client request to be able to use gdrive folder that is per user mounted in their respective home dirs.

iLLiCiTiT:
> The use case is client request to be able to use gdrive folder that is per user mounted in their respective home dirs.

I'd say that this is what should be used.
image

Why I'm against: This will open can of worms. Allowing to do this "easily" looks like you can safely do it, but there is a ton of caviads related to this approach that might make us look bad.

I might be wrong, summoning @BigRoy for opinion.

antirotor:
> 'd say that this is what should be used.

This is really not practical for whole studio setup. You would have to do it for every site.

BigRoy:
> Why I'm against: This will open can of worms. Allowing to do this "easily" looks like you can safely do it, but there is a ton of caviads related to this approach that might make us look bad.

I agree with this. The tricky thing is that these expanded root paths and their 'context' are stored in a variety of places, e.g. in publishes, suddenly we're passing the `context` as `root[work]` with the user location instead of the actual root of the current platform. Or even making a rootless path for something from another user's machine will not succeed (not sure if relevant to anything.)

But regarding using "site root overrides":

> This is really not practical for whole studio setup. You would have to do it for every site.

This is also agree with. Which makes me think what technically differs between a site root override and this? (Except for the fact that the expanded path is set somewhere?)
The USD resolver would now also need to trigger the expandvars (assuming USD can't load directly from a path set as `~/something`?

---

@iLLiCiTiT from a pipeline/technical perspective code-wise. How much does site specific root overrides really differ from expanding the vars?

I think this will need thorough testing, cross-platform with multiple users/home directories - and possibly even with the USD resolver resolving to a published product that uses `~`?

iLLiCiTiT:
> How much does site specific root overrides really differ from expanding the vars?

At the end the important bit is that it can be resolved server side, service side, or in USD resolver.

> assuming USD can't load directly from a path set as ~/something

Also `${HOME}/something` or `$HOME/something` and `%HOME%/something` on windows because of `expandvars` and `{HOME}/something` because of `format_map(os.environ)`.

antirotor:
While I agree that this might be dangerous and might break things - there are many more similar things around pipeline. My point is that nothig is forcing you to do it. It just gives you a possibility. If this is documented properly, it is up to the studio to go that way or not.

Linux allows you to type `sudo rm -rf /` and it has its own uses, but noone force you to write it.

antirotor:
> %HOME%/something

this won't work as it is not expanded by python

iLLiCiTiT:
> this won't work as it is not expanded by python

It is. I just chose wrong env key which is not set by windows. `%USERPROFILE%/something` woult be expanded.

The thing is that with this is looks like you can safely use it, which is not true, you'll block yourself from some features, and it might not work all the time, and it will look like our fault.

client/ayon_core/pipeline/anatomy/roots.py
```diff
@@ -42,17 +43,24 @@ def __init__(self, parent, root_raw_data, name):
         #   as production safe. Some features may not work as expected, for
         #   example USD resolver or site sync.
         try:
-            self.value = lowered_platform_keys[current_platform].format_map(
-                os.environ
-            )
-        except KeyError:
+            self.value = (
+                os.path.expandvars(
```
iLLiCiTiT, line 47: If we gona do it, I'm against using `expandvars`, only `{ENV}` should be possible.

client/ayon_core/pipeline/anatomy/roots.py
```diff
@@ -225,22 +240,65 @@ def find_root_template_from_path(self, path):
...
         for root_os, root_path in self.cleaned_data.items():
             # Skip empty paths
             if not root_path:
                 continue
 
-            _mod_path = mod_path  # reset to original cleaned value
+            # Expand variables in the stored root path so we can compare it
+            # against an already-expanded input path (and vice-versa).
+            expanded_root_path = self._clean_root(
+                os.path.expandvars(os.path.expanduser(root_path))
```
iLLiCiTiT, line 258: Root paths should be already expanded in `cleaned_data`.

client/ayon_core/pipeline/anatomy/roots.py
```diff
@@ -225,22 +240,65 @@ def find_root_template_from_path(self, path):
         output = str(path)
 
         mod_path = self._clean_path(path)
+        # Expanded version of the input path – used when the stored root value
+        # is already expanded while the caller passed an unexpanded path, or
+        # to normalise both sides consistently.
+        expanded_mod_path = self._clean_path(
+            os.path.expandvars(os.path.expanduser(path))
```
iLLiCiTiT, line 247: We should not allow to pass in not expanded path.

antirotor:
btw this is draft PR for USD Resolver - https://github.com/ynput/ayon-usd-resolver/pull/80

client/ayon_core/pipeline/anatomy/roots.py
antirotor: I am not sure I get it - it is not expanded there, only for current platform. so are you suggesting that the expansion logic should be moved to ... probably `_clean_path()`?

client/ayon_core/pipeline/anatomy/roots.py
antirotor: why? shouldn't one place take care of that intead reimplementing the logic in all places where you want to use this feature before calling `find_root_template_from_path()`?

client/ayon_core/pipeline/anatomy/roots.py
antirotor: what is the reason? My argument is that using paths with `$VAR` inside is pretty standard.

iLLiCiTiT: It is standard, but then you have to support it at every place which is using roots, including USD resolver, it is hard to find out if it is missing, and what exact variable it is (it does not have strictly defined end the the variable name). `$MY_custom_var` is valid env key, but how you find out if it should be `$MY` or `$MY_custom`...

client/ayon_core/pipeline/anatomy/roots.py
iLLiCiTiT: If you call `find_root_template_from_path` you should have full path, not `~/some/file`.

client/ayon_core/pipeline/anatomy/roots.py
antirotor: I guess the behavior is clearly defined. `VAR = "foo"`:

`lorem $VAR`: `lorem foo`
`lorem $VARipsum`: `lorem `
`lorem $VAR ipsum`: `lorem foo ipsum`
`lorem %VAR%`: `lorem foo`
`lorem %VAR%ipsum`: `lorem fooipsum`

`os.path.expandvars()` is taking care of that in python and the resolver needs to implement `{VAR}` format anyway so adding `$VAR` and/or `%VAR%` is trivial at that point. But I don't really mind. @BigRoy ?

client/ayon_core/pipeline/anatomy/roots.py
antirotor: that is what you get now:

```python
ar = AnatomyRoot(
            parent=...,
            root_raw_data={"linux": "/home/user/projects"},
            name="work",
)

success, result = item.find_root_template_from_path("~/projects/shot/file.ma")
```
you'll get `result == "{root[work]}/shot/file.ma"`

it handles matching unexpanded paths to expanded roots in various forms (`~`, `$VAR`, etc.) and only if it cannot match, it will `success == False`.

iLLiCiTiT: And where you'd get the path `"~/projects/shot/file.ma"` from?

client/ayon_core/pipeline/anatomy/roots.py
iLLiCiTiT: You can't easily find out if it is missing (without validating it for each platform) and there are limitations, e.g. you can't use `$VARipsum` (which is missing in your example). I'm 100% against it. I know it is "known", but it is not easily detectable and easy to validate.

antirotor: It is there:

`lorem $VARipsum`: `lorem` 

 But it also supports curly bracket format that is standard too. So if you want this to work, you can:

`lorem ${VAR}ipsum`: `lorem fooipsum`

and all that is supported by `os.path.expandvars()`

iLLiCiTiT: This is unresolved expanding and you have no idea it is unresolved
```
print(os.path.expandvars("$NOTEXISTING"))
>>> $NOTEXISTING
```

This does require to be more specific, something that should be filled is not
```
import os
os.environ["TEST"]="value"
print(os.path.expandvars("$TEST_something"))
>>> $TEST_something
```

Supporting this will only bring issues. At the end it does what the python formatting does, but with less control and more headaches for us.

BigRoy:
@iLLiCiTiT @antirotor do we need to discuss this PR to make it move?

iLLiCiTiT:
I'd say it is in stuck state, I'm against using `expandvars` for various reasons. The most notable issue for me is missing validations and reporting, also complicating the code of roots that should be "just working".

Another issue for me is that we should do the expand only first time and do not repeat, also do not expand paths passed to the functions, they should be already resolved, as far as I know there is no place that would allow to work with unexpanded paths.

BigRoy:
@antirotor given what @iLLiCiTiT says - do you see any other approach or workarounds to this?

antirotor:
I honestly don't know. The initial feature request was to be able to have home directories in root paths. Home can be expanded from `~` but also from `{HOME}`. I can limit it to specifically handling of `{HOME}` token or omit it completely and just run `os.path.expanduser()` for dealing with `~`. 

I think this limits its usefulness while not addressing any points about "validation or complicating the code". You can still break things, you still need to support it in USD asset resolver.

Part of the solution could be just better documentation - for example - recommending `${FOO}` format that is much safer than `$FOO`. We can also add validation to check for presence of `$` in a path that should be resolved.

iLLiCiTiT:
I do agree with `~` and `os.environ` as dict, not with `os.path.expandvars`.

Also `find_root_template_from_path` should not try to resolve the paths, it should expect full path.

Issue #946, opened 2024-10-08
Create Context / Publish: Allow access to export the JSON report from CLI publish


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Allow a headless publish via `CreateContext` to generate a `JSON` report that can also be used in the Publish report viewer. This way, e.g. future headless publishes (or farm publishes) will be able to generate these reports as well.

### How would you imagine the implementation of the feature?

Similar to the Publisher UI now generates as report - generate the same, but even without UI.

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the enhancement request.

BigRoy:
This is set to **Needs Info?** What type of info does it need?

jakubjezek001:
This issue is related to #1011

dee-ynput:
> This is set to **Needs Info?** What type of info does it need?

There has been an automatic "Needs Infos" for a while. Fixed now. Setting it as Ready 👍

BigRoy:
Moving into a cycle (for when there's time on your hands). Particularly the two steps you described would be very welcome:

- Step 1 should be to move it out of Publisher UI. 
- Step 2 would be to have our publish function for headless publishing.

PR #1897, opened 2026-06-15
Move publish report logic to publish api

iLLiCiTiT:
## Changelog Description
Moved publish report and report maker from publisher UI to publish api.

## Additional info
This is very first step for https://github.com/ynput/ayon-core/issues/946 . Refactored the code to remove unnecessary code and clean up the logic a little bit. Use dataclasses instead of dictionaries, which cleaned up the data we're working with a little and removed mid-classes in publisher UI. With that were added type-hints at multiple places in UI because I really had to find out if we're working with list/set/None or something else.

At this moment the report maker is not much usable out of publisher, to be able to do that we have to introduce our publish iteration logic as pyblish iteration logic does skip some steps silently. That is current only in publisher, and moving that part would be another big step.

Because I've changed a lot, this should be tested a lot. Publish report viewer tool should work, publisher should show all information during publishing. Errors and logs should be shown correctly in UI with details. It should be possible to trigger actions (repait etc.).

## Testing notes:
1. Open publisher in any DCC.
2. Try to make it fail, log, error out, run actions, whatever comes to your mind.
3. Open publish report viewer and look for older reports if they do work.


kalisp:
Shouldn't I be seeing something here?
image

iLLiCiTiT:
> Shouldn't I be seeing something here?

You should indeed.

kalisp:
I cannot export report (nor copy it to clipboard).

image

iLLiCiTiT:
@kalisp both issues should be fixed

moonyuet:
I can copy and export the report.
But...for some reason, the report has been disappeared after I closed and re-opened the publish report viewer. I am not sure if it is intended.

https://github.com/user-attachments/assets/c54de394-9906-45c6-9d30-713bde67b889

iLLiCiTiT:
> I am not sure if it is intended.

It is not, validating.

iLLiCiTiT:
Should be fixed now.

moonyuet (approved):
Looks good now

Image

MustafaJafar (commented):
Everything seems normal.
Image

MustafaJafar (approved):

PR #1898, opened 2026-06-17
Add file size column to workfiles in launcher 

mahesh-ynput:
## Changelog Description
The Launcher currently shows no file metadata such as timestamp or file size. Could you please add file size and last modified timestamp allowing artists to quickly spot zero-size or potentially corrupt files before wasting time attempting to open them?

## Additional info
last modified column is already there now implemented other column where includes workfile size

## Testing notes:
1. Open launcher select project->folder->task 
2. See workfiles size section 3rd column


client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -185,7 +188,20 @@ def data(self, index, role=QtCore.Qt.DisplayRole):
...
+                if file_size is not None:
+                    return self._format_file_size(file_size)
+                return "N/A"
+            elif role not in (
+                WORKFILE_ID_ROLE,
+                HOST_NAME_ROLE,
+                ITEM_TYPE_ROLE,
+                FILE_SIZE_ROLE,
+            ):
+                return None
```
BigRoy: Why do we need this?

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -234,6 +250,14 @@ def _get_icon(self, icon_url: Optional[str]) -> QtGui.QIcon:
         self._cached_icons[icon_url] = icon
         return icon
 
+    def _format_file_size(self, size_bytes: int) -> str:
+        """Format file size in human-readable format."""
+        for unit in ['B', 'KB', 'MB', 'GB']:
+            if size_bytes < 1024.0:
+                return f"{size_bytes:.1f} {unit}"
+            size_bytes /= 1024.0
+        return f"{size_bytes:.1f} GB"
```
BigRoy: This is better handled by a delegate - so that all sizes are in bytes (making the column work sortable as well) but the value displayed is human readable.

mahesh-ynput:
Resizing column is in pending.

BigRoy (commented):
Can we also make sure that by default there's no horizontal scrollbar - so we'll need to pick some decent default widths? Or trigger resize to fit contents for the columns whenever we switch - but not sure that's entirely nice 🤔

client/ayon_core/tools/launcher/ui/workfiles_page.py
mahesh-ynput: Without the filter, the size column inheriting icon from column 0 seems visual properties are inheriting, I just used same logic which existed on top just to avoid this.

client/ayon_core/tools/launcher/ui/workfiles_page.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1898/commits/3d0810067aabb572c35500612a4753b6a3e98a04

client/ayon_core/tools/launcher/ui/workfiles_page.py
BigRoy: > I  just used same logic which existed on top just to avoid this.

I see now - that's actually rather confusing to do.

How about just starting off with:

```
if role in {
        WORKFILE_ID_ROLE,
        HOST_NAME_ROLE,
        ITEM_TYPE_ROLE,
        FILE_SIZE_ROLE,
}:
    index = self.index(index.row(), 0, index.parent())
    return super().data(index, role)
```

Then all the braches below do not need to care about it?

mahesh-ynput:
> Can we also make sure that by default there's no horizontal scrollbar - so we'll need to pick some decent default widths? Or trigger resize to fit contents for the columns whenever we switch - but not sure that's entirely nice 🤔

https://github.com/ynput/ayon-core/pull/1898/commits/22a8e4f8724518e5903adc3ca595c6e09b2ef68c

client/ayon_core/tools/launcher/ui/workfiles_page.py
mahesh-ynput: So this I can apply this for both cols right? also roy this suggestion is for better readability? or this can provide bit performance optimization?

BigRoy: This logic would live above the if index.column()` checks because it looks like we _always_ want to make it behave like this with the `if not` branches. Or did I misread the code?

It won't be more optimal (except for maybe if we store this as a `set` potentially increasing the lookup speed, but would be neglible. It's mostly avoiding duplicating code...

Even now the code is already wrong, because the first `if branch` doesn't include the `FILE_SIZE_ROLE` which is then a bug on its own.

mahesh-ynput: @BigRoy , https://github.com/ynput/ayon-core/pull/1898/commits/509b8f489acacdc4daa59251ae785481cf01e1f4 I made changes upon what I understand from above saying. But I would like to connect with you tomorrow and get some clarification on this.

client/ayon_core/tools/launcher/models/workfiles.py
```diff
@@ -66,8 +66,9 @@ def get_workfile_items(
             mod_time = None
             if path and exists:
                 mod_time = os.path.getmtime(path)
-
+                file_size = os.path.getsize(path)
             else:
```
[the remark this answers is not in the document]
mahesh-ynput: It was already handled and in case we greyout full row of respective workfile

BigRoy (approved):
LGTM

client/ayon_core/tools/utils/delegates.py
```diff
@@ -56,6 +56,15 @@ def pretty_date(t, now=None, strftime="%b %d %Y %H:%M"):
     return t.strftime(strftime)
 
 
+def format_file_size(size_bytes: int) -> str:
```
iLLiCiTiT: Duplicate of `file_size_to_string`, please copy that function from workfiles tool instead.

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -270,6 +282,12 @@ def initStyleOption(self, option, index):
                     option.text = pretty
                     return
             option.text = "N/A"
+        elif index.column() == 2:
+            raw = index.data(QtCore.Qt.DisplayRole)
+            if raw is not None:
+                option.text = format_file_size(raw)
+            else:
+                option.text = "N/A"
```
iLLiCiTiT:
```suggestion
                option.text = ""
```

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -175,18 +178,26 @@ def flags(self, index):
...
-                ITEM_TYPE_ROLE,
-            ):
+                index = self.index(index.row(), 0, index.parent())
+            else:
+                return None
+        elif index.column() == 2:
+            if role == QtCore.Qt.DisplayRole:
+                return self.data(index, FILE_SIZE_ROLE)
+            else:
                 return None
```
iLLiCiTiT:
```suggestion
        col = index.column()
        if col != 0:
            if role != QtCore.Qt.DisplayRole:
                return None

            if col == 1:
                role = UPDATED_AT_ROLE
            else:
                role = FILE_SIZE_ROLE
            index = self.index(index.row(), 0, index.parent())
```

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -255,6 +266,7 @@ class WorkfilesDelegate(TreeViewItemDelegate):
 
     Column 0: workfile name with middle-elide.
     Column 1: pretty-printed timestamp (falls back to ``"N/A"``).
+    Column 2: file size in human-readable format (falls back to ``"N/A"``).
```
iLLiCiTiT: I'd say seeing `N/A` twice is not necessary.
```suggestion
    Column 2: file size in human-readable format (falls back to ``""``).
```

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -270,6 +282,12 @@ def initStyleOption(self, option, index):
                     option.text = pretty
                     return
             option.text = "N/A"
+        elif index.column() == 2:
+            raw = index.data(QtCore.Qt.DisplayRole)
```
iLLiCiTiT:
```suggestion
            raw = index.data(FILE_SIZE_ROLE)
```

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -175,18 +178,26 @@ def flags(self, index):
         return super().flags(index)
 
     def data(self, index, role=QtCore.Qt.DisplayRole):
+        if role in {
+            WORKFILE_ID_ROLE,
+            HOST_NAME_ROLE,
+            ITEM_TYPE_ROLE,
+            FILE_SIZE_ROLE
+        }:
+            index = self.index(index.row(), 0, index.parent())
```
iLLiCiTiT:
```suggestion
            if index.column() != 0:
                index = self.index(index.row(), 0, index.parent())
```

client/ayon_core/tools/launcher/models/workfiles.py
```diff
@@ -66,8 +66,9 @@ def get_workfile_items(
             mod_time = None
             if path and exists:
```
iLLiCiTiT, line 68:
```suggestion
            file_size = None
            if path and exists:
```

client/ayon_core/tools/launcher/models/workfiles.py
```diff
@@ -66,8 +66,9 @@ def get_workfile_items(
...
                 mod_time = os.path.getmtime(path)
-
+                file_size = os.path.getsize(path)
             else:
+                file_size = None
```
iLLiCiTiT:
```suggestion
```

moonyuet (commented):
Does testing with the launcher, working as expected. Please take a look at review and do the code changes so that we can possibly merge this PR. Thanks

client/ayon_core/tools/launcher/ui/workfiles_page.py
BigRoy: I'd not do `the else` branch here for the column index. Rather be explicitly matching `elif column == 2` to avoid any odd behavior if we were to expand columns.

client/ayon_core/tools/utils/delegates.py
```diff
@@ -56,6 +56,24 @@ def pretty_date(t, now=None, strftime="%b %d %Y %H:%M"):
...
+    }
+    ending = "B"
+    for _ending, _size in size_ending_mapping.items():
+        if file_size < _size:
+            break
+        size = file_size / _size
+        ending = _ending
+    return "{:.2f} {}".format(size, ending)
+
+
```
BigRoy, line 76: @mahesh-ynput you've now copied/duplicated the code - It's in an 'ok' place here - but you should now also remove the code from `side_panel.py` and make that one import from this one as well, right? Or why is it in multiple places still?

mahesh-ynput: yes roy, I supposed do and resolve comment.

mahesh-ynput: https://github.com/ynput/ayon-core/pull/1898/commits/04850c24b4eff0040f6cc159f6a04dafd41ca611

client/ayon_core/tools/launcher/ui/workfiles_page.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1898/commits/04850c24b4eff0040f6cc159f6a04dafd41ca611

client/ayon_core/tools/utils/delegates.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1898/commits/04850c24b4eff0040f6cc159f6a04dafd41ca611

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -271,6 +283,12 @@ def initStyleOption(self, option, index):
                     option.text = pretty
                     return
             option.text = "N/A"
```
moonyuet:
```suggestion
            option.text = ""
```
Marked this as empty string as it displayed  `N/A` in `Modified Date` in the launcher.
We should align with the empty string shown in the Size.
Image

moonyuet (commented):
See the suggestion below.

client/ayon_core/tools/launcher/ui/workfiles_page.py
BigRoy: Agreed. I prefer this as well.

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -270,7 +282,13 @@ def initStyleOption(self, option, index):
                 if pretty is not None:
                     option.text = pretty
                     return
-            option.text = "N/A"
+            option.text = ""
```
iLLiCiTiT, line 285: Not sure if there was conversation about this, but now `"N/A"` is not showed at all...
```suggestion
            option.text = "N/A"
```

BigRoy: I'm personally fine with that - just seeing <Select a folder> left side should be sufficient. It's just clutter otherwise.

See: 

https://github.com/ynput/ayon-core/pull/1898#discussion_r3443171538

client/ayon_core/tools/launcher/ui/workfiles_page.py
```diff
@@ -256,6 +267,7 @@ class WorkfilesDelegate(TreeViewItemDelegate):
 
     Column 0: workfile name with middle-elide.
     Column 1: pretty-printed timestamp (falls back to ``"N/A"``).
+    Column 2: file size in human-readable format (falls back to ``""``).
```
iLLiCiTiT:
```suggestion
    Column 1: pretty-printed timestamp.
    Column 2: file size in human-readable format.
```

moonyuet:
I did the test on the launcher with the updated commit. looks good. 
Are we gonna merge this or we need to resolve something else?
image

BigRoy:
I think we can merge it. Unless we want to keep `ui-qt-luncher` branch lean for the time being.

Will leave up to @iLLiCiTiT

PR #1906, opened 2026-06-24
Integrate thumbnail: Don't change folder thumbnail

iLLiCiTiT:
## Changelog Description
Do not set folder thumbnail during publishing. And unset the thumbnail on existing folders.

## Additional info
Update of folder thumbnail can trigger recalculation of calculated attributes. Also setting the thumbnail id is not necessary and can be skipped.

To "auto-fix" folders that already do use the thumbnail from version was added a cleanup endpoint that does remove the thumbnail before version thumbnails are pushed to server.

## Testing notes:
1. Thumbnail integration still works and folders do have thumbnails.


server/__init__.py
```diff
@@ -99,3 +116,72 @@ async def execute_action(
...
+        Args:
+            project_name (str): Name of the project.
+            payload (CleanupFolderThumbnailsRequestModel): Request payload
+                containing folder ids.
+
+        """
+        async with Postgres.transaction():
+            if payload.folder_ids is None:
+                res = await Postgres.fetch(
+                    FIND_THUMBNAILS_QUERY.format(project_name=project_name),
```
BigRoy: @martastain should we do more than a plain string formatting here to avoid any form of SQL injects? Or is this technically fine?

Also, should this technically require ACL/Permissions, or are we ok with this?

server/__init__.py
```diff
@@ -99,3 +116,72 @@ async def execute_action(
...
+                f"""
+                UPDATE project_{project_name}.folders
+                SET thumbnail_id = NULL
+                WHERE id = ANY($1);
+                """,
+                fildered_folder_ids,
+            )
+        return CleanupFolderThumbnailsResponseModel(
+            count=len(fildered_folder_ids)
+        )
```
BigRoy:
```suggestion
        filtered_folder_ids = [row["id"] for row in res]
        if filtered_folder_ids:
            await Postgres.execute(
                f"""
                UPDATE project_{project_name}.folders
                SET thumbnail_id = NULL
                WHERE id = ANY($1);
                """,
                filtered_folder_ids,
            )
        return CleanupFolderThumbnailsResponseModel(
            count=len(filtered_folder_ids)
        )
```

client/ayon_core/plugins/publish/integrate_thumbnail.py
```diff
@@ -210,52 +212,61 @@ def _integrate_thumbnails(
...
-                thumbnail_info["entity_type"],
-                entity_id,
-                {"thumbnailId": thumbnail_id}
+        # Trigger thumbnail cleanup for the folders that did use thumbnail
+        #   id of a version. We changed the logic, folders don't need
+        #   to have explicitly set thumbnail id, but we need to auto-fix
+        #   existing folders that already did set the thumbnail.
+        if folder_ids:
+            # There is a bug in public version of 'get_addon_endpoint'
+            # - did not support subpaths correctly
```
BigRoy, line 255: so if there's a bug - what's working around it here?

BigRoy (commented):
Overall looks good to me - just needs the query code check and there's a minor variable name typo. @LiborBatek can you test this with regular artist user and admin user and see if it holds up fine? Run it across two different folders  for both runs and they must both have had publishes with thumbnails from before this PR/branch.

LiborBatek:
Im experiencing some warning info msg when using captured thumbnail...not sure if related or irrelevant, so pls tell me @iLLiCiTiT 

image


When the version does have any thumbnail it correctly get propagated from `version` to the `folder` or rather `product` which is good.

Now what Im not so sure is whenever producing a product version which doesnt have any thumbnail it also propagates to the product and removes any existing thumbnail too...resulting into empty thumbnail on the `product`. Again Im not sure if thats correct or not even tho it sounds logical to me...

Screenshot 2026-07-01 150747


So my conclusion been, if any thumb present on the version it gets correctly propagated, and if not it also propagates "nothing" and thumbnail not present.

On the Folder / Tasks all thumbnails stays intact if any existing...which is correct.

image

LiborBatek (approved):
I have tested using 2 distinct users (admin vs regular artist) and all seems working fine...

There are few questions from me above but I guess they are not relevant to this PR but rather how Ayon works with thumbnails/reviewables...feel free to answer those or even take those into account if connected to this PR.

Thumbnails simply stay intact when already present on a `Folder` and `Task` level and each version propagates accordingly to the product itself (whether its thumbnail or empty aka nothing).

Seems ok to me.

BigRoy:
I think the above warning you mentioned actually indicates an issue - so let's make sure to double check this @iLLiCiTiT

server/__init__.py
```diff
@@ -99,3 +116,72 @@ async def execute_action(
...
+        fildered_folder_ids = [row["id"] for row in res]
+        if fildered_folder_ids:
+            await Postgres.execute(
+                f"""
+                UPDATE project_{project_name}.folders
+                SET thumbnail_id = NULL
+                WHERE id = ANY($1);
+                """,
+                fildered_folder_ids,
+            )
```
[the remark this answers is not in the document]
BigRoy: Haha - I like this feedback.. but the suggestion then is ridiculous to still keep them separate calls, right? Why not just update all with the `WHERE` clause matching the first query? Or probably I'm stupid. 🗡️

client/ayon_core/plugins/publish/integrate_thumbnail.py
```diff
@@ -109,24 +111,24 @@ def _prepare_instances(self, context):
                 )
             )
             if thumbnail_path:
-                self.log.debug((
-                    "Found thumbnail path for instance \"{}\"."
+                self.log.debug(
+                    f"Found thumbnail path for instance \"{instance_label}\"."
                     " Thumbnail path: {}"
-                ).format(instance_label, thumbnail_path))
+                )
```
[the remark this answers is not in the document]
BigRoy:
```suggestion
                self.log.debug(
                    f"Found thumbnail path for instance \"{instance_label}\"."
                    f" Thumbnail path: {thumbnail_path}"
                )
```

server/__init__.py
iLLiCiTiT: Added validations for the project with ACL.

client/ayon_core/plugins/publish/integrate_thumbnail.py
iLLiCiTiT: it should be
```python
endpoint = ayon_api.get_addon_endpoint(
    "core", __version__, "cleanupFolderThumbnails", project_name
)
response = ayon_api.post(
    endpoint,
```

server/__init__.py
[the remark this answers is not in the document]
iLLiCiTiT: Modified to update and collect in one call.

BigRoy (changes requested):
Fails for me:

```
HTTP request error: 500 Server Error: Internal Server Error for url: http://localhost:5000/api/addons/core/1.9.7+dev/cleanupFolderThumbnails/test
{
    "status": 500,
    "detail": "TypeError: string indices must be integers, not 'str'",
    "traceback": "TypeError: string indices must be integers, not 'str'\n\nayon_server/api/context.py:77\nresponse = await call_next(request)\n\n/addons/core/1.9.7+dev/server/__init__.py:154\nfiltered_folder_ids = [row[\"id\"] for row in res]\n\n/addons/core/1.9.7+dev/server/__init__.py:154\nfiltered_folder_ids = [row[\"id\"] for row in res]"
}
```

```
ERROR   logging                    | Unhandled exception: TypeError: string indices must be integers, not 'str'
    traceback:
      TypeError: string indices must be integers, not 'str'

      ayon_server/api/context.py:77
      response = await call_next(request)

      /addons/core/1.9.7+dev/server/__init__.py:154
      filtered_folder_ids = [row["id"] for row in res]

      /addons/core/1.9.7+dev/server/__init__.py:154
      filtered_folder_ids = [row["id"] for row in res]
```

BigRoy:
@iLLiCiTiT can you resolve conflicts?

iLLiCiTiT:
> Fails for me:

Not for me, what are you working with?

PR #1907, opened 2026-06-24
Templated Builder: options to trigger when application launch or always on new scene 

### LINKED ISSUE #1040: AY-7243_Template builder always work on start
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

If we want more explicit control, we may need to clarify the exact settings configuration that makes sense:Workfile Template (enum):
Disabled
Always apply to empty scene (act as 'new file' template)
Apply on application launch if no workfile exists yet
Save first workfile if it does not exist yet
 To me, the 'saving of first workfile' should likely need to be outside of the workfile template loading itself. Because it may very well be that someone does want to directly 'save' their filename on launch just so they can directly press CTRL+S if they start working but have no templates configured to start from.

### How would you imagine the implementation of the enhancemenent?
_No response_

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1313917181888561152/1314146303529193515)

_This issue was automatically created from Clickup ticket [AY-7243](https://app.clickup.com/t/86c18jmtk)_

moonyuet:
## Changelog Description
This PR is to add some more options to allow templated builder to be enabled in several scenarios: 
1. Template Always apply to empty scene (act as 'new file' template)
2. Apply on application launch if no workfile exists yet
## Additional info
Resolve https://github.com/ynput/ayon-core/issues/1040
You can test with blender as starting point: https://github.com/ynput/ayon-blender/pull/313

## Testing notes:
1. Build workfile as followed in the host PR above


client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -532,6 +534,10 @@ def build_template(
                  When set to True, this option initiates the saving of the
                  workfile for an initial version. It will skip saving if
                  a version already exists.
+            apply_to_empty_scene (bool): Whether to apply the template
```
BigRoy: Can we rename this `apply_on_new_file`?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -559,15 +565,21 @@ def build_template(
...
+            if apply_on_app_launch is None:
+                apply_on_app_launch: bool = preset["apply_on_app_launch"]
 
         # Build the template if we are explicitly requesting it or if it's
         # an unsaved "new file".
-        is_new_file = not self.host.get_current_workfile()
-        if is_new_file or explicit_build_requested:
+        is_new_file = (
+            not self.host.get_current_workfile()
+            and apply_on_app_launch
```
BigRoy: This seems like an odd check for `is_new_file`? should this variable be `is_app_launch_empty_file` or whatever?

BigRoy (commented):
@iLLiCiTiT shall we discuss this code? Because I think you had bigger remarks to begin with?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -559,15 +565,21 @@ def build_template(
...
+        )
+        if is_new_file or apply_to_empty_scene or explicit_build_requested:
```
iLLiCiTiT:
```suggestion
        is_new_file = not self.host.get_current_workfile()
        if (
            apply_to_empty_scene
            or explicit_build_requested
            or (apply_on_app_launch and is_new_file)
        ):
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -559,15 +565,21 @@ def build_template(
                 keep_placeholders: bool = preset["keep_placeholder"]
             if create_first_version is None:
                 create_first_version: bool = preset["create_first_version"]
+            if apply_to_empty_scene is None:
+                apply_to_empty_scene: bool = preset["apply_to_empty_scene"]
```
iLLiCiTiT: `apply_to_empty_scene` and `apply_on_app_launch` probably won't be available for old addons?

iLLiCiTiT:
Isn't this trying to resolve same thing as the other PR that did not raise an error if preset does not have file?

Right now we do expect that all values from `get_template_preset` are always returned and validated, at multiple places. As I said at the other PR, we have to refactor the logic to handle that correctly, we can't just return `None` at method that is expected to always return validated values without handling it at all.

When we'd do that I'd rather do it properly.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT:
```suggestion
                apply_to_empty_scene: bool = preset.get("apply_to_empty_scene", False)
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -559,15 +565,22 @@ def build_template(
...
+            if apply_on_app_launch is None:
+                apply_on_app_launch: bool = preset["apply_on_app_launch"]
```
iLLiCiTiT:
```suggestion
                apply_on_app_launch: bool = preset.get("apply_on_app_launch", False)
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -511,6 +511,8 @@ def build_template(
         level_limit=None,
         keep_placeholders=None,
         create_first_version=None,
+        apply_to_empty_scene=None,
+        apply_on_app_launch=None,
```
iLLiCiTiT: I think `apply_on_app_launch` is something this method can't resolve. It probably is duplicate of `workfile_creation_enabled`.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -898,7 +917,9 @@ def get_template_preset(self):
         return {
             "path": resolved_path,
             "keep_placeholder": keep_placeholder,
-            "create_first_version": create_first_version
+            "create_first_version": create_first_version,
+            "apply_to_empty_scene": apply_to_empty_scene,
+            "apply_on_app_launch": apply_on_app_launch,
```
iLLiCiTiT: Idea:
`"apply_to_empty_scene"` and `"apply_on_app_launch"` really don't make sense (from my point of view) to be resolved within the logic of workfile builder.

Rather store the `profile` to the output, and use correct arguments in `build_template` based on host logic:
- `apply_to_empty_scene` is duplicate of `workfile_creation_enabled`.
- `apply_on_app_launch` cannot be resolved in any way shape or form in workfile builder logic, it is used only for the import (again, duplicate of `workfile_creation_enabled`).

```python
return {
    "path": resolved_path,
    "keep_placeholder": keep_placeholder,
    "create_first_version": create_first_version,
    "profile": profile,
}
```

Then you can do 
```python
# Somehow find out
just_lauched = ...

host = registered_host()
builder = TemplateBuilder(host)

preset = builder.get_template_preset()
profile = preset["profile"]
is_new_workfile = False
is_new_file = not self.host.get_current_workfile()
if just_lauched:
    is_new_workfile = profile["apply_to_empty_scene"]
elif is_new_file:
    is_new_workfile = profile["apply_on_app_launch"]
    
builder.build_template(
    template_path=preset["path"],
    keep_placeholder=preset["keep_placeholder"],
    create_first_version=preset["create_first_version"],
    workfile_creation_enabled=is_new_file,
)
```

I know this means the code has to be duplicated in each host that would want to do that, but this is the least ~~affecting~~ invasive code changes to avoid being even more blocked with future enhancements. Both values do affect the workfile template builder, but are not really related to the workfile builder. And yes we do have already the is_new_file logic, but why to add more to the pile.

moonyuet: I have two questions with this:
1. Isn't that we have discussed with @BigRoy about using dataclass for the return value? Or we dont need this now(Kinda offtrack though)
2. Aren't `keep_placeholder` and `create_first_version` already part of the profiles? So, I believe `apply_to_empty_scene` and `apply_on_app_launch` should be part of the profile settings?
image

iLLiCiTiT: > So, I believe apply_to_empty_scene and apply_on_app_launch should be part of the profile settings?

They can be part of the profile, but it doesn't make sense to have them as part of the workfile build preset. The workfile builder has nothing to do with the values. It is additional values that describe context before the workfile builder is being used.

The point is that the workfile template builder really needs more refactoring and adding more stuff that is at the end not used by the builder will only complicate future possible enhancements.

Apply to empty scene and on application launch cannot be handled by the workfile builder, it even should not be handled by the builder, you as someone who explicitly do trigger the builder know it, and you should pass that information to the build method, for both options and cases there already is argument that does handle that as far as I understand. If the same method is triggered manually from action the values from the preset should have no value, because at that moment it is not triggered on launch, or on empty workfile, and it should not apply the same logic (you can still trigger the builder manually).

LiborBatek (changes requested):
Been tested with both `blender 4.5.x` and `blender 5.1.0`

and as far as I can tell the only working option is `Create First Workfiles` which works for both blender versions too.

But the rest of  options (marked red) does not have any effect...as the building happens always no matter what are these knobs set to...

Image

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -562,18 +562,16 @@ def build_template(
...
-        if is_new_file or explicit_build_requested:
-            self.log.info(f"Building the workfile template: {template_path}")
-            self.import_template(template_path)
-            self.populate_scene_placeholders(
-                level_limit, keep_placeholders)
-
-        # Do not consider saving a first workfile version, if this is not set
-        # to be a "workfile creation" or `create_first_version` is disabled.
-        if explicit_build_requested or not create_first_version:
+        if not explicit_build_requested:
```
iLLiCiTiT: Did this have to change? Can this change can cause that some logic won't work? I have no idea, just asking.

BTW the comment above doesn't make sense.

moonyuet: Definitely. And I would also remove the comment.

moonyuet:
@LiborBatek I have committed the changes on this core PR and its relevant blender PR. Make sure you pull the latest of these two PRs and test. Thanks.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -562,18 +558,16 @@ def build_template(
...
-        if is_new_file or explicit_build_requested:
-            self.log.info(f"Building the workfile template: {template_path}")
-            self.import_template(template_path)
-            self.populate_scene_placeholders(
-                level_limit, keep_placeholders)
-
-        # Do not consider saving a first workfile version, if this is not set
-        # to be a "workfile creation" or `create_first_version` is disabled.
-        if explicit_build_requested or not create_first_version:
+        if not workfile_creation_enabled:
```
iLLiCiTiT: Can we keep the `explicit_build_requested` variable name? It really makes more sense while reading the code.

BTW the import and population of placeholders did happen before and it does not anymore, so anytime the build is triggered from the menu actions, nothing happens...

moonyuet: I have reverted the changes.

LiborBatek:
again as I have already commented in the dedicated blender PR.... Im having suspicion that some settings might have reversed `enabled` and `disabled` states....as it takes effect when disabled opposed to the enabled state (when there is no any effect of such settings)

Please check this one and confirm whether so or not really.

image

moonyuet:
I have done some decent tests in my side. If we need to make the workfile template build works with the relevant PR, we need to change the code below(not clean up version) to make the knobs work.
```python
    def build_template(
        self,
        template_path=None,
        level_limit=None,
        keep_placeholders=None,
        create_first_version=None,
        workfile_creation_enabled=False
    ):
        """Main callback for building workfile from template path.

        Todo:
            Handle report of populated placeholders from
                'populate_scene_placeholders' to be shown to a user.

        Args:
            template_path (str): Path to a template file with placeholders.
                Template from settings 'get_template_preset' used when not
                passed.
            level_limit (int): Limit of populate loops. Related to
                'populate_scene_placeholders' method.
            keep_placeholders (bool): Add flag to placeholder data for
                hosts to decide if they want to remove
                placeholder after it is used.
            create_first_version (bool): Create first version of a workfile.
                 When set to True, this option initiates the saving of the
                 workfile for an initial version. It will skip saving if
                 a version already exists.
            workfile_creation_enabled (bool): Whether the call is part of
                creating a new workfile.
                When True, we only build if the current file is not
                an existing saved workfile but a "new" file. Basically when
                enabled we assume the user tries to load it only into a
                "New File" (unsaved empty workfile).
                When False, the default value, we assume we explicitly want to
                build the template in our current scene regardless of current
                scene state.

        """
        # More accurate variable name
        # - logic related to workfile creation should be moved out in future
        explicit_build_requested = workfile_creation_enabled

        # Get default values if not provided
        if (
            template_path is None
            or keep_placeholders is None
            or create_first_version is None
        ):
            template_path: str = template_path or preset["path"]
            if keep_placeholders is None:
                keep_placeholders: bool = preset["keep_placeholder"]
            if create_first_version is None:
                create_first_version: bool = preset["create_first_version"]

        # Build the template if we are explicitly requesting it or if it's
        # an unsaved "new file".
        is_new_file = not self.host.get_current_workfile()
        if explicit_build_requested:
            self.log.info(f"Building the workfile template: {template_path}")
            self.import_template(template_path)
            self.populate_scene_placeholders(
                level_limit, keep_placeholders)

        # Do not consider saving a first workfile version, if this is not set
        # to be a "workfile creation" or `create_first_version` is disabled.
        if explicit_build_requested or not create_first_version:
            return

        # If there is no existing workfile, save the first version
        workfile_path = self.get_workfile_path()
        if not os.path.exists(workfile_path):
            self.log.info("Saving first workfile: %s", workfile_path)
            self.save_workfile(workfile_path)
        else:
            self.log.info(
                "A workfile already exists. Skipping save of workfile as "
                "initial version.")
```
But this definitely affects other hosts with templated builder for sure. We might need to figure out some other solutions for that.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -898,7 +908,8 @@ def get_template_preset(self):
         return {
             "path": resolved_path,
             "keep_placeholder": keep_placeholder,
-            "create_first_version": create_first_version
+            "create_first_version": create_first_version,
+            "profile": copy.deepcopy(profile),
         }
```
iLLiCiTiT, line 1187:
```python
def trigger_on_app_launch(self, preset: dict[str, Any] | None = None) -> None:
    if preset is None:
        try:
        except:
            return

    if not preset["profile"]["execute_on_app_launch"]:
        return

    self.build_template(
        template_path=preset["path"],
        keep_placeholders=preset["keep_placeholders"],
        create_first_version=preset["create_first_version"],
        workfile_creation_enabled=False,
    )

def trigger_on_new_file(self, preset: dict[str, Any] | None = None) -> None:
    if preset is None:
        try:
        except:
            return

    if not preset["profile"]["execute_on_new_file"]:
        return

    self.build_template(
        template_path=preset["path"],
        keep_placeholders=preset["keep_placeholders"],
        create_first_version=preset["create_first_version"],
        workfile_creation_enabled=False,
    )
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT:
```python
@dataclass
class TemplatePreset:
    path: str | None = None
    keep_placeholder: bool = True
    create_first_version: bool = True
    execute_on_app_launch: bool = False
    execute_on_new_file: bool = False
    # For integration internal logic
    profile: dict[str, Any] | None = None
...


    if not profile:
        return TemplatePreset()
```

```python
def is_valid_preset(self, preset: TemplatePreset) -> bool:
    if not preset.path or not os.path.exists(preset.path):
       return False
   return True
```

iLLiCiTiT:
```suggestion
path = profile["path"]
if path:
    path = self.resolve_template_path(path)

```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -898,7 +908,8 @@ def get_template_preset(self):
         return {
             "path": resolved_path,
```
iLLiCiTiT:
```suggestion
            "path": path,
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT, line 1171:
```suggestion
        create_first_version = profile.get("create_first_version")
        if create_first_version is None:
            create_first_version = False
```

iLLiCiTiT, line 688:
```suggestion

def build_template(self, preset: TemplatePreset) -> None:
    ...


def old_build_template(
    self,
    template_path=None,
    level_limit=None,
    keep_placeholders=None,
    create_first_version=None,
    workfile_creation_enabled=False
):
```

iLLiCiTiT:
```python
    def save_workfile(self, workfile_path: str | None = None) -> None:
        if workfile_path is None;
            workfile_path = self.host.get_current_workfile()
```

iLLiCiTiT:
```suggestion
```

LiborBatek (changes requested):
I have tested together with `Blender` and all of the options works well but last missing one been the `Create First Workfile` which malfunction and has no effect when enabled...

Other than that seems allrighty!

image

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,7 +533,134 @@ def get_placeholders(self):
...
+        @wraps(func)
+        def inner(self, *args, **kwargs):
+            if "preset" in kwargs:
+                preset = kwargs["preset"]
+                if (
+                    (preset is None or isinstance(preset, TemplatePreset))
+                    and len(kwargs) == 1
+                    and not args
+                ):
+                    return func(self, preset=preset)
```
iLLiCiTiT: If preset was passed in, I guess we should just run the preset method
```suggestion
                return func(self, preset=preset)
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,7 +533,134 @@ def get_placeholders(self):
...
+
+            if (
+                len(args) == 1
+                and isinstance(args[0], TemplatePreset)
+                and not kwargs
+            ):
+                return func(self, *args, **kwargs)
+
+            if not args and not kwargs:
```
iLLiCiTiT: I think the discussion was that by default it will run the old code and `preset` in the new method is required.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,7 +533,134 @@ def get_placeholders(self):
...
+        """Build the workfile template from a preset.
+
+        Args:
+            preset (TemplatePreset, optional): Template preset to use. When
+                omitted, the preset is resolved from project settings.
+        """
+        pass
+
+    @overload
+    def build_template(self,
```
iLLiCiTiT:
```suggestion
    def build_template(
        self,
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -554,11 +709,15 @@ def build_template(
             or create_first_version is None
         ):
             preset = self.get_template_preset()
-            template_path: str = template_path or preset["path"]
+            if not self._is_valid_preset(preset):
+                raise TemplateLoadFailed(
+                    f"Template path '{preset.path}' does not exist."
+                )
+            template_path: str = preset.path or preset["path"]
```
iLLiCiTiT:
```suggestion
            template_path: str = preset.path
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -648,12 +810,16 @@ def get_workfile_path(self):
...
+    def save_workfile(self, workfile_path: str | None = None) -> None:
+        """Save workfile in current host.
+
+        Args:
+            workfile_path (str | None): Path to save the workfile. If None,
+                the current workfile path will be used.
+
+        """
+        if workfile_path is None:
+            workfile_path = self.host.get_current_workfile()
```
iLLiCiTiT:
```suggestion
            workfile_path = self.host.get_current_workfile()
            if not workfile_path:
                return
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +988,93 @@ def _get_build_profiles(self):
...
+        Returns:
+            bool: True if the preset is valid, False otherwise.
+
+        """
+        if not preset.path or not os.path.exists(preset.path):
+            return False
+        return True
+
+    def trigger_on_app_launch(
+            self, preset: TemplatePreset | None = None) -> None:
```
iLLiCiTiT:
```suggestion
        self, preset: TemplatePreset | None = None
    ) -> None:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +988,93 @@ def _get_build_profiles(self):
...
+            preset = self.get_template_preset()
+
+        if not preset.execute_on_app_launch:
+            return
+
+        self.build_template(preset=preset)
+        self.save_first_version(preset=preset)
+
+    def trigger_on_new_file(
+            self, preset: TemplatePreset | None = None) -> None:
```
iLLiCiTiT:
```suggestion
        self, preset: TemplatePreset | None = None
    ) -> None:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +988,93 @@ def _get_build_profiles(self):
...
+        if preset is None:
+            preset = self.get_template_preset()
+
+        if not preset.execute_on_new_file:
+            return
+
+        self.build_template(preset=preset)
+
+    def create_first_workfile_version(
+            self, preset: TemplatePreset | None = None) -> None:
```
iLLiCiTiT:
```suggestion
        self, preset: TemplatePreset | None = None
    ) -> None:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +988,93 @@ def _get_build_profiles(self):
...
+                building the workfile template. If not provided, the default
+                template preset will be used.
+
+        """
+        if preset is None:
+            preset = self.get_template_preset()
+        self.build_template(preset=preset)
+        self.save_first_version(preset=preset)
+
+    def save_first_version(self, preset: TemplatePreset) -> None:
```
iLLiCiTiT:
```suggestion
    def save_first_workfile(self, preset: TemplatePreset) -> None:
```

iLLiCiTiT: Why do we need thid method? Should that be part of `create_first_workfile_version`?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +988,93 @@ def _get_build_profiles(self):
...
+        if preset.create_first_version:
+            workfile_path = self.get_workfile_path()
+            if not os.path.exists(workfile_path):
+                self.log.info("Saving first workfile: %s", workfile_path)
+                self.save_workfile(workfile_path)
+            else:
+                self.log.info(
+                    "A workfile already exists. Skipping save of workfile as "
+                    "initial version."
+                )
```
iLLiCiTiT: Is there a reason to not save it all the time?
```suggestion
        if not preset.create_first_version:
            return

        workfile_path = self.get_workfile_path()
        if not workfile_path:
            self.log.info("Workfile path to save is not available.")
            return

        self.log.info("Saving first workfile: %s", workfile_path)
        self.save_workfile(workfile_path)
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: Or, I think, we might call this `save_next_workfile_version`. We should change implementation of `get_workfile_path` to calculate next workfile version instead of using `AYON_LAST_WORKFILE` which might lead to different workfile, in case it is a DCC allowing to have opened multiple workfiles. That would also allow us to use the workfile interface methods instead of just `save_workfile` (which does not create workfile entity).

moonyuet: > Why do we need thid method? Should that be part of `create_first_workfile_version`?

Yes it should be.


> Or, I think, we might call this `save_next_workfile_version`. We should change implementation of `get_workfile_path` to calculate next workfile version instead of using `AYON_LAST_WORKFILE` which might lead to different workfile, in case it is a DCC allowing to have opened multiple workfiles. That would also allow us to use the workfile interface methods instead of just `save_workfile` (which does not create workfile entity).

We should think of this edge case though. This also means we need to change the setting name of `create_first_version`

BigRoy: - Create first version should only save if there's no existing version.
- It should not rely on env var `AYON_LAST_WORKFILE` because technically it's not accurate (only set on launch, not updated on context switches) although in practice you wouldn't easily face this issue since you'd often just switch context by saving to begin with.
- Using `save_next_workfile_version` is fine, if we only ever call it if there's no existing version.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
BigRoy: > Is there a reason to not save it all the time?

Studio preference, you may want to start with built workfile template, but the scene unsaved. It'd be my studio preference admittedly.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: I have committed with this [3783033](https://github.com/ynput/ayon-core/pull/1907/commits/378303365ec4dcabe2334ec2c97d212f9241bb83) but I actually think it is not a decent solution at all. can we just use the `save_next_version` if we enabled `create_first_version` along with that `filtering`? @BigRoy @iLLiCiTiT

BigRoy: Why do we need to copy the implementation and can't we just call `save_next_version` from core?

I'd assume we'd do:

```
if needs_to_save:
    save_next_version()
```

moonyuet: Implemented the change in [82f2006](https://github.com/ynput/ayon-core/pull/1907/commits/82f20067cf9b24e1f03748422fd95100b3d92a1f) and [385d0bf](https://github.com/ynput/ayon-core/pull/1907/commits/385d0bf92086ad80e5e21296afb7aba79735e997). Do we need to check the version of workfile across different workfile extensions?

LiborBatek (changes requested):
The `Create First Workfile` still fails to save the workfile as the very first version but the rest of the options works fine...

moonyuet:
> The `Create First Workfile` still fails to save the workfile as the very first version but the rest of the options works fine...

Fixed in [bcb9413](https://github.com/ynput/ayon-core/pull/1907/commits/bcb94133430b7b8d139022e150ab494b236603fe)

LiborBatek (changes requested):
The menu action works and saves the first workfile...

Image

but the addons default option for first workfile creation fails still

Image

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: ☝️

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -642,23 +808,66 @@ def get_workfile_path(self):
...
+            self.host_name,
+            project_settings=project_settings,
         )
+        file_template = anatomy.get_template_item("work", template_key, "file")
+        template_data = get_template_data(
+            project_entity,
+            folder_entity,
+            task_entity,
+            self.host_name,
+            project_settings,
```
iLLiCiTiT:
```suggestion
            project_settings=project_settings,
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: I think the `save_next_version` should save next version, no matter what. The logic which is triggering the method should decide if it should call it or not.

From my point of view `save_next_version` just saves the file as next version. The `create_first_workfile_version` should have the conditions for `create_first_version` and checks if there are any existing workfiles.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1031,95 @@ def _get_build_profiles(self):
...
+        Args:
+            preset (TemplatePreset | None): The template preset to use for
+                building the workfile template. If not provided, the default
+                template preset will be used.
+
+        """
+        if preset is None:
+            preset = self.get_template_preset()
+        self.build_template(preset=preset)
+        self.save_next_workfile_version(preset=preset)
```
iLLiCiTiT:
```suggestion
        if preset.create_first_version:
            workfile_path = self.get_workfile_path()
            if not os.path.exists(workfile_path):
                self.log.info("Saving first workfile: %s", workfile_path)
                save_next_version()
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: As I'm looking at the code, we don't need `save_next_version` method on the builder at all, just do [this](https://github.com/ynput/ayon-core/pull/1907/changes#r3544181182).

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1031,95 @@ def _get_build_profiles(self):
...
+            return
+
+        workfile_path = self.get_workfile_path()
+        if os.path.exists(workfile_path):
+            self.log.info("Workfile path to save is not available.")
+            return
+
+        self.log.info("Saving first workfile: %s", workfile_path)
+        save_next_version()
+
```
iLLiCiTiT: Not needed with https://github.com/ynput/ayon-core/pull/1907/changes#r3544181182 . The method does not do what the name says and there is `save_next_version` helper function in workfiles api already, so there is no point of having the method here.
```suggestion
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: BTW implementation of `get_workfile_path` did not change. It still looks for `AYON_LAST_WORKFILE`.

moonyuet: > BTW implementation of `get_workfile_path` did not change. It still looks for `AYON_LAST_WORKFILE`.

Is there any getter function or any insight to get workfile path which does not look for `AYON_LAST_WORKFILE`?

iLLiCiTiT: That is a good question. Considering that you might have already opened workfile we should first ask host if current workfile is set (I guess)? If is not then it should probably calculate last workfile for current context (and cache it so it is not recalculated on each call).

The `AYON_LAST_WORKFILE` does work if you just open a DCC and run the build, but if you then open different workfile in different context and you run the build action from menu, the `AYON_LAST_WORKFILE` is for completely different context.

moonyuet: [47794e2](https://github.com/ynput/ayon-core/pull/1907/commits/47794e2766521dc80a43f480c25166b22aaaac8b)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [47794e2](https://github.com/ynput/ayon-core/pull/1907/commits/47794e2766521dc80a43f480c25166b22aaaac8b)

LiborBatek (approved):
Works well together with the `ayon-blender` #313 and also been tested togehter with `Web Actions` from Ayon frontend succesfully!

LGTM

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,6 +541,68 @@ def get_placeholders(self):
...
+            func (Callable): The original build_template method.
+
+    """
+        @wraps(func)
+        def inner(self, *args, **kwargs):
+            if "preset" in kwargs:
+                preset = kwargs["preset"]
+                return func(self, preset=preset)
+
+            if (
```
iLLiCiTiT, line 591: I wrote it already, comment was probably resolved. Wasn't the agreement that the new method will explicitly require preset? Because now you can't call it in backwards compatible way... You will run it always the new way if no arguments are passed in, but the logic changed, so we are changing how this works for all places where not arguments are passed which is in all integrations using template builder

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,6 +541,68 @@ def get_placeholders(self):
...
+                return func(self, args[0])
+
+            # Legacy API fallback.
+            return self.old_build_template(*args, **kwargs)
+
+        setattr(inner, "has_preset", True)
+        return inner
+
+    @overload
+    def build_template(self, preset: TemplatePreset = None) -> None:
```
iLLiCiTiT: To match what `_backwards_compatibility_build_template` does.
```suggestion
    def build_template(self, preset: TemplatePreset) -> None:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,71 @@ def build_template(
...
+                build the template in our current scene regardless of current
+                scene state.
+
+        """
+        pass
+
+    @_backwards_compatibility_build_template
+    def build_template(
+        self,
+        preset: TemplatePreset = None,
```
iLLiCiTiT: To match what `_backwards_compatibility_build_template` does.
```suggestion
        preset: TemplatePreset,
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,71 @@ def build_template(
...
+    ) -> None:
+        """Build the workfile template from a preset.
+
+        Args:
+            preset (TemplatePreset, optional): Template preset to use. When
+                omitted, the preset is resolved from project settings.
```
iLLiCiTiT:
```suggestion
            preset (TemplatePreset): Template preset to use.

```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,6 +541,68 @@ def get_placeholders(self):
...
+        return inner
+
+    @overload
+    def build_template(self, preset: TemplatePreset = None) -> None:
+        """Build the workfile template from a preset.
+
+        Args:
+            preset (TemplatePreset, optional): Template preset to use. When
+                omitted, the preset is resolved from project settings.
+        """
```
iLLiCiTiT:
```suggestion
            preset (TemplatePreset): Template preset to use.

        """
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,6 +541,68 @@ def get_placeholders(self):
...
+                return func(self, args[0])
+
+            # Legacy API fallback.
+            return self.old_build_template(*args, **kwargs)
+
+        setattr(inner, "has_preset", True)
+        return inner
+
+    @overload
+    def build_template(self, preset: TemplatePreset) -> None:
```
iLLiCiTiT:
```suggestion
    def build_template(self, preset: TemplatePreset | None) -> None:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -505,6 +541,68 @@ def get_placeholders(self):
...
+        """Build the workfile template from a preset.
+
+        Args:
+            preset (TemplatePreset): Template preset to use.
```
iLLiCiTiT:
```suggestion
            preset (TemplatePreset | None): Template preset to use. To use
                default implementation use 'preset=None'.
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,71 @@ def build_template(
...
+                build the template in our current scene regardless of current
+                scene state.
+
+        """
+        pass
+
+    @_backwards_compatibility_build_template
+    def build_template(
+        self,
+        preset: TemplatePreset,
```
iLLiCiTiT:
```suggestion
        preset: TemplatePreset | None,
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,71 @@ def build_template(
...
+    ) -> None:
+        """Build the workfile template from a preset.
+
+        Args:
+            preset (TemplatePreset): Template preset to use.
```
iLLiCiTiT:
```suggestion
            preset (TemplatePreset | None): Template preset to use. To use
                default implementation use 'preset=None'.
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,71 @@ def build_template(
...
+
+        """
+        if not preset:
```
iLLiCiTiT:
```suggestion
        if preset is None:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: I removed the explicit check whether the function is without kwargs on [0b936c4](https://github.com/ynput/ayon-core/pull/1907/commits/0b936c454d3890c96e85ecf1f5dee75b34980fed) so that it won't return to the new function when there is no argument.

iLLiCiTiT (approved):

moonyuet:
@BigRoy is it good to merge?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,72 @@ def build_template(
     ):
         """Main callback for building workfile from template path.
```
BigRoy, line 624: Should we describe in the docstring that this legacy/deprecated or?

BigRoy (approved):
Seems to be an improvement, and works with https://github.com/ynput/ayon-blender/pull/313

The logic also seems to work somewhat with the Maya logic but there it's not working in all cases: https://github.com/ynput/ayon-maya/pull/454 so still needs tweaks there.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [6dfaa28](https://github.com/ynput/ayon-core/pull/1907/commits/6dfaa28d7d9ac69d849f5d1630de767b941870aa)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1018,94 @@ def _get_build_profiles(self):
             ["profiles"]
         )
 
-    def get_template_preset(self):
+    def _is_valid_preset(self, preset: TemplatePreset) -> bool:
```
iLLiCiTiT: This can be usefull as public method. Or maybe we can add `has_valid_path` to the preset itself.
```suggestion
    def is_preset_valid(self, preset: TemplatePreset) -> bool:
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -515,6 +613,73 @@ def build_template(
...
+
+        Args:
+            preset (TemplatePreset | None): Template preset to use. To use
+                default implementation use 'preset=None'.
+
+        """
+        if preset is None:
+            preset = self.get_template_preset()
+
+        if not self._is_valid_preset(preset):
```
iLLiCiTiT:
```suggestion
        if not preset.has_valid_path():
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -554,11 +719,15 @@ def build_template(
             or create_first_version is None
         ):
             preset = self.get_template_preset()
-            template_path: str = template_path or preset["path"]
+            if not self._is_valid_preset(preset):
```
iLLiCiTiT:
```suggestion
            if not preset.has_valid_path():
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -618,8 +787,11 @@ def rebuild_template(self):
     def open_template(self):
         """Open template file with registered host."""
         template_preset = self.get_template_preset()
-        template_path = template_preset["path"]
-        self.host.open_workfile(template_path)
+        if not self._is_valid_preset(template_preset):
```
iLLiCiTiT:
```suggestion
        if not template_preset.has_valid_path():
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -109,6 +119,32 @@ class TemplateLoadFailed(Exception):
...
+            settings.
+    """
+    path: str | None = None
+    keep_placeholder: bool = False
+    create_first_version: bool = True
+    execute_on_app_launch: bool = False
+    execute_on_new_file: bool = False
+    # For integration internal logic
+    profile: dict[str, Any] | None = None
+
```
iLLiCiTiT, line 146:
```suggestion

    def has_valid_path(self) -> bool:
        """Check if the template preset has valid path.

        Returns:
            bool: Preset has set path to an existing file.

        """
        return self.path and os.path.exists(self.path)
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [76889ef](https://github.com/ynput/ayon-core/pull/1907/commits/76889ef25c0050a91d4ed57d35108147e42e8104)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [76889ef](https://github.com/ynput/ayon-core/pull/1907/commits/76889ef25c0050a91d4ed57d35108147e42e8104)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [76889ef](https://github.com/ynput/ayon-core/pull/1907/commits/76889ef25c0050a91d4ed57d35108147e42e8104)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [76889ef](https://github.com/ynput/ayon-core/pull/1907/commits/76889ef25c0050a91d4ed57d35108147e42e8104)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: [fa5e75c](https://github.com/ynput/ayon-core/pull/1907/commits/fa5e75c29226823642976bbee55e54385fc2b169)

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1027,94 @@ def _get_build_profiles(self):
...
+            preset (TemplatePreset): The template preset to validate.
+
+        Returns:
+            bool: True if the preset is valid, False otherwise.
+
+        """
+        if not preset.path or not os.path.exists(preset.path):
+            return False
+        return True
+
```
iLLiCiTiT: With the `has_valid_path` is this obsolete.
```suggestion
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1027,94 @@ def _get_build_profiles(self):
...
+                building the workfile template. If not provided, the default
+                template preset will be used.
+
+        """
+        if preset is None:
+            preset = self.get_template_preset()
+
+        if preset.execute_on_new_file:
+            return
+
```
iLLiCiTiT: I don't think this is relevant for this method? I can use one preset for both new file and on app launch, or?
```suggestion
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1027,94 @@ def _get_build_profiles(self):
...
+        if not preset.execute_on_app_launch:
+            return
+
```
iLLiCiTiT, line 1047:
```suggestion

        if not preset.has_valid_path():
            return

```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1027,94 @@ def _get_build_profiles(self):
...
+                building the workfile template. If not provided, the default
+                template preset will be used.
+
+        """
+        if preset is None:
+            preset = self.get_template_preset()
+
+        if not preset.execute_on_new_file:
+            return
+
```
iLLiCiTiT, line 1075:
```suggestion

        if not preset.has_valid_path():
            return

```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -822,7 +1027,94 @@ def _get_build_profiles(self):
...
+        Args:
+            preset (TemplatePreset | None): The template preset to use for
+                building the workfile template. If not provided, the default
+                template preset will be used.
+
+        """
+        if preset is None:
+            preset = self.get_template_preset()
+        self.build_template(preset=preset)
```
iLLiCiTiT, line 1102:
```suggestion
        if not preset.has_valid_path():
            return

        self.build_template(preset=preset)
```

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: I don't know if it is intentional to raise an error if invalid preset is used?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: this is condition to avoid building template twice for just in case users set both `execute_on_new_file` and `execute_on_app_launch` to True in their addon settings.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: > I don't know if it is intentional to raise an error if invalid preset is used?

We need to somehow inform the preset path doesn't exist.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: That is not something the generic logic should handle. That is what host should handle. If I do want to trigger the preset because I can catch on app launch but I can't catch on new file, I don't care what the preset's value for on new file in settings is set to.

moonyuet: Ok fair enough

iLLiCiTiT: We can add an additional kwarg to allow this method to decide about that too?

something like
```python
def trigger_on_app_launch(
    self,
    preset: TemplatePreset | None = None,
    *,
    skip_on_new_file: bool = False,
) -> None:
    ...
    if skip_on_new_file and preset.execute_on_new_file:
        return
    ...
```
@BigRoy opinions?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
iLLiCiTiT: I'd say that makes sense only if you explicitly trigger an action. And all hosts that do that do not show any message even now.

client/ayon_core/pipeline/workfile/workfile_template_builder.py
BigRoy: Agree with Kuba, it's up to the host integration to handle not launching the new trigger on launch.

moonyuet: > Agree with Kuba, it's up to the host integration to handle not launching the new trigger on launch.

I have already adjusted to do so in the hosts instead

PR #1911, opened 2026-06-30
Integrate Input Links: Use bulk endpoint for creating many links at once

BigRoy:
## Changelog Description

Integrate Input Links: Use bulk endpoint for creating many links at once

## Additional info

Should use bulk endpoint starting with server 1.15.8 and up.
This should be much more optimal for many links, especially if you have high ping to the server.

## Testing notes:

1. Integrate links should work, even for high amount of links. So publish e.g. scenes with many containers in it + workfile publish enabled.
2. If you can, also test with older server than 1.15.8.


LiborBatek (approved):
I have tested it with 150+ containers and published render and workfile succesfully.

I have tested with ayon server 1.15.9 and older `1.15.6` both succesfully.

Im enclosing full json reports for comparison (times)

[maya-server-recent-publish-report-260702-11-28.json.txt](https://github.com/user-attachments/files/29592285/maya-server-recent-publish-report-260702-11-28.json.txt)

[maya-server_1_15_6_publish-report-260702-11-52.json.txt](https://github.com/user-attachments/files/29592291/maya-server_1_15_6_publish-report-260702-11-52.json.txt)

BigRoy:
Thanks @LiborBatek looks good - it seems to correctly switch between the API endpoints.

It did only do `6 links` though ;) but that's probably due to whatever's collected in your scene or whatnot. Seems totally fine I think. I assume you had the same product loaded many times?

LiborBatek:
> Thanks @LiborBatek looks good - it seems to correctly switch between the API endpoints.
> 
> It did only do `6 links` though ;) but that's probably due to whatever's collected in your scene or whatnot. Seems totally fine I think. I assume you had the same product loaded many times?

Heh, yeah you are right... few assets (about 6) but loaded many times...didnt think it might be a difference, glad its ok as is

iLLiCiTiT (approved):

PR #1913, opened 2026-07-03
Chore: Use 'PublishError' over 'KnownPublishError'

iLLiCiTiT:
## Changelog Description
Replace usage of `KnownPublishError` with `PublishError`.

## Additional info
`KnowPublishError` is deprecated for some time, we should stop using it. Modified the error messages a little.

## Testing notes:
1. Validate the code changes.
2. If you don't like it, do a suggestion please.


moonyuet (approved):
LGTM

BigRoy (approved):
If these display as the more user-friendly messages and they directly see the message, then I think this is the correct change.

iLLiCiTiT:
I was actually correct, but I don't know how it works.

`KnownPublishError` shows this
image

`PublishError` shows this
image

BigRoy:
Thanks, then Publish Error is the right approach here.

PR #1931, opened 2026-07-08
Workfiles API: Be able to define context change order

iLLiCiTiT:
## Changelog Description
Host can define if the context change does happen before or after workfile open.

## Additional info
There are integrations that do need to have set context before the workfile is openedand then there are integrations that need to set the context after the workfile is opened. By adding `change_context_before_workfile_open` attribute the host can tell the order.

This was added for hosts where context is stored to workfile metadata and the context change before workfile is openedwould actually affect the context of currently opened file instead of the file that will be opened.

## Testing notes:
1. Validate changes.
2. Compromise if it is wrong.

### DIFF
--- a/client/ayon_core/host/interfaces/workfiles.py
+++ b/client/ayon_core/host/interfaces/workfiles.py
@@ -834,6 +834,9 @@ class IWorkfileHost(AbstractHost):
         all host integrations.
 
     """
+    change_context_after_workfile_open = False
+    change_context_after_workfile_save = False
+
     @abstractmethod
     def save_workfile(self, dst_path: Optional[str] = None) -> None:
         """Save the currently opened scene.
@@ -956,15 +959,25 @@ class IWorkfileHost(AbstractHost):
         # Set 'AYON_WORKDIR' environment variable
         os.environ["AYON_WORKDIR"] = workdir
 
-        self.set_current_context(
-            folder_entity,
-            task_entity,
-            reason=ContextChangeReason.workfile_save,
-            project_entity=save_workfile_context.project_entity,
-            anatomy=save_workfile_context.anatomy,
-        )
+        if not self.change_context_after_workfile_save:
+            self.set_current_context(
+                folder_entity,
+                task_entity,
+                reason=ContextChangeReason.workfile_save,
+                project_entity=save_workfile_context.project_entity,
+                anatomy=save_workfile_context.anatomy,
+            )
 
         self.save_workfile(filepath)
+
+        if self.change_context_after_workfile_save:
+            self.set_current_context(
+                folder_entity,
+                task_entity,
+                reason=ContextChangeReason.workfile_save,
+                project_entity=save_workfile_context.project_entity,
+                anatomy=save_workfile_context.anatomy,
+            )
 
         self._save_workfile_entity(
             save_workfile_context,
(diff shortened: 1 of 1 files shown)


BigRoy:
I wouldn't do this admittedly - it'll complicate other things and it doesn't really make sense to me.

If the context needs to be burned in on scene open, you should always do it... not only if context changed after.

What case is there where on scene open you do not want to set your current context on it BUT do require it only if this particular open came with a context change?

iLLiCiTiT:
> it'll complicate other things

What does it complicate?

> If the context needs to be burned in on scene open, you should always do it... not only if context changed after.

You don't have access to context during the open/save methods. In case the host has custom implementation of get/set current context (which they do) then you can't access the context in which the file is being opened/saved. Right now they can't do it reliably exactly because of this order, because it break the workfiles.

> What case is there where on scene open you do not want to set your current context on it BUT do require it only if this particular open came with a context change?

It is about the order. The current default implementation first changes context and then opens the file. But in integrations like TVPaint, Photoshop and others where multiple workfiles can be opened at the same time, you have to first open file and then change the context, because if you would first change the context, you'd write the context to wrong workfile, and the one where it should be stored has nothing.

```
- set context (in those integrations nothing should happen because it would affect currently active workfile, not the one which should be opened)
- open file
```
Or
```
- open file
- set context (and store the context to the file)
```

~~I'm not sure if it makes sense to change context after workfile is being saved, but it is a must have for workfile opening in those integrations.~~

EDITED:
Removed the order for save, as it really doesn't make sense.

BigRoy:
> You don't have access to context during the open/save methods. In case the host has custom implementation of get/set current context (which they do) then you can't access the context in which the file is being opened/saved. Right now they can't do it reliably exactly because of this order, because it break the workfiles.

I think you can, the new context is set before open or save - as such, you always have access to the new context?

> you have to first open file and then change the context, because if you would first change the context, you'd write the context to wrong workfile, and the one where it should be stored has nothing.

I think this has the misconception that the context of an opened file changes during its lifetime, you either open it in a context or save it to another context. Currently we don't allow (from the tooling) to really change the context whilst being _within_ the same file without a save or open being involved. Which I feel has been a sensible decision so far?

As such, changing the context before open or save is the right approach taking that into consideration?

---

In short, you'd write the context on open, not on context change.. so you will be writing the correct context because we're changing context prior to file open.

BigRoy (approved):
Looks good.

It may be good to add a comment to the `change_context_before_workfile_open` variable to describe the use case. Other than that, awesome.

client/ayon_core/host/interfaces/workfiles.py
```diff
@@ -834,6 +834,12 @@ class IWorkfileHost(AbstractHost):
         all host integrations.
 
     """
+    # Can be changed to False if the host integration needs to change context
+    #   after workfile is opened
+    # - usefull in cases when the integration does store context to metadata
+    #   of the file
```
BigRoy:
```suggestion
    # Can be changed to False if the host integration needs to change context
    #   after workfile is opened, useful if the integration stores and retrieves 
    #   context to metadata of the file
```

`usefull` -> `useful`

But also generally reformatted it.

BigRoy (approved):
Aside of comment typo - lgtm.

PR #1933, opened 2026-07-08
Cap rendered frames to prevent ffmpeg over-read

### LINKED ISSUE #1932: YN-0890: Hiero publish - running out of disk space for tmp files
Closing in favor of PR #1933 which adds an explicit frame cap when rendering image sequences to prevent ffmpeg from reading past the intended range.

jakubjezek001:
## Changelog Description
Cap output frames when rendering image sequences and OTIO segments by computing an explicit `frames_to_render` and passing `-frames:v` to ffmpeg. This prevents the ffmpeg image2 demuxer from reading files past the intended range and ensures rendered outputs contain the correct number of frames. Affects the publish reviewer/renderer.

## Additional info
- Implementation: compute `frames_to_render` from sequence collection indexes or OTIO range duration and add `-frames:v <n>` to the ffmpeg command.
- Files: `client/ayon_core/plugins/publish/extract_otio_review.py`.

## Testing notes:
1. Render a review segment from an image sequence that has extra numbered files beyond the intended range.
2. Confirm ffmpeg stops at the intended frame and output duration matches the OTIO segment.
3. Compare outputs with/without this change to observe that over-read is prevented.

## Dependency
Close #1932
[YN-0890](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=ed865d6fe0bf470791083161b01a7932)


client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -505,6 +512,7 @@ def _render_segment(self, sequence=None,
             frame_start = otio_range.start_time.value
             input_fps = otio_range.start_time.rate
             frame_duration = otio_range.duration.value
+            frames_to_render = int(round(frame_duration))
```
BigRoy: Should this variable be named `frames_num` or `frames_count`. Is `frames_to_render` too ambiguous?

client/ayon_core/plugins/publish/extract_otio_review.py
```diff
@@ -466,9 +466,16 @@ def _render_segment(self, sequence=None,
         command = get_ffmpeg_tool_args("ffmpeg")
 
         input_extension = None
+        # Number of frames the output should contain. Explicitly capping
+        # this on the output side protects against ffmpeg reading past
+        # the intended range, e.g. the image2 demuxer has no "end frame"
+        # option and will keep consuming consecutive numbered files on
+        # disk beyond what this segment actually needs.
+        frames_to_render = None
```
BigRoy: Can we make this less ambiguous by enforcing it to be `int`, like have the default be `0`?

Just so we can do:
```
frames_num: int = 0
```
To enforce it's always integer, etc.

BigRoy (commented):
Makes sense - I'm no ffmpeg export so no idea if `-frames:v` has side effects. E.g. what if it's a video that has a variable frame rate? Does that work?

client/ayon_core/plugins/publish/extract_otio_review.py
jakubjezek001: `frame_count` should improve reading

client/ayon_core/plugins/publish/extract_otio_review.py
jakubjezek001: better to add the arg only `if frames_count is not None`

BigRoy (approved):
LGTM

jakubjezek001:
Client confirmed all works fine for them.

BigRoy:
@jakubjezek001 looks like the unit tests are failing actually - there may be a case where we actually broke things? (or the test itself was wrong? 😄) please check it when you have some time.

iLLiCiTiT:
The tests has to be modified to also expect the `-frames:v xxx` in arguments.

BigRoy:
@iLLiCiTiT @jakubjezek001 can you confirm the test changes? :) tests now pass.

iLLiCiTiT (approved):

PR #1935, opened 2026-07-08
Delivery: compute publishedBasename before destination path validation

### LINKED ISSUE #1934: YN-0905: delivery implementing publishedFilename representaion dynamic context key

Currently if mixture of versions with representations which are either having or not having `originalFilename` key and Anatomy template key is used in delivery template, then processing is failing. 

If we introduce new dynamic key which will be added to processing representation context data called `plublishedFilename` we will secure successfull processing and yet will be able to preserve `originalFilenames` for those which are published with it.  

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1514696776681128107/1524323737992957972)

_This issue was automatically created from AYON ticket [YN-0905]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=309136d05254441b9ae5ac4c36ce674c)

jakubjezek001:
## Changelog Description
Delivery now computes the `publishedBasename` template key (via `pathlib.Path`) before validating the destination path, instead of after. Previously, destination validation ran before `publishedBasename` was set, so templates relying on that key always failed the check.

## Additional info
- Sequences: basename is derived from the first source file with the frame number and trailing separators stripped.
- Single files: basename is used as-is.

## Testing notes:
1. Deliver a sequence representation using a template containing `{publishedBasename}` and confirm the destination path resolves and validates correctly.
2. Deliver a single-file representation using a template containing `{publishedBasename}` and confirm the same.

## Dependency
Close #1934
[YN-0905](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=309136d05254441b9ae5ac4c36ce674c)

### DIFF
--- a/client/ayon_core/plugins/loader/delivery.py
+++ b/client/ayon_core/plugins/loader/delivery.py
@@ -1,7 +1,7 @@
 import platform
 from collections import defaultdict
 from typing import Optional, Any
-
+from pathlib import Path
 import ayon_api
 from qtpy import QtWidgets, QtCore, QtGui
 
@@ -248,17 +248,6 @@ class DeliveryOptionsDialog(QtWidgets.QDialog):
             )
 
             template_data = template_data_by_repre_id[repre["id"]]
-            new_report_items = check_destination_path(
-                repre["id"],
-                self.anatomy,
-                template_data,
-                datetime_data,
-                template_name
-            )
-
-            report_items.update(new_report_items)
-            if new_report_items:
-                continue
 
             args = [
                 repre_path,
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/loader/delivery.py
```diff
@@ -285,8 +274,34 @@ def deliver(self):
...
+                published_basename = Path(first_file_path).stem
+            template_data["publishedBasename"] = published_basename
+
+            new_report_items = check_destination_path(
+                repre["id"],
+                self.anatomy,
+                template_data,
+                datetime_data,
+                template_name
+            )
```
BigRoy, line 300: I actually my doubts about still manipulating the `publishedBasename` like stripping of the frame numer.

First `replace(first_frame, "")` is a wrong assumption.

If my shot is 1100 and the first frame happens to be 1100 then suddenly I'm also removing the shot name from the shot.

What's the reason we can't just have `publishedFilename` be the full source filename?

So essentially:

```python
published_filename = os.path.basename(source)
```

And that's it. It doesn't make much sense to use `{publishedFilename}` with any other tokens (unless you want to prefix it perhaps) but at least it's very clear what it does, it's the full filename of the file we're copying over?

jakubjezek001: but how do you deal with frames and what if you want to renumber frames?

BigRoy: In that scenario you wouldn't renumber frames - you'd get the exact filename that lives in the publish directory essentially.

@iLLiCiTiT thoughts? am I oversimplifying?

iLLiCiTiT: I thought we won't use `publishedBasename` but `publishedFilename` -> use it as it is, no renumbering etc. Even for basename it doesn't make sense to handle renumbering, basename is just extensionless name... As Roy mentioned it should be just `os.path.basename` for each file.

BigRoy: Thanks - I agree; it's the easiest route forward here. So let's make sure we implement a key that just takes the source representation file's filename.

BigRoy: @jakubjezek001 I took the liberty to quickly give this a whirl, see: https://github.com/ynput/ayon-core/pull/1945

jakubjezek001:
continuation in #1945

PR #1936, opened 2026-07-09
Added delivery webaction

kalisp:
## Changelog Description
Adds webaction that triggers `DeliveryDialog` in AYON launcher (if installed on user machine).

Could be triggered on single or multiple selected versions (in `Products` tab).

## Additional info
It should trigger this dialog if you haven't closed it before.

image

## Testing notes:
1. select versions in `Products` or `Lists` tab, trigger webaction

image

2. fill options in Dialog (expects filled at least one `delivery` template in `Anatomy`)
image

3. check delivery folders


jakubjezek001:
Tested and it works and doing what it suppose to do.

As we had discussed on the call with @BigRoy it would be also good to include some of the UX cosmetics:
- available representations offering convert to list widget so it is easier to select with SHIFT or CTRL/CMD keys. 
- perhaps add button to select all available representations. 
- my suggestion is to also collapse the section with renumbering and root into **Advanced features**

client/ayon_core/cli.py
```diff
@@ -22,6 +22,7 @@
     compute_env_variables_structure,
     merge_env_variables,
 )
+from ayon_core.tools.delivery.delivery import DeliveryOptionsDialog
```
iLLiCiTiT: Move this import to the function.

Also change the import to
`from ayon_core.tools.delivery import DeliveryOptionsDialog`
https://github.com/ynput/ayon-core/pull/1936/changes#r3551918669

client/ayon_core/tools/delivery/delivery.py
```diff
@@ -0,0 +1,372 @@
+import platform
```
iLLiCiTiT, line 1: Please add `__init__.py` where `DeliveryOptionsDialog` is imported to `__all__`. And rename this file to `window.py` (or `dialog.py`).

client/ayon_core/plugins/loader/delivery.py
```diff
@@ -1,30 +1,12 @@
...
     LoaderActionResult,
 )
-from ayon_core.pipeline.load import get_representation_path_with_anatomy
-from ayon_core.pipeline.delivery import (
-    get_format_dict,
-    check_destination_path,
-    deliver_single_file,
-    get_representations_delivery_template_data,
-)
+from ayon_core.tools.delivery.delivery import DeliveryOptionsDialog
```
iLLiCiTiT:
```suggestion
from ayon_core.tools.delivery import DeliveryOptionsDialog
```
Related to https://github.com/ynput/ayon-core/pull/1936/changes#r3551918669

client/ayon_core/cli.py
```diff
@@ -230,6 +231,49 @@ def create_project_structure(
...
+    "--project",
+    type=str,
+    help="Project name",
+    required=True)
+@click.option(
+    "--version_ids",
+    type=str,
+    help="Version ids",
+    required=True)
+def delivery(
```
iLLiCiTiT:
```suggestion
def deliver(
```

client/ayon_core/cli.py
```diff
@@ -230,6 +231,49 @@ def create_project_structure(
...
+        get_qt_app()
+        version_ids = version_ids.split(",")
+        dialog = DeliveryOptionsDialog(
+            project, version_ids, log=log
+        )
+        dialog.exec_()
+    except Exception:
+        print("Failed to deliver versions.")
+        exc_info = sys.exc_info()
+        traceback.print_exception(*exc_info)
```
iLLiCiTiT:
```suggestion
        log.warning("Failed to deliver versions.", exc_info=True)
```

client/ayon_core/cli.py
```diff
@@ -230,6 +231,49 @@ def create_project_structure(
...
+
+    print(f">>> Launching browser for Delivery action '{project}'.")
+
+    log = Logger.get_logger("delivery")
+
+    try:
+        # must be here because no module qargparse
+        from ayon_core.tools.utils.lib import get_qt_app
+
+        get_qt_app()
```
iLLiCiTiT:
```suggestion
        _app = get_qt_app()
```

client/ayon_core/cli.py
```diff
@@ -230,6 +231,49 @@ def create_project_structure(
     create_project_folders(project)
 
 
+@main_cli.command()
```
iLLiCiTiT, line 234: For the webaction purposes, I'd say we might need to expect event id where are version ids listed. There is a limit for length of executed command, I can imagine that with multiselection of versions or list, it can easily hit the limit of characters.

With that I'd rename the command to `deliver_ui_from_action` (called as `deliver-ui-from-action`) as the current logic does not have mother usage.

So webaction would create event where would be stored all information needed for delivery command, then start launcher as `["deliver-ui-from-action", "--project", "<project>", "--event", "<the event id>"]`.

kalisp: Make sense, implemented emitting event and querying it.

jakubjezek001:
> * available representations offering convert to list widget so it is easier to select with SHIFT or CTRL/CMD keys.
> * perhaps add button to select all available representations.
> * my suggestion is to also collapse the section with renumbering and root into **Advanced features**

suggestion here https://github.com/ynput/ayon-core/pull/1937

client/ayon_core/cli.py
```diff
@@ -230,6 +230,68 @@ def create_project_structure(
...
+        return
+
+    try:
+        from ayon_core.tools.delivery.delivery import DeliveryOptionsDialog
+        # must be here because no module qargparse
+        from ayon_core.tools.utils.lib import get_qt_app
+
+        _app = get_qt_app()
+        if event_id:
+            import ayon_api
```
iLLiCiTiT: ayon api can be imported at the top.

iLLiCiTiT (approved):
The webaction part looks ok (didn't test). I would personally do the dialog enhancements in different PR, but I'll let you decide if it will be here too.

BigRoy (approved):
Works for me - very nice!

I think we should update the delivery dialog itself a bit to be more informational, but other than that it seems to work.

@iLLiCiTiT merge?

> I would personally do the dialog enhancements in different PR, but I'll let you decide if it will be here too.

Agreed. Started here: https://github.com/ynput/ayon-core/pull/1937

client/ayon_core/cli.py
BigRoy: done with https://github.com/ynput/ayon-core/pull/1936/commits/3e3b0b58a5fef42ccb8e685f08cc0c047ed72a19

client/ayon_core/cli.py
BigRoy: That's actually a very good point. So we'd just store event id and store the version ids in the payload to retrieve from the event id. I like it.

Should we make the cli command take either `event_id` or `version_ids`. From a CLI perspective it'd be nice if you can quickly pass version ids if you wan to run e.g. manually, then there's no pre-existing event. Or am I overthinking it and for now should we focus on web action use case only?

server/__init__.py
```diff
@@ -60,6 +61,20 @@ async def get_simple_actions(
                     allow_multiselection=False,
                 )
             )
+            output.append(
+                SimpleActionManifest(
+                    identifier="core.delivery",
+                    label="Start Delivery Action in Launcher",
```
jakubjezek001:
```suggestion
                    label="Delivery Versions (launcher action)",
```

jakubjezek001: @BigRoy what about this?

BigRoy: If we do change this, I'd make it "Deliver versions" and possibly "requires launcher".

```suggestion
                    label="Deliver versions (requires launcher)",
```

PR #1937, opened 2026-07-09
Delivery tool: replace representation checkboxes with a multi-select list

jakubjezek001:
## Changelog Description
Delivery tool's representation selection now uses a `QListWidget` with multi-selection and a "Select All" button instead of a checkbox per representation. Advanced options (renumber frame, start frame, root) are moved into a collapsible "Advanced options" section, hidden by default, to simplify the default view.

## Additional info
Single file changed: `client/ayon_core/tools/delivery/delivery.py`. Representation names are now sorted for consistent ordering, and `_get_selected_repres` reads from `repre_list.selectedItems()` instead of iterating checkboxes.

## Testing notes:
1. Open the Delivery tool and confirm representations list appears sorted, with multi-select and "Select All" working.
2. Confirm delivery still triggers correctly for selected representations.
3. Expand/collapse "Advanced options" and verify renumber/root fields still function as before.

image

image

## Dependency
#1936

### DIFF
--- a/client/ayon_core/tools/delivery/delivery.py
+++ b/client/ayon_core/tools/delivery/delivery.py
@@ -33,7 +33,7 @@ class DeliveryOptionsDialog(QtWidgets.QDialog):
     ):
         super().__init__(parent=parent)
 
-        self.setWindowTitle("AYON - Deliver versions")
+        self.setWindowTitle(f"Deliver {len(version_ids)} versions")
         icon = QtGui.QIcon(resources.get_ayon_icon_filepath())
         self.setWindowIcon(icon)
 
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Works.

BigRoy:
- Can we make the title of the dialog show `Deliver {len(version_ids)} versions` so at least you have a hint of how big the selection was - this becomes increasingly important when ran from web action to confirm your selection came through. (Without redoing the full dialog to actually show what you had selected, which I think we may need to do later)

- We'll need to resolve conflicts after PR https://github.com/ynput/ayon-core/pull/1936

jakubjezek001:
> * Can we make the title of the dialog show `Deliver {len(version_ids)} versions` so at least you have a hint of how big the selection was - this becomes increasingly important when ran from web action to confirm your selection came through. (Without redoing the full dialog to actually show what you had selected, which I think we may need to do later)

just like this, right?
image

jakubjezek001:
@BigRoy it seems to me the Selected Represenations is also not correctly displaying data. we should also add len of selected representations, right?

jakubjezek001:
Well, I could not resist to add the repre count ;)
image

BigRoy (approved):
Works for me.

PR #1947, opened 2026-07-13
Review Desktop: Submit the user tag correctly to the server

moonyuet:
## Changelog Description
This PR is to fix the bug on the user mentions in the review desktop when users submitting the comment with the user tags.
The `(User)[user:user]` would be submitted as when user add @ User to the comment section.

## Additional info
Resolve https://github.com/ynput/ayon-review-desktop/issues/31

## Testing notes:
1. Launch Review Desktop
2. Test with users mention
3. Should be submitting to the server


BigRoy (changes requested):
Doesn't seem to work for me.

Applying the autocomplete for a user tag I get:

Image

Which should show as `@Johny Bravo`
Upon submitting, checking the web I see:

Image

Which should also, have been a regular tag.
When the tag is made on the web frontend in a comment, then it displays like this in the activity panel:

Image

Which should really be the nicer color - and preferably when clicked or hovered be more useful and provide more details to the user. If that's too complex, then for now let's just print that this link is not yet implemented instead of getting e.g.

Image

The easiest is most likely that the `@user` tag just remains direct text like that inside the comment field but only on submission gets converted to the relevant user markdown link, but I'll leave the implementation up to you.

---

Also, the styling on the user autocomplete pop-up is essentially unreadable:

Image

I think we should fix that too.

moonyuet:
@BigRoy As we discussed in the call, we already fixed the link issue regarding to [f1fa973](https://github.com/ynput/ayon-core/pull/1947/commits/f1fa97362afd7479f7716665f5fb95841caee14a).

I have fixed the popup style issue by [7b2abbc](https://github.com/ynput/ayon-core/pull/1947/commits/7b2abbcb23398bdce273889cf237dc4cda7ae4d9) to increase the readability.
image

It appears that we may need to do much more or even have our own QTextDocument as we default to use github style of which the user tag @ user would be `[user](user:name)` , which means the url in github style. It is okay in the first step when we parsed the word edit `[user](user:name)` back to the @ user as user tags in this PR. Nevertheless, we need to consider the case when we have the version tag and task tag. Are we gonna use the same approach such as for version tags: ([version](version: version_number, version:id)) or task tags: ([task](task:task_id)))?

moonyuet:
I have fixed the first part of issue for the user tags.
image
[8d67798](https://github.com/ynput/ayon-core/pull/1947/commits/8d677983b2fedc93de1a4fb5d3037df863edf124)

I will fix the styling issue faced during the edit text.

client/ayon_core/ui/components/comment.py
```diff
@@ -322,12 +330,172 @@ def set_markdown(self, md: str) -> None:
...
+
+            split_result = self._split_label_around_user_token(
+                label, full_name, user_id
+            )
+            if split_result is None:
+                return f"[@{label}](user:{user_id})"
+
+            prefix, suffix = split_result
+            return f"{prefix}[@{full_name}](user:{user_id}){suffix}"
+
```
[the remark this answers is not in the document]
moonyuet: without [@Admin](user:admin), it would lose the tag on the server. So reverted the changes

moonyuet:
@BigRoy can you test from your side? 
We can essentially to the popup of user during hovering on the user tags in a separate PR.

BigRoy:
Much better!

---

- [x] Fix popup colors

Can we get the pop-up background colors to match what the frontend shows (incl. the highlighted row)

image

This is what I get now:

image

---

- [x] Fix tag directly displaying as `@User` on completion of popup

As soon as we get the autocomplete finished (from the popup) can we display the user tag directly as `@User` instead of `[User](user:user)`?

image

client/ayon_core/ui/components/comment.py
```diff
@@ -322,12 +333,177 @@ def set_markdown(self, md: str) -> None:
...
+        label: str,
+        full_name: str,
+        user_id: str,
+    ) -> tuple[str, str] | None:
+        """Split label into prefix/suffix around a user mention token.
+        Supports malformed labels like:
+        - "1234 @Joe"
+        - "1234 @joe"
+        - "Joe 1234"
+        - "Joe"
```
BigRoy: Why would there be malformed labels? What's this solving?

client/ayon_core/ui/components/comment_completion.py
```diff
@@ -323,8 +337,8 @@ class MentionHighlighter(QSyntaxHighlighter):
     Patterns highlighted:
 
     - Fenced code blocks (```\\`\\`\\` ... \\`\\`\\```) spanning multiple
-      lines - black background, white monospace text.
-      Block state ``1`` tracks whether the current block is inside a fence.
+      lines — black background, white monospace text.  Block state ``1``
```
BigRoy:
```suggestion
      lines - black background, white monospace text.  Block state ``1``
```

client/ayon_core/ui/components/comment.py
```diff
@@ -322,12 +333,177 @@ def set_markdown(self, md: str) -> None:
                 self._adjust_height_to_content()
             return
```
BigRoy, line 328: Bit surprised that the user mention works fine with checkboxes inside the markdown. Which makes me wonder if the `_inject_user_mention_display` is even still needed... but it seems to still be needed otherwise comment submissions with user tags do not work. 🤔

BigRoy (approved):
Technically this works. Code-wise it's a bit of a hassle, but it looks like it's ok within the surrounding code. 

I'm always surprised with the lengths you need to go to to make something like this work in Qt. I feel like there must be a way to be able to do this hat doesn't require the actual string manipulation itself on submission, etc. so that it's purely a visual rendering thing, but I do not have the knowledge on how that'd work.

moonyuet:
> Technically this works. Code-wise it's a bit of a hassle, but it looks like it's ok within the surrounding code. 
> 
> I'm always surprised with the lengths you need to go to to make something like this work in Qt. I feel like there must be a way to be able to do this hat doesn't require the actual string manipulation itself on submission, etc. so that it's purely a visual rendering thing, but I do not have the knowledge on how that'd work.

I don't think these are the neat solutions either. I am thinking if we can reuse the regex we used in the mentionhighlight class and do the conversion.

client/ayon_core/ui/components/comment.py
moonyuet: I do some cleanup. 
`strip_user_mention_display`: is to convert @mention to [mention] ( user:name )
`_inject_user_mention_display` is to convert [mention] ( user:name ) to @mention

client/ayon_core/ui/components/comment.py
```diff
@@ -322,12 +324,47 @@ def set_markdown(self, md: str) -> None:
...
+        Args:
+            user_id: User ID to look up.
+
+        Returns:
+            Full name of the user if found, otherwise None.
+        """
+        for user in self._user_list:
+            if user.name == user_id:
+                return user.full_name.strip()
+        return None
```
BigRoy: The user list may be large; which makes me thing. Should this really be a dict `self._user_by_id` so that we can just do: `return self._user_by_id.get(user_id)` which will be more performant?

Same goes for the other `find_user_for_mention` calls, would be much more performant if we were to lookup the reverse way `user_by_full_name` dict so that it's o1 lookup.

With say 5 tags and 800 users on the server, we may potentially be doing 5x800 iterations whereas otherwise we'd just do the 5x lookup in dicts. Although even 4000 iterations wouldn't be crazy slow with this lookup, so probably still unnoticable in the majority of scenarios.

BigRoy (approved):
Still works, I do think we can improve the user id / full name lookups still though.

client/ayon_core/ui/components/comment.py
moonyuet: [485d572](https://github.com/ynput/ayon-core/pull/1947/commits/485d572273906566d7d0963b4c3aa17ac0e5e9c7) and [e2d7018](https://github.com/ynput/ayon-core/pull/1947/commits/e2d701857bbe1a700ac5e33ab82b2ead301c6d92)

client/ayon_core/ui/components/comment.py
```diff
@@ -239,6 +241,9 @@ def __init__(
         self._num_lines = num_lines
         self._read_only: bool = read_only
         self._user_list: list[User] = user_list or []
+        self._user_full_name_by_user_id = {
+            user.name: user.full_name for user in self._user_list
+        }
```
BigRoy: My guess is that it may actually be more efficient memory-wise if this were just `user_by_user_id`. Because the `user` dataclass would be stored by reference (mutable object) but the full name string is immutable and a copy of the string would get stored as value - potentially taking up more memory because of it.

I think we can ignore it for now - but something to be aware of when doing things like this.

BigRoy (approved):
Nice. Looks good to me.

@iLLiCiTiT can you check and merge if ok with you?

client/ayon_core/ui/components/text_box.py
```diff
@@ -1,5 +1,4 @@
 from __future__ import annotations
-
```
iLLiCiTiT, line 2: I do believe it is recommended to keep the new line between the imports from future and rest of the imports.

client/ayon_core/ui/components/text_box.py
```diff
@@ -272,7 +272,8 @@ def as_markdown(self) -> str:
         """
         if self._checkbox_handler and self._checkbox_handler.has_checkboxes():
             return self._checkbox_handler.to_markdown()
-        return self.document().toMarkdown(MD_DIALECT)
+        rendered_md = self.document().toMarkdown(MD_DIALECT)
+        return strip_user_mention_display(rendered_md, self._user_list)
```
iLLiCiTiT, line 277: This is `AYTextEditor`, a text editor component. Why it should by default strip user mentions?

client/ayon_core/ui/components/comment.py
```diff
@@ -239,6 +241,9 @@ def __init__(
         self._num_lines = num_lines
         self._read_only: bool = read_only
         self._user_list: list[User] = user_list or []
+        self._user_full_name_by_user_id = {
```
iLLiCiTiT:
```suggestion
        self._user_full_name_by_username = {
```

client/ayon_core/ui/components/comment.py
```diff
@@ -322,12 +327,33 @@ def set_markdown(self, md: str) -> None:
...
+        Args:
+            md: Markdown text with links like [label](user:id)
+
+        Returns:
+            str: Text with links replaced by @full_name (or the original link
+                if the user is not found).
+        """
+        def repl(match) -> str:
+            user_id = match.group("id")
+            full_name = self._user_full_name_by_user_id.get(user_id)
```
iLLiCiTiT:
```suggestion
            username = match.group("username")
            full_name = self._user_full_name_by_username.get(username)
```

client/ayon_core/ui/components/comment_completion.py
```diff
@@ -30,9 +30,77 @@
...
 
+# Match plain user mentions like "@Joe" or "@Joe Smith".
+# The optional second token must start with a letter to avoid swallowing
+# trailing numbers (e.g. "@Joe 1234" should only match "@Joe").
+USER_MENTION_PATTERN = re.compile(
+    r"(?<![\w\[])@(?!@)\w+(?: [^\W\d_][\w'-]*)?"
+)
+# Match stored user links like "[Joe](user:admin)".
+USER_MENTION_LINK_PATTERN = re.compile(
+    r"\[(?P<label>[^\]]+)\]\(user:(?P<id>[^)]+)\)"
```
iLLiCiTiT:
```suggestion
    r"\[(?P<label>[^\]]+)\]\(user:(?P<username>[^)]+)\)"
```

client/ayon_core/ui/components/comment.py
iLLiCiTiT: I'd not call it `user_id` if it is `username`.

client/ayon_core/ui/components/text_box.py
moonyuet: It needs to convert display @ joe back to storage links [joe] (user: joe) so that it would show @ joe in frontend

client/ayon_core/ui/components/comment_completion.py
```diff
@@ -30,9 +30,77 @@
...
+)
+
+
+def strip_user_mention_display(md: str, user_list: list[User]) -> str:
```
iLLiCiTiT, line 49: I have a note to this. I think this would work, but I'd recommend to find users in the text intead of finding them by matching regex.

```python
    # Sort users by longest full name
    for user in sorted(user_list, key=lambda u: len(u.full_name), reverse=True):
        mention = f"@{user.full_name}"
        if mention not in md:
            continue
        md = md.replace(mention, f"[{user.full_name}](user:{user.name})")
```

I didn't test it so it might not work. But it might be actually more predictable??

client/ayon_core/ui/components/text_box.py
iLLiCiTiT: I know why it was added, but this class is meant as replacement for `QTextEditor`. With this change it is targeted class for comments, so we can't do that.

iLLiCiTiT: Oh my god, it is embeded into the component. Nevermind. Problems of future us I guess...

client/ayon_core/ui/components/comment_completion.py
```diff
@@ -56,9 +124,11 @@ def paint(
 
         # Draw background
         if option.state & QStyle.StateFlag.State_Selected:
-            painter.fillRect(option.rect, option.palette.light())
+            painter.fillRect(option.rect, COMPLETER_ROW_SELECTED_BG)
```
iLLiCiTiT: Was there a reason why these colors were hardcoded?

I do believe the idea was that the style does define the colors, so we should either fix that or add TODO comment here explaining why this was needed so it can be fixed in future.

client/ayon_core/ui/components/text_box.py
BigRoy: I think the name here should really be `AYCommentEditor` :) but I guess indeed for some refactoring later. Not a new issue from this PR I think.

client/ayon_core/ui/components/comment_completion.py
BigRoy: Are we sure we really want to call this `replace` call for EVERY user? because that'd do the `str.replace` match PER user? Seems rather slow if I have 800 users?

client/ayon_core/ui/components/comment_completion.py
moonyuet: [d2b6b0e](https://github.com/ynput/ayon-core/pull/1947/commits/d2b6b0e4482dd36559408f2831e7058a38925de4)

client/ayon_core/ui/components/comment_completion.py
moonyuet: > Are we sure we really want to call this `replace` call for EVERY user? because that'd do the regex match PER user?

we can test and do the comparison: But I am pretty sure that this would be more accurate for using the regex match. [d2b6b0e](https://github.com/ynput/ayon-core/pull/1947/commits/d2b6b0e4482dd36559408f2831e7058a38925de4)

iLLiCiTiT: I did some tests and it takes `0.040305376052856445` for 1000 users. All of them were in the text, so it did 1000 search and replacements.

If each user was mentioned 5x it took `0.08743906021118164`.

BigRoy: 1. Why the sorting?
2. We should really do one lookup over the full markdown text. We don't want a potentially 1000s of characters text to be `if x in y` checked and then `str.replace()` with 100s if not 1000s of users. It's just a waste.

I'd actually do this if you want exact user full name matches (probably still with Kuba's sorting):
```python
match = re.search("|".join(f"@{re.escape(user.full_name)}" for user in users))
```

A single regex - matching all user name tags exactly.

And that could be `re.sub` directly if you used a replacement function that'd just return the simple response from full name -> `[{user.full_name}](user:{user.name})`

iLLiCiTiT: > Why the sorting?

The sorting is because `@User A` and `@User America` are 2 different users and `@User America` should be replaced first.

> We should really do one lookup over the full markdown text. We don't want a potentially 1000s of characters text to be if x in y checked and then str.replace() with 100s if not 1000s of users. It's just a waste.

Regex with a lots of possible values is actually slower, and it does not handle the example above as it ends match logic when one of them is matching. So if the text would contain `@Jane P` and `@Jane Polite` only `@Jane P` would be marked.

iLLiCiTiT: I tried it and just regex find did take `0.03`, with replacement it was `0.19` (maybe I did inefficient replacement?).

BigRoy: So it works fine with the regex too, with this test run:
```python
import re
import time


users = ["Roy Nieterau"] * 1000
users.append("Roy Nieterauuu")
users.append("Roy Nieterauuuuu")
md = "a" * 2000 + "@Roy Nieterauuu"


# Sort users by longest full name
tmp_md = md
s = time.time()
for user in sorted(users, reverse=True):
    mention = f"@{user}"
    if mention not in tmp_md:
        continue
    tmp_md  = tmp_md.replace(mention, f"[{user}](user:{user})")
e = time.time()
result_a = tmp_md
print(e-s)
print(result_a)

s = time.time()

def _sub(match):
    user = match.group(0)[1:]
    return f"[{user}](user:{user})"

regex = "|".join(f"@{re.escape(user)}" for user in sorted(users, reverse=True))
result_b = re.sub(regex, _sub, md)
e = time.time()
print(e-s)
print(result_b)
print(result_a == result_b)
```

But, the regex compilation in this case is rather lengthy. Assuming the users list is non-static would mean this is slower. Other than that the performance difference is neglible.

I wonder if we should at least do one `if @ in md:` check to avoid all these things to happen if there's no tag to begin with?

Other than that, seems fine with your suggestion 👍

iLLiCiTiT (approved):

client/ayon_core/ui/components/comment_completion.py
iLLiCiTiT: > I wonder if we should at least do one if @ in md: check to avoid all these things to happen if there's no tag to begin with?

Very good suggestion actually. Before anything, check if there is `@` in the string,

moonyuet: [d523778](https://github.com/ynput/ayon-core/pull/1947/commits/d5237785afef292684a92a98e201552155fbb22a)

BigRoy (changes requested):
@moonyuet can we still fix the highlight font color in the popup?

Image

Compared with Frontend:

Image

Other than that, all looks good.

BigRoy (approved):
🚀

PR #1582, opened 2025-12-05
Enhancement: thumbnail presets

kalisp:
## Changelog Description
This PR adds possibility to control `ExtractThumbnail` by profiles.

## Additional info
This was requested as a feature allowing different colorspaces for different tasks.

Contains conversion from older Settings format.

## Testing notes:
1. setup some overrides in `ayon+settings://core/publish/ExtractThumbnail` before deploying this one
2. use `Copy Setting from...` on `core` after deployment
3. Publish in Nuke/AE/Houdini/Fusion

PR #1953, opened 2026-07-14
Fix Extract Thumbnail settings conversions appending more profiles on copy settings

BigRoy:
## Changelog Description

Fix Extract Thumbnail settings conversions appending more profiles on copy settings

## Additional info

Bug was introduced with #1582

## Testing notes:
1. Duplicate profiles should be removed if copied from core <=1.9.8
2. Settings with overrides to Extract Thumbnail should correctly convert from core <1.7.0, without creating duplicates when copying settings again and again. Settings newer than 1.7.0 copied from should also not cause duplicates.

### DIFF
--- a/server/settings/conversion.py
+++ b/server/settings/conversion.py
@@ -1,8 +1,10 @@
 import re
 import copy
+import json
 from typing import Any
 
 from semver import VersionInfo
+from server.settings.conversion import parse_version
 
 from .publish_plugins import DEFAULT_PUBLISH_VALUES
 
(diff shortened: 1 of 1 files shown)


server/settings/conversion.py
```diff
@@ -376,4 +417,5 @@ def convert_settings_overrides(
     _convert_product_base_types_1_8_0(overrides)
     _convert_unify_profile_keys_1_8_0(overrides)
     _convert_burnin_offset_1_8_6(overrides, version)
+    _fix_duplicates_extract_thumbnail(source_version, overrides)
```
iLLiCiTiT: You can use `version` which is `VersionInfo`.
```suggestion
    _fix_duplicates_extract_thumbnail(version, overrides)
```

iLLiCiTiT: Or the other way around and use only `parse_version`.

iLLiCiTiT: Wait, where is `parse_version` comming from?

BigRoy: Haha yes - I pushed to quickly, sorry. Really have to leave and just wanted to make a draft and then noticed it pushed the half-broken commit. How about now?

server/settings/conversion.py
```diff
@@ -362,6 +374,35 @@ def _convert_burnin_offset_1_8_6(
...
+
+    processed_profiles: set[str] = set()
+    non_duplicate_profiles = []
+    for profile in list(profiles):
+        profile_hash = json.dumps(profile, sort_keys=True)
+        if profile_hash in processed_profiles:
+            # Skip duplicate
+            continue
+        processed_profiles.add(profile_hash)
+        non_duplicate_profiles.append(profile)
```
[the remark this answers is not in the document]
BigRoy: Settings really are JSON encodable by design - so this should never be the case.

moonyuet (approved):
Tested successfully in copying overrides in the ayon server

iLLiCiTiT (approved):

PR #1955, opened 2026-07-16
Chore: Remove avalon mentions

iLLiCiTiT:
## Changelog Description
Remove avalon mentions and backwards compatibilities.

## Additional info
I do believe we don't need to support `AVALON_` env variables anymore. Some of the modified methods are deprecated anyways.

NOTE: There are still places that do use avalon (more scary to remove)...

## Testing notes:
1. Validate code changes.

### DIFF
--- a/client/ayon_core/pipeline/actions/launcher.py
+++ b/client/ayon_core/pipeline/actions/launcher.py
@@ -66,11 +66,11 @@ class LauncherActionSelection:
             ),
             category=DeprecationWarning
         )
-        if key in {"AYON_PROJECT_NAME", "AVALON_PROJECT"}:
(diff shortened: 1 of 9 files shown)


BigRoy (approved):
LGTM - and works for me.

> NOTE: There are still places that do use avalon (more scary to remove)...

Should we make a follow-up draft PR to discuss those?

iLLiCiTiT:
> Should we make a follow-up draft PR to discuss those?

Those are potentially scene breaking, like `AVALON_INSTANCE_ID`, or container schema naming (where is also openpype).

I kept some of the mentions in explanation docstrings. Those are there from OpenPype to explain difference from avalon, probably should be replaced by proper explanation instead of comparison....

PR #1957, opened 2026-07-16
Create: Add type-hints to create plugins

iLLiCiTiT:
## Changelog Description
Add type hints into create plugin in create api.

## Testing notes:
1. Validate code changes, hopefully no logic was changed.

### DIFF
--- a/client/ayon_core/pipeline/create/creator_plugins.py
+++ b/client/ayon_core/pipeline/create/creator_plugins.py
@@ -907,7 +907,7 @@ class Creator(BaseCreator):
 
     def get_default_variant(
         self, only_explicit: bool = False
-    ) -> str:
+    ) -> str | None:
         """Default variant value that will be used to prefill variant input.
 
         This is for user input and value may not be content of result from
@@ -922,7 +922,7 @@ class Creator(BaseCreator):
                 variant from '_default_variant' will be returned.
 
         Returns:
-            str: Variant value.
+            str | None: Variant value.
 
         """
         if only_explicit or self._default_variant:
--- a/client/ayon_core/pipeline/create/creator_plugins.py
+++ b/client/ayon_core/pipeline/create/creator_plugins.py
@@ -29,7 +29,7 @@ from .structures import CreatedInstance, ProductTypeItem
 if TYPE_CHECKING:
     import logging
 
-    from ayon_core.host import HostBase
+    from ayon_core.host import IPublishHost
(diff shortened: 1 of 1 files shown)


client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -62,27 +65,28 @@ class ProductConvertorPlugin(ABC):
...
             logging.Logger: Logger with name of the plugin.
+
         """
         if self._log is None:
             self._log = Logger.get_logger(self.__class__.__name__)
         return self._log
 
     @property
-    def host(self):
+    def host(self) -> HostBase:
```
antirotor: MyPy is bitching about this one - declared here as `HostBase` but `host` *CreateContext* is  actually `IPublishHost`. Shouldn't it be the latter (or both?)

```suggestion
    def host(self) -> HostBase | IPublishHost:
```

client/ayon_core/pipeline/create/creator_plugins.py
```diff
@@ -844,10 +901,13 @@ def get_default_variants(self):
...
         Returns:
             list[str]: Whisper variants for user input.
+
         """
         return copy.deepcopy(self.default_variants)
 
-    def get_default_variant(self, only_explicit=False):
+    def get_default_variant(
+        self, only_explicit: bool = False
+    ) -> str:
```
antirotor:
```suggestion
    ) -> str | None:
```
since `self._default_variant` is annotated as `str | None` too?

antirotor (changes requested):
Few small notes, otherwise it looks good. MyPy is bitching about incompatible method signatures but that will align once we type hint super classes too.

Tested it with Maya and found no issues (I have just 2027 available now, might be worth running it in something older)

BigRoy (approved):
LGTM - pending the notes by Ondrej. I just did code review by glancing over it though since Ondrej already went level mypy 🗡️

client/ayon_core/pipeline/create/creator_plugins.py
iLLiCiTiT: I think we can remove the `only_explicit` as it is not used anymore, but probably in different PR.

client/ayon_core/pipeline/create/creator_plugins.py
iLLiCiTiT: Changed to just `IPublishHost`.

PR #1958, opened 2026-07-16
Create: Product base type is required on create plugins

iLLiCiTiT:
## Changelog Description
Mark `product_base_type` as abstractmethod for create plugins and remove `product_type` from create plugins. Also define compatible versions of other addons to the versions that should respect the product base type.

## Additional info
This was planned change in create plugins marking an end of era when both `product_type` and `product_base_type` on create plugins have to be defined and both can be optionally set on instance. From now on all instances should have filled `"productBaseType"` and `"productType"`. Create plugins don't have to define `product_type` (they have to mark compatible core with release when this PR is made).

Property `product_type` now does warn about deprecation in case it is used, but because all existing create plugins do define it it won't be warned until all create plugins remove the definition.

## Testing notes:
1. All affects hosts and integrations do work.


BigRoy:
>  Create plugins don't have to define product_type (they have to mark compatible core with release when this PR is made).

In which way? Older addons wouldn't be able to mark incompatibility of course. But what core version would we start adding as minimal requirement moving forward then?

iLLiCiTiT:
To the version where this is merged 🙂 . Can be filled only when the release of core is made. If we make it to next minor release then `1.10.0`.

kalisp (approved):
LGTM

Tested in Adobe products

iLLiCiTiT:
NOTE: Will be merged after next release of core.

BigRoy:
> To the version where this is merged 🙂 . Can be filled only when the release of core is made. If we make it to next minor release then `1.10.0`.

Just to note, there's quite a few DCC integrations that still touch `self.product_type` on the Creator.
e.g. a quick search reveals:
```
ayon_blender/plugins/create/create_workfile.py
ayon_core/tools/publisher/models/create.py
ayon_comfyui/plugins/create/create_video.py
ayon_comfyui/plugins/create/create_model.py
ayon_comfyui/plugins/create/create_image.py
ayon_cinema4d/plugins/create/create_workfile.py
ayon_cinema4d/plugins/create/create_render.py
ayon_flame/plugins/create/create_workfile.py
ayon_fusion/plugins/create/create_workfile.py
ayon_harmony/plugins/create/create_render_layers.py
ayon_houdini/plugins/create/create_workfile.py
ayon_houdini/plugins/create/create_usd_componentbuilder.py
ayon_hiero/plugins/create/create_workfile.py
ayon_hiero/plugins/create/create_editorial_pkg.py
ayon_katana/plugins/create/create_workfile.py
ayon_loki/plugins/create/create_workfile.py
ayon_openrv/plugins/create/create_workfile.py
ayon_mari/plugins/create/create_workfile.py
ayon_marvelousdesigner/plugins/create/create_workfile.py
ayon_maya/plugins/create/create_render.py
ayon_resolve/plugins/create/create_workfile.py
ayon_substancepainter/plugins/create/create_workfile.py
ayon_tvpaint/plugins/create/create_render.py
ayon_traypublisher/plugins/create/create_movie_batch.py
ayon_traypublisher/plugins/create/create_csv_ingest.py
```
Does this mean these should generally still be in need of refactor? and should we do so now in preparation to avoid Creators just getting spammy on these warnings for the users?

iLLiCiTiT:
> Just to note, there's quite a few DCC integrations that still touch self.product_type on the Creator.
e.g. a quick search reveals:

They should be changed, yes, but should not be needed for this PR to be merged.

> Does this mean these should generally still be in need of refactor? and should we do so now in preparation to avoid Creators just getting spammy on these warnings for the users?

Yes, not now. We first have to remove requirement to define `product_type` (this PR). Then over time we can remove that from the integrations, they'd have to bump required core version to be able to remove `product_type` definition from create plugins.

BTW Workfile create plugins usually do have just `product_type` instead of product type items because they are autocreators, so they will probably still use `self.product_type`.

BigRoy:
Although this works - let's hold off a bit (maybe few weeks) before merging. (Sorry!) This pushes so many required compatibilities that it can be rather annoying if some are so close to their release of the other addons. I'd rather give some more breathing room - and since this is not urgent to merge, I think we can wait for a bit.

PR #1960, opened 2026-07-17
Delivery: support per-representation template overrides

### LINKED ISSUE #1961: YN-0942: Deliver Version tool representation template override rules

Client needs rule based workflow to override template per representation. 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0942]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=05101f3b4072477cad4d865793e7410c)

jakubjezek001:
## Changelog Description
Adds configurable overrides for delivery templates on a per-representation basis. Studios can now define override presets in project settings that swap the directory/file template for specific representation names, instead of being locked to a single delivery template. Also adds support for using `{publishedFilename}` without its extension in templates.

## Additional info
- New settings models (`DeliveryToolModel`, `DeliveryTemplateOverrideModel`, `DeliveryRepresentationRuleModel`) expose override presets under Delivery tool settings, matched by representation name.
- `check_destination_path` and `deliver_single_file` accept an optional `explicit_template_obj` to bypass the named anatomy template lookup.
- Delivery UI builds the override template object per representation, merging `{directory}`/`{file}` tokens from the original template when only one side is overridden.
- Strips stray frame-number segments from the destination folder when `{publishedFilename}` (without extension) is used in a directory token.

image

image

image

## Testing notes:
1. Configure a delivery override preset in project settings (AYON Core > Delivery) with a representation rule overriding directory/file template.
2. Run the Delivery tool against a representation matching the rule and confirm files land at the overridden path.
3. Verify representations without a matching rule still use the default delivery template.

## Dependency
Close https://github.com/ynput/ayon-core/issues/1961

## Related support tickets
[YN-0942](https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=05101f3b4072477cad4d865793e7410c)

### DIFF
--- a/client/ayon_core/tools/delivery/delivery.py
+++ b/client/ayon_core/tools/delivery/delivery.py
@@ -393,7 +393,6 @@ class DeliveryOptionsDialog(QtWidgets.QDialog):
                 m_directory_tmpl = m_directory_tmpl.replace(
                     "{file}", original_file
                 )
-        print(m_directory_tmpl)
 
         m_file_tmpl = (
             matching_rule["template_file"] or original_file
--- a/client/ayon_core/tools/delivery/delivery.py
+++ b/client/ayon_core/tools/delivery/delivery.py
@@ -1,26 +1,26 @@
+import logging
 import os
 import platform
-import logging
 from collections import defaultdict
 
 import ayon_api
-from qtpy import QtWidgets, QtCore, QtGui
+from qtpy import QtCore, QtGui, QtWidgets
 
 from ayon_core import resources, style
 from ayon_core.lib import (
-    format_file_size,
     collect_frames,
+    format_file_size,
     get_datetime_data,
 )
 from ayon_core.pipeline import Anatomy
-
-from ayon_core.pipeline.load import get_representation_path_with_anatomy
 from ayon_core.pipeline.delivery import (
-    get_format_dict,
     check_destination_path,
     deliver_single_file,
+    get_format_dict,
     get_representations_delivery_template_data,
 )
+from ayon_core.pipeline.load import get_representation_path_with_anatomy
+from ayon_core.settings import get_project_settings
 
 
 class DeliveryOptionsDialog(QtWidgets.QDialog):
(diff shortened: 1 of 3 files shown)


client/ayon_core/tools/delivery/delivery.py
```diff
@@ -283,11 +301,13 @@ def deliver(self):
 
             for src_path, frame in sources_and_frames.items():
                 # Support {publishedFilename} token
-                template_data["publishedFilename"] = os.path.basename(
+                publish_basename = os.path.basename(
```
iLLiCiTiT, line 304: Why this? This is supposed to be filename, not basename. Filename has extension. It doesn't make sense to change extension of a file during delivery.

iLLiCiTiT:
This PR adds multiple things at the same time. I would split it.

First of all if someone needs to use the filename as a directory -> without frame and extensions, then don't change `publishedFilename`. We should add new keys that do what it needs to do instead of changing what is now explicitly defined and well defined. It might be useless, but don't store extensionless and frameless filename into `publishedFilename`, it is not `publishedFilename` if it is modified.

I really don't like that we do add different delivery templates to the settings. Rather add an enum of delivery templates from project anatomy, with that you don't need the `explicit_template_obj` argument, which is kinda weird, and the whole logic to resolve the templates from settings is super confusing.

BigRoy (commented):
I somewhat agree with Kuba here. This is just overcomplicating things and I really wonder whether we're not stretching it too far. And with the amount of complication this adds, I feel like it doesn't solve enough. The pure single representation name switching I think doesn't suffice.

Would people maybe want to differentiate certain representations within a single delivery template, sure. Would it be up to representation name only - I don't think so. Also, what if you have two representation names you now want to apply this delivery template to, etc. (or what if `camera_exr`, `exr` and `raw_exr` or whatever special representation names you may have from ingests. The design doesn't scale well. It'd almost need to be profile filtering based, etc. (but we all know how complex that can get to set up in settings)

It's really getting to the point where it just starts making more sense for the delivery tool to be able to specify custom delivery rules/scripts, like a "use this script to define the template" where you can essentially put in your own custom Python script to evaluate it to something custom, etc. because we're seeing the granularity that studios want their deliveries to behave like for their needs. Especially with AI/LLMs being able to write some custom logic would be trivial? We could have a custom delivery format where you get access to the "context" and e.g. just return the template you want to apply. Complexity for the admin? yes, but allows you to get very granular and move the responsibility to the one maintaining the delivery system not having to dive deeper and deeper into growingly scattered and complex settings?

In short, I wonder - are we solving this the right way?

BigRoy:
Perhaps the "batch delivery" 'script' approach we discussed could apply here too @jakubjezek001 ?

PR #1963, opened 2026-07-20
Validate Workfile: Hosts filtering

iLLiCiTiT:
## Changelog Description
Remove hosts filters from `ValidateCurrentSaveFile` and run the logic only for hosts that do inherit from `IWorkfileHost`.

## Additional info
This is a proposal of the change, I'm not sure if this logic is 100% right as there were not all hosts listed in the plugin's `hosts` attribute.

Because it is context plugin, and the validation is not related to any instance we'd have to enable the plugin from host using predefined key on `context.data`.

Also replaced deprecated function `version_up_current_workfile` with `save_next_version`.

### Notes
- The plugin does not validate if the workfile is saved, it does validate if is set.
- There are hosts that do not inherit from `IWorkfileHost` but do set the `"currentFile"`, those would have to validate the key on their own -> which is probably ok.

## Testing notes:
1. Validate code changes.
2. This probably should be test in all hosts that can have a workfile.

### DIFF
--- a/client/ayon_core/plugins/publish/validate_file_saved.py
+++ b/client/ayon_core/plugins/publish/validate_file_saved.py
@@ -2,6 +2,8 @@ import inspect
 
 import pyblish.api
 
+from ayon_core.host import IWorkfileHost
+from ayon_core.pipeline import registered_host
 from ayon_core.pipeline.publish import PublishValidationError
 from ayon_core.tools.utils.host_tools import show_workfiles
 from ayon_core.pipeline.context_tools import version_up_current_workfile
@@ -42,6 +44,9 @@ class ValidateCurrentSaveFile(pyblish.api.ContextPlugin):
     actions = [SaveByVersionUpAction, ShowWorkfilesAction]
 
     def process(self, context):
+        host = registered_host()
+        if not isinstance(host, IWorkfileHost):
+            return
 
         current_file = context.data["currentFile"]
         if not current_file:
--- a/client/ayon_core/plugins/publish/validate_file_saved.py
+++ b/client/ayon_core/plugins/publish/validate_file_saved.py
@@ -5,8 +5,8 @@ import pyblish.api
 from ayon_core.host import IWorkfileHost
 from ayon_core.pipeline import registered_host
 from ayon_core.pipeline.publish import PublishValidationError
+from ayon_core.pipeline.workfile import save_next_version
 from ayon_core.tools.utils.host_tools import show_workfiles
-from ayon_core.pipeline.context_tools import version_up_current_workfile
 
 
 class SaveByVersionUpAction(pyblish.api.Action):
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/publish/validate_file_saved.py
```diff
@@ -36,12 +38,12 @@ class ValidateCurrentSaveFile(pyblish.api.ContextPlugin):
...
     label = "Validate File Saved"
     order = pyblish.api.ValidatorOrder - 0.1
-    hosts = ["fusion", "houdini", "max", "maya", "nuke", "substancepainter",
-             "cinema4d", "silhouette", "gaffer", "blender", "loki",
-             "photoshop"]
     actions = [SaveByVersionUpAction, ShowWorkfilesAction]
 
     def process(self, context):
+        host = registered_host()
+        if not isinstance(host, IWorkfileHost):
```
BigRoy, line 45: Should we debug log when skipping?
```suggestion
        if not isinstance(host, IWorkfileHost):
            self.log.debug(
                "Skipping file save validation because host does "
                "not implement workfiles."
            )
```

BigRoy (approved):
Much better. I hope there aren't any hosts out there that may get hit by side effects from this, 

I was rather surprised to not see `hiero`, `harmony`, `flame`, `aftereffects`, `resolve`, `premiere`, `marvelousdesigner`, `equalizer` and `zbrush` in that original hosts list. So there's definitely a few more now where this _will_ trigger but I think it is the correct approach to get this validation everywhere.

iLLiCiTiT:
@LiborBatek could go trough some of the hosts and trigger publish validations?
- [ ] `hiero`
- [x] `harmony`
- [x] `aftereffects`
- [x] `resolve`
- [x] `premiere`
- [x] `marvelousdesigner`
- [ ] `equalizer`
- [x] `zbrush`

Don't have to be full publising, just validation, technically just open the host, save workfile and hit validate in publisher.

If you don't have some of them write them down. I guess we'll need @rdelillo for flame.

kalisp:
I tested Adobe products and Harmony. Opened DCCs, resaved workfiles and ran validations. No issue noticed.

kalisp (approved):

rdelillo:
@iLLiCiTiT  just checking, do we need any testing on Flame ?
I'm not sure. As far as I can see it is not an `IWorkfileHost` : https://github.com/ynput/ayon-flame/blob/develop/client/ayon_flame/api/pipeline.py#L39

iLLiCiTiT:
> As far as I can see it is not an IWorkfileHost

Ah, ok, then no 🙂

BigRoy:
Works in Resolve 👍

iLLiCiTiT:
I checked code in the rest, I think all hosts that do inherit from `IWorkfileHost` do fill `"currentFile"` and this validation should just work.

moonyuet:
@BigRoy asked me to test if there is any issue on Zbrush and marvelous designer and I can double confirm that it is working with thest two hosts.
[publish-report-260722-19-36.json](https://github.com/user-attachments/files/30266707/publish-report-260722-19-36.json)
[publish-report-260722-19-39.json](https://github.com/user-attachments/files/30266739/publish-report-260722-19-39.json)

PR #1965, opened 2026-07-21
[YN-0943] Support slateFrame key for backward compatibility in ExtractReviewSlate

### LINKED ISSUE #333: YN-0943: slate generator processing trigger

client requires to be able to trigger slate processing from nuke write node without need for nuke node in the scene 

### Additional context:

_This issue was automatically created from AYON ticket [YN-0943]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=42d0f38afd1347caa40c374f1d154d4f)

jakubjezek001:
## Changelog Description
The `ExtractReviewSlate` plugin now supports both `slateFrames` and `slateFrame` keys when extracting data from upstream, enabling backward compatibility with the Slates Generator plugin. When no slate frame data is provided at all (neither key exists or contains empty values), the plugin returns early instead of crashing. This prevents runtime crashes when downstream tasks receive no slate information from upstream plugins.

## Additional info
The change modifies `client/ayon_core/plugins/publish/extract_review_slate.py` to:
1. Support reading from both `slateFrames` and `slateFrame` dictionary keys
2. Perform an early return when neither key has data, avoiding crashes on the downstream side

## Testing notes:
1. Ensure ExtractReviewSlate runs without error when upstream provides slate frames in either format
2. Verify no crash occurs when upstream provides no slate frame data at all

## Dependency
Close [#333](https://github.com/ynput/ayon-nuke/issues/333)

## Related support tickets
[YN-0943](https://support.ayon.app/projects/Active_Clients/overview/projects/Active_Clients/overview/projects/Active_Clients?project=Active_Clients&type=task&id=f2d9f886affb455a9dbaa9ecc0221cca)

### DIFF
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -514,7 +514,7 @@ def prepare_representations(
         if ext in skip_integration_repre_list:
             rep["tags"].append("delete")
 
-        if ext in slate_representation_ext:
+        if ext == slate_representation_ext:
             rep["tags"].append("slate-frame")
 
         if skeleton_data.get("multipartExr", False):
--- a/client/ayon_core/pipeline/farm/pyblish_functions.py
+++ b/client/ayon_core/pipeline/farm/pyblish_functions.py
@@ -360,6 +360,15 @@ def create_skeleton_instance(
         if item in instance.data.get("families", []):
             instance_skeleton_data["families"] += [item]
 
+    slate_representation_ext = instance.data.get(
+        "slateRepresentationExt")
+    if (
+        "slate" in families_transfer
+        and slate_representation_ext
+    ):
+        instance_skeleton_data["slateRepresentationExt"] = \
+            slate_representation_ext
+
     # transfer specific properties from original instance based on
     # mapping dictionary `instance_transfer`
     for key, values in instance_transfer.items():
@@ -428,6 +437,7 @@ def prepare_representations(
 
     """
     representations = []
+    slate_representation_ext = skeleton_data.get("slateRepresentationExt", [])
     host_name = os.environ.get("AYON_HOST_NAME", "")
     collections, remainders = clique.assemble(exp_files)
 
@@ -503,6 +513,9 @@ def prepare_representations(
         # poor man exclusion
         if ext in skip_integration_repre_list:
             rep["tags"].append("delete")
+
+        if ext in slate_representation_ext:
+            rep["tags"].append("slate-frame")
 
         if skeleton_data.get("multipartExr", False):
             rep["tags"].append("multipartExr")
--- a/client/ayon_core/plugins/publish/extract_review_slate.py
+++ b/client/ayon_core/plugins/publish/extract_review_slate.py
@@ -40,8 +40,8 @@ class ExtractReviewSlate(publish.Extractor):
 
         # get slates frame from upstream
         slates_data = inst_data.get("slateFrames")
-        slate_frame = inst_data.get("slateFrame")
-        if not slates_data and slate_frame is not None:
+        slate_frame = inst_data.get("slateFrame", 0)
+        if not slates_data and not slate_frame:
             # make it backward compatible and open for slates generator
             # premium plugin
             slates_data = {
--- a/client/ayon_core/plugins/publish/extract_review_slate.py
+++ b/client/ayon_core/plugins/publish/extract_review_slate.py
@@ -41,11 +41,11 @@ class ExtractReviewSlate(publish.Extractor):
         # get slates frame from upstream
         slates_data = inst_data.get("slateFrames")
         slate_frame = inst_data.get("slateFrame")
-        if not slates_data and slate_frame:
+        if not slates_data and slate_frame is not None:
             # make it backward compatible and open for slates generator
             # premium plugin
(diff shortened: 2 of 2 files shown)


jakubjezek001:
Main purpose for this PR @iLLiCiTiT is to provide required functionality for Current slate generator. This plugin was manly developed fro Nuke slate workflow but since we need to also enable Nuke workflow with Nuke slate generated trigger, we are not able to provide either of those keys and this plugin needs to be skipped. We are only sending instances with **slate** family and representations with **slate-frame** tags from Nuke - rest of processing and frame integration is done within the Slate generator plugin.

client/ayon_core/plugins/publish/extract_review_slate.py
```diff
@@ -40,12 +40,15 @@ def process(self, instance):
 
         # get slates frame from upstream
         slates_data = inst_data.get("slateFrames")
-        if not slates_data:
+        slate_frame = inst_data.get("slateFrame")
+        if not slates_data and slate_frame:
```
iLLiCiTiT:
```suggestion
        if not slates_data and slate_frame is not None:
```

client/ayon_core/plugins/publish/extract_review_slate.py
```diff
@@ -40,12 +40,15 @@ def process(self, instance):
...
             # make it backward compatible and open for slates generator
             # premium plugin
             slates_data = {
                 "*": inst_data["slateFrame"]
```
BigRoy:
```suggestion
                "*": slate_frame
```

client/ayon_core/plugins/publish/extract_review_slate.py
BigRoy: ☝️ Need to potentially consider `slateFrame` being `0`.

BigRoy (dismissed):
Applied the code suggestions - @jakubjezek001 confirm you're happy with it and then good to go I'd say.

Code looks ok to me.

client/ayon_core/plugins/publish/extract_review_slate.py
```diff
@@ -40,12 +40,15 @@ def process(self, instance):
 
         # get slates frame from upstream
         slates_data = inst_data.get("slateFrames")
-        if not slates_data:
+        slate_frame = inst_data.get("slateFrame")
+        if not slates_data and slate_frame is not None:
```
jakubjezek001:
```suggestion
        slate_frame = inst_data.get("slateFrame", 0)
        if not slates_data and not slate_frame:
```

jakubjezek001: i would perhaps add the `0` as you suggested @BigRoy

BigRoy: Does `slateFrame` mean the amount of frames? or at which frame the slate exists?

jakubjezek001: it is numeric value reflecting amount of slate frames preceding. This was old "slateFrame" was obsolete key value usually set to `1` if slate was needed.

BigRoy: In that case, this suggestion is fine with me.

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -504,6 +514,9 @@ def prepare_representations(
         if ext in skip_integration_repre_list:
             rep["tags"].append("delete")
 
+        if ext in slate_representation_ext:
+            rep["tags"].append("slate-frame")
```
BigRoy, line 518: Ok, now thing are starting to get confusing for me.

Is this still a backwards compatibility thing?
And why do we need this `slateRepresentationExt` handling?

jakubjezek001: yes it should be backward compatible. This needs to be added for Render farm also process the slate workflow. There is no other way to do this. We have discussed it with @iLLiCiTiT. Related changes in Nuke https://github.com/ynput/ayon-nuke/pull/335/changes#diff-27fe00275e66e51a967e0a35beb54de61049fd37ffbea8e1dea6917c10c4b276R409-R415

client/ayon_core/pipeline/farm/pyblish_functions.py
```diff
@@ -504,6 +514,9 @@ def prepare_representations(
         if ext in skip_integration_repre_list:
             rep["tags"].append("delete")
 
+        if ext in slate_representation_ext:
```
BigRoy:
```suggestion
        if ext == slate_representation_ext:
```

client/ayon_core/pipeline/farm/pyblish_functions.py
BigRoy: Had a chat with @iLLiCiTiT about this as well. It's far from the code we want to have, but it seems at this point in time we can't do it better a different way so it's the only solution.

So I guess, that's it. Do please change the representation ext check to be an exact check instead of in though.

Then if @iLLiCiTiT is happy, I'm ok with it for now as well.

jakubjezek001:
@iLLiCiTiT @BigRoy if anybody can approve it please?

BigRoy (approved):

PR #1967, opened 2026-07-21
Editorial plugins: Plugins without hosts filters

iLLiCiTiT:
## Changelog Description
Editorial plugins have 2 variants now. One plugin variant has old filters and the order has just families filter with very targeted families filtering to just one family.

## Additional info
The goal is that hosts that do want to actually use any logic from the plugins should add the family to instance. That way the host is in charge if logic from core will be triggered instead of core knowing it. That way any change in existing integrations or new editorial host should not need to change core.

This is first step, second step would be to add the families in the host integrations and last step will be to remove the old plugins, that will require to bump compatible versions in package.py.

### Affected hosts
- [x] Flame https://github.com/ynput/ayon-flame/pull/142
- [x] Hiero https://github.com/ynput/ayon-hiero/pull/159
- [x] Resolve https://github.com/ynput/ayon-resolve/pull/109
- [x] Traypublisher https://github.com/ynput/ayon-traypublisher/pull/181

---

### CollectOtioReview
- renamed to <b>CollectOTIOReviewTrack</b>
- added one plugin with family `"otio.review.track"`
- kept old for backwards compatibility
- <b>'otioClip' is NOT automatically used if there is "review" in families !!!</b>
- <b>'otioReviewClips' can be pre-filled -> host integration can fill it to use `otioClip` !!!</b>

### CollectOtioRanges
- added one plugin with family `"otio.clip.ranges"`
- kept old for backwards compatibility

### CollectOtioSubsetResources
- renamed to <b>CollectOTIOProductResources</b>
- added one plugin with family `"otio.clip.resources"`
- kept old for backwards compatibility

### ExtractOTIOFile
- added one plugin with family `"otio.timeline.workfile"`
- kept old for backwards compatibility

### ExtractOTIOReview and ExtractOTIOTrimmingVideo
- both just removed `hosts` and changed `families` to `"otio.trim.review"` `"otio.trim.video"`
- the families are added by `CollectOTIOProductResources`

## Testing notes:
Nothing should change with this PR.

Someone should look at the new families and make sure it makes sense.

The way how host integration tells core what to use for track review changed too.

We should try each host in this combinations:
- core develop & host develop - to validate everything works
- core PR & host develop - a debug log in each old plugin will be visible at the start
- core develop & host PR - nothing that can be visualy validated
- core PR & host PR - a debug log should not be visible


client/ayon_core/plugins/publish/collect_otio_review.py
rdelillo, line 193: This is here to handle `ReviewableSource: [Clip's Media]`

Do we want to clean up the `review` family here ?
We add the support an additional `otio.review.clip` family here to make it more specific.

rdelillo (commented):
I've tested it all, and all combinaison in Flame. It works great including trimming and review. 

Just added some additional debug logs.
Also see my comment regarding the remaning `review` family usage in `collect_otio_review`.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -208,3 +213,16 @@ def _collect_retimed_ranges(self, instance, otio_clip):
...
+
+
+class CollectOTIORangesOld(CollectOTIORanges):
+    label = "Collect OTIO Ranges (old)"
+    order = CollectOTIORanges.order + 0.000001
+    families = ["shot", "clip"]
+
+    def process(self, instance):
+        if instance.data.pop(FW_KEY, False) is True:
+#            self.log.debug("Skipping, CollectOTIORanges has run.")
```
BigRoy: Should be cleaned?

client/ayon_core/plugins/publish/collect_otio_review.py
iLLiCiTiT: I think it will happen automatically because of condition on line 106? Or you mean to avoid doubled `"review"` family?

iLLiCiTiT: > Do we want to clean up the review family here ?
We add the support an additional otio.review.clip family here to make it more specific.

I understand now what you mean, sorry, I'm slow. I can't reliably say yes or no, need more context.

This logic is confusing a little bit for me. It does create review of whole clip if review track is not specified, I guess?
The question for me is what is reason for the current logic.

I think the best solution is to avoid any families check. That can be done if we can rely only on `"reviewTrack"`. But the question is if we can set `"reviewTrack"` to what is in `"otioClip"` from the hosts?

iLLiCiTiT:
Logic of <b>CollectOTIOReviewTrack</b> was changed to be different from original <b>CollectOtioReview</b>. It does not auto-use `otioClip` for review if review track is not set and there is `"review"` family in families, instead the host can pre-fill `"otioReviewClips"` to use it.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -57,14 +59,17 @@ class CollectOtioRanges(pyblish.api.InstancePlugin):
 
     label = "Collect OTIO Ranges"
     order = pyblish.api.CollectorOrder - 0.08
-    families = ["shot", "clip"]
+    families = ["otio.clip.ranges"]
```
BigRoy, line 60: This particular plug-in could just keep support for the old `families` for the time being? (Although you wouldn't be getting the old log then?)

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
```diff
@@ -10,6 +10,8 @@
 
 import pyblish.api
 
+FW_KEY = "__OTIO_CLIP_RANGES_PROCESSED__"
```
BigRoy: The `FW_KEY` variable name doesn't seem to directly ring a bell. It's likely "Forward compatibility"? But admittedly, without the context of this PR it makes little sense as an abbreviation?

Not sure what other short name would make sense though.

BigRoy (commented):
Change to more explicit families name I think is a good improvement.

I wish we could clean up the backwards compatibility some day though. 🗡️ Because that'd make this much prettier.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
iLLiCiTiT: > (Although you wouldn't be getting the old log then?)

Yep. Also it would not be really clear that "Hey stop using those, use this instead". I think the families changes should be as clear as possible and having one plugin that has to be removed is a clear message what is the plan, supporting both new and old in one plugin doesn't give any message. Also a little bit harder to validate if it actually worked.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
iLLiCiTiT: Opened to suggestions, I can add a comment... The variable is used only in this file, does not have to be super api friendly.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
iLLiCiTiT: I would actually keep it. It might help ~~resolve~~ find out possible issues with publish report.

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
BigRoy: It's only used in these files - yes.

What if we put the variable on the class itself as `HAS_RUN_KEY` so that the if statement on the other plug-in also reads as `if context.data.get(OtherPlugin.HAS_RUN_KEY, False) is True:`?

Hmm, feel odd too?

iLLiCiTiT: It is instance specific key. If we did forget about a place that creates instances that should run through this plugin we should find out from the logs. But it can be `HAS_RUN_KEY` on the plugin.

iLLiCiTiT: We could maybe also store it to context (or on the class). So if the new did run but the old is triggered on an instance it would log a warning instead of debug?

BigRoy: It being instance specific key is fine for me - I mistakenly used `context.data` then. I was mostly describing how to in code clarify what `FW_KEY` is supposed to be `and that `OtherPlugin.HAS_RUN_KEY` may read like clearer intent?

So that it reads as:
- if the other thing ran, then skip it.

iLLiCiTiT: Was arelady added 🙂

BigRoy: Sorry, missed that - looks great! ❤️

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
BigRoy: I meant mostly the commented out `self.log.debug("Skipping, CollectOTIORanges has run.")` line that also appears to be overindented by one space (it isn't because the `#` is at the start of the line, but still)

client/ayon_core/plugins/publish/collect_otio_product_resources.py
```diff
@@ -37,11 +36,20 @@ def process(self, instance):
...
         if not product_base_type:
             product_base_type = instance.data["productType"]
 
-        if "audio" in product_base_type:
+        # TODO remove when 'CollectOtioSubsetResources'
+        # - if instance has family 'otio.resources' it should be
+        #   processed
+        if (
+            "otio.resources" in instance.data["families"]
+            and "audio" in product_base_type
```
iLLiCiTiT:
```suggestion
            "audio" in product_base_type
            and "otio.resources" not in instance.data["families"]
```

client/ayon_core/plugins/publish/collect_otio_frame_ranges.py
iLLiCiTiT: Oh yeah, the commented lines were uncommented to log it.

BigRoy: Sometimes I hate Github diffs :') thanks!

BigRoy (approved):
Code-wise looks OK to me. Needs more testing across various integrations for editorial + also testing for backwards compatibility there.

iLLiCiTiT:
@rdelillo could you find time to test some of the other hosts too?

rdelillo:
Update, all good for Resolve as well. The debug log when when plugins are skipped helps a lot.

rdelillo:
Update, tested in Hiero. Works as expected.

rdelillo (approved):
Tested with all editorial host. LGTM !

PR #1970, opened 2026-07-24
UI: Browser - Support views

### LINKED ISSUE #1806: YN-0483: default filters for Loader

we need similar defaults (studio/project) we have in web UI for filters to define for Loader.

  

So for example we can hide omitted products automatialy. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1469286655859888281/1469286661832708190)

_This issue was automatically created from AYON ticket [YN-0483]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=7147d9f0034f11f1a1ac21fece2f6b68)

mahesh-ynput:
## Changelog Description
Fix all bugs in "views" functionality in redesigned-browser

## Additional info
Local browser "views" functionality should match with web 

Fix https://github.com/ynput/ayon-core/issues/1806

## Testing notes:
1. Create different kind of views from products page and share it with multiple users with different permissions
2. Login different users from launcher and cross check complete "views" functionality should match with web 
3. Sorting, hiding columns, order of columns should be same as we see in web view to local view with respective views 

TODO:

- [x] - 1) Clicking on "working view" should unselect the view(match with web) and close the menu and update the working view
- [x] - 2) maintain only "desktop.browser"
- [x] - 3) save,load and change modified state on modifications of card/table view and any changes customize menu as same as currently doing for filters
- [x] - 4) Fix default settings loading(studio/project)
- [x] - 5) When loading any default settings(studio/project or clicking on "reset" should unselect if has any view selection and set to selection to working view)
- [x] - 6) For some scenario icon highlight is not removing 
- [x] - 7) use 'share api' for user access permission control
- [x] - 8) Make 'working view' as always in selection unless if any view loaded and it's modified
- [x] - 9) Fix "All projects" in scope combo in view editor
- [x] - 10) In editor add icons for users, match it with web


mahesh-ynput:
- [x]  Remove unwanted filters.
- [x]  Update **Filter Names** match with web.
- [x]  Update **Filter edit** When try to edit filter instead of showing it attributes it's only showing search option Eg: to modify status filter if click on it it should show status list but it showing  search option. Note: This behaviour can see only when filters are applied with any view selection
- [x]  Fix **Group By** so clearing the selection removes the applied grouping.
- [x]  Rename **Working** to **Working View**, and on action should clear the applied filters.
- [x]  Fix **Star Icon Working** needs to remove from view, (@BigRoy any idea why this here?)
- [x]  Add a **Save** option to persist changes made to the current view.
- [x]  Update **Edit View** Scope should be disabled old views only should enable for new view creation
- [x]  Update **New View**  Scope to display the project name and include an **All Projects** 
- [x]  Update **Edit View** & **New View** Fetch data for "People with access"
- [x]  Fix **Create New View** functionality, so new views are created successfully.
- [x]  Separate **My Views** and **Shared Views** into same like in web currently it showing all in "My views"
- [ ]  Add **Columns** order should be with user view? @BigRoy 
- [ ]  Custom attribute values are not populating to table and filter

BigRoy:
> * Fix **Star Icon Working** needs to remove from view, (@BigRoy any idea why this here?)

No idea - it should really behave like the web frontend. I don't see a reason for that 'star'.

> Add Columns order should be with user view? @BigRoy

The columns + filter should be included in the view, the slicer I believe should not. The slicer mode should be included, but not the slicer selection. (So it should remember whether it's on "Hierarchy" or "Reviews")

Also, note that the top "Folders" and "Reviews" tabs are still there only for development purposes, so that you can compare easily with the old loader - upon release we will NOT have that separation (those tabs won't be there).

Columns being hidable and reorderable, please in **separate PR**

---

It's also important that the views logic is mostly a "state manager" and preferably this state that it manages is not hardcoded into the "Views" logic. Because we may get similar Views support in e.g. Launcher or other tools. As such, it'd be great to keep in mind that the views purely store a named state and manage that, but it's up to what this particular view manager is connected to it to adjust based on it so that it can be used in a variety of tools as the "view state manager" etc.

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -43,7 +43,23 @@
 
 log = Logger.get_logger(__name__)
 
-VIEW_TYPE_VERSIONS = "versions"
+VIEW_TYPE_HIERARCHY = "desktop.browser"
+VIEW_TYPE_REVIEWS = "desktop.reviews"
```
BigRoy: The views aren't PER slicer. The hierarchy mode and review mode is just a different slicer, they save to the same views.

The Browser itself has views, the state of the slicer (which mode is chosen) is part of that saved view!
So these two constants make no sense. The browser has a singular views type, it contains the full state of the browser (except for the selection within the slicer, but the dropdown it does include (e.g. whether it's hierarchy or review, but it doesn't store the selection made inside the hierarchy)

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -945,6 +964,7 @@ def _make_real() -> LazyThumbnailWidget | None:
                 "folderName",
                 "Folder Name",
                 width=_w("Folder Name"),
+                filterable=False,
```
BigRoy, line 982: What does it mean if the column is not filterable? and why?

Frontend also doesn't filter by folder name, so fine by me. 👍

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -264,6 +274,18 @@ def _make_row(self, view: View) -> AYContainer:
             edit_btn.clicked.connect(
                 lambda _checked=False, v=view: self._on_edit_clicked(v)
             )
+            if hasattr(self._current_view, "id") and self._view_filters_modified:
```
BigRoy: Should `id` attribute always exist on `_current_view`?

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -395,6 +421,36 @@ def _close_menu(self) -> None:
...
+            if view is None
+            else (c for c in view.settings.filter.conditions if isinstance(c, dict))
+        ) if (view or self._bindings.filter_bar) else ()
+        return tuple(sorted(
+            (k, s, tuple(sorted(map(str, v))))
+            for i in items
+            if (k := str(i.get("key") if view else getattr(i, "key", "")))
+            and (v := i.get("values") if view else getattr(i, "values", ()))
+            for s in [bool(i.get("useSubstring") if view else getattr(i, "use_substring", False))]
+        ))
```
BigRoy: Can we write this in a different way, this is rather unreadable to me :)

client/ayon_core/ui/components/buttons.py
```diff
@@ -118,6 +92,56 @@ def __init__(
...
+            self._icon_hover_color = compute_color_for_contrast(
+                QColor(icon_hover_bg).toTuple(),
+                self._icon_color.toTuple(),
+                min_contrast_ratio=7,
+            )
+
+        # get the disabled opacity
+        self._disabled_opacity = self._style_data.get("disabled", {}).get(
+            "opacity", 0.5
+        )
```
BigRoy, line 127: Which icon, where? Why do we need all this complexity? 🤔

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: When we pass this it creates the columns but not add into filters.

client/ayon_core/ui/components/views/view_selector.py
mahesh-ynput: Actually, I wrote initially with helpers but used copilot to keep it into lesser code, here we just checking here current filter state so it can check whether filters have been modified compared to a actual loaded view. let's add same into doc or want me to rewrite into more readable?

client/ayon_core/ui/components/buttons.py
mahesh-ynput: we have view button primarily with different variant but we're changing it to other variant so visually it indicates filters are modified, but here i see AYButton icon state colors can I rename the method and add a short doc to make this intent clearer? or figure out more better way to handle?

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: I see - I figured since we had separate filters available that aren't columns (at least frontend sort of has that, right?), like "has reviewables" that the filters weren't based on the columns themselves, anyway. 👍 all good then.

client/ayon_core/ui/components/views/view_selector.py
BigRoy: Let' turn this into something WAY more readable. There's no reason for anything remotely this complex.

We're writing readable/understandable code, we're not trying to impress with one-liners where you'll later regret that you did and have no clue what it's doing reading it again in two years.

client/ayon_core/ui/components/buttons.py
BigRoy: I see, we're solely exposing `set_variant` and we need this refresh logic to make sure things look the way they should. 👍 I really hate that this much logic is involved with what should just be a change of color, but I guess it's the way this was built into the component itself to begin with. 👍

client/ayon_core/ui/components/views/view_selector.py
BigRoy: > here we just checking here current filter state so it can check whether filters have been modified compared to a actual loaded view. let's add same into doc or want me to rewrite into more readable?

Admittedly - I'd just keep a check whether we triggered any change to the view since last time, if we were to change, and then do another change that makes it match the view again completely I'm fine with it still saying it's unsaved. It's probably much quicker and simpler that way, no?

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: Reason for having two One view is for Review(Review sessions page) and other for browser(Same as products page). Still we should consider as single? I just matched with web UX where it has different views for both. What you're thoughts on this?

client/ayon_core/ui/components/views/view_selector.py
mahesh-ynput: > > here we just checking here current filter state so it can check whether filters have been modified compared to a actual loaded view. let's add same into doc or want me to rewrite into more readable?
> 
> Admittedly - I'd just keep a check whether we triggered any change to the view since last time, if we were to change, and then do another change that makes it match the view again completely I'm fine with it still saying it's unsaved. It's probably much quicker and simpler that way, no?

Roy, I'm bit confused here let's connect today when you're available.

BigRoy: What I'm saying is that `_on_filters_changed` you just mark it dirty and highlight the button perhaps? So that you know when something is unsaved. When choosing a dedicated view or saving a view (and you're in that view) then we can just set 'no changes'.

That's what I'm saying ;)

mahesh-ynput: Better I should change the name of _on_filters_changed to _on_filters_clicked, Currently after calling the _on_filters_changed then it decides weather current state of values are vary from like before even clicked filters.

mahesh-ynput: There's other way to do is , check if any new filter added and check if any exiting filter modified upon any filter user clicks, But I thought it more complex just just taking values from before clicking on filters and comparing them

mahesh-ynput: Better I'll handle this in ViewBindings

mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/d4c6441d27dabbfb216856d6bffe26ca90f6d9ed

mahesh-ynput:
> In the View lists:
> * The save button should appear on any row if you hover over it, but it should be there by default (even when not hovering) and highlighted on the row of your list active view from which you started making changes, e.g.
>   
>   * Enter view "X"
>   * Make some edits
>   * You're now in a working view, but on view "X" you see the highlighted save button.
>   * However, if you hover another existing view you also get the save button so you can always save your working view into any existing views.

https://github.com/ynput/ayon-core/pull/1970/commits/21dd7f3a5d31eb93b27e7f69f1ee11adf1b501ac

mahesh-ynput:
@BigRoy , Currently I see on clicking of "working view" reset in web it just looking for any default view we set at project or studio else clearing all filters. Can you please confirm the same logic we should apply here in local browser side too?

BigRoy:
> @BigRoy , Currently I see on clicking of "working view" reset in web it just looking for any default view we set at project or studio else clearing all filters. Can you please confirm the same logic we should apply here in local browser side too?

Yes - if we can support studio/project level defaults then that's the way it should work:

Reset to default means:
- reset to project default if set
- reset to studio default if set
- reset to AYON default

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -408,10 +440,26 @@ def _delete_view(self, view: View) -> None:
...
+        # reset to AYON default
         self._current_view = None
         self._sync_view_filters_modified_state()
         self._close_menu()
 
+    def _on_set_default_view_clicked(self, button: AYButton) -> None:
+        """Save to the default view for the given *scope*"""
+        #TODO: Implement the actual logic to save
+        # the default view for the given scope (studio or project).
+        if button._variant_str == "surface":
```
BigRoy: I don't think we should be guessing the scope based on the button variant string.

1. We should not be accessing its private attributes.
2. We should not be guessing.

Instead, one option is let the method take a scope enum. (note that below I explain why this may not even be needed.) 
```python
class DefaultViewScope(enum.Enum):
    Studio = "studio"
    Project = "project"
```

We either pass the button explicitly, or we preserve the buttons on self and access them via e.g. `self.default_view_project_button` or alike.

```py
def _on_set_default_view_clicked(self, scope: DefaultViewScope, button: AYButton):
    ...
```

Then you can attach the callbacks via `functools.partial`
```py
project_btn.clicked.connect(
    lambda _checked=False: partial(
        self._on_set_default_view_clicked,
        DefaultViewScope.Project
        project_btn
    )
)
```
But one could argue that because this method just does `if ..else` anyway, you can just as well write separate methods:

```py
def _on_set_default_project_view(...):

def _on_set_default_studio_view(...):
```

And you don't need to mind the whole scope to begin with.

mahesh-ynput: Noted!

mahesh-ynput:
@BigRoy, UI part is ready, need few enhancements but I need right end points to sharing view with access, set studio/project defaults

mahesh-ynput:
image
I can only modify settings but not 'access': {}  of any view , eg:  'access': {'__everyone__': 0, 'user:mahesh_lead_01': 10},

@BigRoy, I'm wrong here

BigRoy:
> image I can only modify settings but not 'access': {} of any view , eg: 'access': {'__everyone__': 0, 'user:mahesh_lead_01': 10},
> @BigRoy, I'm wrong here

Looks like you edited the comment later with "I'm wrong here." so you figured it out? Does that mean you were able to finish the things alongside the UI tweaks?

BigRoy:
I'm still seeing an `(unnamed view)` entry which I don't understand why it's there.

image

This entry also shows no options nor save icon. It should really not be there at all, right?

The views list is also lacking higlighting of which is my currently active view:
- If I select a view, that should turn to be the active one.
- Otherwise, it's the working view that should be active (highlighted) like the frontend does
-

---

required:
- [x] Remove `(unnamed view)`? not sure why I have an unnamed view that I can't enter/save?
- [x] Add highlighting of active view
- [x] Fix reset of working view (should activate working view and set it to the active default (e.g. project, studio or AYON default)
- [x] Studio/Project default views should still be fixed to actually work (doesn't do anything currently?)
    - Note that on the web frontend the studio/project view can also be clicked to enter that view aside of just 'adding/deleting' it. I think we'll need similar UX where we can activate the studio or project view or delete it
- [x] Store the view mode: table or rows into the view as well. Currently it doesn't switch to e.g. card view if I'm going to a saved view.


low prio:
- [x] Upon hovering (tooltip) of the view button display the active view, like frontend does:


image
image

- [x] Can you move the "Views" button to the LEFT of the Filter Bar so that it becomes

```
       [ view ]  [ filter bar ]  [ group by ]  [ table/cards ]  [ customize ]
```

BigRoy:
- [x] When in a named view, clicking "Working view" doesn't close the dialog, nor deselects the current named view.
- [x] Working view "reset button" does nothing it seems either?
- [x] When Studio / Project defaults are set, we should be able to click these to still go into those view and have that separate clickable area as an "X" on that default view (similar to web frontend.)

BigRoy:
When I click on a view it activates it, then I change e.g. card sizes - it remains the active entry and I can't save it. Preferably it turns into this:

image
_This is on frontend, my active view was "TEST"; as soon as I change something I'm in a "working view" and my last named view was "TEST" so we indicate that we'd likely want to change our changes there._

- It should make the Working View the active one. Because I've now changed something I'm working in a 'working view'
- The save icon remains visible and highlighted on the row that LAST was our active view. So that it hints where you'd like to save it into.
- A bonus would be if in that scenario we could highlight the "Views" button as well to indicate you have unsaved changes:
- When user explicitly clicks on Working View I think we should unlink from the last named view so that the highlight of the Views button is removed.

So: if currently in working view and last view was X then highlight save button on X and highlight the views button.

mahesh-ynput:
> When I click on a view it activates it, then I change e.g. card sizes - it remains the active entry and I can't save it. Preferably it turns into this:
> 
> image _This is on frontend, my active view was "TEST"; as soon as I change something I'm in a "working view" and my last named view was "TEST" so we indicate that we'd likely want to change our changes there._
> * It should make the Working View the active one. Because I've now changed something I'm working in a 'working view'
> * The save icon remains visible and highlighted on the row that LAST was our active view. So that it hints where you'd like to save it into.
> * A bonus would be if in that scenario we could highlight the "Views" button as well to indicate you have unsaved changes:
> * When user explicitly clicks on Working View I think we should unlink from the last named view so that the highlight of the Views button is removed.
> 
> So: if currently in working view and last view was X then highlight save button on X and highlight the views button.


https://github.com/user-attachments/assets/c7eb93cf-ce6c-46a3-aadd-d25345b4119e

@BigRoy , here also has same logic right? What I'm missing here

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: Should be one key only, as explained during call.

BigRoy:
> here also has same logic right? What I'm missing here

I wasn't changing the filters, but changing e.g. the card size ;) which it doesn't respond to currently. It should move me to working view as soon as I change anything that we would save into the view.

mahesh-ynput:
> * [ ]   Add **Columns** order should be with user view.
> * [ ]   Custom attribute values are not populating to table and filter.
> * [ ]   hide/unhide columns
> * [ ]   Fetch default filter values from project settings instead of just filling from from versions data
      Eg: currently if we load any project where has single versionit only show filters of status with that version
****
@BigRoy, Can we have separate PR for table fixes

BigRoy:
> > * [ ]    Add **Columns** order should be with user view.
> > * [ ]    Custom attribute values are not populating to table and filter.
> > * [ ]    hide/unhide columns
> > * [ ]    Fetch default filter values from project settings instead of just filling from from versions data
> >   Eg: currently if we load any project where has single versionit only show filters of status with that version
> > * [ ]    In view filters, whenever any filters are applying, instead of appending values, append the icon when possible to match it with web.


> @BigRoy, Can we have separate PR for table fixes

Yes please. Let's focus this PR as much as possible on the views themselves. Expanding the table + filters we can do in separate (smaller) PRs.

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -188,22 +189,48 @@ def _card_mapper(row_data: dict) -> dict:
...
+            )
+        )
+        self._display_type.display_type_changed.connect(
+            self._view_selector.notify_view_modified
+        )
+        self._group_by_menu.group_by_changed.connect(
+            self._view_selector.notify_view_modified
+        )
+        self._customize.show_empty_groups_changed.connect(
+            self._view_selector.notify_view_modified
```
BigRoy, line 220: This callback is rather slow (it's still comparing the view state?). Which is clearly noticable when you are dragging the card size slider in that it's rather laggy.

I really feel like there's no need to do so, as soon as the user changes the state in any way, we should just mark it as "there are changes" which makes the callback logic just the setting of a boolean state of "yes, I'm dirty."

Essentially:
```
def on_change():
    self.unsaved_view_changes = True
```

I don't think we'll need to capture the full view state, do a comparison of values, etc. Sure, it's nice if you were to match it again with the exact view that it may not ask you to save it again, but admittedly I wouldn't bother.

OR if there's a good reason we MUST compare states, then I think we should really be delaying this for these interactive changes because currently it's just too laggy and we may want to put it behind a timer then, so that it doesn't trigger as you interactively drag the card size, etc.

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -453,6 +481,11 @@ def _apply_view_extras(self, extra: dict[str, Any]) -> None:
         in ``extra`` after a server round-trip — they are applied from
         ``view.settings.grouping`` inside :meth:`_on_view_applied`.
 
+        Supported loader-specific extras:
+            ``gridHeight`` (card width),
+            ``featuredVersionOrder`` (controller order),
```
BigRoy:
```suggestion
            ``featuredVersionOrder`` (hero/latest version order),
```

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -481,6 +514,10 @@ def _apply_view_extras(self, extra: dict[str, Any]) -> None:
                     log.debug(
                         "Failed to set featuredVersionOrder: %r", order
                     )
+        if "displayType" in extra:
+            display_type = str(extra["displayType"] or "").strip().lower()
+            if display_type in {"table", "grid"}:
+                self._display_type.set_display_type(display_type, emit=True)
```
BigRoy: We really should just be expecting these values to be correct - when set they should be one of the two.
Which means we need no strip, no lower call, no `or ""` fallback. 

```suggestion
        if "displayType" in extra:
            display_type: Literal["table", "grid"] = extra["displayType"]
            self._display_type.set_display_type(display_type, emit=True)
```

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: When would there not be an order? Isn't there always one?

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -626,6 +667,8 @@ def _iter_visible_proxy_indices(
             return
 
         row_height = self._table_row_height()
+        if row_height <= 0:
+            row_height = 32
```
BigRoy: What's this for?

client/ayon_core/tools/loader/ui/_review_toolbar.py
```diff
@@ -242,6 +242,29 @@ def _build(self) -> None:
     def display_type(self) -> str:
         return self._display_type
 
+    def set_display_type(self, display_type: str, emit: bool = False) -> None:
```
BigRoy: Either make this a `Literal` type hint or do something else to make this very explicit, like using some enum.
```suggestion
    def set_display_type(self, display_type: Literal["table", "grid"], emit: bool = False) -> None:
```

BigRoy: When would this not emit? also, I see a single usage (or I'm wrong?) as such isn't the `emit=False` redundant?

client/ayon_core/tools/loader/ui/reviews_widget.py
```diff
@@ -73,6 +75,12 @@ def __init__(
         )
         self._inspector = ReviewInspector(self._controller)
         self._table.display_type_changed.connect(self._inspector.set_view)
+        self._table.default_view_message.connect(
+            lambda message, success: self.default_view_message.emit(
+                message,
+                success,
+            )
+        )
```
BigRoy, line 83: This feels rather messy how we need to pass on this message.

client/ayon_core/ui/components/views/data_models.py
```diff
@@ -387,10 +372,12 @@ def to_payload(self) -> dict[str, Any]:
 # View
 # ---------------------------------------------------------------------------
 
+VIEWER_ACCESS_LEVEL: int = 10
```
BigRoy, line 375: Are we lacking a constant for the no access?

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,367 @@
...
+            else self.project_default_view
+        )
+        if default_view is None or not default_view.id:
+            return False
+
+        if not self._confirm_unset_default(scope):
+            return False
+
+        try:
+            with sel._suspend_auto_apply():
```
BigRoy: I'm a bit surprised what this is trying to do, why does a save even TRY to change anything?

All it should do is:
- get the current state, write to database
- unhighlight the view button (and its menu states)

It should in by no means touch any UI widgets themselves for which we're tracking state.
As such, why would need to bypass ANYTHING?

The same goes for e.g. when deleting a view - it'd just mean that if were in that view, that our current view would become the working view. We are not CHANGING the UIs.

I think there's something fundamentally wrong if these are accidentally changing UI states.

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,367 @@
...
+    # ------------------------------------------------------------------
+    # Persistence helpers
+    # ------------------------------------------------------------------
+
+    def _fetch_default_views(self) -> tuple[View | None, View | None]:
+        """Pull the current studio and project default views from the manager."""
+        sel = self._selector
+        try:
+            self.studio_default_view = sel._manager.get_default_studio_view(sel._view_type)
+            self.project_default_view = sel._manager.get_default_project_view(sel._view_type)
```
BigRoy, line 200: We shouldn't be touching private method of something that is not `self`

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -157,17 +251,48 @@ def list_views(self, view_type: str) -> list[View]:
...
+                # summary values from the list response.
+                for key, value in detail_payload.items():
+                    if key == "settings" and not value:
+                        continue
+                    if (
+                        key in {"id", "label", "viewType"}
+                        and (value is None or value == "")
+                    ):
+                        continue
+                    candidate[key] = value
```
BigRoy: What's this hack and why is it needed? shouldn't we just be getting the data back that we stored?

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -177,6 +302,7 @@ def list_views(self, view_type: str) -> list[View]:
         for v in views:
             if v.id:
                 self._id_to_type[v.id] = view_type
+                self._id_to_scope[v.id] = v.scope
```
BigRoy: Doesn't it just make more sense to store the views per id instead, instead of flattening out certain attributes per id, etc.

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -324,6 +517,61 @@ def _extract_data(resp: Any) -> Any:
...
+    @staticmethod
+    def _extract_view_dict(payload: Any) -> dict[str, Any] | None:
+        """Return a single-view dict from common detail response shapes."""
+        if not isinstance(payload, dict):
+            return None
+        if isinstance(payload.get("view"), dict):
+            return payload["view"]
+        if isinstance(payload.get("data"), dict):
+            return payload["data"]
+        return payload
```
BigRoy: Please do not 'guess' where data may live, the server should respond with a consistent output - and we should get that data as such. We should not be checking view, then data, etc.

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -324,6 +517,61 @@ def _extract_data(resp: Any) -> Any:
...
+                scope.value,
+                view_type,
+                view_payload.get("detail"),
+            )
+            return None
+
+        candidate = dict(view_payload)
+        candidate["viewType"] = str(candidate.get("viewType") or view_type)
+        candidate["scope"] = str(candidate.get("scope") or scope.value)
+        candidate["label"] = str(candidate.get("label") or "__base__")
```
BigRoy: Why do we need all these fallbacks? Shouldn't we be able to rely on the data we're writing out?

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -343,3 +591,45 @@ def _upsert_cache(self, view: View) -> None:
...
+        normalized: dict[str, int] = {}
+        for key, value in raw_access.items():
+            name = str(key).strip()
+            if not name:
+                continue
+            try:
+                normalized[name] = int(value)
+            except (TypeError, ValueError):
+                continue
+        return normalized
```
BigRoy, line 510: Why does this need normalizing? Shouldn't we get the data accordingly back from the server?

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -343,3 +591,45 @@ def _upsert_cache(self, view: View) -> None:
...
+            return None
+
+        for addon in bundle_info.addons:
+            if addon.name != _POWERPACK_ADDON_NAME:
+                continue
+            if addon.version:
+                self._powerpack_version = addon.version
+                return self._powerpack_version
+
+        return None
```
BigRoy, line 535: We may want to set some dedicated value if it has no power-pack, we don't want to query bundle information every time because we didn't cache/store that we didn't have power-pack. So we may need some constant or state to indicate (we don't have powerpack, and cache that state)

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -115,62 +130,57 @@ def _build_ui(self) -> None:
                 "short_text": "Project",
             }
         ]
+        #TODO: its not working as expected, need to check
```
BigRoy, line 133: Need to check what? :)

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -115,62 +130,57 @@ def _build_ui(self) -> None:
...
-        self._user_access_edit = AYLineEdit(
-            placeholder="Add people or access groups"
+        self._form.add_row("Scope", self._scope_combo)
+
+        # User/group selector dropdown
+        self._user_selector = AYComboBox()
+        self._update_user_dropdown()
+        self._form.add_row(
+            "Add people or access groups",
+            self._user_selector
```
BigRoy, line 146: Can we make this a full-width search field with auto complete or something along those lines, to make it behave a bit more like the frontend. Now it looks like the rest of the form and it doesn't look like something that'd expand the list below it, etc.

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +204,141 @@ def _load_from_view(self, view: View) -> None:
...
+        if self._user_list:
+            for user in self._user_list:
+                user_key = f"user:{user}"
+                if user_key not in added_keys:
+                    items.append({"text": f"{user}"})
+
+            # Add group option if not already added
+            group_key = "group:artist"
+            if group_key not in added_keys:
+                items.append({"text": "group:artist"})
```
BigRoy, line 225: We should not be hardcoding the user group - we should be listing all groups.

client/ayon_core/ui/components/views/server_view_manager.py
BigRoy, line 112: This whole storing of `_known_types` seems odd. Shouldn't we really just be clearing things when switching project and just assume we can't auto-query these then apparently? Why not just let another `list_views()` call just populate them again when needed. Seems odd to emit this?

If we were to switch project, shouldn't it just remain in the currently active view. (Sure, the new project may not have the active list view name, but it'd essentially make that the working view then. right?)

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -113,6 +123,91 @@ def set_project(self, project_name: str) -> None:
...
+        if not self._project_name:
+            return None
+
+        try:
+            resp = ayon_api.get(
+                f"views/{view_type}/base",
+                project_name=self._project_name,
+            )
+        except Exception as exc:  # noqa: BLE001
+            log.exception("Failed to fetch project default view for %s", view_type)
```
BigRoy, line 164: Should this include the stack trace? `exc_info=True`?

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -157,17 +251,48 @@ def list_views(self, view_type: str) -> list[View]:
...
+                    continue
+
+                view_id = str(item.get("id") or "").strip()
+                if not view_id:
+                    continue
+
+                candidate: dict[str, Any] = dict(item)
+                is_default_candidate = (
+                    str(candidate.get("label") or "") == "__base__"
+                )
```
BigRoy: Not a fan of all this `get()` with fallbacks etc. We should really be much more consistent with what we put in `items` list so we don't need all this "oh boy data may be wrong". Right?

client/ayon_core/ui/components/buttons.py
```diff
@@ -325,4 +354,7 @@ def _on_button_clicked(self) -> None:
     def _on_popup_closed(self) -> None:
         """Update state when the popup signals it has closed."""
         self._menu_open = False
+        if QtWidgets.QApplication.mouseButtons() & Qt.MouseButton.LeftButton:
+            local_pos = self.mapFromGlobal(QtGui.QCursor.pos())
+            self._suppress_reopen_on_next_click = self.rect().contains(local_pos)
```
BigRoy, line 359: Why is this needed?

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: > if there's a good reason we MUST compare states

For view selector button variant toggle(highlighted/filled) we do need some change detection. but instead of capturing and comparing the full view state on every UI event, maybe we can track with self.changed_filters={}
/dirty flag and compare only that. comparison is still needed, but this should be better than full state capture.
what are your thoughts on this ?

Can you comment out this signals and see the difference, if it still too laggy then I can add timer.

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: this freezing the UI If the table hasn't rendered yet, row_height could be 0 which is causing an infinite loop with below while loop

client/ayon_core/tools/loader/ui/_review_toolbar.py
mahesh-ynput: No needed, we can remove this, https://github.com/ynput/ayon-core/pull/1970/commits/5b1adee8d2a03bf68503bb0fc5b52b0c4f4431be

mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/ab9a21064b6596488a6013173fb23a4f007de968

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: No, needed. order is always initialized with default values. https://github.com/ynput/ayon-core/pull/1970/commits/ab9a21064b6596488a6013173fb23a4f007de968

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/ab9a21064b6596488a6013173fb23a4f007de968

client/ayon_core/ui/components/views/default_view_control.py
mahesh-ynput: Yes, we no needed while inspecting UI freezing I added this just to debug but it was fixed with other root cause https://github.com/ynput/ayon-core/pull/1970#discussion_r3701721181

mahesh-ynput: We can remove them compeletly

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: I see - if the table hasn;'t rendered yet can't we just skip yielding anything since there's nothing rendered yet, and hence nothing is visible? What would that break?

client/ayon_core/ui/components/views/default_view_control.py
mahesh-ynput: yes, I was suppose to ask you same is here, any better approach to handle DefaultViewControl for AYViewSelector, or should I rename every private methods in use

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: > For view selector button variant toggle(highlighted/filled) we do need some change detection. but instead of capturing and comparing the full view state on every UI event, maybe we can track with self.changed_filters={}
/dirty flag and compare only that. comparison is still needed, but this should be better than full state capture.
what are your thoughts on this ?

I think even for that we don't care - we just highlight as soon as we've made a change. If someone makes a change that just happens to make it match exactly the saved state, it'd remain highlighted. We don't care at that point.

client/ayon_core/ui/components/views/data_models.py
mahesh-ynput: Not even needed here, we already defined  in view_editor.py anyhow can we create constants.py put everything there @BigRoy ?

BigRoy: I'll leave that preference to @iLLiCiTiT

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: Noted roy, then I'll just  go with 
def on_change():
    self.unsaved_view_changes = True

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/default_view_control.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/33899d76f0ac8a5e66d4484bd81a8db6de868f85

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -157,17 +160,43 @@ def list_views(self, view_type: str) -> list[View]:
...
                 except Exception as exc:  # noqa: BLE001
-                    log.exception("Failed to fetch view %s", item.get("id"))
+                    log.exception("Failed to fetch view %s", view_id)
                     self.error.emit(f"Failed to fetch view: {exc}")
                     continue
-                full_item = resp.data
-                views.append(View.from_payload(full_item))
+
+                raw_detail = self._extract_data(resp)
+                detail_payload = self._extract_view_dict(raw_detail)
```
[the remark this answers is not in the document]
mahesh-ynput: fixed in code refactored https://github.com/ynput/ayon-core/pull/1970/commits/1bd10d4ebc8e9a0627a580b0f5b3aeaaf7b94a6d

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -188,22 +189,48 @@ def _card_mapper(row_data: dict) -> dict:
...
             manager=self._view_manager,
-            view_type=VIEW_TYPE_VERSIONS,
+            view_type=BROWSER_VIEW_TYPE,
             current_user=_get_current_user(),
             user_access_level=50,
             allow_studio_scope=False,
             parent=self,
         )
+        self._view_selector.setToolTip("Views")
+        self._view_selector._icon_size = 20
```
BigRoy, line 199: It looks like this icon size fix does make the button the same size, but it also makes the actual icon bigger/thicker than the others. The real issue here is that a button without text has less height than one with text, meaning that likely the text makes it thicker. Which would make it a bug/issue on the Button component itself that actually buttons with solely icons and no text should be taller? (not by increasing the icon size, but by increasing the padding/margins around the icons with 2 pixels perhaps?)

Let's remove this `_icon_size` fix. And then it'd be great if you can find out how to solve the button height itself. Or the alternative is to reduce the padding around the text so that even though when text is added on a single line, the button remains the same height.

(I don't believe it's an issue of `AYViewSelector` but of `AYButton` is my guess)

client/ayon_core/ui/components/views/view_editor.py
mahesh-ynput: Can I create a custom widget where it can also include user avatar similar like web?

BigRoy: I don't think that'd need a custom widget - it'd be the "user model" in the combobox providing the icons right? However, the combobox with search functionality may need a custom component on its own?

client/ayon_core/ui/components/views/view_editor.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/808d2fabe7dd22811464cc3377c155643fa78620

client/ayon_core/ui/components/buttons.py
mahesh-ynput: to prevent the menu from reopening on the same click that closed it on menu click

client/ayon_core/ui/components/views/view_editor.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/5ba790ba6721f0f0cff5db43d57eb73702c2fa98 Previously when user try to create new view "All projects" not listed in. It's fixe

client/ayon_core/ui/components/views/view_editor.py
mahesh-ynput: I'll create new component which support combobox with search functionality

client/ayon_core/ui/components/views/default_view_control.py
BigRoy: The AYViewSelector should give us access to what we need to touch. Not _every_ private method, but we should make clean decisions about what is the API for the data that we need to pass through. In this case I'd need to read up what is the Selector, and then its Manager, and then its default views, etc.

I'm not sure why the separation is as it is, but we should at least avoid needing to touch private methods of other objects.

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +204,141 @@ def _load_from_view(self, view: View) -> None:
...
+        # Clear existing rows
+        for widget in self._access_row_widgets.values():
+            widget.deleteLater()
+        self._access_row_widgets.clear()
+
+        # Clear access-related rows that were added dynamically
+        # (Owner, Everyone, and user/group rows)
+        layout = self._form.layout()
+        while layout.rowCount() > 3:
+            layout.removeRow(3)
```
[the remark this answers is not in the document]
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/76f58160ed60c6a7b5add04be99b7d9ee3fb5189

client/ayon_core/ui/components/views/view_editor.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/76f58160ed60c6a7b5add04be99b7d9ee3fb5189

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/19d620f54715238881b81fdfe13b5764c8d4f252

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: log.exception captures the stack trace, right? why again exc_info=True?

client/ayon_core/tools/loader/ui/reviews_widget.py
mahesh-ynput: I just matched with how we used in Launcher and Loader, Any suggestions for better way of using it?

client/ayon_core/ui/components/views/default_view_control.py
mahesh-ynput: We can handle DefaultViewControl within AYViewSelector, but I separated as it has some different UI logic and other helpers.

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/b207d6bde13f0b9121c6ae3c7eff94d9f2f5bd51

client/ayon_core/ui/components/views/server_view_manager.py
BigRoy: Ah you're right - missed that.

client/ayon_core/ui/components/buttons.py
BigRoy: Shouldn't it really be that the button itself acts as toggle, like when clicked if menu is open, then close it... else show it? I don't get why we need this - sorry, this points to an issue/bug elsewhere, no?

client/ayon_core/ui/components/searchable_combo_box.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def _on_text_changed(self, text: str) -> None:
+        self._dropdown.filter_items(text)
+        if not self._dropdown.isVisible():
+            self._show_dropdown()
+
+    def _on_item_selected(self, key: str) -> None:
+        self._line_edit.clear()
+        self._dropdown.filter_items("")
+        self.item_selected.emit(key)
```
BigRoy:
```suggestion
        self.item_selected.emit(key)

```

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: @mahesh-ynput still pending.

client/ayon_core/tools/loader/ui/_review_table.py
```diff
@@ -493,15 +529,15 @@ def _capture_view_extras(self) -> dict[str, Any]:
...
         """
-        extra: dict[str, Any] = {
-            "gridHeight": int(self._card_view.card_width),
-        }
-        order = self._controller.featured_version_order
-        if order:
-            extra["featuredVersionOrder"] = order
+        extra: dict[str, Any] = {"gridHeight": int(self._card_view.card_width),
+                                 "displayType": self._display_type.display_type,
+                                 "featuredVersionOrder": self._controller.featured_version_order}
```
BigRoy:
```suggestion
        extra: dict[str, Any] = {
            "gridHeight": int(self._card_view.card_width),
            "displayType": self._display_type.display_type,
            "featuredVersionOrder": self._controller.featured_version_order
        }
```

client/ayon_core/tools/loader/ui/reviews_widget.py
BigRoy: Can you give an example of where we do that?

client/ayon_core/ui/components/views/default_view_control.py
BigRoy: @iLLiCiTiT can you give some recommendation here?

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -343,3 +475,46 @@ def _upsert_cache(self, view: View) -> None:
         else:
             view_list.append(view)
         view_list.sort(key=lambda v: (v.position, v.label.lower()))
+
+    @staticmethod
+    def _normalize_access_payload(raw_access: Any) -> dict[str, int]:
```
BigRoy: Why do we need this? Shouldn't `raw_access` be `dict[str, int]` to begin with, or?

can we give better type hint for `raw_access` argument?

BigRoy (changes requested):
- When not hovering the save icon on a view row it should show the icon, but not have bright "hover" color over it.

desktop:
Image
web:

Image

Image

- When no project/studio view is set resetting the working view should clear the filters and states to some predefined default that we'd consider the browser's "default" instead of not being able to reset. Prepare the model so that we can set some defaults in the code for that.

client/ayon_core/ui/components/views/data_models.py
iLLiCiTiT: As far as I see, the contants are not used for anything, it is just for reference, I guess?

I'm fine keeping it in this file. But if we do start to use it elsewhere we should move it to some `constants.py`. At least put it at the top above classes and functions. Define constants in the middle of file makes it really hard to find.

client/ayon_core/ui/components/views/default_view_control.py
iLLiCiTiT: Add `get_default_studio_view` to `AYViewSelector`.

client/ayon_core/tools/loader/ui/reviews_widget.py
mahesh-ynput: >This feels rather messy how we need to pass on this message. 

sorry @BigRoy , I misunderstand here 

I can change it to:
 self._table.default_view_message.connect(
            self.default_view_message
        ) 
?

mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/3932710202c76ab437eecb79b879dc70b1e49594

client/ayon_core/ui/components/views/data_models.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/e24a8da4d2026a5fb82819d9c57033b6b12d724b

mahesh-ynput:
> * When no project/studio view is set resetting the working view should clear the filters and states to some predefined default that we'd consider the browser's "default" instead of not being able to reset. Prepare the model so that we can set some defaults in the code for that.

@BigRoy , But in web currently it's just clearing can we bring same behavior here also or we really needed some defaults?

client/ayon_core/ui/components/searchable_combo_box.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/33200d32bc6692880776b339b684dd4613824c8b

BigRoy:
> But in web currently it's just clearing can we bring same behavior here also or we really needed some defaults?

It may look like only it's "clearing", because those are the defaults. Note how e.g. on web Overview, when resetting working view if there's no studio/project default that it doesn't show ALL columns, it orders them a certain way, the row height resets to 34, group by turns to Hierarchy. "Reset to default view" means to reset to a certain default, and we need that here as well. It doesn't just flat out show all columns, etc. - it resets to an "AYON default" product team prepared there. We'll want to do the same here, but for the loader.

client/ayon_core/ui/components/views/server_view_manager.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/f796f08917b41fdaba83c9bbf9df7fb7797c7b70

mahesh-ynput:
> * When not hovering the save icon on a view row it should show the icon, but not have bright "hover" color over it.

https://github.com/ynput/ayon-core/pull/1970/commits/0c3ec39305dc267c6e3cf54192dac468ed724f3e, Still entire view row needs to have "hover" color maybe I'll match it later

client/ayon_core/ui/components/buttons.py
mahesh-ynput: class AYDropdownPopup(AYFrame):
    def __init__():
        self.setWindowFlags(

>             Qt.WindowType.Popup

 | Qt.WindowType.FramelessWindowHint
            | Qt.WindowType.NoDropShadowWindowHint
        )
Qt.Popup auto closes on the button click before the button clicked() signal emits, so by the time the toggle runs the menu already looking closed and immediately reopening and I'm not sure excatly is this root cause. can you correct me here @BigRoy

mahesh-ynput:
> * Prepare the model so that we can set some defaults in the code for that.

https://github.com/ynput/ayon-core/pull/1970/commits/d73323406de28ed17c3959d2275437eb8d248ac6

client/ayon_core/ui/components/views/data_models.py
```diff
@@ -265,6 +253,16 @@ def is_empty(self) -> bool:
...
+DEFAULT_SORT_BY = None
+DEFAULT_SORT_DESC = False
+DEFAULT_ROW_HEIGHT = 34
+DEFAULT_GROUPING = GroupingDef()
+DEFAULT_FILTER = FilterDef()
+DEFAULT_EXTRA = {
+    "gridHeight": 230,
+    "displayType": "table",
+    "featuredVersionOrder": ["latestDone", "latest", "hero"],
+}
```
iLLiCiTiT: Why these were added? Re-using the same object for default value is dangerous. Having these as constants doesn't make sense at all (at least I don't see any reason).

Only thing that needs custom defaults is `extra`, and that one should use function instead of stored value too.

```python
def _get_default_extra() -> dict[str, Any]:
    return {
        "gridHeight": 230,
        "displayType": "table",
        "featuredVersionOrder": ["latestDone", "latest", "hero"],
    }
```
To be used as `field(default_factory=_get_default_extra)`.

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,380 @@
...
+
+    # ------------------------------------------------------------------
+    # Row builder
+    # ------------------------------------------------------------------
+
+    def build_row(self) -> AYContainer:
+        """Build and return the full Studio + Project default-view row."""
+        self.studio_default_view, self.project_default_view = (
+            self._fetch_default_views()
+        )
```
iLLiCiTiT:
```suggestion
        self._fetch_default_views()
```

client/ayon_core/tools/loader/ui/_review_table.py
mahesh-ynput: > It looks like this icon size fix does make the button the same size, but it also makes the actual icon bigger/thicker than the others. The real issue here is that a button without text has less height than one with text, meaning that likely the text makes it thicker. Which would make it a bug/issue on the Button component itself that actually buttons with solely icons and no text should be taller? (not by increasing the icon size, but by increasing the padding/margins around the icons with 2 pixels perhaps?)
> 
> Let's remove this `_icon_size` fix. And then it'd be great if you can find out how to solve the button height itself. Or the alternative is to reduce the padding around the text so that even though when text is added on a single line, the button remains the same height.
> 
> (I don't believe it's an issue of `AYViewSelector` but of `AYButton` is my guess)

https://github.com/ynput/ayon-core/pull/1970/commits/1ef71bd18fe2883f948a36f3c0b2266630c2a9db

I couldn't find a better way to handle this style overrides, @BigRoy have any suggestions here?

client/ayon_core/ui/components/views/data_models.py
mahesh-ynput: https://github.com/ynput/ayon-core/pull/1970/commits/942e05577d89bf78859fc074edf25b7c1343af7c, Can we do this way?

client/ayon_core/tools/loader/ui/_review_table.py
BigRoy: In that case, ignore fixing it - just remove the size override.

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,398 @@
...
+        if control is None:
+            return
+        layout = control.layout()
+        if layout is None:
+            return
+        while layout.count():
+            item = layout.takeAt(0)
+            widget = item.widget() if item is not None else None
+            if widget is not None:
+                widget.deleteLater()
```
iLLiCiTiT, line 111:
```suggestion
                widget.setVisible(False)
                widget.deleteLater()
```

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,398 @@
...
+    and ``_suspend_auto_apply``.
+    """
+
+    def __init__(self, selector: "AYViewSelector") -> None:
+        self._selector = selector
+
+        self.studio_default_view: View | None = None
+        self.project_default_view: View | None = None
+
+        self._studio_control: AYContainer | None = None
```
iLLiCiTiT, line 61: Why it is called `control`?

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,398 @@
...
+    def load_studio_default_view(self, _checked: bool = False) -> None:
+        self._fetch_default_views()
+        if self.studio_default_view is None:
+            self._emit_action_message(
+                "No studio default view configured.",
+                success=False,
+            )
+            return
+        try:
+            self._selector._apply_view(self.studio_default_view, emit=True)
```
iLLiCiTiT, line 312: This probably should be public method.

client/ayon_core/ui/components/views/default_view_control.py
```diff
@@ -0,0 +1,398 @@
...
+            sort_desc=DEFAULT_SORT_DESC,
+            row_height=DEFAULT_ROW_HEIGHT,
+            grouping=DEFAULT_GROUPING,
+            filter=DEFAULT_FILTER,
+            extra=DEFAULT_EXTRA,
+        )
+
+        self._selector._bindings.apply(default_view)
+        self._selector._clear_modified()
+        self._selector._close_menu()
```
iLLiCiTiT, line 371: Eh? This should be one public method on selector.

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -69,12 +74,11 @@ def __init__(
...
         self._cache: dict[str, list[View]] = {}
-        # All view_types ever requested via list_views (retained after
-        # cache clear so set_project can emit per-known-type).
-        self._known_types: set[str] = set()
-        # view_id -> view_type (survives list_views; used by delete_view
-        # so it doesn't require a populated per-type cache).
-        self._id_to_type: dict[str, str] = {}
+        # view_id -> (view_type, scope) (survives list_views; used by
+        # delete_view so it doesn't require a populated per-type cache).
+        self._id_to_view_attributes: dict[str, tuple[str, Scope]] = {}
```
iLLiCiTiT, line 79: I would rename it to `_view_attributes_by_id`.

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -237,13 +307,69 @@ def save_view(self, view: View) -> View:
...
         self._upsert_cache(saved)
 
+        # Access grants are managed by the dedicated share endpoint.
+        self.patch_view_access(saved.id, access_data, should_share_access)
+
         self.view_saved.emit(saved.id)
         self.views_changed.emit(saved.view_type)
         return saved
 
+    def patch_view_access(self, view_id: str, access_data: dict[str, Any], should_share_access) -> None:
```
iLLiCiTiT, line 320:
```suggestion
    def patch_view_access(
        self, view_id: str,
        access_data: dict[str, Any],
        should_share_access: bool,
    ) -> None:
```

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -237,13 +307,69 @@ def save_view(self, view: View) -> View:
...
+
+        payload_access = self._normalize_access_payload(access_data)
+        if not payload_access:
+            return
+
+        powerpack_version = self._get_powerpack_version()
+        if not powerpack_version:
+            self.error.emit("Could not resolve powerpack addon version")
+            return
+        visibility = Visibility.PUBLIC.value if should_share_access else Visibility.PRIVATE.value
```
iLLiCiTiT, line 349:
```suggestion
        visibility = (
            Visibility.PUBLIC.value
            if should_share_access
            else Visibility.PRIVATE.value
        )
```

client/ayon_core/ui/components/views/server_view_manager.py
```diff
@@ -261,23 +387,29 @@ def delete_view(self, view_id: str) -> None:
...
-                f"views/{view_type}/{view_id}",
-                project_name=self._project_name,
-            )
+            if is_studio:
+                ayon_api.delete(f"views/{view_type}/{view_id}")
+            else:
+                ayon_api.delete(
+                    f"views/{view_type}/{view_id}",
+                    project_name=self._project_name,
+                )
```
iLLiCiTiT, line 405:
```suggestion
            body = {}
            if not is_studio:
                body = dict(project_name=self._project_name)
            ayon_api.delete(f"views/{view_type}/{view_id}", **body)
```

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +272,289 @@ def _load_from_view(self, view: View) -> None:
...
+        # Load access data from view
+        self._access_dict = dict(view.access) if view.access else {}
+
+        # Update Owner
+        owner_name = self._view.owner or self._current_user or "-"
+        self._owner_label.setText(owner_name)
+        
+        # Recreate owner avatar with updated name
+        self._owner_avatar._name = owner_name
+        self._owner_avatar._full_name = owner_name
```
iLLiCiTiT, line 284: Shouldn't this be one public method?

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +272,289 @@ def _load_from_view(self, view: View) -> None:
...
+        row.add_widget(remove_btn)
+
+        self._access_rows_container.add_widget(row)
+        self._access_row_widgets[key] = row
+
+    def _populate_access_rows(self) -> None:
+        """Populate access control rows from _access_dict."""
+        # Clear existing rows
+        for widget in self._access_row_widgets.values():
+            widget.deleteLater()
```
iLLiCiTiT, line 410:
```suggestion
            widget.setVisible(False)
            widget.deleteLater()
```

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -230,48 +270,99 @@ def _make_current_row(self, view: View) -> AYContainer:
...
+
+    def _make_row(self, view: View) -> AYContainer:
+        """Build one selectable row for *view*."""
+        row = AYContainer(
+            layout=AYContainer.Layout.HBox,
+            layout_spacing=4,
+            layout_margin=0
+        )
+
+        not_modified = (
```
iLLiCiTiT, line 335: Could we swap the name to avoid `not_*` that is then checked as `if not not_modified:`...
```suggestion
        modified = not (
```

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -296,15 +387,28 @@ def refresh(self) -> None:
...
-            if working is not None:
-                self._apply_view(working, emit=True)
+        working = self._ensure_working_view_exists()
+        if (
+            working is not None
+            and self._current_view is None
+            and not self._suppress_auto_apply
+        ):
+            self._apply_view(working, emit=True)
+            return
```
iLLiCiTiT, line 397:
```suggestion
```

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -296,15 +387,28 @@ def refresh(self) -> None:
             log.exception("Failed to list views for %r", self._view_type)
             self._views = []
 
-        if self._current_view is None and not self._suppress_auto_apply:
-            working = next((v for v in self._views if v.working), None)
-            if working is not None:
-                self._apply_view(working, emit=True)
+        working = self._ensure_working_view_exists()
```
iLLiCiTiT, line 390:
```suggestion
        working_view = self._ensure_working_view_exists()
```

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +272,289 @@ def _load_from_view(self, view: View) -> None:
...
+        Returns:
+            AYContainer with formatted widgets for the access row.
+        """
+        if key.startswith("user:"):
+            # Extract username from key
+            user_name = key.replace("user:", "")
+
+            # Find the full name from usernames_and_groups
+            user_full_name = user_name
+            for user in self.usernames_and_groups.get("users", []):
```
iLLiCiTiT, line 479:
```suggestion
            for user in self.usernames_and_groups["users"]:
```

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +272,289 @@ def _load_from_view(self, view: View) -> None:
...
+                if user.get("name") == user_name:
```
iLLiCiTiT, line 480:
```suggestion
                if user["name"] == user_name:
```

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -317,10 +421,15 @@ def _on_view_selected(self, view: View) -> None:
     def _on_edit_clicked(self, view: View) -> None:
         """Open the editor for an existing view."""
         self._close_menu()
+        editable = View.from_payload(view.to_payload())
```
iLLiCiTiT, line 424: I guess the `editable` is a view to edit? I would call it `view_copy`.

client/ayon_core/ui/components/views/view_selector.py
```diff
@@ -362,27 +474,87 @@ def _delete_view(self, view: View) -> None:
...
+
+    def _get_usernames_and_groups(self) -> dict[str, list]:
+        """Fetch active project users and return unique usernames."""
+        project_name = getattr(self._manager, "project_name", "")
+        if not project_name:
+            return {"users": [], "groups": []}
+        users = []
+        groups = []
+        try:
+            users = ayon_api.get_users(project_name=project_name) or []
```
iLLiCiTiT, line 531:
```suggestion
            users = ayon_api.get_users(project_name=project_name)
```

client/ayon_core/ui/components/views/view_editor.py
```diff
@@ -194,6 +272,289 @@ def _load_from_view(self, view: View) -> None:
...
+                    user_full_name = user.get("fullName", "") or user_name
```
iLLiCiTiT, line 481:
```suggestion
                    user_full_name = user["fullName"] or user_name
```

client/ayon_core/ui/components/buttons.py
```diff
@@ -325,4 +463,9 @@ def _on_button_clicked(self) -> None:
     def _on_popup_closed(self) -> None:
         """Update state when the popup signals it has closed."""
         self._menu_open = False
+        if QtWidgets.QApplication.mouseButtons() & Qt.MouseButton.LeftButton:
+            local_pos = self.mapFromGlobal(QtGui.QCursor.pos())
```
iLLiCiTiT, line 467:
```suggestion
            local_pos = self.mapFromGlobal(QtGui.QCursor.pos())
            if isinstance(local_pos, QtCore.QPointF):
                local_poc = local_pos.toPoint()
```

client/ayon_core/ui/components/searchable_combo_box.py
```diff
@@ -0,0 +1,356 @@
...
+
+        Args:
+            items: A list of dicts with ``key`` and ``label`` keys, or
+                plain strings (used as both key and label).
+        """
+        while self._list_layout.count():
+            item = self._list_layout.takeAt(0)
+            w = item.widget()
+            if w:
+                w.deleteLater()
```
iLLiCiTiT, line 169:
```suggestion
                w.setVisible(False)
                w.deleteLater()
```

client/ayon_core/ui/components/searchable_combo_box.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def filter_items(self, text: str) -> None:
+        """Show only rows whose key or label contains *text*.
+
+        The comparison is case-insensitive.
+
+        Args:
+            text: Search string typed into the line edit.
+        """
+        needle = text.lower()
```
iLLiCiTiT, line 195: `needle`?

BigRoy: The needle you search for in a haystack :)

client/ayon_core/ui/components/views/default_view_control.py
mahesh-ynput: control(AYContainer) which using to set and unset project and studio defaults when we set any defaults it modify the widgets inside

client/ayon_core/ui/components/views/default_view_control.py
mahesh-ynput: There are many private methods from self._selector are in private methods. I'm bit unclear which and all needs to make public.

client/ayon_core/ui/components/searchable_combo_box.py
iLLiCiTiT: Ok... I see... Could it be `text_filter` instead?

BigRoy:
Closing in favor of https://github.com/ynput/ayon-core/pull/2031

Issue #1950, opened 2026-07-14
Frame/udim information on representation files


iLLiCiTiT:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

It would be good idea to store information about representation file to the file value.

### What we usually do miss
- frame it represents
- frame it has in filename, in some cases this might be different e.g. traypublisher but we need the frame e.g. for ffmpeg processing -> maybe we can find better way how to use the filepath for ffmpeg and in that case we might not need the information?
- udim value
- ??staging dir/filepath??
- ??file template?? `sh0100_0100.exr` -> `sh0100_{frame:0>4}.exr` same for udim. This might be too much... Or we might change `file_frame` to `str` to avoid the `0>4` (but udims are still `int`)?
- ..there might be more (modify the description or write a comment to add more)

This is also connected to multiple other issues, e.g. single frame sequences.

### How would you imagine the implementation of the feature?

Because `"files"` has to be `list[str]` or `str` we could introduce new `str` type that has the additional attributes available to not break backwards compatibility. The best would be to use dataclasses but I can't imagine how would the backwards compatibility lookl like.

At first stages we should start set the values in a host of our choice, and use it in some general plugin to validate if it makes sense and is actually usefull. We might find out that the `file_frame` doesn't make sense because we can handle things other way?

Sketch of the type
```python
class RepreFile(str):
    # File template with frame and udim keys
    template: str
    # Frame value in filename ('0001' in 'test.0001.exr')
    frame: str | None = None
    # Frame it represents during publishing (not related to filename 1001)
    repre_frame: int | None = None
    # Udim it represents during publishing
    udim: str | None = None

    def __new__(cls, *args, **kwargs):
        output = super().__new__(cls, *args, **kwargs)
        output.template = str(output)
        return output

    @classmethod
        def create_frame(
        cls,
        filename: str,
        file_frame: str,
        udim: str | None = None,
        repre_frame: int | None = None,
        template: str | None = None,
    ) -> "RepreFile":
        if repre_frame is None:
            repre_frame = int(file_frame)

        if template is None:
            if len(filename.split(file_frame)) > 2:
                raise ValueError(
                    f"Can't auto-fill template: {filename}."
                    f" Frame '{file_frame}' appears more than once"
                    f" in filename."
                )

            template = filename.replace(file_frame, "{frame}")
            if udim is not None:
                if len(template.split(udim)) > 2:
[...]

Issue #1855, opened 2026-05-22
prepare_representations: clique.assemble() O(N²) hang on large frame ranges (50k frames -> 12h CPU burn)


funkygfx79:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior:

`ayon_core.pipeline.farm.pyblish_functions.prepare_representations` calls `clique.assemble(exp_files)` (line 429) on the complete list of expected render frames. With large frame counts and long file paths, clique's sequence-discovery scales super-linearly and can hang for **hours**.

In our case a 50,000-frame After Effects render submission spent **~12h22m of pure CPU time** inside `clique.assemble` before `ProcessSubmittedJobOnFarm` could post the dependent publish job to Deadline.
(Same workfile, same comp, with the fast-path workaround applied: full publish (both Deadline jobs submitted) completes in ~10 seconds.)

External symptoms while hanging:
- AE WebSocket bridge keeps pinging every 30s — Python is alive, MainThread holds GIL but does no I/O
- No HTTP traffic in `urllib3` logs (nothing logged in any plugin)
- No DCC RPC calls
- Pyblish Publisher UI shows the plugin running indefinitely with no progress
- Once the hang ends, all subsequent plugins finish in seconds

py-spy dump captured during the hang (MainThread, `active+gil`):

​```
assemble (clique\__init__.py:131)
__init__ (clique\collection.py:50)
_update_expression (clique\collection.py:79)
compile (re.py:252)
_compile (re.py:304)
compile (sre_compile.py:788)
parse (sre_parse.py:955)
_parse_sub (sre_parse.py:444)
_parse (sre_parse.py:841)
_parse_sub (sre_parse.py:444)
_parse (sre_parse.py:846)
prepare_representations (ayon_core\pipeline\farm\pyblish_functions.py:429)
process (submit_publish_job.py:362)
​```

A second dump minutes later showed the stack still inside `clique.assemble` but at line 149 of `clique\__init__.py` — the function makes progress, it's not deadlocked, it's genuinely doing O(N²) work.

A 60s flamegraph confirmed ~100% CPU time inside `clique.assemble:147` / `:149` and `Collection.head` (`clique\collection.py:58`).

### Expected Behavior:

`prepare_representations` should complete in seconds, not hours, regardless of frame count. The expected-files list passed in is, by construction, a clean, contiguous, uniformly-padded sequence — `clique.assemble`'s discovery work is unnecessary for this shape of input.

[...]

PR #1973, opened 2026-07-27
Implementation of representation dataclass

moonyuet:
## Changelog Description
This PR is to implement representation dataclass to deal with the files data type issue.
```python
from ayon_core.pipeline.publish import Representation
representation_data = {
    "name": "exr",
    "exr" : "exr",
    "files": files,
    # or more....
}
representation = Representation()
representation.add_representation(representation_data)
```

## Additional info
This is implemented from the ongoing discussion. Very experimental. We would change back to the initial method by https://github.com/ynput/ayon-core/issues/1950. 

I dont know the changes would be good idea as it vastly lowers down the readability for the backward compatibility.
Links https://github.com/ynput/ayon-core/issues/1855 and Related to https://github.com/ynput/ayon-core/issues/1950

## Testing notes:
1. start with this step
2. follow this step


client/ayon_core/pipeline/publish/representation.py
```diff
@@ -0,0 +1,112 @@
...
+        key (str): Key to get value from.
+        default (Any): Default value if key/attribute is missing.
+
+    Returns:
+        Any: Value from representation.
+    """
+    if isinstance(repre, dict):
+        return repre.get(key, default)
+
+    return getattr(repre, key, default)
```
BigRoy: Shouldn't we be more explicit?

```
    if isinstance(repre, dict):
        return repre.get(key, default)
    elif isinstance(repre, Representation):
        return getattr(repre, key, default)
    raise TypeError(...)
```

Same for `repre_set`.

client/ayon_core/pipeline/publish/representation.py
```diff
@@ -0,0 +1,112 @@
...
+        ...
+
+    def set_sequence(self, files: list[str]) -> None:
+        """Set sequence of files for representation.
+
+        Args:
+            files (list[str]): List of file paths.
+        """
+        if len(files) == 1:
+            self.set_single_file(files[0])
```
BigRoy, line 50: This is wrong. The whole point is to support single frame sequences... and not enforce them to be "a single file"  as `str` so that we can differentiate between what is a single frame sequence and what is a single file.

Also, I don't see the `Sequence[str]` handled separately, as such I think the typing you're trying to do is `files: list[str] | Sequence[str]` and isn't `list[str]` actually a `Sequence[str]` to begin with?

I actually think we can fix the whole `list[str]` for single frame already, without having to refactor to the whole dataclass stuff to begin with. 🤔 Looking at this PR, that may make more sense to start with.

client/ayon_core/pipeline/publish/representation.py
```diff
@@ -0,0 +1,112 @@
...
+        else:
+            self.files = list(files)
+            self.sequence_files = list(files)
+
+    def set_colorspace_data(self, colorspaceData: dict) -> None:
```
BigRoy, line 55: This should really be more explicit about what data is supposed to be in `colorspaceData` - a `dict` is way too generic.

BigRoy (commented):
Oh boy - we may need to just discuss this @iLLiCiTiT before continuing more efforts here.

client/ayon_core/pipeline/publish/representation.py
moonyuet: I am okay with that but does it mean we also need to convert the colorspaceData as dataclass?

moonyuet:
> Oh boy - we may need to just discuss this @iLLiCiTiT before continuing more efforts here.

I actually agree with this, It is just way too much changes I am not sure if it is worth for that.

iLLiCiTiT:
If this PR is to solve the sequence issues, then representation class won't solve anything. We still have exactly same issue, there are files and we don't know anything about it. If we don't know anything about it, it is just guess game that might or might not work.

I honestly do believe that only correct fix is to store more information about files.

Considering representation as a class, that is different discussion, I think that needs much more discussion to think about some of the data we do store there. Clean it up a little, some of that is also related to the files themselved, because if we do store udim/frame etc. to the files then we don't need that information on representation. It is also something that will take months to fully process, it is as with product base types. It should be as backwards compatible as possible, but with the enhancements in mind, which this PR is not, this just changed keys to attributes -> that's not helpfull or solving any issue.

NOTE: There already is `Representation` class for traits...

moonyuet:
> If this PR is to solve the sequence issues, then representation class won't solve anything. We still have exactly same issue, there are files and we don't know anything about it. If we don't know anything about it, it is just guess game that might or might not work.
> 
> I honestly do believe that only correct fix is to store more information about files.
> 
> Considering representation as a class, that is different discussion, I think that needs much more discussion to think about some of the data we do store there. Clean it up a little, some of that is also related to the files themselved, because if we do store udim/frame etc. to the files then we don't need that information on representation. It is also something that will take months to fully process, it is as with product base types. It should be as backwards compatible as possible, but with the enhancements in mind, which this PR is not, this just changed keys to attributes -> that's not helpfull or solving any issue.
> 
> NOTE: There already is `Representation` class for traits...

I am thinking if it is the best-case to use the traits which is initiated by @antirotor. If there are some cravats, we can redesign or refactor it instead.  @BigRoy @iLLiCiTiT @antirotor thoughts?

If I dont get it wrong, the representation trait is only used in marvelous designer.

iLLiCiTiT:
I'd like to say yes, but I don't know how do traits solve any of the issues we have with files. We can store additional information to the representation, but not to the files themselves. It is nice to know that I have representation with sequence `1001,1003-1004,1010` but no information about the filename template, frame padding in the file, or the case that I have some frame range in filename but different frame range it does represent on instance, or it has udim. So I still have to use clique to guess it and do a lot of guessing work to find out what real frame it is for the current case. Traits is good concept for integration, not for in-process data.

PR #1976, opened 2026-07-28
Loader: Use selected rows for representations

iLLiCiTiT:
## Changelog Description
Use `selectedRows` instead of `selectedIndexes` to get selections.

## Additional info
Selected indexes does return all indexes from different columns (one row can return multiple indexes), using `selectedRows` is used the same but only one index per row.

## Testing notes:
1. Actions on representations do work no matter which column is used for selection.


BigRoy (approved):
Didn't test it (sorry) but code changes make sense. So ✅ 👍

BigRoy (approved):
Tested. Works fine, ready to merge.

PR #1978, opened 2026-07-30
Fix: Don't process hero version without trait representations.

antirotor:
## Changelog Description
Fix cases where *Integrate Hero Versions with traits* was triggered every time, messing with regular hero version integration.

## Testing notes:
Publishing of regular hero version should work.

### DIFF
--- a/client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
+++ b/client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
@@ -113,6 +113,22 @@ class IntegrateHeroVersionTraits(
                 "traits. Skipping")
             return
 
+        # don't allow both representations
+        # with traits and standard representations
+        if (
+                has_trait_representations(instance)
+                and instance.data.get("representations")
+        ):
+            msg = (
+                f"Instance '{instance.name}' has representations with traits "
+                "but also has standard representations. This is not allowed. "
+                "Please use either representations with traits or "
+                "standard representations, not both."
+            )
+            raise PublishError(
+                msg
+            )
+
         anatomy: Anatomy = instance.context.data["anatomy"]
         project_name = anatomy.project_name
 
--- a/client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
+++ b/client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
@@ -115,10 +115,7 @@ class IntegrateHeroVersionTraits(
 
         # don't allow both representations
         # with traits and standard representations
-        if (
-                has_trait_representations(instance)
-                and instance.data.get("representations")
-        ):
+        if instance.data.get("representations"):
             msg = (
                 f"Instance '{instance.name}' has representations with traits "
                 "but also has standard representations. This is not allowed. "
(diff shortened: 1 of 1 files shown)


LiborBatek (approved):
It works like charm...`Hero version` been published while `Integrate hero version with repre traits` been gracefully skipped.

LGTM

Also tested and loaded the resulting hero product back into scene. All good!

client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
```diff
@@ -106,6 +107,12 @@ def process(self, instance: pyblish.api.Instance) -> None:
         if not self.is_active(instance.data):
             return
 
+        if not has_trait_representations(instance):
```
BigRoy, line 110: Can we add a comment above this that this still assumes that an instance (and with that a version) will never mix representations with traits with those without traits. Because then it'd still do the wrong thing, right? If we were to mix, it would still cause the non-traits representations to get wiped away?

BigRoy (approved):
I think this is fine for now, But I did add a comment that technically this isn't safe and sound if we mix the representations. So maybe we'll want to elaborate that in a comment as described?

client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
colinhacks: Fixed in `29d9318c`: the OSS application path still repairs authoritative owner account type, but no longer starts an organization Pro trial. Trial creation remains at genuine Pullfrog installation/runtime entry points, so a repo maintainer cannot consume the organization’s one-time trial before install.

BigRoy: Wtf? :)

antirotor: wtf indeed. Anyway I've added error to stop the integration when both types are used. I think it is safer that way (even though you cannot undone results of the first integration because publishing isn't atomic)

client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
```diff
@@ -106,6 +107,28 @@ def process(self, instance: pyblish.api.Instance) -> None:
...
+                f"Instance '{instance.name}' has no representations with "
+                "traits. Skipping")
+            return
+
+        # don't allow both representations
+        # with traits and standard representations
+        if (
+                has_trait_representations(instance)
+                and instance.data.get("representations")
+        ):
```
BigRoy: No need to check has trait representations again?
```suggestion
        # don't allow both representations
        # with traits and standard representations
        elif instance.data.get("representations"):
```

client/ayon_core/plugins/publish/integrate_hero_version_with_traits.py
```diff
@@ -106,6 +107,28 @@ def process(self, instance: pyblish.api.Instance) -> None:
...
+            msg = (
+                f"Instance '{instance.name}' has representations with traits "
+                "but also has standard representations. This is not allowed. "
+                "Please use either representations with traits or "
+                "standard representations, not both."
+            )
+            raise PublishError(
+                msg
+            )
```
BigRoy, line 127:
```suggestion
            raise PublishError(
                f"Instance '{instance.name}' has representations with traits "
                "but also has standard representations. This is not allowed. "
                "Please use either representations with traits or "
                "standard representations, not both."
            )
```

antirotor: This is because of https://docs.astral.sh/ruff/rules/raw-string-in-exception/

PR #1983, opened 2026-08-05
Handle missing colorspace from display and view

jakubjezek001:
## Summary
Derive the colorspace template value from `display` and `view` when `colorspace` is unavailable, with a `view`-only fallback.

## Testing
ideally test with https://github.com/ynput/ayon-nuke/pull/347

### DIFF
--- a/client/ayon_core/plugins/publish/integrate.py
+++ b/client/ayon_core/plugins/publish/integrate.py
@@ -623,7 +623,13 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
 
         # add template data for colorspaceData
         if repre.get("colorspaceData"):
-            colorspace = repre["colorspaceData"]["colorspace"]
+            colorspace = repre["colorspaceData"].get("colorspace")
+            display = repre["colorspaceData"].get("display")
+            view = repre["colorspaceData"].get("view")
+            if not colorspace and (display and view):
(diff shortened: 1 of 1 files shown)


client/ayon_core/plugins/publish/integrate.py
iLLiCiTiT, line 637:
```suggestion
        repre_colorspace_data = repre.get("colorspaceData")
        if repre_colorspace_data:
            colorspace = repre_colorspace_data.get("colorspace")
            if not colorspace:
                display = repre_colorspace_data.get("display")
                view = repre_colorspace_data.get("view")
                if display and view:
                    colorspace = f"{display}_{view}"
                elif view:
                    colorspace = view

            # replace spaces with underscores
            # pipeline.colorspace.parse_colorspace_from_filepath
            # is checking it with underscores too
            if colorspace:
                colorspace = colorspace.replace(" ", "_")
```

BigRoy:
This is incorrect, right? As far as I know we can't just assume display+view is also settable as colorspace.

Display/View != Colorspace.

iLLiCiTiT:
If this is needed then we should require to also fill `colorpsace` in the data, or?

And actually [this](https://github.com/ynput/ayon-core/pull/1983#issuecomment-5191801441) is good question.

PR #1985, opened 2026-08-06
Explore here action fix Windows behavior when path has spaces.

BigRoy:
## Changelog Description

Explore here action fix Windows behavior when path has spaces.

## Additional info

The `start` command allows to pass the window title and then the app to launch. When a path with spaces was passed then Windows started to think the value was supposed to be the title because it was quoted instead of the executable. By forcefully passing the initial quotes Windows never makes the wrong assumption.

Reported here: https://community.ynput.io/t/explore-here-not-working-with-custom-path/3247

## Testing notes:

1. Explore here should work on path with spaces

### DIFF
--- a/client/ayon_core/plugins/actions/open_file_explorer.py
+++ b/client/ayon_core/plugins/actions/open_file_explorer.py
@@ -97,7 +97,8 @@ class OpenTaskPath(LauncherAction):
     def open_in_explorer(path):
         platform_name = platform.system().lower()
         if platform_name == "windows":
-            args = ["start", "", path]
+            os.startfile(path)
+            return
         elif platform_name == "darwin":
             args = ["open", "-R", path]
         elif platform_name == "linux":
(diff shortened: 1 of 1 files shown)


iLLiCiTiT (approved):

iLLiCiTiT:
We could just use `explorer.exe`...

BigRoy:
> We could just use `explorer.exe`...

I believe it's the case that if you have a custom app as 'explorer' for file paths or equivalent that it may work with `start` and not with `explorer` but I'm not entirely sure. But that thinking was the sole reason I decided not to change it around.

iLLiCiTiT:
Or `os.startfile(path)` which is using windows native functions (and is windows only).

BigRoy:
> Or `os.startfile(path)` which is using windows native functions (and is windows only).

Done with https://github.com/ynput/ayon-core/pull/1985/commits/d3465a2c20622875f8e0f3dcacdad8f558da551f

PR #1986, opened 2026-08-06
Move scene version collector earlier

moonyuet:
## Changelog Description
Move `CollectSceneVersion` earlier regarding to https://github.com/ynput/ayon-aftereffects/pull/90

## Additional info
Draft PR for ongoing discussion.

## Testing notes:
1. Add the host into `ayon+settings://core/publish/CollectSceneVersion` if it is not the default host.
2. Testing publishing or any kind

### DIFF
--- a/client/ayon_core/plugins/publish/collect_scene_version.py
+++ b/client/ayon_core/plugins/publish/collect_scene_version.py
@@ -11,7 +11,7 @@ class CollectSceneVersion(pyblish.api.ContextPlugin):
         version (int, optional): version number of the publish
     """
 
-    order = pyblish.api.CollectorOrder - 0.45
+    order = pyblish.api.CollectorOrder - 0.5
     label = "Collect Scene Version"
     # configurable in Settings
     hosts = ["*"]
--- a/client/ayon_core/plugins/publish/collect_scene_version.py
+++ b/client/ayon_core/plugins/publish/collect_scene_version.py
@@ -11,7 +11,7 @@ class CollectSceneVersion(pyblish.api.ContextPlugin):
         version (int, optional): version number of the publish
     """
 
-    order = pyblish.api.CollectorOrder - 0.5
+    order = pyblish.api.CollectorOrder - 0.39
     label = "Collect Scene Version"
     # configurable in Settings
     hosts = ["*"]
--- a/client/ayon_core/plugins/publish/collect_scene_version.py
+++ b/client/ayon_core/plugins/publish/collect_scene_version.py
@@ -11,7 +11,7 @@ class CollectSceneVersion(pyblish.api.ContextPlugin):
         version (int, optional): version number of the publish
     """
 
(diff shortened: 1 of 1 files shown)


BigRoy:
Note that -0.45 would break things for: TVPaint, Zbrush, Flame, Fusion, ComfyUI. See the list here: https://github.com/ynput/ayon-aftereffects/pull/90#discussion_r3459786967

moonyuet:
> Note that -0.45 would break things for: TVPaint, Zbrush, Flame, Fusion, ComfyUI. See the list here: [ynput/ayon-aftereffects#90 (comment)](https://github.com/ynput/ayon-aftereffects/pull/90#discussion_r3459786967)

[5bf0447](https://github.com/ynput/ayon-core/pull/1986/commits/5bf04475d14c1defb2f28ad18bcf740d460929be)

BigRoy:
> Note that -0.45 would break things for: TVPaint, Zbrush, Flame, Fusion, ComfyUI. See the list here: [ynput/ayon-aftereffects#90 (comment)](https://github.com/ynput/ayon-aftereffects/pull/90#discussion_r3459786967)

Based on this comment @moonyuet it should've been later, not earlier ;)

As stated there:
> **Meaning that we can nudge CollectSceneVersion to -0.39**. Yet I also think pushing those host that can to -0.5 or at least -0.49 could be nice, just so CollectSceneVersion could perhaps be fine at -0.45

Until we start moving the `currentFile` data to be collected earlier in those DCCs.

iLLiCiTiT:
If I understand correcly, this PR was added because the AE plugin does use `currentFile` and `version`? Is it used for any logic, or just as an "metadata"? If it is used as a metadata then just fill it later during publishing...

We need created instances with filled product base type, basic families and correct folder and task. Everything else can be collected later.

moonyuet:
> If I understand correcly, this PR was added because the AE plugin does use `currentFile` and `version`? Is it used for any logic, or just as an "metadata"? If it is used as a metadata then just fill it later during publishing...

Yes It is moved earlier because "version" from `CollectAERender` is using "version" from context data and they use it to create `AERenderInstance`

iLLiCiTiT:
> Yes It is moved earlier because "version" from CollectAERender is using "version" from context data and they use it to create AERenderInstance

So it is used only as a metadata and does not really have to filled in the plugin.

moonyuet:
> > Yes It is moved earlier because "version" from CollectAERender is using "version" from context data and they use it to create AERenderInstance
> 
> So it is used only as a metadata and does not really have to filled in the plugin.

I would say yes. It is only used for filling version attribute lasted in RenderInstance. If we dont fill in the version data I assume it would be errored out for RenderInstance as the version attribute doesn't default to any value.
Guess it would be good enough to just move the ordering from `CollectRender` in AE side? @BigRoy @iLLiCiTiT

PR #1990, opened 2026-08-06
Tools: Update and enhance abstract methods

iLLiCiTiT:
## Changelog Description
Fill missing abstract methods for tools and change how task sorting is received from controller.

## Additional info
Some abstract classes were missing abstract methods that are mandatory for tools to work.

Changed how task sorting is received from controller, use dedicated method instead of getting full settings dictionary from the controller. This step was made so we do avoid creating copy of huge dictionary just to get boolean (with [this](https://github.com/ynput/ayon-core/pull/1987) in mind).

## Testing notes:
All tools can be showed, are working and tasks are sorted based on sorting from settings.
- [ ] Context selection - No idea where this is used
- [x] Launcher
- [x] Loader
- [x] Publisher
- [x] Workfiles


kalisp:
Fiddled with Settings in:
- Launcher - OK
- Publisher - OK
- Browser - didn't react on Settings changes
- Workfiles - OK

Screenshot 2026-08-07 162102

image

(sorted alphabetically, not by task types)

iLLiCiTiT:
> Browser - didn't react on Settings changes

Should be fixed now.

kalisp (approved):
Works in Browser too.

PR #1991, opened 2026-08-07
Backward compatibility of build template without profile won't return preset path None.

moonyuet:
## Changelog Description
This PR is to make sure the backward compatibility of build template won't trigger the template-related error below when there is no profile 
Links to https://github.com/ynput/ayon-core/pull/1907
```
#   File "D:\ayon-core\client\ayon_core\lib\events.py", line 338, in process_event
#     callback()
#     ~~~~~~~~^^
#   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\maya_0.6.11+dev\ayon_maya\api\pipeline.py", line 703, in on_new
#     create_first_workfile_template()
#     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
#   File "C:\Users\Kayla\AppData\Local\Ynput\AYON\addons\maya_0.6.11+dev\ayon_maya\api\workfile_template_builder.py", line 254, in create_first_workfile_template
#     builder.build_template(workfile_creation_enabled=True)
#     ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "D:\ayon-core\client\ayon_core\pipeline\workfile\workfile_template_builder.py", line 598, in inner
#     return self.old_build_template(*args, **kwargs)
#            ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
#   File "D:\ayon-core\client\ayon_core\pipeline\workfile\workfile_template_builder.py", line 732, in old_build_template
#     raise TemplateLoadFailed(
#         f"Template path '{preset.path}' does not exist."
#     )
# ayon_core.pipeline.workfile.workfile_template_builder.TemplateLoadFailed: Template path 'None' does not exist.
```

## Additional info

## Testing notes:
1. Launch Nuke and Houdini with this built branch in Core
2. The error log above should be gone

### DIFF
--- a/client/ayon_core/pipeline/workfile/workfile_template_builder.py
+++ b/client/ayon_core/pipeline/workfile/workfile_template_builder.py
@@ -729,7 +729,7 @@ class AbstractTemplateBuilder(ABC):
         ):
             preset = self.get_template_preset()
             # Path may e.g. be none if there's no active workfile template
-            # set for current text, which would indicate it's just disabled.
+            # set for current context, which would indicate it's just disabled.
             # In that case, do nothing
             if preset.path is None:
                 return
--- a/client/ayon_core/pipeline/workfile/workfile_template_builder.py
+++ b/client/ayon_core/pipeline/workfile/workfile_template_builder.py
@@ -151,7 +151,7 @@ class TemplatePreset:
             bool: Preset has set path to an existing file.
 
         """
-        return self.path and os.path.exists(self.path)
+        return self.path is not None and os.path.exists(self.path)
 
 
 class AbstractTemplateBuilder(ABC):
@@ -729,8 +729,9 @@ class AbstractTemplateBuilder(ABC):
         ):
             preset = self.get_template_preset()
             if not preset.has_valid_path():
-                return
-
+                raise TemplateLoadFailed(
+                    f"Template path '{preset.path}' does not exist."
+                )
             template_path: str = preset.path
             if keep_placeholders is None:
                 keep_placeholders: bool = preset.keep_placeholder
--- a/client/ayon_core/pipeline/workfile/workfile_template_builder.py
+++ b/client/ayon_core/pipeline/workfile/workfile_template_builder.py
@@ -728,10 +728,10 @@ class AbstractTemplateBuilder(ABC):
             or create_first_version is None
         ):
             preset = self.get_template_preset()
-            if not preset.has_valid_path():
-                if preset.path is None:
-                    return
+            if preset.path is None:
+                return
 
+            if not preset.has_valid_path():
                 raise TemplateLoadFailed(
                     f"Template path '{preset.path}' does not exist."
                 )
--- a/client/ayon_core/pipeline/workfile/workfile_template_builder.py
+++ b/client/ayon_core/pipeline/workfile/workfile_template_builder.py
@@ -729,6 +729,8 @@ class AbstractTemplateBuilder(ABC):
         ):
             preset = self.get_template_preset()
             if not preset.has_valid_path():
+                self.log.info(
+                    f"Template path '{preset.path}' does not exist. ")
                 return
             template_path: str = preset.path
             if keep_placeholders is None:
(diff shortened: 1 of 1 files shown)


antirotor:
It doesn't crash anymore which is fine! One bit:

```
>>> [ Template path 'None' does not exist.  ] 
```

I would handle that error message more user friendly :)

if it is `None`, it shouldn't even run the check

BigRoy:
Agree with @antirotor here. If there is not preset path, which essentially means there's no matching profile? Then it means profile template build is _actually_ disabled completely and we don't need to log anything. We could debug log that "Workfile template build is disabled." but admittedly, even that's quite verbose.

antirotor:
but thinking of it, I don't think the whole stuff should even run by default, no? I would think that if I have this in settings:
image
it should not invoke any template builder logic at all.

BigRoy:
> but thinking of it, I don't think the whole stuff should even run by default, no? I would think that if I have this in settings: image it should not invoke any template builder logic at all.

The DCCs are triggering the workfile template builder for the core template builder to choose whether it should be running or not, based on the settings. So:

- DCC says: Hello, this is the moment workfile template build can run if it needs to.
- Core replies with: OK, nothing to do because it's disabled.
- OR Core replies with: There's a matching profile, will build it! 

So we do 'trigger' the workfile template builder in core, for the core logic to then tell us whether it should be running. Right? 🤔 And if it shouldn't be running, there's nothing to log really.

moonyuet:
@BigRoy @antirotor I have just removed the log message. [aa78c94](https://github.com/ynput/ayon-core/pull/1991/commits/aa78c94e4db7edba68732f8fcb95164fbfd79b3c)

BigRoy:
You now removed the error if there is an actual path set but it's not valid 🤔 which I think is a valid case to error on?

moonyuet:
> You now removed the error if there is an actual path set but it's not valid 🤔 which I think is a valid case to error on?

Friday Brain, commit twice: [2ece055](https://github.com/ynput/ayon-core/pull/1991/commits/2ece055722b2c56a94fdee065a976d4630c09fbb) It should only raise exception when preset.path is not None and not exists

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -729,6 +729,9 @@ def old_build_template(
         ):
             preset = self.get_template_preset()
             if not preset.has_valid_path():
+                if preset.path is None:
```
BigRoy: Does this need clarification?
```suggestion
                # Path may e.g. be none if there's no active workfile template
                # set for current text, which would indicate it's just disabled.
                # In that case, do nothing
                if preset.path is None:
```

BigRoy (commented):
LGTM. Needs testing in the different DCCs with older code.

May need a comment to explain the reason why we ignore it.

Also,. @moonyuet should we do this `if preset.path is None` check even before we do `if not preset.has_valid_path()` because in practice it'll always be invalid if it's `None`?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
moonyuet: We should move before `.has_valid_path()`

moonyuet:
@antirotor can you test?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -728,6 +728,12 @@ def old_build_template(
             or create_first_version is None
         ):
             preset = self.get_template_preset()
+            # Path may e.g. be none if there's no active workfile template
+            # set for current text, which would indicate it's just disabled.
```
BigRoy:
```suggestion
            # set for current context, which would indicate it's just disabled.
```

jakubjezek001 (approved):
works for me in Nuke

antirotor (approved):
did quick test in Nuke and Houdini and it doesn't crash as before

PR #1992, opened 2026-08-07
Integrator: Refactor to `ContextPlugin` and add type hints

BigRoy:
## Changelog Description

Integrator: Refactor to `ContextPlugin` and add type hints

## Additional info

Turning this into an integrator allows us to optimize some queries, like the prefetching of products, version data, etc. and also join the file transactions into one threaded bulk and the OperationsSession into fewer bulks. This should make big ingests with many instances faster. But timings will still need to be made.

**Draft for now, because it needs more performance tests but also general discussion**

## Testing notes:
1. Publishing should work from all DCCs

### DIFF
--- a/client/ayon_core/plugins/publish/integrate.py
+++ b/client/ayon_core/plugins/publish/integrate.py
@@ -67,6 +67,7 @@ class _PreparedRepresentation:
 class _PreparedInstance:
     instance: pyblish.api.Instance
     prepared_representations: list[_PreparedRepresentation]
+    operations: OperationsSession
     existing_repres_by_name: dict[str, dict[str, Any]]
     resource_destinations: set[str]
     anatomy: Anatomy
@@ -213,18 +214,17 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
             log=self.log,
             allow_queue_replacements=False,
         )
-        operations = OperationsSession()
         try:
             prepared_instances = self._prepare_instances(
-                context, instances, operations, file_transactions
+                context, instances, file_transactions
             )
-            operations.commit()
+            self._commit_instance_operations(prepared_instances)
             self.log.debug("Integrating source files to destination ...")
             file_transactions.process()
             self._prepare_representation_operations(
-                context, prepared_instances, operations
+                context, prepared_instances
             )
-            operations.commit()
+            self._commit_instance_operations(prepared_instances)
         except DuplicateDestinationError as exc:
             file_transactions.rollback()
             raise PublishError(str(exc)).with_traceback(sys.exc_info()[2])
@@ -235,9 +235,35 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
 
         file_transactions.finalize()
 
-    def _commit_if_full(self, operations: OperationsSession) -> None:
-        if len(operations) >= self.operations_chunk_size:
-            operations.commit()
+    def _commit_instance_operations(
+        self, prepared_instances: list[_PreparedInstance]
+    ) -> None:
+        """Commit all operations sessions in chunks to avoid payload size
+        issues. However, enforce that all operations of a single instance are
+        committed together to avoid partial commits.
+        """
+        combined = OperationsSession()
+        for prepared_instance in prepared_instances:
+            operations = prepared_instance.operations
+            if not len(operations):
+                continue
+
+            # If the single instance has more operations than the chunk
+            # size, commit it directly
+            if len(operations) > self.operations_chunk_size:
+                operations.commit()
+                continue
+
+            # If the combined session is too large, commit existing stack of
+            # operations before adding new operations
+            if (
+                len(combined) + len(operations) > self.operations_chunk_size
+            ):
+                combined.commit()
+
+            combined.extend(operations._operations)
+            operations.clear()
+        combined.commit()
 
     def _prepare_instances(
         self,
@@ -245,7 +271,6 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
         instances: list[tuple[
             pyblish.api.Instance, list[PublishRepresentation]
         ]],
-        operations: OperationsSession,
         file_transactions: FileTransaction,
     ) -> list[_PreparedInstance]:
         project_name = context.data["projectName"]
@@ -264,20 +289,29 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
             for product in products
         }
 
+        # Keep track of operations per instance, so if we need to chunk the
+        # OperationsSession we can ensure to still commit all operations for a
+        # single instance together
+        operations_by_instance_id: dict[str, OperationsSession] = {
+            instance.id: OperationsSession()
+            for instance, _ in instances
+        }
         product_entities = {}
+        product_operations: dict[tuple[str, str], OperationsSession] = {}
         for instance, _ in instances:
             key = (
                 instance.data["folderEntity"]["id"],
                 instance.data["productName"],
             )
             if key not in product_entities:
+                operations = operations_by_instance_id[instance.id]
                 product_entities[key] = self.prepare_product(
                     instance,
                     operations,
                     project_name,
                     existing_product_entity=products_by_key.get(key),
                 )
-                self._commit_if_full(operations)
+                product_operations[key] = operations
 
         product_ids = {product["id"] for product in products}
         version_numbers = {
@@ -304,6 +338,7 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
             product_entity = product_entities[product_key]
             version_key = (product_entity["id"], instance.data["version"])
             if version_key not in version_entities:
+                operations = product_operations[product_key]
                 version_entities[version_key] = self.prepare_version(
                     instance,
                     operations,
@@ -311,7 +346,6 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
                     project_name,
                     existing_version=versions_by_key.get(version_key),
                 )
-                self._commit_if_full(operations)
 
         existing_version_ids = {version["id"] for version in versions}
         representations = []
@@ -343,6 +377,7 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
                 filtered_repres,
                 file_transactions,
                 repres_by_version.get(version_entity["id"], {}),
+                operations_by_instance_id[instance.id],
             ))
         return prepared_instances
 
@@ -352,6 +387,7 @@ class IntegrateAsset(pyblish.api.ContextPlugin):
         filtered_repres: list[PublishRepresentation],
         file_transactions: FileTransaction,
         existing_repres_by_name: dict[str, dict[str, Any]],
+        operations: OperationsSession,
     ) -> _PreparedInstance:
         instance_stagingdir = instance.data.get("stagingDir")
         template_name = self.get_template_name(instance)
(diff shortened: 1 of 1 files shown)


BigRoy:
@iLLiCiTiT can you give me your thoughts on this?

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -800,11 +910,13 @@ def prepare_representation(
             # NOTE: Frame would not be available only if template does not
             #   contain '{frame}' in template -> Do we want support it?
             if not is_udim:
+                assert repre_context is not None
```
iLLiCiTiT, line 912: Why are the asserts here?

BigRoy: They are just to please the IDE's linter. Because if not `repre_context` may be `None` further down and linter will be like "hey, this _could_ be `None`" because `files` may have been empty.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -134,254 +186,305 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
...
         file_transactions.finalize()
 
-    def filter_representations(self, instance):
-        """Filter representations to be integrated."""
-        repres = instance.data.get("representations")
-        if not repres:
-            return []
+    def _commit_if_full(self, operations: OperationsSession) -> None:
+        if len(operations) >= self.operations_chunk_size:
+            operations.commit()
```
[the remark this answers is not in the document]
BigRoy: I actually think the whole idea of the chunking should be avoidable by just allowing the OperationsSession to handle massive datasets. The danger is that you may try to commit 10000s of versions and representations, and suddenly we're sending a massive JSON to the server when we are trying to initiate the background operation. @iLLiCiTiT any ideas or thoughts?

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -736,6 +844,7 @@ def prepare_representation(
                     repre_context = dst.used_values
 
             if not is_udim and first_index_padded is not None:
+                assert repre_context is not None
                 repre_context["frame"] = first_index_padded
```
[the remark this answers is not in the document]
BigRoy: They are just to please the IDE's linter honestly.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -134,254 +186,305 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
...
-        # Finalizing can't rollback safely so no use for moving it to
-        # the try, except.
         file_transactions.finalize()
 
-    def filter_representations(self, instance):
-        """Filter representations to be integrated."""
-        repres = instance.data.get("representations")
-        if not repres:
-            return []
+    def _commit_if_full(self, operations: OperationsSession) -> None:
```
iLLiCiTiT, line 238: Could we instead collect the operations per instance and then make sure one instance operations are sent in one batch?
```python
instance_operations = [
    # Instance A
    [ProductOp, VersionOp, RepreOp1, RepreOp2],
    # Instance B
    [ProductOp, VersionOp, RepreOp1, RepreOp2, RepreOp3, RepreOp4, RepreOp5, RepreOp6],
    ...
]

operations = OperationsSession()
for instance_ops in instance_opeartions:
    if len(operations) + len(instance_ops) > self.operations_chunk_size:
        operations.commit()

    for op in instance_ops:
        operations.add(op)

if len(operations) > 0:
    operations.commit()    
```

client/ayon_core/plugins/publish/integrate.py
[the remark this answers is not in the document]
iLLiCiTiT: You're right but I don't think that is a job of `OperationsSession` to handle it. It can happen that the operations are related to each other and chunking that automatically is not good idea. If you submit 2 background operations and the 2nd batch is based on changes from the 1st batch you might hit a race condition issues.

client/ayon_core/plugins/publish/integrate.py
iLLiCiTiT: Keeping the asserts is dangerous and confusing. We should check it once and raise PublishError (if that makes IDE happy).

BigRoy: I'll tweak them then.

client/ayon_core/plugins/publish/integrate.py
BigRoy: I think we can yes. 🤔 Might complicate the code even further hehe.

iLLiCiTiT: We can create `OperationsSession` per instance, and them merge it at the end.

BigRoy: https://github.com/ynput/ayon-core/pull/1992/commits/6c889b83853a7b79f57615ed9d224d8b6f61c349

client/ayon_core/plugins/publish/integrate.py
BigRoy: https://github.com/ynput/ayon-core/pull/1992/commits/2d748f569787b1aff28fee639e9990ebd01049e2

BigRoy:
May be good to do some performance testing with this. Maybe @antirotor loves to do that, or @jakubjezek001 comparing it with latest `develop` on a large CSV ingest or alike? :)

moonyuet:
I agree that we need to do some performance testing and my concern is whether prefetching data in this PR rapidly increasing memory usage. 

1. Multiple instances
2. Single very large dataset (CSV ingest etc.)
3. Extremely large instance count

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -4,14 +4,15 @@
 import logging
 import sys
 import copy
+from dataclasses import dataclass
```
iLLiCiTiT, line 7: Dataclasses can't be used. Same issue as here https://github.com/ynput/ayon-core/pull/2050

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -134,254 +187,337 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
...
             )
 
-        # Filter representations
-        filtered_repres = []
-        for repre in repres:
-            if "delete" in repre.get("tags", []):
-                continue
-            filtered_repres.append(repre)
+        products = list(get_products(
+            project_name, names_by_folder_ids=names_by_folder_ids
```
iLLiCiTiT, line 285: I might be wrong, but it looks like we need only id, folder id and name?
```suggestion
            project_name,
            names_by_folder_ids=names_by_folder_ids,
            fields={"id", "folderId", "name"},
```

iLLiCiTiT: Not true, we need full body.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -134,254 +187,337 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
...
-                "@ {0}".format(instance_stagingdir)
+        existing_version_ids = {version["id"] for version in versions}
+        representations = []
+        if existing_version_ids:
+            representations = list(get_representations(
+                project_name, version_ids=existing_version_ids
+            ))
+        repres_by_version: dict[
+            str, dict[str, dict[str, Any]]
+        ] = {}
```
iLLiCiTiT:
```suggestion
        ] = defaultdict(dict)
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -134,254 +187,337 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
...
+        for repre in representations:
+            repres_by_version.setdefault(repre["versionId"], {})[
+                repre["name"].lower()
+            ] = repre
```
iLLiCiTiT:
```suggestion
            version_id = repre["versionId"]
            repre_name = repre["name"].lower()
            repres_by_version[version_id][repre_name] = repre
```

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -134,254 +187,337 @@ class IntegrateAsset(pyblish.api.InstancePlugin):
...
+        for instance, filtered_repres in instances:
+            product_key = (
+                instance.data["folderEntity"]["id"],
+                instance.data["productName"],
             )
+            version_entity = version_entities[(
+                product_entities[product_key]["id"],
+                instance.data["version"],
+            )]
+            instance.data["versionEntity"] = version_entity
```
iLLiCiTiT:
```suggestion
            product_entity = product_entities[product_key]
            version_key = (product_entity["id"], instance.data["version"])
            version_entity = version_entities[version_key]
            instance.data["versionEntity"] = version_entity
```

client/ayon_core/plugins/publish/integrate.py
iLLiCiTiT, line 931: NOTE: In ideal case scenario we should collect file info now. The `template_filled` object has rootless path with root that was 100% used instead of guessing it later on.

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -891,29 +1040,31 @@ def prepare_representation(
...
+        if existing is not None:
             update_data = prepare_changes(existing, repre_doc)
 
-        return {
-            "representation": repre_doc,
-            "repre_update_data": update_data,
-            "anatomy_data": template_data,
-            "transfers": transfers,
+        return _PreparedRepresentation(
+            representation=repre_doc,
```
iLLiCiTiT, line 1054: Why the hell is the variable still named `repre_doc`? 😆

client/ayon_core/plugins/publish/integrate.py
```diff
@@ -891,29 +1040,31 @@ def prepare_representation(
...
+            repre_update_data=update_data,
+            anatomy_data=template_data,
+            transfers=transfers,
             # todo: avoid the need for 'published_files' used by Integrate Hero
             # backwards compatibility
-            "published_files": [transfer[1] for transfer in transfers]
-        }
+            published_files=[transfer[1] for transfer in transfers],
```
iLLiCiTiT:
```suggestion
            published_files=[dst_path for _, dst_path in transfers],
```

client/ayon_core/plugins/publish/integrate.py
BigRoy: They can, if you don't have `from __future__ import annotations` in the file. ;)
As such, do we still care?

Also, given how painful this is it makes me wonder if we should have a `dataclass` decorator wrapper that avoids the bug, just so we can use dataclasses. Because this stuff gets painful easily otherwise.

iLLiCiTiT: I have it in my TODO list to fix the issue.

BigRoy: It's fixed in `pyblish`  https://github.com/pyblish/pyblish-base/pull/403 btw ;) there's just no new pyblish release since.

iLLiCiTiT: BTW I think `from __future__ import annotations` is issue with core's import logic, but with pyblish import logic it fails completelly event without the import.

BigRoy: Pretty sure the pyblish one also ONLY happened if that from future import was there. @antirotor worked on that fix at the time.

iLLiCiTiT: Added PR https://github.com/ynput/ayon-core/pull/2062 to fix the dataclasses issue.

iLLiCiTiT: > Pretty sure the pyblish one also ONLY happened if that from future import was there. @antirotor worked on that fix at the time.

Oh, yeah, but that is not released, and 100% is not in any dependency package we have. So in practice we can't rely on it.

PR #1996, opened 2026-08-12
Use enum instead of bool for USD contribution variant

### LINKED ISSUE #1666: YN-0448: USD contribution: set as default variant allow to not set at all

Currently in the AYON USD Contribution workflow there's an instance attribute for "Set as default variant selection" that will mark that variant as the active variant for the asset. However, when this setting is disabled - it will still set it as the default variant selection **IF** currently there is not actively set variant by default. However, there are cases where you may want to author a variant but explicitly not mark it as the active by default.

  

As such, most likely the better case here would be to have an enum that allows you to:

- Do not set
- Set as default if no current default
- Set as default

  

Where the 2nd option matches the current behavior, and hence for backwards compatibility potentially should default to that? 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1465709289980297298)

_This issue was automatically created from AYON ticket [YN-0448]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=dc129010fc2c11f0927a07090ab79798)

tadeas-hejnic:
## Changelog Description
This PR is adding an option to `Set as default variant selection` during the publishing. Before there was only toggle that has this behaviour:
     - `True` : Always set the variant as default
     - `False` : Set as the default only if there is no default
     
So it wasn't possible to `Don't set the variant as default if the default is not set yet`. The current implementation is adding this option and maps also the previous values to the new enum. 

## Additional info
The conversion is done this way to keep the behaviour of the previous implementation the same:
1. `False` -> `Set as default if no current default`
2. `True` -> `Set as default`

The new option is `Do not set`, which really doesn't set the default even if there is no default yet.

## Testing notes:
1. test publish with all possible choices in `Set as default variant selection` and check if it matches the settings
2. test conversion from older settings:
    - `False` -> `Set as default if no current default`
    - `True` -> `Set as default`

### DIFF
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -835,12 +835,10 @@ class ExtractUSDLayerContribution(publish.Extractor):
                 # Set default variant selection
                 variant_set_name = contribution.variant_set_name
                 variant_name = contribution.variant_name
-                if (
-                    contribution.variant_default_policy
-                    == "always"
+                policy = contribution.variant_default_policy
+                if (policy == "always"
                     or (
-                        contribution.variant_default_policy
-                        == "if_missing"
+                        policy == "if_missing"
                         and variant_set_name not in prim_spec.variantSelections
                     )
                 ):
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -361,16 +361,6 @@ class CollectUSDLayerContributions(pyblish.api.InstancePlugin,
             variant_default_policy = attr_values.get(
                 "contribution_variant_default_policy"
             )
-            if variant_default_policy is None:
-                # Support data created before the boolean was replaced by the
-                # policy enum.
-                variant_default_policy = (
-                    "always"
-                    if attr_values.get(
-                        "contribution_variant_is_default", False
-                    )
-                    else "if_not_set"
-                )
 
             contribution = VariantContribution(
                 instance=instance,
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -358,9 +358,8 @@ class CollectUSDLayerContributions(pyblish.api.InstancePlugin,
         # Define contribution
         in_layer_order: int = attr_values.get("contribution_in_layer_order", 0)
         if attr_values["contribution_apply_as_variant"]:
-            variant_default_policy = attr_values.get(
-                "contribution_variant_default_policy"
-            )
+            variant_default_policy = attr_values[
+                "contribution_variant_default_policy"]
 
             contribution = VariantContribution(
                 instance=instance,
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -4,7 +4,7 @@ import os
 import platform
 from collections import defaultdict
 from operator import attrgetter
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Literal
 
 import pyblish.api
 try:
@@ -74,7 +74,9 @@ class VariantContribution(_BaseContribution):
     # Variant
     variant_set_name: str
     variant_name: str
-    variant_default_policy: str  # Policy controlling variant selection opinion
+    variant_default_policy: Literal[
+        "if_not_set", "always", "never"
+    ]  # Policy controlling variant selection opinion
 
 
 CONTRIBUTION_VARIANT_DEFAULT_POLICY = {
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -362,13 +362,13 @@ class CollectUSDLayerContributions(pyblish.api.InstancePlugin,
             if variant_default_policy is None:
                 # Support data created before the boolean was replaced by the
                 # policy enum.
-                variant_default_policy = CONTRIBUTION_VARIANT_DEFAULT_POLICY[
+                variant_default_policy = (
                     "always"
                     if attr_values.get(
                         "contribution_variant_is_default", False
                     )
                     else "if_missing"
-                ]
+                )
 
             contribution = VariantContribution(
                 instance=instance,
@@ -567,18 +567,16 @@ class CollectUSDLayerContributions(pyblish.api.InstancePlugin,
                 "contribution_variant_set_name": "{layer}",
                 "contribution_variant": "{variant}",
                 "contribution_variant_default_policy":
-                    CONTRIBUTION_VARIANT_DEFAULT_POLICY["if_missing"],
+                    "if_missing",
             }
         elif "contribution_variant_default_policy" not in profile:
             # Support profiles created before the boolean was replaced by the
             # policy enum.
             profile = profile.copy()
             profile["contribution_variant_default_policy"] = (
-                CONTRIBUTION_VARIANT_DEFAULT_POLICY[
-                    "always"
-                    if profile.get("contribution_variant_is_default", False)
-                    else "if_missing"
-                ]
+                "always"
+                if profile.get("contribution_variant_is_default", False)
+                else "if_missing"
             )
 
         # Define defaults
@@ -706,10 +704,10 @@ class CollectUSDLayerContributions(pyblish.api.InstancePlugin,
                     f"'{CONTRIBUTION_VARIANT_DEFAULT_POLICY['always']}', "
                     "the final result depends on their contribution order."
                 ),
-                items=list(CONTRIBUTION_VARIANT_DEFAULT_POLICY.values()),
+                items=CONTRIBUTION_VARIANT_DEFAULT_POLICY,
                 default=profile.get(
                     "contribution_variant_default_policy",
-                    CONTRIBUTION_VARIANT_DEFAULT_POLICY["if_missing"],
+                    "if_missing",
                 ),
                 visible=variant_visible,
             ),
@@ -839,10 +837,10 @@ class ExtractUSDLayerContribution(publish.Extractor):
                 variant_name = contribution.variant_name
                 if (
                     contribution.variant_default_policy
-                    == CONTRIBUTION_VARIANT_DEFAULT_POLICY["always"]
+                    == "always"
                     or (
                         contribution.variant_default_policy
-                        == CONTRIBUTION_VARIANT_DEFAULT_POLICY["if_missing"]
+                        == "if_missing"
                         and variant_set_name not in prim_spec.variantSelections
                     )
                 ):
--- a/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
+++ b/client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
@@ -575,6 +575,7 @@ class CollectUSDLayerContributions(pyblish.api.InstancePlugin,
         publish_attributes = instance["publish_attributes"].get(
             cls.__name__, {})
 
+        # Convert legacy attribute value
         if "contribution_variant_is_default" in publish_attributes:
             contribution_variant_is_default = publish_attributes.pop(
                 "contribution_variant_is_default"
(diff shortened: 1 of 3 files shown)


server/settings/publish_plugins.py
```diff
@@ -1704,7 +1726,8 @@ class PublishPuginsModel(BaseSettingsModel):
                 "contribution_apply_as_variant": False,
                 "contribution_variant_set_name": "{layer}",
                 "contribution_variant": "{variant}",
-                "contribution_variant_is_default": False,
+                "contribution_variant_default_policy":
+                    CONTRIBUTION_VARIANT_DEFAULT_POLICY["if_missing"],
```
BigRoy: This is wrong, right - it should just be "if_missing". The default value is not the label, but the value... and the value is `if_missing`?

server/settings/publish_plugins.py
```diff
@@ -261,16 +276,19 @@ class CollectUSDLayerContributionsProfileModel(BaseSettingsModel):
             "The default variant name for instances matching this profile."
         ),
     )
-    contribution_variant_is_default: bool = SettingsField(
-        False,
+    contribution_variant_default_policy: str = SettingsField(
+        CONTRIBUTION_VARIANT_DEFAULT_POLICY["if_missing"],
```
BigRoy: Same as https://github.com/ynput/ayon-core/pull/1996/changes#r3766588397

server/settings/conversion.py
```diff
@@ -287,11 +290,36 @@ def _convert_oiio_transcode_0_4_5(publish_overrides):
...
+
+        if "contribution_variant_is_default" not in profile:
+            continue
+
+        is_default = profile.pop("contribution_variant_is_default")
+        profile["contribution_variant_default_policy"] = (
+            CONTRIBUTION_VARIANT_DEFAULT_POLICY[
+                "always" if is_default else "if_missing"
+            ]
+        )
```
BigRoy, line 311: The value is set directly, not by label. Right?
```suggestion
        profile["contribution_variant_default_policy"] = (
            "always" if is_default else "if_missing"
        )
```

server/settings/publish_plugins.py
tadeas-hejnic: Yes, you are right. It's working now, but it's not stored as it should be... I will fix it on all places in next commit

tadeas-hejnic:
### My test workflow

Tested all combinations of the new default variant selection policy, both with and without an existing variant selection.

- **`never` with no existing selection**
  - Published a new variant to a clean target.
  - Verified that the variant was added.
  - Verified that no variant selection was authored.

- **`if_not_set` with no existing selection**
  - Published a new variant to a clean target.
  - Verified that the variant was added.
  - Verified that it became the selected/default variant.

- **`always` with no existing selection**
  - Published a new variant to a clean target.
  - Verified that the variant was added.
  - Verified that it became the selected/default variant.

- **`never` with an existing selection**
  - Published an additional variant to a target that already had `Base` selected.
  - Verified that the new variant was added.
  - Verified that the existing `Base` selection was preserved.

- **`if_not_set` with an existing selection**
  - Published an additional variant to the same target.
  - Verified that the new variant was added.
  - Verified that the existing `Base` selection was preserved.

- **`always` with an existing selection**
  - Published an additional variant to the same target.
  - Verified that the new variant was added.
  - Verified that the existing selection was overridden and changed to the newly published variant.

All six policy scenarios behaved as expected.

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -74,7 +74,14 @@ class VariantContribution(_BaseContribution):
     # Variant
     variant_set_name: str
     variant_name: str
-    variant_is_default: bool  # Whether to author variant selection opinion
+    variant_default_policy: str  # Policy controlling variant selection opinion
```
BigRoy: Can we give this a `Literal` type hint? `Literal["if_missing", "always", "never"]`

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -790,8 +835,15 @@ def process(self, instance):
...
-                        variant_set_name not in prim_spec.variantSelections:
+                if (
+                    contribution.variant_default_policy
+                    == "always"
+                    or (
+                        contribution.variant_default_policy
+                        == "if_missing"
+                        and variant_set_name not in prim_spec.variantSelections
+                    )
+                ):
```
BigRoy, line 839: is this more readable, or worse?
```suggestion
                policy = contribution.variant_default_policy
                if (policy == "always"
                    or (
                        policy == "if_missing"
                        and variant_set_name not in prim_spec.variantSelections
                    )
                ):
```

BigRoy (commented):
Cool, very solid!

Should we rename key `if_missing` to `if_not_set`?

@MustafaJafar can you give it a spin?

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
tadeas-hejnic: Yes, it's a bit better

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
tadeas-hejnic: Done

server/settings/conversion.py
```diff
@@ -287,11 +287,34 @@ def _convert_oiio_transcode_0_4_5(publish_overrides):
             }
 
 
+def _convert_usd_contribution_variant_default_policy(overrides):
+    """Convert the USD contribution default boolean to a policy enum."""
+    profiles = (
+        overrides
+        .get("publish", {})
+        .get("CollectUSDLayerContributions", {})
+        .get("profiles", [])
```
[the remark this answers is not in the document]
tadeas-hejnic: I don't think it's necessary...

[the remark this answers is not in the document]
BigRoy: Agreed

MustafaJafar (approved):
It works on my side as in https://github.com/ynput/ayon-core/pull/1996#issuecomment-5267728923
Also, copy settings work.

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -349,13 +358,27 @@ def process(self, instance):
...
+            if variant_default_policy is None:
+                # Support data created before the boolean was replaced by the
+                # policy enum.
+                variant_default_policy = (
+                    "always"
+                    if attr_values.get(
+                        "contribution_variant_is_default", False
+                    )
+                    else "if_not_set"
+                )
```
BigRoy: With the publish plug-in now using a new attribute name. Would the `process` ever end up getting the old value still?

@iLLiCiTiT shouldn't we somehow be triggering a conversion on the plug-in to take whatever old value there was an push it directly into the new attribute value? Is that possible from a plug-in like this?

BigRoy (commented):
@iLLiCiTiT can you do code review and merge if happy. Please double check my question comment.

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
iLLiCiTiT: > With the publish plug-in now using a new attribute name. Would the process ever end up getting the old value still?

I don't think so.

On publish plugin the conversion can happen during `get_attr_defs_for_instance`.

BigRoy: @tadeas-hejnic could you look into that and see also if needed?

tadeas-hejnic: It's already done in `get_attr_defs_for_instance`, so I will remove it here.

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -544,8 +558,18 @@ def get_attr_defs_for_instance(cls, create_context, instance):
...
             }
+        elif "contribution_variant_default_policy" not in profile:
+            # Support profiles created before the boolean was replaced by the
+            # policy enum.
+            profile = profile.copy()
+            profile["contribution_variant_default_policy"] = (
+                "always"
+                if profile.get("contribution_variant_is_default", False)
+                else "if_not_set"
+            )
```
iLLiCiTiT: Always expect the settings are using correct model. It is really hard to maintain old settings conversions in the codebase.
```suggestion
```

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -790,8 +827,13 @@ def process(self, instance):
                 # Set default variant selection
                 variant_set_name = contribution.variant_set_name
                 variant_name = contribution.variant_name
-                if contribution.variant_is_default or \
-                        variant_set_name not in prim_spec.variantSelections:
+                policy = contribution.variant_default_policy
+                if (policy == "always"
```
iLLiCiTiT:
```suggestion
                if (
                    policy == "always"
```

server/settings/conversion.py
```diff
@@ -287,11 +287,34 @@ def _convert_oiio_transcode_0_4_5(publish_overrides):
...
+    )
+
```
iLLiCiTiT:
```suggestion
        .get("profiles")
    )
    if not profiles:
        return

```

server/settings/conversion.py
```diff
@@ -287,11 +287,34 @@ def _convert_oiio_transcode_0_4_5(publish_overrides):
...
+            "always" if is_default else "if_not_set"
+        )
+
+
 def _convert_publish_plugins(overrides):
     if "publish" not in overrides:
         return
     _convert_validate_version_0_3_3(overrides["publish"])
     _convert_oiio_transcode_0_4_5(overrides["publish"])
+    _convert_usd_contribution_variant_default_policy(overrides)
```
iLLiCiTiT:
```suggestion
    _convert_usd_contribution_variant_default_policy_1_9_11(overrides)
```

server/settings/conversion.py
```diff
@@ -287,11 +287,34 @@ def _convert_oiio_transcode_0_4_5(publish_overrides):
             }
 
 
+def _convert_usd_contribution_variant_default_policy(overrides):
```
iLLiCiTiT:
```suggestion
def _convert_usd_contribution_variant_default_policy_1_9_11(overrides):
```

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
iLLiCiTiT, line 588: This is converting values of instances to the new value
```suggestion
            if "contribution_variant_is_default" in publish_attributes:
                contribution_variant_is_default = publish_attributes.pop(
                    "contribution_variant_is_default"
                )

                publish_attributes["contribution_variant_default_policy"] = (
                    "always"
                    if contribution_variant_is_default
                    else "if_not_set"
                )

```

tadeas-hejnic:
Tested after the changes suggested by @iLLiCiTiT and everything works.

iLLiCiTiT (approved):

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -349,13 +358,17 @@ def process(self, instance):
         # Define contribution
         in_layer_order: int = attr_values.get("contribution_in_layer_order", 0)
         if attr_values["contribution_apply_as_variant"]:
+            variant_default_policy = attr_values.get(
+                "contribution_variant_default_policy"
+            )
```
BigRoy: This using `.get` technically means it doesn't adhere to the `Literal` type hint because it's not allowed to be `None`. This `.get()` should either fall back to a default value or not use `.get` but directly index the attribute.

client/ayon_core/plugins/publish/extract_usd_layer_contributions.py
```diff
@@ -560,6 +575,17 @@ def get_attr_defs_for_instance(cls, create_context, instance):
         publish_attributes = instance["publish_attributes"].get(
             cls.__name__, {})
 
+        if "contribution_variant_is_default" in publish_attributes:
```
BigRoy, line 578: Can we add a comment here?

```suggestion
        # Convert legacy attribute value
        if "contribution_variant_is_default" in publish_attributes:
```

BigRoy (approved):
LGTM - just some cosmetic notes only left.

iLLiCiTiT (approved):

PR #1998, opened 2026-08-12
Optimization: use `str.format_map(data)` over `str.format(**data)`

BigRoy:
## Changelog Description

Optimization: use `str.format_map(data)` over `str.format(**data)`

## Additional info

Double check whether we're only doing these to `str` and not to any of our custom "template" classes that do not inherit from `str`.

Here's some numbers:
```
=== Small dict (2 keys) ===
format(**data)   : 0.4824s for 1,000,000 calls  (482.4 ns/call)
format_map(data) : 0.3197s for 1,000,000 calls  (319.7 ns/call)

=== Large dict (52 keys) ===
format(**data)   : 0.8103s for 1,000,000 calls  (810.3 ns/call)
format_map(data) : 0.4332s for 1,000,000 calls  (433.2 ns/call)

=== Repeat runs for consistency (large dict) ===
format(**data)    run0: 0.7982s for 1,000,000 calls  (798.2 ns/call)
format_map(data)  run0: 0.4288s for 1,000,000 calls  (428.8 ns/call)
format(**data)    run1: 0.8166s for 1,000,000 calls  (816.6 ns/call)
format_map(data)  run1: 0.4307s for 1,000,000 calls  (430.7 ns/call)
format(**data)    run2: 0.8488s for 1,000,000 calls  (848.8 ns/call)
format_map(data)  run2: 0.4439s for 1,000,000 calls  (443.9 ns/call)
```
We're talking very small differences. Do note that for small dict the old way seems that it's technically slightly faster.
However, it should also use less memory, less garbage collection, etc. so there's other benefits too.

However, given this ☝️ we may want to not swap the `{root` ones because it has a single key. So those are now written as `format(root=...)` to avoid even having to create the `dict`, etc.

## Testing notes:

1. Test general pipeline usage.
2. Test publishing.
3. Test extract review with burnins.

### DIFF
--- a/client/ayon_core/lib/path_templates.py
+++ b/client/ayon_core/lib/path_templates.py
@@ -760,7 +760,7 @@ class FormattingPart:
             parent_fill_data[used_keys[-1]] = value
 
         template = f"{{{field_name}{self._format_spec}{self._conversion}}}"
-        formatted_value = template.format(**root_fill_data)
+        formatted_value = template.format_map(root_fill_data)
         used_key = key
         if keys_to_value is not None:
             used_key = self.keys_to_template_base(keys_to_value)
--- a/client/ayon_core/lib/path_templates.py
+++ b/client/ayon_core/lib/path_templates.py
@@ -191,6 +191,19 @@ class StringTemplate:
             invalid_types
         )
 
+    def format_map(self, data: dict[str, Any]) -> "TemplateResult":
+        """Format the template using a mapping of replacement values.
+
+        This mirrors :meth:`str.format_map`.
+
+        Args:
+            data (dict): Containing keys to be filled into the template.
+
+        Returns:
+            TemplateResult: Filled or partially filled template.
+        """
+        return self.format(data)
+
     def format_strict(self, data: dict[str, Any]) -> "TemplateResult":
         result = self.format(data)
         result.validate()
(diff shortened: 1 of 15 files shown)


client/ayon_core/lib/path_templates.py
iLLiCiTiT, line 1: We should implement `format_map` on `StringTemplate` too, so if you work with `StringTemplate` object you actually get the same result. I think you just broke all the places that are using `StringTemplate` (or anatomy subclass).

BigRoy: [`StringTemplate`'s `format` method](https://github.com/ynput/ayon-backend/blob/d42f863b02d3b143a880bd7743a7e464f92c25b6/api/resolve/templating.py#L519) actually seems to work already exactly like `format_map` and not like `format`.

So are you saying to just do `format_map = format` on the `StringTemplate` or just doing:
```python
    def format_map(self, data: dict[str, Any]) -> "TemplateResult":
        return self.format(data)
```

Because it already behaves differently - I'm not too keen on adding `format_map` next to it 🤔 Unless we're thinking of deprecating the old, etc. etc.?

iLLiCiTiT: > So are you saying to just do format_map = format on the StringTemplate or just doing:

I would just do what you did in the example, yes.

Few of the changed places do work with `StringTemplate` and calling `format_map` on them will actually use the default str implementation not returning `TemplateResult`.

BigRoy: https://github.com/ynput/ayon-core/pull/1998/commits/28a417b4a7996c887d6ed180e8a41754b45170ca

iLLiCiTiT (approved):

PR #2004, opened 2026-08-18
YN-1000 - Fix audio offset ffmpeg filter on extract_review.

rdelillo:
## Changelog Description

Changes:
* [x] Fix dead code for audio offset in `extract_review` plugins
* [x] Fix positive audio offset using [`adelay` filter](https://stackoverflow.com/questions/52446539/how-to-add-offset-or-delay-to-audio-file-with-ffmpeg) instead of `-itsoffset` which is for video 

## Testing notes:
1. Create a review instance with an audio offset (positive or negative)
2. Ensure the audio track of the resulting review media match audio offset.

### DIFF
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1304,21 +1304,24 @@ class ExtractReview(pyblish.api.InstancePlugin):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
-            offset_seconds = audio.get("offset") or 0
+            offset_seconds = audio.get("offset", 0)
+            # Delay audio start to match its timeline position.
             if offset_seconds > 0:
+                # "all=1" applies to every channel (mono, stereo, 5.1…).
                 delay_ms = int(round(offset_seconds * 1000))
-                audio_filters.append("adelay={}|{}".format(delay_ms, delay_ms))
-
+                audio_filters.append(
+                    "adelay={}:all=1".format(delay_ms)
+                )
+            # Crop audio file input.
             elif offset_seconds < 0:
                 audio_in_args.append(
                     "-ss {}".format(abs(offset_seconds))
                 )
 
-            # Audio duration is offset from `-ss`
-            audio_duration = duration_seconds + offset_seconds
-
-            # Set audio duration
-            audio_in_args.append("-to {:0.10f}".format(audio_duration))
+            # Audio duration
+            audio_in_args.append(
+                "-to {:0.10f}".format(duration_seconds - offset_seconds)
+            )
 
             # Ignore video data from audio input
             audio_in_args.append("-vn")
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1304,6 +1304,9 @@ class ExtractReview(pyblish.api.InstancePlugin):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
+            # 'offset' was set only by maya addon and it used frame offset
+            #     instead of seconds offset. To prevent issues the value is
+            #     ignored and logged as skipped.
             if audio.get("offset"):
                 self.log.warning(
                     "Ignored deprecated audio input attribute 'offset'. "
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1338,6 +1338,23 @@ class ExtractReview(pyblish.api.InstancePlugin):
             audio_out_args.append("-filter_complex amerge")
             audio_out_args.append("-ac {}".format(len(audio_inputs)))
 
+        # Resample audio to ensure block duration from AAC
+        # encoder never exceeds one video frame duration.
+        # Pick the lowest video standard rate satisfying this constraint.
+        # For non-AAC codecs, this has no negative effect.
+
+        # TODO: Replace with once https://trac.ffmpeg.org/ticket/11668 is fixed
+        # audio_filters.append("apad")
+        # audio_out_args.append("-shortest")
+
+        standard_audio_rates = (48000, 96000, 192000)
+        min_audio_rate = 1024 * temp_data.fps  # 1 AAC block = 1024 samples
+        sample_rate = next(
+            (rate for rate in standard_audio_rates if rate > min_audio_rate),
+            standard_audio_rates[-1]
+        )
+        audio_out_args.append("-ar {}".format(sample_rate))
+
         return audio_in_args, audio_filters, audio_out_args
 
     def get_letterbox_filters(
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1304,7 +1304,7 @@ class ExtractReview(pyblish.api.InstancePlugin):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
-            if not audio.get("offset"):
+            if audio.get("offset"):
                 self.log.warning(
                     "Ignored deprecated audio input attribute 'offset'. "
                     "Use 'offset_in_seconds' instead."
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1304,15 +1304,14 @@ class ExtractReview(pyblish.api.InstancePlugin):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
-            offset_seconds = 0
+            offset_seconds = audio.get("offset") or 0
             if offset_seconds > 0:
-                audio_in_args.append(
-                    "-ss {}".format(offset_seconds)
-                )
+                delay_ms = int(round(offset_seconds * 1000))
+                audio_filters.append("adelay={}|{}".format(delay_ms, delay_ms))
 
             elif offset_seconds < 0:
                 audio_in_args.append(
-                    "-itsoffset {}".format(abs(offset_seconds))
+                    "-ss {}".format(abs(offset_seconds))
                 )
 
             # Audio duration is offset from `-ss`
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1359,7 +1359,7 @@ class ExtractReview(pyblish.api.InstancePlugin):
             (rate for rate in standard_audio_rates if rate > min_audio_rate),
             standard_audio_rates[-1]
         )
-        audio_out_args.append("-ar {}".format(sample_rate))
+        audio_out_args.append(f"-ar {sample_rate}")
 
         return audio_in_args, audio_filters, audio_out_args
 
--- a/client/ayon_core/plugins/publish/extract_review.py
+++ b/client/ayon_core/plugins/publish/extract_review.py
@@ -1304,7 +1304,13 @@ class ExtractReview(pyblish.api.InstancePlugin):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
-            offset_seconds = audio.get("offset", 0)
+            if not audio.get("offset"):
+                self.log.warning(
+                    "Ignored deprecated audio input attribute 'offset'. "
+                    "Use 'offset_in_seconds' instead."
+                )
+
+            offset_seconds = audio.get("offset_in_seconds", 0)
             # Delay audio start to match its timeline position.
             if offset_seconds > 0:
                 # "all=1" applies to every channel (mono, stereo, 5.1…).


client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1304,15 +1304,14 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
...
 
         for audio in audio_inputs:
-            offset_seconds = 0
+            offset_seconds = audio.get("offset") or 0
             if offset_seconds > 0:
-                audio_in_args.append(
-                    "-ss {}".format(offset_seconds)
-                )
+                delay_ms = int(round(offset_seconds * 1000))
+                audio_filters.append("adelay={}|{}".format(delay_ms, delay_ms))
```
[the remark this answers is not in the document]
rdelillo: Fixed.

moonyuet (approved):
Tested successfully in maya 2027 and it is published with the audio correctly to the server

https://github.com/user-attachments/assets/090c1538-9da8-406f-bd29-0996130e2599

BigRoy:
> Tested successfully in maya 2027 and it is published with the audio correctly to the server
> 
>  Recording.2026-08-19.163922.mp4

@moonyuet did you specifically test offsetting the audio track by e.g. 10 frames ahead, publish - then publish with 10 frames subtracted

- [x] no offset
- [x] test with positive time offset in maya
- [ ] test with negative time offset in maya

Audio track in the published video should match exactly the point in time from within Maya.
can you share publish reports for these three cases.

You may need to test with an audio file that is not monotone ;) so you can actually check it's sync as intended.

moonyuet:
I used another audio example for testing, still not a good one.
I hit some issue when testing the negative offset. The sounds of the review from the server is the same as the maya time editor shows.

https://github.com/user-attachments/assets/afd14a56-538a-4ac3-b292-bd6bf8f2cf22
[publish-report-260819-18-36.json](https://github.com/user-attachments/files/31219449/publish-report-260819-18-36.json)
I also hit some issue when publishing the offset. it is adding the total frames. 
I have -5 as offset and then 30 frames for publishing, the total frames are 33(should be 30).


https://github.com/user-attachments/assets/87c4b93e-0d09-44e6-87c8-b949f7922ca3


Same thing to the positive offset, I have 20 frames for publishing and have positive offsets but there is one extra frame added to total frame.

rdelillo:
@moonyuet thanks for your feedback, could you please check my fix ?

I found out why it did not work on your case.
The audio file you've used was `11025Hz` which is way lower the standard `48000Hz`.
```
Input #1, wav, from 'D:/sh_boss_proj/interior_lenny/lib/work/art/Time Editor/Clip Exports/carpentry.wav':
  Duration: 00:00:09.13, bitrate: 88 kb/s
  Stream #1:0: Audio: pcm_u8 ([1][0][0][0] / 0x0001), 11025 Hz, mono, u8, 88 kb/s
  ```
  
What happened is that the AAC encoding was forcing the audio bitrate at 192Kbps with **`-b:a 192k`** while keeping the original audio sample rate of `11KHz`. With such a low rate, 1 AAC block was more than 1 frame duration hence the additional "audio-only" frames in the resulting MP4 container.

Ultimately it's great we've spotted that.
The fix I've added enforce an audio resample to a minimal 48Khz standard audio rate for video while adjusting it for very hight fps.

**Review by moonyuet — CHANGES_REQUESTED** [review]
I published no offset and negative offset successfully.
But I fail to publish the positive offset.

[publish-report-260820-15-03.json](https://github.com/user-attachments/files/31253171/publish-report-260820-15-03.json)

**Comment by rdelillo** [comment]
> I published no offset and negative offset successfully. But I fail to publish the positive offset.
> 
> [publish-report-260820-15-03.json](https://github.com/user-attachments/files/31253171/publish-report-260820-15-03.json)

@moonyuet 
Can you confirmed this was tested with https://github.com/ynput/ayon-maya/pull/462 ?
It feels to me it might have not.
The FFmpeg command line is wrong because the audio offset was provided as frame number (1006) instead of offset in second (1006-1001/25.0) = 0.2secs which is exactly the purpose of the maya PR.

See: `"adelay=1006000:all=1"` instead of `"adelay=2000:all=1"`
```
C:\Users\Kayla\AppData\Local\Ynput\AYON\addons_resources\ayon_third_party\ffmpeg_windows_3d2b8241\bin\ffmpeg.exe -apply_trc gamma22 -start_number 1001 -framerate 25.0 -to 0.8000000000 -i C:\Users\Kayla\AppData\Local\Temp\ay_tmp_m23kb5lj\reviewAudioTest.%d.png -1005.2000000000 -vn -i "D:/sh_boss_proj/interior_lenny/lib/work/art/Time Editor/Clip Exports/carpentry.wav" -filter:a "adelay=1006000:all=1" -pix_fmt yuv420p -crf 18 -c:a aac -b:a 192k -g 1 -movflags faststart -color_range tv -colorspace bt709 -color_primaries bt709 -color_trc bt709 -ar 48000 -y C:\Users\Kayla\AppData\Local\Temp\ay_tmp_m23kb5lj\reviewAudioTest_h264.mp4
```

**Review by moonyuet — APPROVED** [review]
It is working now for all three scenarios, my bad as I am at the wrong branch:)

**Comment by BigRoy** [comment]
> Can you confirmed this was tested with [ynput/ayon-maya#462](https://github.com/ynput/ayon-maya/pull/462) ?
> It feels to me it might have not.
> The FFmpeg command line is wrong because the audio offset was provided as frame number (1006) instead of offset in second (1006-1001/25.0) = 0.2secs which is exactly the purpose of the maya PR.

If this fails with older ayon-maya or vice versa we may need to set up package requirements for these new releases between core/maya.

@iLLiCiTiT 

**Or** @rdelillo we may want to indeed change the data model to make it more explicit these are seconds, and essentially retain old logic if older data (or warn and skip?), but use the seconds if it's the newly supported data?

so essentially old maya integration would send old data, we'd warn about it... new ayon-maya would send new data which this new core knows how to handle?

**Comment by rdelillo** [comment]
> **Or** @rdelillo we may want to indeed change the data model to make it more explicit these are seconds, and essentially retain old logic if older data (or warn and skip?), but use the seconds if it's the newly supported data?
> 
> so essentially old maya integration would send old data, we'd warn about it... new ayon-maya would send new data which this new core knows how to handle?

The `offset` audio attribute is deadcode in current `ayon_core` (does nothing).
I have made the fix used a new `offset_in_seconds` in `ayon_maya` and `ayon_core`:
* hopefully `offset_in_seconds` is clearer than `offset`
* added a warning about deprecated `offset` attribute
* this remaing backward-compatible, if any addon version mismatch it does nothing as before

Would that work for you ?

File: client/ayon_core/plugins/publish/extract_review.py [hunk 2]
```diff
@@ -1304,22 +1304,30 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
-            offset_seconds = 0
-            if offset_seconds > 0:
-                audio_in_args.append(
-                    "-ss {}".format(offset_seconds)
+            if not audio.get("offset"):
```
BigRoy: Shouldn't this be reversed?
```suggestion
            if audio.get("offset"):
```

BigRoy (commented):
Aside of the one remark, looks ok to me.

@iLLiCiTiT thoughts?

- `offset` was swapped with `offset_in_seconds` to avoid anything that would be passing the old data to suddenly be thought of as being seconds (previously they were frames e.g. from maya). 
- The old code was actually dead code because it hardcoded `offset_seconds = 0`.
- It makes more sense to support as seconds for audio tracks than frames.

client/ayon_core/plugins/publish/extract_review.py
rdelillo: Totally, thanks for that !

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1304,22 +1304,30 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
...
                 )
 
-            # Audio duration is offset from `-ss`
-            audio_duration = duration_seconds + offset_seconds
-
-            # Set audio duration
-            audio_in_args.append("-to {:0.10f}".format(audio_duration))
+            # Audio duration
+            audio_in_args.append(
+                "-to {:0.10f}".format(duration_seconds - offset_seconds)
```
iLLiCiTiT, line 1332: If `offset_seconds` is negative then this will make it longer...

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1304,22 +1304,30 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
...
-            if offset_seconds > 0:
-                audio_in_args.append(
-                    "-ss {}".format(offset_seconds)
+            if audio.get("offset"):
+                self.log.warning(
+                    "Ignored deprecated audio input attribute 'offset'. "
+                    "Use 'offset_in_seconds' instead."
                 )
 
+            offset_seconds = audio.get("offset_in_seconds", 0)
```
iLLiCiTiT, line 1316: Why don't keep both?
```suggestion
            offset_seconds = audio.get("offset_in_seconds")
            if offset_seconds is None:
                offset_seconds = audio.get("offset")
                if offset_seconds is not None:
                     self.log.warning(
                        "Ignored deprecated audio input attribute 'offset'. "
                        "Use 'offset_in_seconds' instead."
                    )

            if offset_seconds is None:
                offset_seconds = 0.0
```

Can be moved to dedicated method
```python
    def _get_audio_offset(self, audio: dict[str, Any]) -> float:
        offset_seconds = audio.get("offset_in_seconds")
        if offset_seconds is not None:
            return offset_seconds

        offset_seconds = audio.get("offset")
        if offset_seconds is not None:
            self.log.warning(
                "Using deprecated audio input attribute 'offset'. "
                "Use 'offset_in_seconds' instead."
            )
            return offset_seconds
        return 0.0
```

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1304,22 +1304,30 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
...
+                # "all=1" applies to every channel (mono, stereo, 5.1…).
+                delay_ms = int(round(offset_seconds * 1000))
+                audio_filters.append(
+                    "adelay={}:all=1".format(delay_ms)
+                )
+            # Crop audio file input.
             elif offset_seconds < 0:
                 audio_in_args.append(
-                    "-itsoffset {}".format(abs(offset_seconds))
+                    "-ss {}".format(abs(offset_seconds))
```
iLLiCiTiT, line 1327: I do remember that this caused some issues. I think it did extend the video duration.

client/ayon_core/plugins/publish/extract_review.py
iLLiCiTiT: I think it has to be like this
```python
to_value = duration_seconds
if offset_seconds > 0.0:
    to_value -= offset_seconds
    ...
elif offset_seconds < 0.0:
    to_value += offset_seconds
    ...

audio_in_args.append(f"-to {to_value:0.10f}")
```

iLLiCiTiT: Tested and current logic is actually correct, sorry.

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1336,6 +1344,23 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
...
+        # audio_filters.append("apad")
+        # audio_out_args.append("-shortest")
+
+        standard_audio_rates = (48000, 96000, 192000)
+        min_audio_rate = 1024 * temp_data.fps  # 1 AAC block = 1024 samples
+        sample_rate = next(
+            (rate for rate in standard_audio_rates if rate > min_audio_rate),
+            standard_audio_rates[-1]
+        )
+        audio_out_args.append("-ar {}".format(sample_rate))
```
iLLiCiTiT:
```suggestion
        audio_out_args.append(f"-ar {sample_rate}")
```

client/ayon_core/plugins/publish/extract_review.py
rdelillo: Current `ayon_core` was not applying `offset` and current `offset` value in `ayon_maya` is wrong (returned as frame number instead of time offset).

I'm fine with this but if we do it this way, then we might need to update the dependencies on `ayon_maya` and handle ffmpeg failing in `ayon-core` as it will suddenly break with previous maya addon versions and potentially other integrations.

client/ayon_core/plugins/publish/extract_review.py
```diff
@@ -1304,22 +1304,30 @@ def audio_args(self, instance, temp_data: TempData, duration_seconds):
             return audio_in_args, audio_filters, audio_out_args
 
         for audio in audio_inputs:
-            offset_seconds = 0
-            if offset_seconds > 0:
-                audio_in_args.append(
-                    "-ss {}".format(offset_seconds)
+            if audio.get("offset"):
```
iLLiCiTiT, line 1310:
```suggestion
            # 'offset' was set only by maya addon and it used frame offset
            #     instead of seconds offset. To prevent issues the value is
            #     ignored and logged as skipped.
            if audio.get("offset"):
```

client/ayon_core/plugins/publish/extract_review.py
iLLiCiTiT: Makes sense, added suggestion https://github.com/ynput/ayon-core/pull/2004/changes#r3853952082

iLLiCiTiT (approved):
Just this comment https://github.com/ynput/ayon-core/pull/2004/changes#r3853952082 (different words can be used).

Tried all possible offsets and all of them worked.

PR #2005, opened 2026-08-18
FFmpeg: Use `-/filter` for ffmpeg-9.x+

vincentullmann:
## Changelog Description
use `-/filter` for ffmpeg 9.x

## Additional info
`-filter_script` has been removed in ffmpeg 9.x. ([relevant commit](https://github.com/FFmpeg/FFmpeg/commit/07407fff6142f14dcb21b8a06d0d15db0e31135e#diff-b646b1a71ed55a47856727c59940ee9579254bdf1db57c33f2d229516beb3fc4L1800-L1804)) 
It was marked as deprecated for a while.

I've added a check and decided to support both, since `-/filter` is apparently only supported as of ffmpeg 7.x.
Removing it would break things for anyone running ffmpeg 1.x - 5.x

## Testing notes:
1. install ffmpeg 9.x
2. try to publish a product that uses otio_burnin

### DIFF
--- a/client/ayon_core/lib/vendor_bin_utils.py
+++ b/client/ayon_core/lib/vendor_bin_utils.py
@@ -1,3 +1,4 @@
+import functools
 import os
 import logging
 import platform
@@ -463,6 +464,48 @@ def get_ffmpeg_tool_args(tool_name, *extra_args):
     raise ToolNotFoundError(
         "FFmpeg '{}' tool not found.".format(tool_name)
     )
+
+
+@functools.lru_cache
+def get_ffmpeg_supported_options(mode: str = "long") -> set[str]:
+    """Get all the options supported by the current FFmpeg version.
+
+    Args:
+        mode: Which version of the help message to parse
+            Can be "long", "full" or "" (for short)
+            Default is "long"
+
+    Returns:
+        set[str]: All the options supported by the current FFmpeg version.
+
+    """
+    ffmpeg = get_ffmpeg_tool_path("ffmpeg")
+    result = subprocess.run(
+        [ffmpeg, "-h", mode],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+        check=False,
+    )
+
+    options: set[str] = set()
+    for line in result.stdout.splitlines():
+        option = line.lstrip()
+        if not option.startswith("-"):
+            continue
+        option = option.split(" ", maxsplit=1)[0]  # remove description
+        option = option.split("[", maxsplit=1)[0]  # remove stream_specifier
+        if not option:
+            continue
+
+        options.add(option)
+
+    return options
+
+
+def is_ffmpeg_supported_option(option: str) -> bool:
+    """True if the current version of ffmpeg supports the given option."""
+    return option in get_ffmpeg_supported_options()
 
 
 def is_oiio_supported():
--- a/client/ayon_core/lib/vendor_bin_utils.py
+++ b/client/ayon_core/lib/vendor_bin_utils.py
@@ -479,9 +479,8 @@ def get_ffmpeg_supported_options(mode: str = "long") -> set[str]:
         set[str]: All the options supported by the current FFmpeg version.
 
     """
-    ffmpeg = get_ffmpeg_tool_path("ffmpeg")
     result = subprocess.run(
-        [ffmpeg, "-h", mode],
+        get_ffmpeg_tool_args("ffmpeg", "-h", mode),
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
         text=True,
--- a/client/ayon_core/lib/vendor_bin_utils.py
+++ b/client/ayon_core/lib/vendor_bin_utils.py
@@ -495,6 +495,7 @@ def get_ffmpeg_supported_options(mode: str = "long") -> set[str]:
             continue
         option = option.split(" ", maxsplit=1)[0]  # remove description
         option = option.split("[", maxsplit=1)[0]  # remove stream_specifier
+        option = option.lstrip("-")
         if not option:
             continue
 
--- a/client/ayon_core/lib/vendor_bin_utils.py
+++ b/client/ayon_core/lib/vendor_bin_utils.py
@@ -503,7 +503,7 @@ def get_ffmpeg_supported_options(mode: str = "long") -> set[str]:
     return options
 
 
-def is_ffmpeg_supported_option(option: str) -> bool:
+def is_ffmpeg_option_supported(option: str) -> bool:
     """True if the current version of ffmpeg supports the given option."""
     return option in get_ffmpeg_supported_options()
 
--- a/client/ayon_core/scripts/otio_burnin.py
+++ b/client/ayon_core/scripts/otio_burnin.py
@@ -27,7 +27,7 @@ from ayon_core.lib import (
     get_ffmpeg_codec_args,
     get_ffmpeg_format_args,
     convert_ffprobe_fps_value,
-    is_ffmpeg_supported_option,
+    is_ffmpeg_option_supported,
     StringTemplate,
 )
 
@@ -649,7 +649,7 @@ class ModifiedBurnins(ffmpeg_burnins.Burnins):
                 filters_path = temp.name
 
             # "-filter_script" was removed in FFmpeg 9 in favor of "-/filter"
-            if is_ffmpeg_supported_option("filter_script"):
+            if is_ffmpeg_option_supported("filter_script"):
                 filters = f'-filter_script:v "{filters_path}"'
             else:
                 filters = f'-/filter:v "{filters_path}"'
--- a/client/ayon_core/scripts/otio_burnin.py
+++ b/client/ayon_core/scripts/otio_burnin.py
@@ -650,7 +650,7 @@ class ModifiedBurnins(ffmpeg_burnins.Burnins):
                 filters_path = temp.name
 
             # "-filter_script" was removed in FFmpeg 9 in favor of "-/filter"
-            if is_ffmpeg_supported_option("-filter_script"):
+            if is_ffmpeg_supported_option("filter_script"):
                 filters = f'-filter_script:v "{filters_path}"'
             else:
                 filters = f'-/filter:v "{filters_path}"'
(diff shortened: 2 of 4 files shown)


BigRoy (commented):
Looks like `-/filter` was added in FFMPEG 6.0 - do we need to support even that far back to back-support 5.x and below? If not, I'd just swap the flags?

@iLLiCiTiT

vincentullmann:
> Looks like `-/filter` was added in FFMPEG 6.0 - do we need to support even that far back to back-support 5.x and below? If not, I'd just swap the flags?
> 
> @iLLiCiTiT

just to note: There are 2 different related features:

in 6.0:  https://github.com/FFmpeg/FFmpeg/blob/4b47405e1c251405bfa5196e79aa395c59ad2323/Changelog#L276-L277 ([commit](https://github.com/FFmpeg/FFmpeg/commit/78f46065d861eb8487124e92ab46a459426eb89c))
> - filtergraph syntax in ffmpeg CLI now supports passing file contents as option values, by prefixing option name with '/'
 ```sh
ffmpeg -i <INPUT> -vf drawtext=/text=/tmp/some_text <OUTPUT>
```
 
 
in 7.0: https://github.com/FFmpeg/FFmpeg/blob/4b47405e1c251405bfa5196e79aa395c59ad2323/Changelog#L168-L169 ([commit](https://github.com/FFmpeg/FFmpeg/commit/6d17991b7e1bf1a5d104c8a6261709f7e6640d97))
> - ffmpeg CLI options may now be used as -/opt <path>, which is equivalent to -opt <contents of file <path>>
```sh
ffmpeg -i INPUT -/filter:v filter.script OUTPUT
```

from what I understand the feature from 6.0 supports loading individual filters form disk where as the one in 7.0 allows to load the entire graph.

client/ayon_core/scripts/otio_burnin.py
```diff
@@ -48,6 +49,52 @@
 SOURCE_TIMECODE_KEY = "{source_timecode}"
 
 
+@functools.lru_cache
+def get_supported_ffmpeg_options(mode: str = "long") -> set[str]:
```
iLLiCiTiT: Does the `mode` argument make any sense?

iLLiCiTiT:
The PR is made in a way that it does not matter, so can probably keep it as it is?

So far we did not have requirement on ffmpeg and we probably still do support ffmpeg 4. But we probably should raise the requirement to at least 7.

BigRoy:
> The PR is made in a way that it does not matter, so can probably keep it as it is?

It's mostly that there've been cases where just calling anything on ffmpeg takes up to a few seconds for some. This just means that each process would at least the first time cause an extra few seconds of slowness. 

Anyway, I feel like we may benefit from this logic in other places anyway where we can decide on whether something's available or not - so I guess let's not care.

client/ayon_core/scripts/otio_burnin.py
BigRoy: I guess it's a default ffmpeg argument - so it may be useful. Might be nice at least to use `Literal` type hint to be more exact here?

vincentullmann: I've added it mostly because it felt like the `full` help did take a tiny bit longer amount to load, but I wanted to make sure we have the option to increase it if necessary in the future.


I just ran some tests to compare.
```sh
$ time ffmpeg -hide_banner -h | grep "filter_script"

real	0m0.014s
user	0m0.008s
sys	0m0.006s
```
```sh
$ time ffmpeg -hide_banner -h long | grep "filter_script"
-filter_script[:<stream_spec>] <filename>  deprecated, use -/filter

real	0m0.013s
user	0m0.010s
sys	0m0.004s
```
```sh
$ time ffmpeg -hide_banner -h full | grep "filter_script"
-filter_script[:<stream_spec>] <filename>  deprecated, use -/filter

real	0m0.040s
user	0m0.029s
sys	0m0.010s
```

The difference is smaller than what it appeared last night
without the `grep` its around 0m0.016s for full and 0m0.014s for long

So if you want I recon we can drop the `mode` argument, hardcode it to use `long` for now,
and reintroduce it / or swap it to `full` if ever needed in the future, given that the difference is not that big

BigRoy: This `ffmpeg` is local on your disk? or shared over a network access? Many teams configure it over their network storage, so the single run may be much slower in certain scenarios.

vincentullmann: those timings where done using a ffmpeg installed on our network drive.

Our NFS does a good amount of caching tough (which can be seen here when I first hit ffmpeg 7.1)
```
$ time /mnt/pipeline/.../ffmpeg/ffmpeg_4.4/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.140s
user	0m0.016s
sys	    0m0.010s

$ time /mnt/pipeline/.../ffmpeg/ffmpeg_7.1/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.490s
user	0m0.021s
sys	    0m0.008s

$ time /mnt/pipeline/.../ffmpeg/ffmpeg_7.1/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.025s
user	0m0.020s
sys	    0m0.003s

$ time /mnt/pipeline/.../ffmpeg/ffmpeg_7.1/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.028s
user	0m0.019s
sys	    0m0.005s

$ time /mnt/pipeline/.../ffmpeg/ffmpeg_7.1/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.025s
user	0m0.018s
sys	    0m0.006s

$ time /mnt/pipeline/.../ffmpeg/ffmpeg_8.0/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.039s
user	0m0.029s
sys	    0m0.007s

$ time /mnt/pipeline/.../ffmpeg/ffmpeg_8.0/ffmpeg -hide_banner -h full>/dev/null 2>&1
real	0m0.039s
user	0m0.028s
sys	    0m0.009s
```

Interestingly the network drive on our work network outperforms my old SSD in my pc at home:
```
$ time /usr/bin/ffmpeg  -hide_banner -h full >/dev/null 2>&1
real    0m0.085s
user    0m0.065s
sys     0m0.019s

$ time /usr/bin/ffmpeg  -hide_banner -h full >/dev/null 2>&1
real    0m0.093s
user    0m0.076s
sys     0m0.016s

$ time /usr/bin/ffmpeg  -hide_banner -h full >/dev/null 2>&1
real    0m0.087s
user    0m0.066s
sys     0m0.020s

$ time /usr/bin/ffmpeg  -hide_banner -h full >/dev/null 2>&1
real    0m0.128s
user    0m0.109s
sys     0m0.018s

$ time /usr/bin/ffmpeg  -hide_banner -h full >/dev/null 2>&1
real    0m0.127s
user    0m0.100s
sys     0m0.026s

```

BigRoy: 0.5 seconds - sucks a tiny bit, but it's nothing crazy or off the charts I suppose, since we cache the response anyway.

BigRoy (approved):
LGTM - @antirotor can you test with a few ffmpeg versions?

client/ayon_core/scripts/otio_burnin.py
iLLiCiTiT: Well, the question is still same, does it make sense to have the argument? But probably not that important as it lives in burnins script.

BigRoy: I'm fine with it either way - so fine to keep for now.

client/ayon_core/scripts/otio_burnin.py
```diff
@@ -646,7 +693,13 @@ def command(self, output=None, args=None, overwrite=False):
             ) as temp:
                 temp.write(filter_string)
                 filters_path = temp.name
-            filters = '-filter_script:v "{}"'.format(filters_path)
+
+            # "-filter_script" was removed in FFmpeg 9 in favor of "-/filter"
+            if ffmpeg_supports_option("-filter_script"):
+                filters = f'-filter_script:v "{filters_path}"'
+            else:
+                filters = f'-/filter:v "{filters_path}"'
```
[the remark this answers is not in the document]
BigRoy: @vincentullmann can you confirm this perhaps?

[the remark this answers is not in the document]
vincentullmann: `-filter_complex_script` 
was indeed removed too ([line 1821](https://github.com/FFmpeg/FFmpeg/commit/07407fff6142f14dcb21b8a06d0d15db0e31135e#diff-b646b1a71ed55a47856727c59940ee9579254bdf1db57c33f2d229516beb3fc4L1821)) so yes, I think this PR should address this too (good that we already have a method in place to check which flags ffmpeg supports :wink: )

I'll update the PR later

client/ayon_core/scripts/otio_burnin.py
```diff
@@ -48,6 +49,52 @@
...
+        set[str]: All the options supported by the current FFmpeg version.
+
+    """
+    result = subprocess.run(
+        [
+            *FFMPEG_EXE_ARGS,
+            "-hide_banner",
+            "-h",
+            mode,
+        ],
```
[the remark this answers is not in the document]
vincentullmann: > Looks like `-hide_banner` was added in FFMPEG 6.0 - do we need to support even that far back to back-support 5.x and below? If not, I'd just swap the flags?

guess it doesn't hurt to remove the flag. Doubt it will make much of a difference

BigRoy:
@vincentullmann any chance you're able to make those last tweaks - like removing the hiding of the banner (if it doesn't make much difference in performance) and fixing the `-filter_complex_script` calls?

vincentullmann:
> @vincentullmann any chance you're able to make those last tweaks - like removing the hiding of the banner (if it doesn't make much difference in performance) and fixing the `-filter_complex_script` calls?

both done.

client/ayon_core/lib/vendor_bin_utils.py
```diff
@@ -465,6 +466,48 @@ def get_ffmpeg_tool_args(tool_name, *extra_args):
...
+        option = option.lstrip("-")
+        if not option:
+            continue
+
+        options.add(option)
+
+    return options
+
+
+def is_ffmpeg_supported_option(option: str) -> bool:
```
iLLiCiTiT:
```suggestion
def is_ffmpeg_option_supported(option: str) -> bool:
```

iLLiCiTiT (approved):

PR #2011, opened 2026-08-25
Publish: Capture more logs

iLLiCiTiT:
## Changelog Description
Correctly capture logs if log to console is enabled and capture logs from AYON's `Logger` logger objects.

## Additional info
Root logger always does register publish handler to capture logs and plugin's log has added handler only if does not propagate records to parent logger.

To simplify work with the loggers in `Logger` a root logger was added. The root logger is the source for log level, formatting and record handlers.

With root logger for Logger loggers it is easy to add publish log handler for them. It actually also simplified management of AYON loggers `get_logger` function.

Also added deprecation for `get_logger` method, from now on the `name` has to be passed in. Not passing it triggers warning message. As far as I know there is no place that does not pass the name.

### Why is it needed
1. `PublishLogic` did not capture all logs by default. Running the same publish with publisher and PublishLogic might produce different reports.
2. We do pass along logger objects at many places to capture the logs for publishing report. With this change we don't really have to do that.

## Testing notes:
1. Logs triggered by any `Logger` object will be captured in publish report.
3. It is important that the logs are not duplicated.

### DIFF
--- a/client/ayon_core/lib/log.py
+++ b/client/ayon_core/lib/log.py
@@ -1,14 +1,14 @@
 from __future__ import annotations
 
-import os
-import sys
+import copy
 import getpass
 import logging
+import os
 import platform
 import socket
+import sys
 import time
 import threading
-import copy
 import warnings
 
 from . import Terminal
--- a/client/ayon_core/lib/log.py
+++ b/client/ayon_core/lib/log.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import os
 import sys
 import getpass
@@ -113,6 +115,7 @@ class Logger:
 
     # Logging level - AYON_LOG_LEVEL
     log_level = None
+    logger_names: set[str] = set()
 
     # Data same for all record documents
     process_data = None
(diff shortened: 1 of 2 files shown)


client/ayon_core/lib/log.py
```diff
@@ -120,28 +137,23 @@ class Logger:
     _process_name = None
 
     @classmethod
-    def get_logger(cls, name=None):
+    @_deprecated_getter
+    def get_logger(cls, name: str) -> logging.Logger:
```
BigRoy, line 141: Why not allow `name=None` anymore? 🤔 `logging.getLogger` also allows that and has it as default arg. Why won't we match?

BigRoy (approved):
Works for me, on farm and in DCC.

client/ayon_core/lib/log.py
iLLiCiTiT: `logging.getLogger` with `None` does return root logger, we do prefill it with `__main__` to avoid returning the root logger. It is not about matching `logging` but to give api sense for developers using the `Logger`.

PR #2012, opened 2026-08-25
User model: Better type hints

iLLiCiTiT:
## Changelog Description
Added better type hints to user model in tool api.

## Testing notes:
No logic changed, just added/changed type-hints.
1. Validate code changes.

### DIFF
--- a/client/ayon_core/tools/common_models/users.py
+++ b/client/ayon_core/tools/common_models/users.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Any, Optional
+from typing import Any
 
 import ayon_api
 
@@ -48,7 +48,7 @@ class UsersModel:
         self._controller = controller
         self._users_cache = NestedCacheItem(default_factory=list)
 
-    def get_current_username(self) -> Optional[str]:
+    def get_current_username(self) -> str | None:
         if self._current_username is NOT_SET:
             self._current_username = get_ayon_username()
         return self._current_username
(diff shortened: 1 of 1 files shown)


rdelillo (approved):

client/ayon_core/tools/common_models/users.py
```diff
@@ -48,47 +48,52 @@ def __init__(self, controller):
...
-            List[UserItem]: List of user items.
+            list[UserItem]: List of user items.
 
         """
         self._invalidate_cache(project_name)
         return self._users_cache[project_name].get_data()
 
-    def get_user_items_by_name(self, project_name):
+    def get_user_items_by_name(
+        self, project_name: str | None
```
BigRoy, line 70: Why is `project_name` allowed to be `None`?

client/ayon_core/tools/common_models/users.py
```diff
@@ -97,7 +102,7 @@ def get_user_item_by_username(self, project_name, username):
                 return user_item
         return None
 
-    def _invalidate_cache(self, project_name):
+    def _invalidate_cache(self, project_name: str | None) -> None:
```
BigRoy, line 105: Why is `project_name` allowed to be `None`?

BigRoy (approved):
Looks ok to me, bit surprised why `project_name` can be `None`.

client/ayon_core/tools/common_models/users.py
iLLiCiTiT: Because it can happen now. PR is adding typehints to know what we do pass in, not how it should be.

To make it better I would have to go through all the places using the methods and change them. I can, but that is even more time spent on it.

iLLiCiTiT: Damn, I merged the PR by accident, should I do followup PR?

BigRoy: Nothing critical for now. May be nice one day.

PR #2015, opened 2026-08-25
Projects model: Better type hints

iLLiCiTiT:
## Changelog Description
Added better type hints to projects model in common models.

## Additional info
Added some safeguard in case project name is passed as None to some of the methods.

## Testing notes:
Technically nothing changed.
1. Validate code changes.
2. Open tool which uses projects model: Launcher / Loader.
3. Projects do show correctly, folder types, task types, statuses and tags are available and loook correctly.

### DIFF
--- a/client/ayon_core/tools/common_models/projects.py
+++ b/client/ayon_core/tools/common_models/projects.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
-import json
-import contextlib
 from abc import ABC, abstractmethod
-from typing import Any, Optional
+import contextlib
 from dataclasses import dataclass
+import json
+import typing
+from typing import Any, Generator, Callable
 
 import ayon_api
 from ayon_api.graphql_queries import projects_graphql_query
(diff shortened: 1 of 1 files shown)


client/ayon_core/tools/common_models/projects.py
```diff
@@ -103,15 +116,17 @@ class FolderTypeItem:
     short: str
     icon: str
 
-    def to_data(self):
+    def to_data(self) -> dict[str, str]:
```
BigRoy, line 119: I wonder if there's any trivial way to generate a correct Type Hint for these dataclasses on their `to_data` and `from_data` because technically we can make an exact `TypedDict`? 🤔

BigRoy (approved):
Works for me.

moonyuet (approved):
Works for me too

client/ayon_core/tools/common_models/projects.py
iLLiCiTiT: If you define `TypedDict` it is expected that you wil return object of the `TypedDict` -> it is not just about typehint. Also defining typed dict for every dataclass we have is just waste of time, practically doubling every class with second typed dict class, with same keys and typehints -> double maintanance and amount of code. The only purpose of these is for inter process communication.

PR #2017, opened 2026-08-26
Workfiles tool: disable file extension combobox in "Copy & Open"

### LINKED ISSUE #1118: Workfiles tool > Published > Copy & Open - invalidly prompts for save file extension

### Current Behavior:

When doing a Copy & Open then the user gets the save prompt with allowing the user to check the extension. However - what it'll do is just take the published file, and copy it with the different extension - in this case it'd e.g. just copy the `.zip` as ` .sfx` instead of actually "resaving" it into a different format. Worse even, it does not **default** to the extension of the published file, but to the 'first' file extension defined for the host.

Image

Image

I think "Copy & Open" should **not** allow changing the extension. So the copy should always use the extension from the published file.

### Expected Behavior:

Matching the extension seems most reasonable - disallowing the extension change in the prompt.

### Version

1.0.14

Windows

### Steps To Reproduce:

1. In Silhouette integration, publish workfile as .zip
2. Use "Copy and Open" from published section in workfiles tool
3. It'll prompt to default to `.sfx` (and just copy the `.zip` as `.sfx`) which is the incorrect format.
4. Hence Silhouette will be unable to open the file.

### Relevant log output:

```shell

```

### Additional context:

Came up internally in a private chat [here](https://discord.com/channels/517362899170230292/773127690453516298/1335920531765067798)

tadeas-hejnic:
## Changelog Description
This PR disables file extension change during `Copy & Open` action in work files tool. 

It is done via adding new argument `enable_ext_selection` and disabling the combobox based on this passed argument value. 

## Additional info
N/A

## Testing notes:
1. Open the workfiles tool in DCC
2. Click on `Copy & Open` button


client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -314,8 +321,10 @@ def _on_task_changed(self, event):
         self._update_published_btns_state()
         self._update_workarea_btns_state()
 
-    def _on_published_save_clicked(self):
-        result = self._exec_save_as_dialog()
+    def _on_published_save_clicked(self, enable_ext_selection=True):
+        result = self._exec_save_as_dialog(
+            enable_ext_selection=enable_ext_selection
+        )
```
iLLiCiTiT: When this method is called it always uses copy workfile representation, so it always should disable the extensions.
```suggestion
    def _on_published_save_clicked(self):
        result = self._exec_save_as_dialog(allow_ext_selection=False)
```

client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -99,7 +99,10 @@ def __init__(self, controller, parent):
 
         published_widget.save_as_requested.connect(self._on_save_as_request)
         published_btn_copy_n_open.clicked.connect(
-            self._on_published_save_clicked)
+            lambda: self._on_published_save_clicked(
+                enanble_ext_selection=False
+            )
+        )
```
iLLiCiTiT:
```suggestion
            self._on_published_save_clicked
        )
```

client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -166,14 +169,18 @@ def set_text_filter(self, text_filter):
...
 
         Returns:
             Union[dict[str, Any], None]: Result of the dialog.
         """
 
-        dialog = SaveAsDialog(self._controller, self)
+        dialog = SaveAsDialog(
+            self._controller,
+            self,
+            enable_ext_selection=enable_ext_selection
```
iLLiCiTiT:
```suggestion
            allow_ext_selection=allow_ext_selection,
```

client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -166,14 +169,18 @@ def set_text_filter(self, text_filter):
         self._workarea_widget.set_text_filter(text_filter)
         self._published_widget.set_text_filter(text_filter)
 
-    def _exec_save_as_dialog(self):
+    def _exec_save_as_dialog(self, enable_ext_selection=True):
```
iLLiCiTiT:
```suggestion
    def _exec_save_as_dialog(self, allow_ext_selection=True):
```

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -93,7 +93,7 @@ class SaveAsDialog(QtWidgets.QDialog):
         controller (AbstractWorkfilesFrontend): The control object.
     """
 
-    def __init__(self, controller, parent):
+    def __init__(self, controller, parent, enable_ext_selection=True):
```
iLLiCiTiT: I'd say enabled is saying what will happen in UI instead of what is needed.
```suggestion
    def __init__(self, controller, parent, allow_ext_selection=True):
```

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -196,6 +196,8 @@ def __init__(self, controller, parent):
         subversion_input.text_changed.connect(self._on_comment_change)
         extension_combobox.currentIndexChanged.connect(
             self._on_extension_change)
+        if enable_ext_selection is False:
+            extension_combobox.setEnabled(False)
```
iLLiCiTiT:
```suggestion
        extension_combobox.setEnabled(allow_ext_selection)
```

iLLiCiTiT: BTW I'm not sure if this is good idea. This will still show the combobox with possibly wrong extension which might be confusing, and actually cause wrong output filepath.

So I guess we should allow to pass in allowed extensions, in that case the extensions from data won't be used. With that we can pass only the extension from published workfile, so no need to disable anything as there will be just one option to choose from.

I don't know the underlaying logic, it might be easier to pass in selected representation id which would be passed to the getter method, not sure I'd have to go through the codebase.

tadeas-hejnic: OK, got it. Now it should be better solution, based on your suggestion.

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -263,7 +262,11 @@ def update_context(self):
         self._last_version = data["last_version"]
 
         self._extension_combobox.clear()
-        self._extension_combobox.addItems(data["extensions"])
+        if self._allow_ext_selection:
+            self._extension_combobox.addItems(data["extensions"])
+        else:
+            # allow only the extension from the copied workfile
+            self._extension_combobox.addItems([data["ext"]])
```
iLLiCiTiT: The `"ext"` from `get_workarea_save_as_data` uses "current workfile" extension, not the published representation extension. The save dialog has no way how to get the extension with information we do pass in it.

client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -165,15 +165,18 @@ def set_text_filter(self, text_filter):
         self._workarea_widget.set_text_filter(text_filter)
         self._published_widget.set_text_filter(text_filter)
 
-    def _exec_save_as_dialog(self, allow_ext_selection=True):
+    def _exec_save_as_dialog(self, allow_ext_selection=True, extension=None):
```
iLLiCiTiT: The `allow_ext_selection` can be removed.

tadeas-hejnic: Yes, removed already in the previous commit.

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -93,11 +93,14 @@ class SaveAsDialog(QtWidgets.QDialog):
         controller (AbstractWorkfilesFrontend): The control object.
     """
 
-    def __init__(self, controller, parent):
+    def __init__(
+        self, controller, parent, allowed_extension=None
+    ):
```
iLLiCiTiT: Just `extension` is sufficient.
```suggestion
        controller (AbstractWorkfilesFrontend): The control object.
        parent (QtWidgets.QWidget): Parent widget.
        extension (str | None): Limit extensions to specific one.
        
    """
    def __init__(self, controller, parent, extension=None):
```

client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -98,8 +98,7 @@ def __init__(self, controller, parent):
         workarea_btn_save.clicked.connect(self._on_workarea_save_clicked)
 
         published_widget.save_as_requested.connect(self._on_save_as_request)
-        published_btn_copy_n_open.clicked.connect(
-            self._on_published_save_clicked)
+        published_btn_copy_n_open.clicked.connect(self._on_published_save_clicked)
```
iLLiCiTiT:
```suggestion
        published_btn_copy_n_open.clicked.connect(
            self._on_published_save_clicked)
```

client/ayon_core/tools/workfiles/widgets/files_widget.py
```diff
@@ -166,14 +166,18 @@ def set_text_filter(self, text_filter):
...
 
         Returns:
             Union[dict[str, Any], None]: Result of the dialog.
         """
 
-        dialog = SaveAsDialog(self._controller, self)
+        dialog = SaveAsDialog(
+            self._controller,
+            self,
+            allowed_extension=extension,
```
iLLiCiTiT:
```suggestion
            extension=extension,
```

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -89,15 +89,16 @@ class SaveAsDialog(QtWidgets.QDialog):
...
+        extension (str | None): Limit extensions to specific one.
     """
-
-    def __init__(self, controller, parent):
+    def __init__(self, controller, parent, extension=None):
         super(SaveAsDialog, self).__init__(parent=parent)
         self.setWindowFlags(self.windowFlags() | QtCore.Qt.FramelessWindowHint)
 
         self._controller = controller
+        self._allowed_extension = extension
```
iLLiCiTiT:
```suggestion
        self._extension = extension
```

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -263,7 +263,11 @@ def update_context(self):
         self._last_version = data["last_version"]
 
         self._extension_combobox.clear()
-        self._extension_combobox.addItems(data["extensions"])
+        if self._allowed_extension:
+            # allow only the extension from the copied workfile
+            self._extension_combobox.addItems([self._allowed_extension])
+        else:
+            self._extension_combobox.addItems(data["extensions"])
```
iLLiCiTiT:
```suggestion
        extensions = data["extensions"]
        if self._extension:
            extensions = [self._extension])

        self._extension_combobox.addItems(extensions)
```

iLLiCiTiT (approved):

PR #2018, opened 2026-08-26
Workfile build: Normalized the preset path when checking the valid path.

moonyuet:
## Changelog Description
This PR is to normalize the preset.path for `has_valid_path` function. This ensures the case of escaped filepath like `D:/studio/nuke/template.nk` can be checked if it existed correctly.

## Additional info
n/a

## Testing notes:
1. Go to AYON menu
2. Open template

### DIFF
--- a/client/ayon_core/pipeline/workfile/workfile_template_builder.py
+++ b/client/ayon_core/pipeline/workfile/workfile_template_builder.py
@@ -151,7 +151,7 @@ class TemplatePreset:
             bool: Preset has set path to an existing file.
 
         """
-        return self.path and os.path.exists(self.path)
+        return self.path and os.path.exists(os.path.normpath(self.path))
 
 
 class AbstractTemplateBuilder(ABC):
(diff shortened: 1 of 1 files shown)


iLLiCiTiT:
Should the path be normalized before it is passed to the preset object?

client/ayon_core/pipeline/workfile/workfile_template_builder.py
```diff
@@ -1191,7 +1191,7 @@ def get_template_preset(self) -> TemplatePreset:
             create_first_version = True
 
         return TemplatePreset(
-            path=resolved_path,
+            path=os.path.normpath(resolved_path),
```
iLLiCiTiT: In case `resolved_path` would be empty string the normpath would change it to `"."` which is "valid" path.
```suggestion
        if resolved_path:
             resolved_path = os.path.normpath(resolved_path)

        return TemplatePreset(
            path=resolved_path,
```

moonyuet: [4b0577f](https://github.com/ynput/ayon-core/pull/2018/commits/4b0577f4627b8d54f1a7c0f867952a7b96cdeb02)

iLLiCiTiT (approved):

BigRoy:
> This ensures the case of escaped filepath like D:/studio/nuke/template.nk can be checked if it existed correctly.

Out of curiosity. By what means? As in from user logs? Because `os.path.exists` handles such a path just fine?

PR #2019, opened 2026-08-27
Loader: Fix filtering

iLLiCiTiT:
## Changelog Description
Fix role constant values used for filtering.

## Testing notes:
1. Filtering by statuses or tags does work in loader.


BigRoy (approved):
Does seem to fix it, yes.

Status, Version Tags and Task Tags work as intended now.

PR #2021, opened 2026-08-27
Create Context: Add type hints

iLLiCiTiT:
## Changelog Description
Added type hints to methods in CreateContext.

## Additional info
Added warning to `bulk_instances_collection` as it is marked as deprecated. Added `AYONPublishPluginType` type to mark plugin inheriting from both `pyblish.api.Plugin` and `AYONPyblishPluginMixin`.

## Testing notes:
1. Validate code changes.
2. Try to open publisher in a DCC with oldest supported python you have.

### DIFF
--- a/client/ayon_core/pipeline/create/context.py
+++ b/client/ayon_core/pipeline/create/context.py
@@ -56,7 +56,11 @@ if typing.TYPE_CHECKING:
     from ayon_core.lib import AbstractAttrDef
     from ayon_core.lib.events import EventCallback, Event
     from ayon_core.host import AbstractHost
-    from ayon_core.pipeline.publish.typing import PublishPluginType
+    from ayon_core.pipeline import AYONPyblishPluginMixin
+    from ayon_core.pipeline.publish.typing import (
+        PublishPluginType,
+        AYONPublishPluginType,
+    )
 
     from .structures import CreatedInstance
     from .creator_plugins import BaseCreator, ProductConvertorPlugin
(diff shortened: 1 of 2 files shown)


moonyuet (approved):

client/ayon_core/pipeline/create/context.py
```diff
@@ -196,57 +205,57 @@ def __init__(
...
+        self.convertor_items_by_id: dict[str, ConvertorItem] = {}
 
-        self.publish_discover_result: Optional[DiscoverResult] = None
-        self.publish_plugins_mismatch_targets = []
-        self.publish_plugins = []
-        self.plugins_with_defs = []
+        self.publish_discover_result: DiscoverResult | None = None
+        self.publish_plugins_mismatch_targets: list[PublishPluginType] = []
+        self.publish_plugins: list[PublishPluginType] = []
+        self.plugins_with_defs: list[AYONPublishPluginType] = []
```
BigRoy, line 252: So I wonder if technically from a type hint perspective whether this is actually correct....

Because:
```
class AYONPublishPlugin(pyblish.api.Plugin, AYONPyblishPluginMixin):
    pass


class MyPlugin(pyblish.api.Plugin, AYONPyblishPluginMixin):
    pass
```
In this scenario (the bottom is what most of our plug-ins do) the `MyPlugin` is not of type `AYONPublishPlugin`
Pretty sure a type checker would tell you it's wrong essentially.

I believe this must technically inherit also from `Protocol`?

```
class AYONPublishPluginProtocol(pyblish.api.Plugin, AYONPyblishPluginMixin, Protocol):
```

For the typing then to be `Type[AYONPublishPluginProtocol]`?

iLLiCiTiT: That would be true if `pyblish.api.Plugin` and `AYONPyblishPluginMixin` would inherit from `Protocol` too. What is there now is the best we can do with what we have 😉 .

I've added the type so IDE can show what is available for methods. It still does not recognize some of the pyblish plugin things because they are "appended" to the class instead of being available on the base class.

BigRoy: I see - probably best we can do then indeed.

PR #2022, opened 2026-08-27
Core: Modules discovery returns objects

iLLiCiTiT:
## Changelog Description
Function `modules_from_path` does return objects instead of tuples of lists of tuples.

## Additional info
The function does return dataclasses with attributes. For backwards compatibility all dataclasses do implement `__iter__` to be able to unwrap the object into tuple.

I can add warning to the `__iter__` method to warn developers about the change. There are multiple addons using the function so it might help to reveal the change.

With the change type hints were added to the file.

NOTE: Looks like the only other addon using this function is applications addon.

## Testing notes:
1. Validate code changes.
2. Discovery of plugins and actions does work -> Actions, Loader actions, Loader plugins, Publish plugins, Create plugins.


rdelillo (approved):
Code changes looks good and confirm it works as exepcted for me (tested from Hiero).

> I can add warning to the __iter__ method to warn developers about the change. There are multiple addons using the function so it might help to reveal the change.

A `DeprecationWarning` would make sense.
_(I know we had discussions in the past with @BigRoy that developper-oriented `logging.warning` are not ideal)_
Side discussion, IMHO I would enforce the launcher to show `DeprecationWarning` in dev bundles. It does not seem to be the case right now.

PR #2024, opened 2026-08-27
Workfiles tool: Add type hints

iLLiCiTiT:
## Changelog Description
Added typehints to controller and base classes for controller in workfiles tool.

## Additional info
This does add typehints as (I think) are passed in right now. There is room for enhancements of the code, but I wanted to avoid that as that requires refactor of the code and this is already changing few places.

Changed few classes to dataclasses.

## Testing notes:
1. Validate code changes.
2. Workfiles tool should work (in any DCC).


BigRoy (approved):
Works fine.

PR #2028, opened 2026-08-28
Launcher: Fix uri and icon handling from webactions

iLLiCiTiT:
## Changelog Description
Icon is converted to correct data for UI and uri without scheme it is expected that the uri leads to AYON.

## Testing notes:
1. Webactions with form, having custom icons are correctly shown.
2. Uri leading relative to AYON server do work.


BigRoy:
Do we know where/how to best test this? As in, do we know of an example action?

iLLiCiTiT:
> Do we know where/how to best test this? As in, do we know of an example action?

The uri is in `Open in Review`. Don't know which action does define the the icon.

PR #2029, opened 2026-08-28
Add structured logging support

antirotor:
## Changelog Description
Add structured logging support. For description see ynput/ayon-launcher#326 (that is required by this)


client/ayon_core/pipeline/publish/logic.py
```diff
@@ -1018,7 +1017,6 @@ def _inner_publish_iter(self) -> PublishIterGen:
     @contextmanager
     def _log_manager(self, plugin: PluginType):
         root = logging.getLogger()
-        ayon_root = Logger.get_root_logger()
```
iLLiCiTiT, line 1021: Why was this removed? We added this so we can see logs in publish report viewer.

pyproject.toml
```diff
@@ -33,7 +33,10 @@ dependencies = [
     "numpy >=2.4.3",
     "qtpy >=2.4.3",
     "pyside6 >=6.8.3",
-    "Pillow ==9.5.0"
+    "Pillow ==9.5.0",
+    "structlog>=26.1.0",
+    "rich>=15.0.0",
+    "requests>=2.32.5",
```
iLLiCiTiT, line 39: Requests is defined by launcher, how usefull will this be?

client/ayon_core/lib/log.py
```diff
@@ -3,16 +3,161 @@
...
+            )
+        except Exception:
+            self.handleError(record)
+
+    def close(self):
+        self._session.close()
+        super().close()
+
+
+def configure_logger() -> None:
```
iLLiCiTiT, line 96: Why do we need to call this explicitly? ~~Not all processes are executed through cli.~~ Most of processes are not executed through cli.

client/ayon_core/lib/log.py
```diff
@@ -138,10 +283,15 @@ class Logger:
...
     @_deprecated_getter
-    def get_logger(cls, name: str) -> logging.Logger:
+    def get_logger(cls, name: str) -> structlog.BoundLogger | logging.Logger:
         if not cls.initialized:
             cls.initialize()
 
+        # Delegate to structlog when configured so records share the same
+        # processors (e.g. 'site_id', timestamps) as the rest of the app.
+        if structlog.is_configured():
+            return structlog.get_logger(name or "__main__")
```
iLLiCiTiT, line 340:
```suggestion
            return structlog.get_logger(name)
```

client/ayon_core/lib/log.py
```diff
@@ -138,10 +283,15 @@ class Logger:
...
     @classmethod
     @_deprecated_getter
-    def get_logger(cls, name: str) -> logging.Logger:
+    def get_logger(cls, name: str) -> structlog.BoundLogger | logging.Logger:
         if not cls.initialized:
             cls.initialize()
 
+        # Delegate to structlog when configured so records share the same
+        # processors (e.g. 'site_id', timestamps) as the rest of the app.
+        if structlog.is_configured():
```
iLLiCiTiT: This does completelly skip the existing AYON logic. I just don't understand?

pyproject.toml
antirotor: when you run tests for example, they work with the `.venv` defined by the addon, not the launcher venv.

client/ayon_core/lib/log.py
antirotor: you need to configure the logger somewhere. It needs some entrypoint. The goal is to find out all the entrypoints and add this there.

client/ayon_core/pipeline/publish/logic.py
antirotor: good catch, this is misfire. I tried to figure out double-logging. In any way - if we start using structlog, we could rewrite the publisher loging as another sink with it's own processors. I'll revert the changes.

client/ayon_core/lib/log.py
antirotor: this followed the same logic already in place but even there `"__main__"` shouldn't happen.

client/ayon_core/lib/log.py
antirotor: yes, it use structlog configured logger as root logger. Eventually we can completely remove `ayon_core.lib.Logger`

client/ayon_core/host/host.py
```diff
@@ -231,7 +245,12 @@ def set_current_context(
         self._before_context_change(context_change_data)
         self._set_current_context(context_change_data)
         self._after_context_change(context_change_data)
-
+        unbind_contextvars("ayon_context")
+        bind_contextvars(ayon_context=AyonLogContext(
+            project=project_name,
+            folder=folder_path,
+            task=task_name,
+        ))
```
BigRoy, line 252: `set_current_context` is not triggered on app launch, right? Only when switching context, as such - are we lacking a `ayon_context` bound var until we actually switched?

antirotor: that is something I am trying to figure out. I am also adding host and project on `host.install()` but we don't have single entrypoint and handling all cases is tricky. For example - in current state you get host bound correctly in parts of the code, but some parts are skipping that and are initializing in different way... Funny as structlog is wrapping standard python logging, so in theory, you wouldn't have to make any changes in code, but ...

antirotor:
with the latest change, it will work even with ayon-launcher without structlog (so it is backwards compatible).

client/ayon_core/lib/log.py
```diff
@@ -3,17 +3,209 @@
...
+# If structlog is missing (ayon-launcher is outdated),
+# the library will fall back to standard logging.
+
+structlog: Any = None
+try:
+    import structlog as _structlog
+except ImportError:
+    pass
+else:
+    structlog = _structlog
```
iLLiCiTiT, line 31:
```suggestion
try:
    import structlog
except ImportError:
    structlog = None
```

client/ayon_core/lib/log.py
iLLiCiTiT: > this followed the same logic already in place

That logic was removed and marked as deprecated.

client/ayon_core/lib/log.py
iLLiCiTiT: Well, we already have initialization logic in Logger, why we need a second one that is against what is happening there?

antirotor:
I've restored the original logger nesting. Added file logging and I've also hardened the Vector logging (fixes for possible leaks, added retries and proper timeouts). For usage, see description on ayon-launcher PR.

BigRoy:
Just for sake of testing it - this does work for me:

image

I think we'll need at least a unique process id per log and possibly the current process's executable name or `AYON_APP_NAME` or equivalent to at least get some sense of _where we're running_ and filter to e.g. a particular launch, or a particular Maya session. 

Also, grafana is a beast to configure.

client/ayon_core/lib/log.py
```diff
@@ -138,18 +302,39 @@ class Logger:
...
+        name = name or "__main__"
+
+        logger = logging.getLogger(name)
+        if logger is not cls._root_logger:
+            logger.parent = cls._root_logger
+
+        # Delegate to structlog when configured so records share the same
+        # processors (e.g. 'site_id', timestamps) as the rest of the app.
+        if structlog is not None and structlog.is_configured():
+            return structlog.get_logger(name)
```
BigRoy, line 339: Could this potentially return a different logger, if so - should this be above the parenting of the logger and do:

```python
if structlog ...:
    logger = structlog.get_logger(name)
else:
    logger = logging.getLogger(name)
```
Why re-request the logger?

client/ayon_core/lib/log.py
```diff
@@ -138,18 +302,39 @@ class Logger:
...
 
         return logger
 
     @classmethod
-    def get_root_logger(cls) -> logging.Logger:
+    def get_root_logger(cls) -> logging.Logger | None:
```
BigRoy: Why would this ever return `None`? The whole point of it triggering initialize means it must then NOT be `None` afterwards, right?:

client/ayon_core/lib/log.py
```diff
@@ -264,3 +462,129 @@ def get_process_name(cls):
...
+        if os.getenv("AYON_DEBUG") is not None:
+            root_logger.setLevel(logging.DEBUG)
+
+        # 'Logger' (ayon_core.lib.log) may have attached its own fallback
+        # console handler to the "AYON" logger before structlog was configured.
+        # Drop it and let records propagate to the root logger instead, which
+        # now owns the shared handlers - avoids logging each record twice.
+        ayon_logger = logging.getLogger("AYON")
+        for old_handler in list(ayon_logger.handlers):
+            ayon_logger.removeHandler(old_handler)
```
BigRoy: Why would this be the case? We have full control over the logger, can't we just avoid it configuring wrong at first?

antirotor: this is old note from previous state where it could happen. I'll remove it.

client/ayon_core/lib/log.py
antirotor: honestly it is there to calm static type checkers. Problem is that since this is class method and the `_root_logger` is defined as `None`, it doesn't figure out `cls.initialize()` will do that. Since it is depending on the state, we should perhaps not use static/classmethods.

BigRoy: You can internally enforce a type hint to the variable as well. E.g. I believe you can technically silence it like this?
```python
value: logging.Logger = self._root_logger
return value
```

antirotor:
I've added to launcher some support for session id - how does it work:

UUID is generated for new session and stored as `AYON_SESSION_ID` and as a contextvar to the logger. If `AYON_SESSION_ID` already exists, it will append shortened one to tag the process chain - `a68bc729-49df-4022-b85a-28d1ca6ac08a:4f48ebd6:a1ef64b0`. This is capped so the ID won't grow endlessly. When hitting the cap, `.` is appended to mark it. There is constant to handle the chain length.

Ideally, this should go along the traces in the future, to connect everything together.

client/ayon_core/lib/log.py
antirotor: Seems like mypy is more clever than that so I opted for ignoring the error

Issue #1830, opened 2026-05-13
Browser: Refactor the table view filter to align it with the FE.


philippe-ynput:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Please describe the feature you have in mind and explain what the current shortcomings are?

- [ ] The front end uses hard-coded meta filters that we need to have too.
  * version: "featured","hero", "latest", "latest done" (exclusive)
  * author: "no author", "has author" (additional)
  * tags: "no tags", "has tags" (additional)
  * has reviewable: "yes", "no" (exclusive)
- [ ] Remove the Exclude button (it is a paid feature)
- [ ] Display a search field with completion instead of a list of existing options in the db.

### How would you imagine the implementation of the feature?

pretty well.

### Are there any labels you wish to add?

- [x] I have added the relevant labels to the enhancement request.

BigRoy:
> * author: "no author", "has author" (additional)
> * tags: "no tags", "has tags" (additional)
> * has reviewable: "yes", "no" (exclusive)

Adding these with https://github.com/ynput/ayon-core/pull/2054

PR #2031, opened 2026-09-01
Refactor Browser

### LINKED ISSUE #1611: Support folder type colors in desktop tools

Similar to Task Type colors the backend now also provides Folder Type colors - and matching to https://github.com/ynput/ayon-frontend/issues/1719 it'd be good to also get the colors to be used in the desktop tools.

Show the Folder colors in the UI. Similar to what https://github.com/ynput/ayon-core/pull/1438 did for Task Type colors.

_No response_

### Additional context:

Folder type colors exist in backend since https://github.com/ynput/ayon-backend/releases/tag/1.13.3

### LINKED ISSUE #1831: Browser: Add "my tasks" toggle

The current loader has "my tasks" toggle to filter the view.

- [ ] get a design from @mkolar and @Innders 
- [ ] implement that design, probably as a "stealth" view filter.

that remains to be seen.

_No response_

### Additional context:

_No response_

BigRoy:
## Changelog Description

- Separate legacy loader from new browser
- Show hero versions in browser
- Allow to switch versions in the row directly by double clicking version cell (like old loader)
- Add explicit working view defaults
- Add an interface contract so that custom addons can add new columns/filers, e.g. Site Sync.
- Add columns behavior now matches more with latest frontend
- Filter bar has been tweaked to match more with latest frontend
- Optimize queries for group by where possible
- Add informational overlays when nothing is selected.
- Add "In Scene" column functionality to show which versions are already loaded
- Add "Go to current context" button when in a DCC (when you have a context)
- Allow to hide/show columns + reorder columns
- Set `AYON_USE_LEGACY_LOADER=1` env var to use the legacy loader or toggle it in settings
- Refactor 'Review' terminology to 'Browser'

## Additional info

This won't be an easy one to review because well, it's massive. And it still requires a lot of cleanup - but I've focused on just throwing things together to get the UI to a functional state with admittedly spending a small amount of time only on cleanup.

It'd be great to review it by:
- Testing it
- Then do some glancing over the code and ask questions about things that are entirely stupid.
- Only once we're getting to a state that we're all somewhat happy I think we should sit down and do a thorough review of everything.

Also, I'm sure the tests are shit ;) I've not read them at all and I let AI do that along the side just to give it some wheels.

And, I'm so sorry!

Fix https://github.com/ynput/ayon-core/issues/1830
Fix https://github.com/ynput/ayon-core/issues/1831

### TODO

- [x] Filter bar should not extend the UI size
- [x] Add Columns menu has issues with the focus (doesn’t auto-expand)
- [x] Group By:
   - [x] Improve querying: probably ~~not query groups separately,~~ just query the versions and group them UI side.
   - [x] Show empty groups (does not filter out the empty groups?)
- [x] Fix invalid query being done on selecting a group (for the side panel)
- [x] Toggle between old or new browser via setting in core
    - [x] Allow an env var to override the setting
- [x] Add "My Tasks" filtering to the slicer?
- [x] Loader inside a DCC is not respecting Library attribute and shows all projects by default
- [x]  When Loader/Browser is launched for the very first time, it is empty (no folders). Closing it and running it again will populate the data, or selecting a different project and going back will refresh the data
- [x] ~~LoaderController + BrowserController should not both exist in the browser~~ separated between BrowserController & BrowserWidgetController (QObject for signals) for now; needs more @iLLiCiTiT review.
- [x] ~~Change the columns interface around so that addon doesn't need to expose a method that requires Qt - avoid the Qt requirement + maybe change it around for something else to 'register something with core', so there some `core.x.y.register_column_handler()`~~ Removed the public interface API for now
- [x] Add Count on group by on table view
- [x] views list UI cosmetics
- [x] Test site sync integration with custom columns
- [x] ~~Add a hardcore limit of never displaying more than e.g. 1000 versions with a little notice to recommend the user to perform tighter filtering to avoid tons of paginated queries?~~ Skipping for now because we're paginating when we can.
- [x] Test Power pack-less views access groups? --> https://github.com/ynput/ayon-core/pull/2056
    - [x] ~~Perhaps add additional license check for power-pack?~~ Tests for if it's available in bundle.
- [x] Editing of view does not work.  --> https://github.com/ynput/ayon-core/pull/2056
- [x] Fix Tasks "name" filter disappearing on folder switch due to tasks widget bottom left
- [x] Ensure all relevant toggles, like e.g. "My Tasks" and "Group By" are stored within the Views, and when resetting the working view to "default" that it also resets both of these.
- [ ] ~~Add a means to group by 'productGroup'~~ Possibly better with a server-side migration script that moves it to a regular attribute.
- [ ] Folder list in browser should not populate per folder expansion (to avoid issues with latency to server); instead spawn of background thread to query full rest folder hierarchy.
- [ ] Confirm Optional action still works in legacy loader (and in new loader)

## Testing notes:

1. Cry for a bit that you've been assigned reviewer here.
2. Test it functionally
3. Glance over code and ask questions, point out obvious things.

Only then after this PR has gone through initial tests, then we should go over thorough code review.


BigRoy (commented):
Getting this in Maya 2026:
```
#   File "C:\roydev\dev\ayon-core\client\ayon_core\ui\components\scroll_area.py", line 52, in paintEvent
#     get_ayon_style().drawComplexControl(
#   File "C:\roydev\dev\ayon-core\client\ayon_core\ui\style.py", line 282, in drawComplexControl
#     return super().drawComplexControl(cc, opt, p, w)
#            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "C:\roydev\dev\ayon-core\client\ayon_core\ui\style.py", line 352, in subControlRect
#     return sizer(cc, opt, sc, w)
#            ^^^^^^^^^^^^^^^^^^^^^
#   File "C:\roydev\dev\ayon-core\client\ayon_core\ui\drawers\scrollbar.py", line 192, in get_size
#     slider = self.get_size(
#              ^^^^^^^^^^^^^^
#   File "C:\roydev\dev\ayon-core\client\ayon_core\ui\drawers\scrollbar.py", line 146, in get_size
#     opt.minimum,
#     ^^^^^^^^^^^
# AttributeError: 'PySide6.QtWidgets.QStyleOptionComplex' object has no attribute 'minimum'
```

Getting this in Fusion with PySide2:
```
  File "C:\roydev\dev\ayon-core\client\ayon_core\ui\style.py", line 239, in polish
    self.style_widget(widget)
  File "C:\roydev\dev\ayon-core\client\ayon_core\ui\style.py", line 185, in style_widget
    widget.set_palette(
  File "C:\roydev\dev\ayon-core\client\ayon_core\ui\components\label.py", line 142, in set_palette
    self._configure_palette()
  File "C:\roydev\dev\ayon-core\client\ayon_core\ui\components\label.py", line 295, in _configure_palette
    for role in QPalette.ColorRole:
TypeError: 'Shiboken.EnumType' object is not iterable
```
And:
```

AttributeError: 'PySide2.QtWidgets.QStyleOptionComplex' object has no attribute 'minimum'
  File "C:\roydev\dev\ayon-core\client\ayon_core\ui\style.py", line 352, in subControlRect
    return sizer(cc, opt, sc, w)
  File "C:\roydev\dev\ayon-core\client\ayon_core\ui\drawers\scrollbar.py", line 146, in get_size
    opt.minimum,
```
Some of those PySide2 related issues came up in the past here: https://github.com/ynput/ayon-ui-qt/pull/53

mahesh-ynput:
- [x] For initial browser opening we should have <select project>  in project combo and project pin is not displaying.

client/ayon_core/addon/interfaces.py
```diff
@@ -39,6 +43,27 @@ class AYONInterface(metaclass=_AYONInterfaceMeta):
     log = None
 
 
+class IBrowserColumnAddon(AYONInterface):
```
iLLiCiTiT: I would remove this interface.

BigRoy:
> * For initial browser opening we should have   in project combo and project pin is not displaying.

Sorry, can you elaborate, bit hard to understand.

1. For initial browser opening we should have in project combo, what? It should auto-select the current context already if you launch it from a DCC?
2. We're lacking the "Pinned projects" in the projects combobox, that's correct. I'll fix that.

mahesh-ynput:
@BigRoy , With server 1.15.16+202607171343

It giving error:

> ERROR:ayon_core.ui.components.task_queue:Task fetch_page_root_0 failed: [{'message': "Unknown argument 'folderFilter' on field 'ProjectNode.versions'. Did you mean 'folderIds'?", 'locations': [{'line': 29, 'column': 7}]}, {'message': "Unknown argument 'latestPerFolder' on field 'ProjectNode.versions'.", 'locations': [{'line': 31, 'column': 7}]}]
>   File "C:\Users\mahesh\Desktop\contribute_to_ayon\PR\ayon-core\client\ayon_core\ui\components\task_queue.py", line 379, in _run_task_in_pool
>     result = task.function()
>              ^^^^^^^^^^^^^^^
>   File "C:\Users\mahesh\Desktop\contribute_to_ayon\PR\ayon-core\client\ayon_core\ui\components\table_model.py", line 1066, in <lambda>
>     function=lambda: self._fetch_page(
>                      ^^^^^^^^^^^^^^^^^
>   File "C:\Users\mahesh\Desktop\contribute_to_ayon\PR\ayon-core\client\ayon_core\tools\browser\ui\browser_controller.py", line 670, in fetch_versions_page
>     rows = self._fetch_versions_page_base(
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "C:\Users\mahesh\Desktop\contribute_to_ayon\PR\ayon-core\client\ayon_core\tools\browser\ui\browser_controller.py", line 902, in _fetch_versions_page_base
>     edges, page_info = self._get_versions_page(
>                        ^^^^^^^^^^^^^^^^^^^^^^^^
>   File "C:\Users\mahesh\Desktop\contribute_to_ayon\PR\ayon-core\client\ayon_core\tools\browser\ui\browser_controller.py", line 2342, in _get_versions_page
>     raise RuntimeError(resp.errors)
> RuntimeError: [{'message': "Unknown argument 'folderFilter' on field 'ProjectNode.versions'. Did you mean 'folderIds'?", 'locations': [{'line': 29, 'column': 7}]}, {'message': "Unknown argument 'latestPerFolder' on field 'ProjectNode.versions'.", 'locations': [{'line': 31, 'column': 7}]}]

~~But when I update to latest server I don't see this error that means this supposed have latest version of server?~~

Something wrong from my side it seems this only happening with specific project.

client/ayon_core/ui/components/filterable_list.py
```diff
@@ -107,8 +92,21 @@ def add_stretch(self) -> None:
...
+        for widget in widgets:
+            widget.hide()
+            self._items_layout.removeWidget(widget)
         self._items.clear()
-        self._items_layout.clear()
+
+        while self._items_layout.count():
+            item = self._items_layout.takeAt(0)
+            del item
+        return widgets
```
iLLiCiTiT:
```suggestion
        """Remove widgets."""
        self._items.clear()
        while self._items_layout.count():
            item = self._items_layout.takeAt(0)
            widget = item.widget()
            if widget is not None:
                widget.setVisible(False)
                widget.deleteLater()
```

1. `take_items` has no usage.
2. Current implementation of `take_items` actually does delete the widgets. If you call `removeWidget` it does internally call `delete itemAt(*widget pos*)` which does delete the widget from memory, you remain with only python object.

It is kinda useless to remove widgets you have in `self._items` and then remove other items from items layout in separate way.

client/ayon_core/ui/components/combo_box.py
```diff
@@ -416,6 +439,60 @@ def __init__(
...
+        press inside the popup distinguishes those two cases.
+        """
+        viewport = self.view().viewport()
+        if watched is viewport:
+            event_type = event.type()
+            press_types = {
+                QtCore.QEvent.Type.MouseButtonPress,
+                QtCore.QEvent.Type.MouseButtonDblClick,
+            }
+            if event_type in press_types:
```
iLLiCiTiT: This is actually faster as creating set is slow operation compare to tuple, it still has to go through each value.
```suggestion
            if event_type in (
                QtCore.QEvent.Type.MouseButtonPress,
                QtCore.QEvent.Type.MouseButtonDblClick,
            ):
```

client/ayon_core/ui/components/combo_box.py
```diff
@@ -416,6 +439,60 @@ def __init__(
...
+
+            elif event_type == QtCore.QEvent.Type.MouseButtonRelease:
+                index = self.view().indexAt(event.pos())
+                pressed_index = self._popup_pressed_index
+                self._popup_pressed_index = QtCore.QPersistentModelIndex()
+                if (
+                    pressed_index.isValid()
+                    and index.isValid()
+                    and pressed_index == index
+                ):
```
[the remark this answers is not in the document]
iLLiCiTiT: ☝️

client/ayon_core/ui/components/entity_thumbnail.py
```diff
@@ -602,10 +638,17 @@ def paintEvent(self, arg__1: QPaintEvent) -> None:
...
+            if not self.icon().isNull():
+                self.icon().paint(
+                    p,
+                    QRect(0, 0, size.width(), size.height()),
+                    Qt.AlignmentFlag.AlignCenter,
+                )
+        else:
+            get_ayon_style().drawControl(
+                QStyle.ControlElement.CE_PushButton, option, p, self
+            )
```
iLLiCiTiT:
```suggestion
        if not self._transparent:
            get_ayon_style().drawControl(
                QStyle.ControlElement.CE_PushButton, option, p, self
            )

        elif not self.icon().isNull():
            self.icon().paint(
                p,
                QRect(0, 0, size.width(), size.height()),
                Qt.AlignmentFlag.AlignCenter,
            )
```

client/ayon_core/ui/components/filter.py
```diff
@@ -504,7 +504,10 @@ def _remove_tag(self, key: str) -> None:
             return
 
         tag = self._tags.pop(key)
-        tag.setParent(None)
+        tag.hide()
+        tags_layout = self._tags_container.layout()
+        if tags_layout is not None:
+            tags_layout.removeWidget(tag)
         tag.deleteLater()
```
iLLiCiTiT:
```suggestion
        tag.setVisible(False)
        tags_layout = self._tags_container.layout()
        if tags_layout is not None:
            tags_layout.removeWidget(tag)
        else:
            tag.deleteLater()
```
Calling `removeWidget` does delete the widget (which is potential danger BTW because it does not handle possible signal/slot connections).
I do recommend using `setVisible(False)` instead of `hide()`. A lo easier to see and read intention.

client/ayon_core/ui/components/label.py
```diff
@@ -172,7 +184,9 @@ def set_icon(self, icon: str | None = None, color: str = "") -> None:
             elif same_as_bg:
                 icon_color = self._style_data["base"].get("color")
 
-            if self._icon == "none":
+            if isinstance(self._icon, QIcon):
+                icn = QIcon(self._icon)
```
iLLiCiTiT:
```suggestion
                icn = self._icon
```

client/ayon_core/ui/components/option_action.py
```diff
@@ -53,102 +50,191 @@ def __init__(
             fixed_width=False,
         )
 
+    def is_hovered(self, global_pos: QtCore.QPoint) -> bool:
+        """Return whether a global cursor position is inside the button."""
+        return self.rect().contains(self.mapFromGlobal(global_pos))
```
iLLiCiTiT:
```suggestion
        pos = self.mapFromGlobal(global_pos)
        if isinstance(pos, QtCore.QPointF):
            pos = pos.toPoint()
        return self.rect().contains(pos)
```

client/ayon_core/ui/components/combo_box.py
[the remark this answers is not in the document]
BigRoy: Why are all these copilot suggestions broken :/ why does it mess up the indentation. Now I need to go over all them manually to fix them - what a waste of time.

client/ayon_core/tools/browser/ui/window.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def _on_load_finished(self, event):
+        error_info = event["error_info"]
+        if not error_info:
+            self._controller.invalidate_loaded_containers()
+            self.browser_widget.refresh_loaded_state()
+            return
+
+        box = LoadErrorMessageBox(error_info, self)
+        box.show()
```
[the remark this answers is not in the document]
BigRoy: Ignoring this for now, it's same as old loader.

m-u-r-p-h-y:
Was testing it today and here are my observations

- When Loader/Browser is launched for the very first time, it is empty (no folders). Closing it and running it again will populate the data, or selecting a different project and going back will refresh the data
image

- Loader inside a DCC is not respecting Library attribute and shows all projects by default

- It only shows data on one level. It is quite confusing. You need to select assets/characters/robot to see something. If you select assets/characters it is empty

- Referencing model from Loader in Maya is not doing anything. No error given

BigRoy:
> * When Loader/Browser is launched for the very first time, it is empty (no folders). Closing it and running it again will populate the data, or selecting a different project and going back will refresh the data

Noticed that as well, that's because it doesn't actually initialize to the first project even though it "looks" selected. Will fix.

> Loader inside a DCC is not respecting Library attribute and shows all projects by default

Thanks, yes.

> It only shows data on one level. It is quite confusing. You need to select assets/characters/robot to see something. If you select assets/characters it is empty

Correct, this matches the OLD loader behavior - if you want to include the products of the children you must enable it in the Customize section. This is intentional so that we try to keep the majority of the queries as light as possible with the **default** behavior.

> Referencing model from Loader in Maya is not doing anything. No error given

Fixed ✅

client/ayon_core/tools/browser/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+if TYPE_CHECKING:
+    from .ui import BrowserWindow, LoaderWindow
+
+
+__all__ = (
+    "BrowserWindow",
+    "BrowserController",
+    "LoaderController",
+    "LoaderWindow",
```
iLLiCiTiT: We should not have LoaderController or LoaderWindow here at all. That's what loader tool has, this should have only BrowserController and BrowserWindow.

Also not sure why `__getattr__` is used here?

client/ayon_core/tools/browser/models/actions.py
```diff
@@ -0,0 +1,1113 @@
...
+
+        for repre_entity in repre_entities:
+            version_id = repre_entity["versionId"]
+            version_entity = version_entities_by_id[version_id]
+            product_id = version_entity["productId"]
+            product_entity = product_entities_by_id[product_id]
+            folder_id = product_entity["folderId"]
+            folder_entity = folder_entities_by_id[folder_id]
+
+            repre_context_by_id[repre_entity["id"]] = {
```
moonyuet, line 583:
```suggestion
            repre_id = repre_entity["id"]
            repre_context_by_id[repre_id] = {
```

client/ayon_core/ui/components/order.py
```diff
@@ -218,6 +218,30 @@ def current_order(self) -> list[str]:
...
+        """Apply an order without emitting :attr:`order_changed`."""
+        if order == self.current_order():
+            return
+        options_by_text = {
+            option._text: option
+            for option in self._options
+        }
+        if (
+            len(order) != len(options_by_text)
+            or set(order) != set(options_by_text)
```
moonyuet, line 231:
```suggestion
            set(order) != set(options_by_text)
```

client/ayon_core/ui/drawers/menu.py
```diff
@@ -369,18 +374,28 @@ def draw_menu_item(
...
-            check_icon.paint(painter, check_rect)
+            if option.checked or check_style != "checkmark":
+                check_icon = get_icon(
+                    (
+                        "check"
+                        if check_style == "checkmark"
+                        else "check_box"
+                        if option.checked
+                        else "check_box_outline_blank"
+                    ),
```
moonyuet: maybe we can use variable to store this or use `if...elif..else` store it But it is already many `if..elif..elif...elif..else` in this PR.
```python
icon_name =  (
	      "check"
	      if check_style == "checkmark"
	      else "check_box"
	      if option.checked
	      else "check_box_outline_blank"
)
```
or
```python
if check_style == "checkmark":
    icon_name = "check"
elif option.checked:
    icon_name = "check_box"
else:
    icon_name = "check_box_outline_blank"
```

client/ayon_core/tools/browser/ui/browser_controller.py
```diff
@@ -0,0 +1,2683 @@
...
+                key_aliases.get(str(item.key), str(item.key)),
+                [str(value) for value in item.values],
+                bool(item.use_substring),
+            )
+            for item in criteria
+            if item.values
+        ]
+        self._reset_pagination()
+
+    def _get_query_filters(self) -> dict[str, Any]:
```
moonyuet, line 603: @BigRoy Guess we should come up with some clean up with this function to make it more readable?

client/ayon_core/tools/browser/ui/browser_queries.py
```diff
@@ -0,0 +1,324 @@
...
+        task_fields.append("status")
+    if "taskTags" in keys:
+        task_fields.append("tags")
+    if "folderName" in keys:
+        folder_fields.append("label")
+    if "folderStatus" in keys:
+        folder_fields.append("status")
+
+    if keys.intersection({"product/version", "path"}):
+        version_fields.append("path")
```
moonyuet, line 162: Could it be
```suggestion
    fields_mapping = {
        "thumb": ("version", "thumbnailId"),
        "status": ("version", "status"),
        "author": ("version", "author"),
        "createdAt": ("version", "createdAt"),
        "updatedAt": ("version", "updatedAt"),
        "tags": ("version", "tags"),
        "featuredVersionType": ("version", "featuredVersionType"),
        "productType": ("product", "productType"),
        "productBaseType": ("product", "productBaseType"),
        "productStatus": ("product", "status"),
        "taskType": ("task", "taskType"),
        "taskStatus": ("task", "status"),
        "taskTags": ("task", "tags"),
        "folderName": ("folder", "label"),
        "folderStatus": ("folder", "status"),
    }
    for key_values, (scope, field_name) in fields_mapping.items():
        if key_values in keys:
            if scope == "version":
                version_fields.append(field_name)
            elif scope == "product":
                product_fields.append(field_name)
            elif scope == "task":
                task_fields.append(field_name)
            elif scope == "folder":
                folder_fields.append(field_name)
            elif key_values.intersection({"product/version", "path"}):
                version_fields.append("path")

```

BigRoy: I understand from a readability perspective - but to me it doesn't make a massive improvement on readability but it would technically be more performant the way it's now since we only create the string values on a match instead of prepping a full dict, and then still doing similar check - it's not such a major cosmetic cleanup that I'd really care. So will keep as is for now, unless we find an even more streamlined approach.

client/ayon_core/tools/browser/ui/browser_controller.py
BigRoy: At some point - yes, but not a priority honestly - more important that it works and that it's fast. Might do that once we get to the thorough code cleanup here though. Needs the features/bugs resolved first :D

client/ayon_core/ui/components/order.py
BigRoy: This was actually added due to copilot describing there's an edge case if we don't. Admittedly I have a hard time finding those Copilot comments now. I'll need to look that up again 🗡️

client/ayon_core/tools/browser/models/actions.py
BigRoy: Value is not used elsewhere below, so I'm fine with the way it is.

client/ayon_core/ui/components/buttons.py
```diff
@@ -193,7 +230,7 @@ def set_icon(self, icon_name: str):
                 icon_name_off=self._icon,
                 color_off=self._icon_color,
                 icon_name_on=self._icon_on,
-                color_on=self._icon_color,
+                color_on=self._icon_checked_color,
```
iLLiCiTiT: Shouldn't that be named `._icon_color_on`?

client/ayon_core/ui/components/buttons.py
```diff
@@ -298,14 +360,35 @@ def __init__(
...
+            self._dropdown = ButtonMenuDropdown(
+                self, container_margin=dropdown_margin
+            )
+        else:
+            self._dropdown = ButtonMenuDropdown(
+                self,
+                frame_variant=dropdown_variant,
+                container_variant=dropdown_variant,
+                container_margin=dropdown_margin,
+            )
```
iLLiCiTiT:
```suggestion
        kwargs = {}
        if dropdown_variant is not None:
            kwargs = dict(
                frame_variant=dropdown_variant,
                container_variant=dropdown_variant,
            )
        self._dropdown = ButtonMenuDropdown(
            self,
            container_margin=dropdown_margin,
            **kwargs
        )
```

client/ayon_core/ui/components/buttons.py
```diff
@@ -325,4 +412,9 @@ def _on_button_clicked(self) -> None:
     def _on_popup_closed(self) -> None:
         """Update state when the popup signals it has closed."""
         self._menu_open = False
+        if QtWidgets.QApplication.mouseButtons() & Qt.MouseButton.LeftButton:
+            local_pos = self.mapFromGlobal(QtGui.QCursor.pos())
```
iLLiCiTiT, line 416:
```suggestion
            local_pos = self.mapFromGlobal(QtGui.QCursor.pos())
            if isinstance(local_pos, QtCore.QPointF):
                local_pos = local_pos.toPoint()
```

client/ayon_core/ui/components/card_view.py
```diff
@@ -266,6 +293,20 @@ def _source_model(self) -> QtCore.QAbstractItemModel | None:
...
         return model
 
+    def _fetch_more(self, index: QModelIndex) -> None:
+        """Fetch through the source model after card expansion approval."""
+        model = self.model()
+        if isinstance(model, QSortFilterProxyModel):
+            source = model.sourceModel()
+            if isinstance(source, PaginatedTableModel):
+                source.fetch_more(model.mapToSource(index))
+                return
```
iLLiCiTiT, line 305:
```suggestion
        while isinstance(model, QSortFilterProxyModel):
            model = model.sourceModel()
```

client/ayon_core/ui/components/card_view.py
```diff
@@ -514,7 +567,13 @@ def _calculate_tree_layout(
             )
             label_color = meta_idx.data(Qt.ItemDataRole.ForegroundRole)
             label_icon = meta_idx.data(Qt.ItemDataRole.DecorationRole)
-            child_count = model.rowCount(group_idx)
+            row_data = group_idx.data(Qt.ItemDataRole.UserRole) or {}
```
iLLiCiTiT, line 572: Can I use `AYCardView` with custom model? If yes, then why it's using `Qt.ItemDataRole.UserRole` to get UI related data?

That is potential danger for cusom models using `UserRole` for different purposes. Also how do we make sure that the data are actually stored there?

I didn't dig enough to find out what `child_count`, `group_percentage` and `group_metrics` are comming from and what are representing, but this doesn't look good unless the model is strictly defined.

client/ayon_core/ui/components/card_view.py
```diff
@@ -1040,7 +1123,7 @@ def refresh_visible_editors(self) -> None:
             self._active_editor_pmis.discard(QPersistentModelIndex(idx))
         self._schedule_editor_sync()
 
-    def _maybe_fetch_more(self) -> None:
+    def fetch_more_if_needed(self) -> None:
```
iLLiCiTiT, line 1128: I don't think this should be publicly available.

client/ayon_core/ui/components/card_view.py
```diff
@@ -1069,7 +1152,7 @@ def _maybe_fetch_more(self) -> None:
         root_pmi = QPersistentModelIndex(root)
         if near_bottom and root_pmi not in fetched:
             if model.canFetchMore(root):
-                model.fetchMore(root)
+                self._fetch_more(root)
```
iLLiCiTiT, line 1157: Why to change this to `_fetch_more` instead of `fetchMore` defined by Qt's `QAbstractItemModel`?

client/ayon_core/ui/components/filterable_list.py
iLLiCiTiT:
```suggestion
        self._items.clear()
        while self._items_layout.count():
            item = self._items_layout.takeAt(0)
            widget = item.widget()
            if widget is not None:
                widget.setVisible(False)
                widget.deleteLater()
```
I think this was already suggested.

client/ayon_core/ui/components/frame.py
```diff
@@ -50,3 +55,74 @@ def get_bg_color(self, base_color: str):
...
+            widget.installEventFilter(self)
+        return self
+
+    def eventFilter(
+        self, obj: QtCore.QObject, event: QtCore.QEvent
+    ) -> bool:
+        if event.type() == QtCore.QEvent.Type.Enter:
+            self._frame.set_hovered(True)
+        elif event.type() == QtCore.QEvent.Type.Leave:
+            local = self._frame.mapFromGlobal(QtGui.QCursor.pos())
```
iLLiCiTiT, line 164:
```suggestion
            local = self._frame.mapFromGlobal(QtGui.QCursor.pos())
            if isinstance(local, QtCore.QPointF):
                local = local.toPoint()
```

client/ayon_core/ui/components/frame.py
```diff
@@ -50,3 +55,74 @@ def get_bg_color(self, base_color: str):
...
+            if not self._frame.rect().contains(local):
+                self._frame.set_hovered(False)
+        return False  # never consume the event
```
iLLiCiTiT:
```suggestion
        # never consume the event
        return False
```

client/ayon_core/ui/components/frame.py
```diff
@@ -18,20 +18,25 @@ def __init__(
...
         bg_tint="",
+        hover_enabled=False,
         **kwargs,
     ):
         # Convert enum to string if needed
         self._bg: bool = bg
         self._variant_str = variant.value
         self._bg_tint = bg_tint
         self._bg_color = None
+        self._hover_enabled = hover_enabled
```
iLLiCiTiT, line 31: No reason to store the value.

Considering what it does, I'd rather have helper method that would do that instead of achieve it using args.

client/ayon_core/ui/drawers/frame.py
```diff
@@ -95,6 +108,12 @@ def draw_frame(
             )
 
         if radius:
+            # Clamp to a single value derived from the shorter side so
+            # both corner axes stay circular. Qt's own clamping caps
+            # each axis independently against half its own dimension,
+            # which turns an oversized radius on a wide-but-short rect
+            # (e.g. a pill chip) into an ellipse instead of a stadium.
+            radius = min(radius, draw_rect.width() / 2.0, draw_rect.height() / 2.0)
```
iLLiCiTiT: Multiply is faster then divide
```suggestion
            radius = min(radius, draw_rect.width() * 0.5, draw_rect.height() * 0.5)
```

client/ayon_core/ui/color_utils.py
```diff
@@ -131,5 +142,35 @@ def compute_color_for_contrast(
             if white_ratio > black_ratio
             else QColor(0, 0, 0, original_alpha)
         )
-    # print(f"Final ratio: {final_ratio}: {result.name()}")
-    return result
+    return tuple(result.toTuple())
```
iLLiCiTiT, line 145: This could just return the color? The `compute_color_for_contrast` takes the tuple and converts it to QColor.

client/ayon_core/ui/utils.py
```diff
@@ -195,10 +195,8 @@ def clear_layout(layout):
         if widget:
             # Recursively clear any layouts this widget might have
```
iLLiCiTiT: Outdated comment

client/ayon_core/tools/utils/host_tools.py
```diff
@@ -84,27 +85,45 @@ def show_workfiles(
 
         return workfiles_tool
 
-    def get_loader_tool(self, parent):
+    def get_loader_tool(self, parent, *, use_context: bool = False):
```
iLLiCiTiT: Why was `use_context` added?

client/ayon_core/tools/utils/host_tools.py
```diff
@@ -540,3 +562,31 @@ def get_pyblish_icon():
...
+
+        project_name = get_current_project_name()
+        if project_name:
+            settings = get_project_settings(project_name)
+        else:
+            settings = get_studio_settings()
+        use_legacy = settings["core"]["tools"]["loader"].get(
+            "use_legacy_loader", False
+        )
+    return use_legacy
```
iLLiCiTiT:
```suggestion
    if use_legacy is not None:
        return use_legacy

    from ayon_core.pipeline import get_current_project_name
    from ayon_core.settings import (
        get_project_settings,
        get_studio_settings,
    )

    project_name = get_current_project_name()
    if project_name:
        settings = get_project_settings(project_name)
    else:
        settings = get_studio_settings()
    return settings["core"]["tools"]["loader"].get(
        "use_legacy_loader", False
    )
```

client/ayon_core/tools/utils/projects_widget.py
```diff
@@ -567,10 +567,11 @@ def __init__(
         controller: AbstractProjectController,
         parent: QtWidgets.QWidget,
         handle_expected_selection: bool = False,
+        variant: AYComboBox.Variants = AYComboBox.Variants.Default,
```
iLLiCiTiT, line 573: Does it make sense to allow change of the variant on project combobox?

client/ayon_core/ui/components/table_view.py
```diff
@@ -73,8 +73,28 @@ def __init__(
         self._style_model = style_model
         self._variant_str = variant
         self._toggle_btn: QToolButton | None = None
+        # Keys of columns the user has flagged as *pinned* (kept at the
+        # left in the visual order).  Phase 1 only roundtrips this set;
+        # visual freeze rendering is added in Phase 3.
+        self._pinned_keys: set[str] = set()
```
iLLiCiTiT, line 79: What is this for? The comments describes "column" the variable is name "keys" and `AYTableHeader` does store it but does not use it.

moonyuet:
The sorting works now. 
image
image
image

client/ayon_core/tools/utils/projects_widget.py
BigRoy: Not sure - but there currently is a place where we use the variant override to force a different style - so we do have  use case. Browser slicer uses the Low variant.

client/ayon_core/ui/color_utils.py
BigRoy: It's merely for the `lru_cache` to cache the result to something immutable. Because the `QColor` being returned otherwise CAN change in-place making the cached result essentially wrong.

client/ayon_core/ui/components/table_view.py
BigRoy: I think this was for the pinning in the table view (I suspect it came from Philippe's preparation work) where you can stick columns so the left side in the view. I'd say this is much less relevant for the desktop tool so I'm inclined to just remove it instead of finishing the work now, we can add it in separate PR later. Thoughts?

client/ayon_core/ui/components/frame.py
BigRoy: It's actually read here: https://github.com/ynput/ayon-core/pull/2031/changes/BASE..d26b140aa72ccc7a2c9828ea57583ad71e33aacd#diff-712c6aec1802d7eb8bec4b1d66b734e80eeddb2f00850ea7fb7ba90ea3a3722bR51

This drawer logic is killing me so I have hard time quickly guessing whether we could live without the attribute. Need me to look into whether that can be simplified? 🤔 At least exposing a helper method for it could be nice though, but we'd likely still set the value on `__init__`, etc.?

client/ayon_core/tools/utils/host_tools.py
BigRoy: Removed again - I initially thought these were passed somehow before, but couldn't get it to work with the browser - then later learned that legacy loader just always seemed to apply the context on start if it had any and I think I forgot to clean this up.

BigRoy: Ah damn it - and now I see in the diff that it did exist on `show_loader` but not on `get_loader_tool` :( bit annoying. So, only keep on `show_loader` then?

client/ayon_core/ui/components/card_view.py
BigRoy: Hmm, valid question - currently I don't think it's fully prepared for that.

These values are the group by section values for showing the percentages and amount of child entries that section has.

client/ayon_core/ui/components/card_view.py
BigRoy: Currently called by the main view: https://github.com/ynput/ayon-core/blob/cbbd0840c564b510c3d1d29cbbd3461033f898de/client/ayon_core/tools/browser/ui/_browser_table.py#L1249

Will need to look into this some more.

client/ayon_core/ui/components/card_view.py
BigRoy: This is calling it on itself which is an item view, which does not have `fetchMore` so it'd almost look like it's a Qt method then but it isn't. This is a helper method that triggers the `fetchMore` call when you're at a spot in the card view where it should start to paginate again (only load more when needed). (Although I think that currently only works reliably in group by mode, but that's the intent.)

Why it stops calling `model.fetchMore()` directly I need to double check.

kalisp:
Tested with SiteSync, looked OK. (Tested upload/download too.)

moonyuet:
When I tested with the latest commit of the browser, I find out that when I select task it would be at the filter, which is good. A when switching to other folder type without the same task name which is already at the filter, it would show nothing, which makes sense. But it is a bit weird in terms of experience. We can just remove filter for task if there is no exactly same task type when switching the folder type.
https://github.com/user-attachments/assets/e0e4a623-bd26-485e-8d0d-c24b55102236

BigRoy:
> When I tested with the latest commit of the browser, I find out that when I select task it would be at the filter, which is good. A when switching to other folder type without the same task name which is already at the filter, it would show nothing, which makes sense. But it is a bit weird in terms of experience. We can just remove filter for task if there is no exactly same task type when switching the folder type.
> [github.com/user-attachments/assets/e0e4a623-bd26-485e-8d0d-c24b55102236](https://github.com/user-attachments/assets/e0e4a623-bd26-485e-8d0d-c24b55102236)

I understand the point of view - but the filters are intended to be static from the context switches. So I'm not sure it's a good idea. I actually intentionally put in effort to ensure it doesn't drop it on context switches.

client/ayon_core/lib/icon_definitions.py
```diff
@@ -59,19 +59,25 @@ def to_data(self) -> dict:
 
 @dataclass
 class MaterialSymbolsIcon(IconBase):
-    """Material Symbols icon."""
+    """Material Symbols icon.
+
+    Icons are outlined by default, which is how the AYON web frontend
+    draws them. Pass ``fill=True`` for the filled variant.
+    """
```
iLLiCiTiT:
```suggestion
    """Material Symbols icon."""
```

client/ayon_core/lib/log.py
```diff
@@ -264,3 +279,50 @@ def get_process_name(cls):
...
+                "TIMER:  %s :: %s took %.3f seconds",
+                func_name,
+                message,
+                elapsed,
+            )
+else:
+    # Dummy context manager if timing is not enabled
+    @contextmanager
+    def log_timing(message: str) -> Generator[None, None, None]:
+        yield
```
iLLiCiTiT, line 321:
```suggestion
# Dedicated logger for timing with a concise format
_log_timing_enabled = env_value_to_bool("AYON_CORE_TIMERS")
_timing_logger = Logger.get_logger("ayon-core-timers")
_timing_logger.setLevel(logging.INFO)

@contextmanager
def log_timing(message: str) -> Generator[None, None, None]:
    """Context manager to log the execution time of a code block.

    Args:
        message (str): Description of the operation being timed.

    Yields:
        None

    Example:
        with log_timing("Loading activities"):
            # Your code here
            data = fetch_data()
    """

    if not _log_timing_enabled:
        yield
        return

    # Get the caller's function name with simple error handling
    func_name = "unknown"
    try:
        frame = inspect.currentframe()
        if frame and frame.f_back and frame.f_back.f_back:
            func_name = frame.f_back.f_back.f_code.co_name
    except (AttributeError, ValueError):
        pass  # Use default "unknown" if inspection fails

    start = time.perf_counter()
    try:
        yield
    finally:
        elapsed = time.perf_counter() - start
        _timing_logger.info(
            "TIMER:  %s :: %s took %.3f seconds",
            func_name,
            message,
            elapsed,
        )
```

BigRoy: Fine with me - feel free to apply.

PR #2034, opened 2026-09-02
Delete Versions: Allow to delete selected versions

### LINKED ISSUE #1804: YN-0686: Loader delete versions

Delete old versions action does suffer from caching too much data into form attribute values. The json string can be shortened with text def so it is cut off meaning that action trigger fails to run because it can't parse the form data.

<br>

Also it is not possible to delete one specific version which might be nice addition. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1491717113851674644)

_This issue was automatically created from AYON ticket [YN-0686]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=359b09c03e4311f195fcdd07044d01bc)

tadeas-hejnic:
## Changelog Description
Adding `Delete Selected Versions` action in browser.

## Additional info
image

## Testing notes:
1. select versions 
2. remove them
3. check if everything went correctly


client/ayon_core/plugins/loader/delete_old_versions.py
```diff
@@ -52,6 +52,15 @@ def get_action_items(
         }
 
         return [
+            LoaderActionItem(
+                label="Delete Selected Versions",
+                order=36,
+                data={
+                    "version_ids": [version["id"] for version in versions],
```
iLLiCiTiT: I would avoid this, there is a limit of characters to store into data (bug). You have access to the selection in execute method, no need to store it.
```suggestion
```

client/ayon_core/plugins/loader/delete_old_versions.py
```diff
@@ -104,12 +123,37 @@ def execute_action(
                 selection,
             )
 
-        if step == "delete-versions":
+        if step in {"delete-versions", "delete-selected-versions"}:
```
iLLiCiTiT: ~~Where will `"delete-versions"` get the `version_ids` from? That needs separate handling.~~

Oh, you changed that place too, ok.

client/ayon_core/plugins/loader/delete_old_versions.py
tadeas-hejnic:
```suggestion
                label="Delete Old Versions"
```

Maybe this should be changed from `Delete Versions` to `Delete Old Versions` to make the action clearer.

client/ayon_core/plugins/loader/delete_old_versions.py
```diff
@@ -325,11 +413,11 @@ def _get_delete_form(
         size: int,
         remove_publish_folder: bool,
         version_ids: list[str],
-        repre_ids_by_version_id: dict[str, list[str]],
-        filepaths_by_repre_id: dict[str, list[str]],
         repeated: bool = False,
+        title: str = "Delete versions",
+        step: str = "delete-versions",
```
iLLiCiTiT: There is no difference in `"delete-selected-versions"` and `"delete-version"` step. I would also keep the same title as `"Delete versions"`.

client/ayon_core/plugins/loader/delete_old_versions.py
iLLiCiTiT: We might also define `group_label` as `"Delete Versions"` and labels would be `"Selected versions"` and `"Old versions"`.

tadeas-hejnic:
image

Group label added. Ready to merge now? @iLLiCiTiT

Or should I put the `Calculate versions size` also to the group?

iLLiCiTiT (approved):
Looks good. @BigRoy can you take a quick look?

BigRoy (approved):
Clicking the entry works for me.

**However** for whatever reason it's not in a submenu inside the context menu for me:

Image

@iLLiCiTiT you may know why?
Would be nice if we can check that.

Tiny side note is that when you trigger it the version remains shown in the loader UI until you refresh. I wonder if there's a way to trigger a refresh to the loader itself. But maybe that's more of a problem to solve with the new browser at some point.

iLLiCiTiT:
> However for whatever reason it's not in a submenu inside the context menu for me:

Looks like you have older commint. The labels are also different.

> Tiny side note is that when you trigger it the version remains shown in the loader UI until you refresh.

I think we'll have the same issue in browser too, but woud not bother with it right now.

PR #2035, opened 2026-09-02
YN-0813 Add "promised_tasks" support in CreatedInstance and Publish UI

### LINKED ISSUE #184: YN-0813: Traypublisher: Editorial Simple — task/folder no longer settable on shots, and publish fails on task lookup

found a regression for the Editorial Simple Creator in the Tray Publisher. We used to be able to set the destination task for the created shots, but it is currently greyed out (see image). This causes issues when the anatomy template contains {task} key, which then errors in Integrate Asset, because the task doesn't exists. The issue seems to have been adding in the 0.4.* release of the traypublisher addon. I've attached the quick debug claude did to isolate the issue. 

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1509927657394016256/1509927730601263154)

_This issue was automatically created from AYON ticket [YN-0813]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=69b08000641d11f189e9676172ae74ce)

rdelillo:
## Changelog Description

fixes: https://github.com/ynput/ayon-traypublisher/issues/184

With these changes, created instance with `has_promised_context` can also provide an optional `promised_tasks` transient data `list[str]` that contains the name of the tasks that will be created in the promised context. This way the user can edit the target task and get this confirmed from the combo box like any other instance.

## Additional info
image

## Testing notes:
1. To be tested with https://github.com/ynput/ayon-traypublisher/pull/196


client/ayon_core/tools/publisher/widgets/product_context.py
```diff
@@ -727,7 +766,7 @@ def _on_submit(self):
         changes_by_id = {}
         for item in self._current_instances_by_id.values():
             # Ignore instances that have promised context
-            if item.has_promised_context:
+            if item.has_promised_context and not item.promised_tasks:
                 continue
 
             instance_changes = {}
```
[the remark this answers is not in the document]
rdelillo: fixed but not related to this changes, but this is a regression introduced in another PR.

client/ayon_core/tools/publisher/widgets/product_context.py
```diff
@@ -716,7 +752,9 @@ def _on_submit(self):
             variant_value = self.variant_input.get_value()[0]
 
         if self.folder_value_widget.has_value_changed():
-            folder_path = self.folder_value_widget.get_selected_items()[0]
+            folder_path = tuple(
+                self.folder_value_widget.get_selected_items()
+            )[0]
```
rdelillo, line 757: Bug. Function return was changes from a list to a set:
https://github.com/ynput/ayon-core/commit/533ad2b6891cd70462e33de28f15ef16a74b15a8#diff-36083c8981633e2463c2fc1b95ed74109601fda539d8e014a5bd6ac40c7b68bb

rdelillo:
Not 100% sure this is the correct approach, let's discuss.

client/ayon_core/tools/publisher/widgets/product_context.py
```diff
@@ -416,29 +431,46 @@ def confirm_value(self, folder_paths):
...
+                folders excluded (their tasks come from
+                'promised_task_lists').
+            promised_task_lists (list[list[str]] | None): Allowed task names
+                per promised-context instance.
         """
         self._proxy_model.set_filter_empty(False)
         self._proxy_model.sort(0)
 
-        if folder_task_combinations is None:
-            folder_task_combinations = []
```
iLLiCiTiT, line 430: Why was this removed and replaced by `or []` everywhere?

client/ayon_core/tools/publisher/widgets/product_context.py
iLLiCiTiT, line 757:
```suggestion
            folder_path = next(iter(
                self.folder_value_widget.get_selected_items()
            ))
```

iLLiCiTiT:
Shouldn't that be a create attribute instead? I'm 99% convinced it would be a lot better than adding another thing to the instance object and publishing shebank. It would be fully in hands of create plugin, also allowing completelly custom task name if it would be text input.

I kinda don't like that we do add another part of the logic to the core publishing that has very specific use-case which has other ways how to handle them. The promised context was added in mind that the create plugin is in full charge of the context, this does break the assumption and it is partial. Also showing tasks in tasks widget that do not exist yet is also not ideal.

iLLiCiTiT:
I also don't fully understand the issue. Is the issue with instances that already have existing folder, or with new instances without existing folders?

Long story short: Is it aboud being able to create new task, or just swap between existingtasks?

jakubjezek001 (approved):
works well

rdelillo:
> Shouldn't that be a create attribute instead? I'm 99% convinced it would be a lot better than adding another thing to the instance object and publishing shebank. It would be fully in hands of create plugin, also allowing completelly custom task name if it would be text input.
> 
> I kinda don't like that we do add another part of the logic to the core publishing that has very specific use-case which has other ways how to handle them. The promised context was added in mind that the create plugin is in full charge of the context, this does break the assumption and it is partial. Also showing tasks in tasks widget that do not exist yet is also not ideal.

All totally fair points, and as I mentioned, is my PR the best approach ? debatable.. 

To detail a bit more, the request/use-case come from editorial creators.
User wants to select dynamically the target task of the `plate`, `review`, `audio` from the Publish UI combo box for shot+task that might not exist yet.

- from an EDL/Timeline/stuff the creator generates a lot of interconnected instances grouped per shot(s)
- `plate`, `audio`, `review`... products are associated to 1 `shot` instance corresponding to a folder that might or might not already exists + an optional task
- a `shot` instance might contain additional list of task to be created along the folder, this comes from the editorial addon settings (this is where the available task list comes from)
- everything is currently flagged as `has_promised_context`, so if a instance (shot+task) is missing on publish, it gets created on the flight, if it's already there we just use it.


> Shouldn't that be a create attribute instead?

Maybe ? How do you see that ? The creator part is done once the `plate`/`review`/`audio` instances are created. Also, we cannot just have 1 text field for everything, the creator works on the all timeline, when the user want to adjust this per instance.

> The promised context was added in mind that the create plugin is in full charge of the context, this does break the assumption and it is partial. 

Exactly that, yes. We would break the read-only promised context and offer user options to edit it on the fly on the instance.
Is that a good idea ?

> Long story short: Is it aboud being able to create new task, or just swap between existing tasks?

Both, 
You ingest a timeline once, nothing exist yet, you want to select the destination task from a list of tasks that do not exist yet.
You re-ingest the same timeline once more, shots/tasks now exist, you want to select a destination task from a list of existing tasks.

iLLiCiTiT:
Ok, I don't think the task input should allow you to choose a task that might not yet exist. I know it is the place where to look to change context but it should show only available options and should be disabled if the promised context is set.

I would instead add task name enum definition to create plugin definitions. You can listen to the changes and update task name (with product name) in the context on the go but using enum defined by the creator.

And sorry to say this, but I would like to avoid complicating the publisher's UI code with this. It's already too much complicated (I'd say).

rdelillo:
> And sorry to say this, but I would like to avoid complicating the publisher's UI code with this. It's already too much complicated (I'd say).

Don't be sorry, it's true.

So you'd prefer we add an enum to select the task in the creator `get_attr_defs_for_instance`, right ?
Implementation-wise this is indeed much simpler and the code can stay in the relevant editorial addons.

I anticipate UI/UX-wise, some people might wonder why this is not the usual task combo box, but I guess we could add a tooltip or something to detail a bit.

This works for me, would be interested in @BigRoy @jakubjezek001 feeling on this.

BigRoy:
I agree with @iLLiCiTiT here and loops back to what I initially proposed briefly @rdelillo to just expose the field as a publish attribute.

The trickiest thing I see here is that our client in this regard wants to select an **existing task name** from a dropdown, instead of typing it manually. But with a promised context, there is no existing task? (and possibly no folder either) as such all we could really off is a text field for someone to choose a task name, but then ... also a task type? And complexity grows easily.

In this case you've already computed the `promised_tasks` for the user to pick from, so likely we could offer the same in an `EnumDef` attribute. However, if the user really always wants to toggle this to another doesn't that actually hint there is a configuration issue for how these are ingested? Like, would we really want to make it the responsibility of the artist to always toggle these? Shouldn't it then just default to the other task name? Which is likely configurable for the editorial workflow?

I'm a bit lost as to why they _sometimes_ want to edit the target task. Anyway, I do agree that having it just as an attribute.

The way I understand:

- The Editorial Creator defines which tasks to add to a target shot (from settings)
- By default however for these instances it seems that for this client it's targeting the wrong task.
- Should this then just be configurable which task it targets?

Wouldn't that already solve it, without their artists needing to manually toggle things?

I do think it'll look confusing if the top section of the publisher UI shows a different task name then it'll actually end up applying to because it's actually being overridden/set by the publish attribute. Which is the only reason that I think this PR's approach is nicer, it's just _slightly nicer_ UX for the artists. Unless the promised context is always displayed empty for these ones (which seems like should be allowed too)

jakubjezek001:
Way I see it from what I was reading here: agree with @iLLiCiTiT, there is no need to add new complexities to core. 

> - The Editorial Creator defines which tasks to add to a target shot (from settings)

We could also add enumerator with multi-selection for admin to select from available products base types. So those will be automatically assigned on creation.

> - By default however for these instances it seems that for this client it's targeting the wrong task.

This could be solved by auto assigning from above. Each {audio, plate, review} product based instance would have available instance attribute, which would be single select enumerator - by default set to what ever was set in the settings. User could override it if he/she needs to. 

> - Should this then just be configurable which task it targets?

Yes, that is main purpose as I understood it.

rdelillo:
Thanks for the feedback. 
Closing this PR as we'll implement the changes directly in the client addon, see: https://github.com/ynput/ayon-traypublisher/pull/198

PR #2040, opened 2026-09-04
Custom frame range: Plugin works with specific family

iLLiCiTiT:
## Changelog Description
The custom frame plugin does work with specific family `"custom.frame.range"`, but also has backwards compatibility for the families that were defined by deadline plugin but only for hosts that did work with deadline addon.

## Additional info
Using this approach it still works for hosts that did work before, but hosts that would like to use this new feature has to add `"custom.frame.range"` to instance to be able to show the option.

## Testing notes:
1. Validate the proposed logic if actually makes sense!!!
2. For hosts and families for which it was already working it should work.
3. Add the `"custom.frame.range"` family to an instance in some other host to validate it shows for it too.

### DIFF
--- a/client/ayon_core/plugins/publish/collect_custom_frame_range.py
+++ b/client/ayon_core/plugins/publish/collect_custom_frame_range.py
@@ -1,30 +1,68 @@
 """Collect custom frame range for render submission."""
+from __future__ import annotations
+
+import typing
+
 import pyblish.api
-from typing import Optional
+
 from ayon_core.lib import EnumDef, TextDef
-from ayon_core.pipeline import KnownPublishError
+from ayon_core.pipeline import KnownPublishError, get_current_host_name
 from ayon_core.pipeline.publish import AYONPyblishPluginMixin
 
+if typing.TYPE_CHECKING:
+    from ayon_core.pipeline import CreatedInstance
 
-class CollectCustomFrameRange(pyblish.api.InstancePlugin,
-                              AYONPyblishPluginMixin):
+# For backwards compatibility with deadline's logic
+# - Combination of families and host names for which the plugin is compatible.
+# - To properly convert the hosts to use the "new way" they have to add
+#   "custom.frame.range" family to created instances.
+# - We can remove host names from the 'FARM_HOST_NAMES' set once the host
+#   integrations are updated to add the family to instances and core has
+#   bumped compatibility version for the host.
+FARM_FAMILIES = {
+    "render", "render.local", "render.farm", "render.frames_farm",
+    "prerender", "prerender.farm", "prerender.frames_farm",
+    "renderlayer", "imagesequence", "image",
+    "vrayscene", "maxrender",
+    "arnold_rop", "mantra_rop",
+    "karma_rop", "vray_rop", "redshift_rop",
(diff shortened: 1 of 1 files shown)


moonyuet:
Tested with https://github.com/ynput/ayon-3dsmax/pull/163, works with me.

iLLiCiTiT:
> Tested with https://github.com/ynput/ayon-3dsmax/pull/163, works with me.

Nice! @BigRoy do you think this is good idea?

BigRoy (approved):
LGTM

I wonder if the family should actually be something like `framerange.custom` instead? 🤔 Fine with it either way - just feel like since we're pushing out more families as we intended that we need some naming rules? 🗡️

LiborBatek:
I have been using it while testing the `ayon-3dsmax` and its been working together well...

iLLiCiTiT:
> just feel like since we're pushing out more families as we intended that we need some naming rules? 🗡️

I 100% agree, that's why I asked for you to take a look 🙂 . My initial idea was to use `customizable.frame.range` which exactly describes what "can" happen instead. Not sure if that is good idea tho...

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -111,12 +148,31 @@ def _is_custom_frames_used(cls, value) -> bool:
...
+            families = set(i_families)
+
+        if "custom.frame.range" in families:
+            return True
+
+        if not IS_FARM_HOST:
+            return False
+
+        families.add(i_product_base_type)
+        return bool(families.intersection(FARM_HOST_NAMES))
```
BigRoy, line 178: Let's mark the below backwards compatibility in code cleary.
```suggestion
        # Backwards compatibility for farm host/families logic
        if not IS_FARM_HOST:
            return False

        families.add(i_product_base_type)
        return bool(families.intersection(FARM_HOST_NAMES))
```
I also wonder whether if this returns True we should log it.

```suggestion
        # Backwards compatibility for farm host/families logic
        if not IS_FARM_HOST:
            return False

        families.add(i_product_base_type)
        legacy_farm_families = families.intersection(FARM_HOST_NAMES)
        if legacy_farm_families:
            cls.log.debug(f"Detected legacy farm families: {legacy_farm_families}")
            return True
        return False
```
Just so that in time we can still notice which are using the old logic and whether we can deprecate them down the line.

client/ayon_core/plugins/publish/collect_custom_frame_range.py
```diff
@@ -111,12 +148,31 @@ def _is_custom_frames_used(cls, value) -> bool:
...
+        else:
+            i_families = instance.get("families")
+            i_product_base_type = instance.product_base_type
+
+        # families = [instance.product_base_type]
+        families = set()
+        if i_families is not None:
+            families = set(i_families)
+
+        if "custom.frame.range" in families:
```
BigRoy, line 171: I'd probably reverse them to something like:
```
framerange.custom_frames`
```
(I think if we were to pick that we'd need camelCase? `framerange.customFrames`?)
Or:
```
framerange.custom
```
@iLLiCiTiT 

I think the `.` between frame and range is redundant, and even somewhat wrong.

PR #2043, opened 2026-09-08
Fix styling on the workfiles save as dialog

BigRoy:
## Changelog Description

Fix styling on the workfiles save as dialog

## Additional info

image

## Testing notes:

1. Workfiles tool works.

### DIFF
--- a/client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
+++ b/client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
@@ -1,28 +1,41 @@
 from qtpy import QtWidgets, QtCore
 
-from ayon_core.tools.utils import PlaceholderPlainTextEdit
 from ayon_core.ui.components import (
     AYButton,
     AYCheckBox,
     AYComboBox,
+    AYContainer,
     AYLabel,
-    AYHBoxLayout,
+    AYMenu,
     AYVBoxLayout,
-    AYGridLayout,
     AYLineEdit,
     AYSpinBox,
+    AYTextEdit,
 )
 
+# Surface variant used by the dialog and by every plain wrapper widget
+# inside it, so the wrappers blend into the dialog surface instead of
+# falling back to the host/legacy stylesheet background.
+SURFACE_VARIANT = AYContainer.Variants.Dialog
 
-class SubversionLineEdit(QtWidgets.QWidget):
+
+class SubversionLineEdit(AYContainer):
     """QLineEdit with QPushButton for drop down selection of list of strings"""
 
     text_changed = QtCore.Signal(str)
 
-    def __init__(self, *args, **kwargs):
-        super(SubversionLineEdit, self).__init__(*args, **kwargs)
+    def __init__(self, parent=None):
+        super().__init__(
+            parent,
+            layout=AYContainer.Layout.HBox,
+            variant=SURFACE_VARIANT,
+            layout_margin=0,
+            layout_spacing=3,
+        )
 
-        input_field = AYLineEdit(parent=self)
+        input_field = AYLineEdit(
+            parent=self, variant=AYLineEdit.Variants.Low
+        )
         menu_btn = AYButton(
             variant=AYButton.Variants.Surface,
             icon="arrow_drop_down",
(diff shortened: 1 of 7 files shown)


client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -1,40 +1,52 @@
...
         menu_btn = AYButton(
             variant=AYButton.Variants.Surface,
             icon="arrow_drop_down",
             parent=self,
         )
         menu_btn.setFixedWidth(18)
 
-        menu = QtWidgets.QMenu(self)
+        menu = AYMenu(self)
         menu_btn.setMenu(menu)
```
BigRoy: Because this menu is parented to the widget - it doesn't have hover over effects it seems like? 🤔 Something to double check?

client/ayon_core/ui/components/spinbox.py
```diff
@@ -65,6 +65,24 @@ def __init__(
...
+            "AYSpinBox[ayonPainted=\"true\"] { background: transparent; "
+            "border: none; padding: 0px; "
+            "selection-background-color: none; selection-color: none; }"
+            "AYSpinBox[ayonPainted=\"true\"]::up-button, "
+            "AYSpinBox[ayonPainted=\"true\"]::down-button { "
+            "background: transparent; border: none; }"
+            "AYSpinBox[ayonPainted=\"true\"]::up-arrow, "
+            "AYSpinBox[ayonPainted=\"true\"]::down-arrow { image: none; }"
+        )
+
```
iLLiCiTiT: Why is this needed? We don't trigger default paint of the style.
```suggestion
```

client/ayon_core/ui/ayon_style.json
```diff
@@ -524,6 +538,9 @@
                 "high": {
                     "background-color": "--md-sys-color-surface-container-dark"
                 },
+                "dialog": {
```
iLLiCiTiT: `"dialog"`? Rather use something like `"high-dark"`.

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
```diff
@@ -167,14 +209,26 @@ def __init__(self, controller, parent):
...
 
-        version_layout = AYHBoxLayout(version_widget, margin=0, spacing=4)
+        version_layout = version_widget.layout()
         version_layout.addWidget(version_input)
         version_layout.addWidget(last_version_check)
 
-        # Artist note widget
-        description_input = PlaceholderPlainTextEdit(inputs_widget)
+        # Artist note widget. AYTextEdit does not paint a background of its
+        # own, so it is framed the same way as the side panel's note field.
```
BigRoy, line 86: Should really just get an `AYTextEdit` flag or variant that does render this, right?

client/ayon_core/tools/workfiles/widgets/save_as_dialog.py
iLLiCiTiT: Image

I also have this issue, it is showing weird. There should be one item.

BigRoy:
Note: I think I failed to correctly refactor `Dialog` to `HighDark`. Or at least, I can't see the commit here.
Want me to try and push again @iLLiCiTiT ? Also, I feel the UI is fine without the HighDark perhaps - just use the regular variant? 🤔 

Also, really stupid this needs the `AYContainer` inside the dialog. I'd feel better if instead we just implement `AYDialog` component - can I push that?

client/ayon_core/ui/variants.py
```diff
@@ -111,6 +114,10 @@ class QStyledItemDelegateVariants(Enum):
     Default = "default"
 
 
+class AYHeaderViewVariants(Enum):
+    Default = "default"
```
BigRoy, line 118: FYI, this is in the diff - because it was in the `ayon_style.json` but the `variants.py` was not rebuilt before it seems.

BigRoy:
@iLLiCiTiT looking really nice. Just need some left padding on the spinbox input field?

image

---

Maya 2026
image

Nuke 16
image

Houdini 22 (which uses QML)
image

Houdini 21:
image

iLLiCiTiT:
I agree with the spinbox, but that is more complicated, that goes deep down to style issues.

iLLiCiTiT (approved):

PR #2048, opened 2026-09-10
Publisher: Fix context change on instances

iLLiCiTiT:
## Changelog Description
Use next iter to get selected values instead of index access.

## Additional info
The index access can fail if the widget does return set or iterable.

## Testing notes:
1. Open publisher in a host.
2. Create an instance.
3. Select the instance, change folder/task and confirm it in the UI.
4. The context should be changed.


LiborBatek (approved):
I have tested with multiple publish types and all been working as expected!

LGTM

BigRoy (approved):

Issue #2046, opened 2026-09-09
YN-1074: Silhouette 2026: Qt 6.5.4 fails to draw qtawesome and qtmaterialsymbols icons


ynbot:
### Please describe the feature you have in mind and explain what the current shortcomings are?

Due to the Qt version used in Silhouette 2026 it seems to fail to draw the icons generated by `qtawesome` and `qtmaterialsymbols`. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1544227602464505906/1547184064879132682)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-1074]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=d926e25f13b24d7cbfcec6b226b9ffb6)

BigRoy:
And some more information about the bug with some AI help to get the findings:

**Root cause: [QTBUG-110502](https://bugreports.qt.io/browse/QTBUG-110502) - "Don't do font merging for PUA characters", fixed in Qt 6.6.0.**

Qt has a "font merging" fallback mechanism: when a character seems unsupported by the requested font, Qt automatically substitutes a glyph from another installed font rather than showing `.notdef`/blank. Prior to Qt 6.6.0, this fallback logic was *also* incorrectly triggered for Private Use Area codepoints - even though PUA is explicitly reserved by the Unicode standard for exactly this use case (custom/icon fonts assign their own meaning to these codepoints, and no other font should ever be treated as a valid substitute for them). So Qt would decide our Material Symbols glyph "needed help," pull in a fallback font instead of trusting the font we explicitly asked for, that fallback naturally has no glyph for `U+E163`, and the result renders as nothing. The fix - landing in **Qt 6.6.0**, per the release notes: *"Font merging (automatic assignment of alternative fonts) is no longer applied for characters in the Private Use Areas of Unicode"* - makes Qt trust the requested font for PUA characters again.

Silhouette 2026 bundles **PySide6 6.5.4** - one minor version behind that fix. Other DCCs that render this fine are almost certainly on Qt 6.6+ (or a Qt5 build that never had this particular regression path).

Notes work workaround:
- **`QRawFont.glyphIndexesForString`/`pathForGlyph` works** - those are low-level, single-font glyph lookups that never invoke font-merging at all, which is exactly why a fix like that, rendering glyphs via `QRawFont` outlines instead of `drawText` avoids the bug.

Sources:
[...]

PR #2049, opened 2026-09-10
Material Symbols: Use alternative way how to draw icons

iLLiCiTiT:
## Changelog Description
Use glyph path instead of text rendering for specific range of Qt versions.

## Additional info
There was a bug in Qt related to painting unicode characters which do return 0 width of the character. To fix the issue a new approach was added which is painting the path directly. This does affect Silhouette and OpenRV applications.

This PR fixes ONLY material symbols, and does not fix Awesome icons, so there still might be missing icons here and there. That will NOT be fixed in this PR.

## Testing notes:
1. Launch Silhouette or OpenRV, and also other DCCs to validate we didn't break them.
2. Open AYON tools that do show the icons (e.g. folders or tasks in workfiles tool).
3. The icons should be visible.

Helps with https://github.com/ynput/ayon-core/issues/2046 but does not fully resolve it.

### DIFF
--- a/client/ayon_core/vendor/python/qtmaterialsymbols/iconic_font.py
+++ b/client/ayon_core/vendor/python/qtmaterialsymbols/iconic_font.py
@@ -17,8 +17,7 @@ import warnings
 from packaging.version import parse
 from typing import Dict, Optional, Union
 
-from qtpy import QtCore, QtGui, QtWidgets
-from qtpy import QT_VERSION
+from qtpy import QtCore, QtGui, QtWidgets, QT_VERSION
 
 from .structures import IconOptions, Position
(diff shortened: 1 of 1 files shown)


BigRoy (approved):
Tested only on Windows, but these now show correct on:

Fixed with this:
- Silhouette 2026 (Qt 6.5.4)
- OpenRV (Qt 6.5.3)

Working ok:
- Silhouette 2025.5
- Nuke 16
- Maya 2026
- Houdini 22, 21
- Fusion 20 (Python 3.9 with PySide2)
- Blender 5.2

Note: only tested on Windows - may warrant a test on Mac/Linux with a few apps.

jakubjezek001 (approved):
tested at Nuke and workfiles looks ok to me.. 

Image

PR #2051, opened 2026-09-13
Fix AYMenu crashing + losing hover effect when parented

BigRoy:
## Changelog Description

Fix AYMenu crashing + losing hover effect when parented

## Additional info

Should likely remove the verbose comments - just keeping them for now so the reasoning for the changes can easily be reviewed.

Verbose explanation:

AYONStyle.style_widget() unconditionally called setWindowFlags() on
every QMenu it styled, and it ran twice per menu: once explicitly from
AYMenu.__init__, and again automatically through Qt's own polish() the
first time the menu was shown. setWindowFlags() tears down and
recreates the widget's native window on every call, whether or not the
flags actually changed.

A menu is frequently shown while already nested inside another open
popup - a submenu, or (as here) a context menu opened from a dropdown.
Recreating its native window while an ancestor still holds the
platform's implicit popup grab corrupts that grab: depending on
timing this either crashes the application outright, or leaves the
menu silently unable to receive the native mouse-move events its
hover highlight depends on - no exception, the highlight just never
lights up.

Fix:
- style_widget() now only touches the window flags once; any later
  call (Qt's automatic polish, or a manual re-style) is a no-op if
  they are already set.
- AYMenu applies the AYON style eagerly at construction, before it is
  ever shown or nested inside anything, instead of leaving it to Qt's
  lazy first-show polish - so the one necessary flag change always
  happens while the menu is still inert.
- Restored the AYMenu usage in the workfiles page's "Group by host"
  context menu, reverting the plain-QMenu workaround that was needed
  for exactly this bug.

Verified: setWindowFlags() now fires exactly once per menu instead of
twice; a menu parented inside an already-open AYDropdownPopup renders
its hover correctly for both a forced active-action state and a real
synthesized mouse-move; the restored workfiles-page menu opens via its
real customContextMenuRequested -> exec_() path without raising, with
a working hover. The existing tests/client/ayon_core/ui/test_visual.py
suite shows the identical 11 pre-existing failures before and after
this change (unrelated font/AA rendering baseline mismatches in this
environment, confirmed by diffing against the pre-fix baseline) - no
regressions introduced.

## Testing notes:

1. Launcher > workfiles > right click menu should work fine with hover effect.
2. Also, executing parented `AYMenu`'s with `popup()` command should now not crash.


client/ayon_core/ui/style.py
```diff
@@ -220,9 +220,27 @@ def style_widget(self, widget: QWidget) -> None:
             widget.setAttribute(
                 Qt.WidgetAttribute.WA_TranslucentBackground, True
             )
-            widget.setWindowFlags(
-                widget.windowFlags() | Qt.WindowType.NoDropShadowWindowHint
-            )
+            # 'style_widget' runs more than once per widget - once
```
iLLiCiTiT: TBH I stopped reading after 3rd line. If this would say: "Add `NoDropShadowWindowHint` only once to avoid reinitialization of the window." I'd be happy enough.

client/ayon_core/ui/style.py
```diff
@@ -391,6 +409,19 @@ def styleHint(
...
+            QStyle.StyleHint.SH_MenuBar_MouseTracking,
+        ):
+            # AYONStyle subclasses 'QCommonStyle', a minimal base that
+            # answers this '0' (off) - real platform styles (Fusion,
+            # WindowsVista, ...) answer '1'. With it off, QMenu only
+            # updates its hover highlight while a mouse button is held
+            # down since the menu was opened; plain mouse movement after
+            # an ordinary click or right-click never highlights anything,
+            # even though the menu paints its rows (including the
+            # highlighted state) correctly once told to.
```
iLLiCiTiT:
```suggestion
            # Make sure the mouse movement is tracked.
```

client/ayon_core/ui/style.py
BigRoy: Totally agree - I kept it here just for the review, I think we can discard them almost completely.

iLLiCiTiT (approved):

PR #2054, opened 2026-09-13
Browser: Add support for "No value / Has value" options in filters + "Yes / No" filters exclusive

### LINKED ISSUE #1830: Browser: Refactor the table view filter to align it with the FE.

BigRoy:
## Changelog Description

Add support for "No value" and "Has value" options in filters + make "Yes/No" bool filters exclusive

## Additional info

Boolean value filters are now mutually exclusive:

image

Adds "Has values" and "No values" filtering for multiselects and text fields:

image

image

Fix https://github.com/ynput/ayon-core/issues/1830

## Testing notes:

1. Filter bar should work correctly with these Has values / No values
2. Behavior should feel similar to AYON frontend.


client/ayon_core/tools/browser/ui/_browser_table.py
```diff
@@ -1933,12 +1933,12 @@ def _build_filters(self) -> list[FilterEntry]:
...
                 "hasReviewables", "Has Reviewables",
                 values=["Yes", "No"],
-                icon="play_circle", entity="Version",
+                icon="play_circle", entity="Version", single_select=True,
             ),
             FilterEntry(
                 "tags", "Tags",
                 values=enum_values["tags"],
-                icon="local_offer", entity="Version",
+                icon="local_offer", entity="Version", allow_empty=True,
```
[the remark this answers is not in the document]
BigRoy: I don't see a reason for author to ever be empty admittedly - and I explicitly avoided adding this argument so it's a decision I made intentionally.

client/ayon_core/tools/browser/ui/browser_controller.py
```diff
@@ -616,13 +618,48 @@ def _get_query_filters(self) -> dict[str, Any]:
...
+                    .get("type", "")
+                )
+                attribute_key = f"attrib.{attribute_name}"
+                condition = None
+                if attribute_type == "boolean" and values:
+                    condition = {
+                        "key": attribute_key,
+                        "value": values[0].lower() == "true",
+                        "operator": "eq",
+                    }
```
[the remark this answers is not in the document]
BigRoy: It's not been out there in the wild yet. So backwards incompatibility, especially if not breaking much like this one is fine for now.

moonyuet (approved):
Tested successfully with browser in tag filtering, comment filtering and reviewable filtering.

LiborBatek (approved):
I had followed testing steps, applied different filtering and compared to AYON frontend and all been looking as matching (same filters applied),

LGTM

client/ayon_core/ui/components/table_model.py
```diff
@@ -135,6 +147,8 @@ class FilterEntry:
     value_icons: dict[str, str] = field(default_factory=dict)
     value_colors: dict[str, str] = field(default_factory=dict)
     text_search: bool = False
+    single_select: bool = False
+    allow_empty: bool = False
```
iLLiCiTiT, line 151: Shouldn't this be `True` by default and you have to specify `False` if you need "different" behavior instead? Looks like the defaults were added in favor of the boolean attribute instead the other way.

client/ayon_core/ui/components/table_model.py
```diff
@@ -116,6 +116,13 @@ class BatchFetchRequest:
     parent_id: str | None
 
 
+#: Criterion value matching rows where the filtered field is empty.
+NO_VALUE = "noValue"
```
iLLiCiTiT, line 120: I don't think these constans belong here, they are not used here.

client/ayon_core/ui/components/table_filter.py
```diff
@@ -242,9 +254,15 @@ def _direct_match(
             else:
                 continue
             cell_str = "" if cell_value is None else str(cell_value).lower()
+            is_empty = cell_str in ("", "[]")
```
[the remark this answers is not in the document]
BigRoy: Fixed with https://github.com/ynput/ayon-core/pull/2054/commits/c0c3f6d6beee7dd107dd3927cef87fc848a1b292

client/ayon_core/ui/components/table_filter.py
```diff
@@ -887,13 +942,27 @@ def _populate_value_page(
...
+        empty_labels: dict[str, str] = {}
+        empty_icons: dict[str, str] = {}
+        if entry is not None and entry.allow_empty:
+            empty_labels = {
+                NO_VALUE: f"No {entry.label}",
+                HAS_VALUE: f"Has {entry.label}",
+            }
+            empty_icons = {NO_VALUE: "unpublished", HAS_VALUE: "check"}
+        self._is_free_text = is_text_search or not distinct
+        options = [*empty_labels, *distinct]
```
iLLiCiTiT, line 963:
```suggestion
        empty_items: dict[str, str] = {}
        if entry is not None and entry.allow_empty:
            empty_items = {
                NO_VALUE: (f"No {entry.label}", "unpublished"),
                HAS_VALUE: (f"Has {entry.label}", "check"),
            }

        self._is_free_text = is_text_search or not distinct
        options = [*empty_items, *distinct]
```

client/ayon_core/ui/components/table_filter.py
```diff
@@ -907,11 +976,14 @@ def _populate_value_page(
...
 
-            for val in distinct:
+            for val in options:
                 value_label = val
                 value_icon = None
                 value_color = None
-                if entry is not None:
+                if val in empty_labels:
+                    value_label = empty_labels[val]
+                    value_icon = empty_icons[val]
```
iLLiCiTiT, line 985:
```suggestion
                    value_label, value_icon = empty_items[val]
```

client/ayon_core/ui/components/table_model.py
BigRoy: Hmm - no difference to me admittedly. I feel like there's plenty of usages both ways. Having it _true_ just means it shows more filter entries (the "Has values / No values") but e.g. for integers it's irrelevant, etc. so the split is likely 50/50. I'm fine with keeping it as is.

client/ayon_core/ui/components/table_filter.py
```diff
@@ -907,11 +976,14 @@ def _populate_value_page(
             inner = AYFrame(variant=AYFrame.Variants.Low)
             inner_layout = AYVBoxLayout(inner, margin=0, spacing=0)
 
-            for val in distinct:
+            for val in options:
                 value_label = val
                 value_icon = None
                 value_color = None
-                if entry is not None:
+                if val in empty_labels:
```
iLLiCiTiT, line 983: Maybe we should resolve the options before this look to pre-defined dataclass objects instead of resolving that in the loop. Right now it looks like you can have both empty labels and distinct, which I do believe you can't?

client/ayon_core/ui/components/table_filter.py
BigRoy: This would make it `dict[str, Tuple[str, str]` - I understand the reasoning but not confident it's that much cleaner :D If it was much clearer typed, maybe - but the tuple doesn't really make it much nicer though. Unless we use maybe namedtuple?

client/ayon_core/ui/components/table_model.py
```diff
@@ -125,6 +132,11 @@ class FilterEntry:
         entity: Entity level used to group the filter in the menu.
         text_search: Whether values are entered as fuzzy text instead of
             selected from a distinct-value list.
+        single_select: Whether the values are mutually exclusive, e.g. a
+            Yes/No filter. The first value is preselected when the filter
+            is picked from the menu.
+        allow_empty: Whether to offer "No {label}" and "Has {label}"
```
iLLiCiTiT, line 138: Is this docstring correct?

client/ayon_core/ui/components/table_filter.py
BigRoy: You can, yes - the empty 'labels' are the filter entries to filter to "Has values" or "No values". They are present together with the actual entries.

client/ayon_core/ui/components/table_model.py
BigRoy: Yes. It does explain the result of it, yes. If a value _can be empty_ then we do show those additional filter items, yes.

So that you can filter also to only empty entries, or only filled entries.

iLLiCiTiT: Oh, is that related to the selection? The dosctring describes the use-case, not the usage.

Shouldn't we instead introduce selection mode? To have one variable that defines combination of these two...

Something like `multi`, `single`, `single_or_none`.

BigRoy: Doesn't really clarify it more than this for me honestly 🤔 I prefer it the way it is.

client/ayon_core/ui/components/table_model.py
BigRoy: Will leave as is for now.

client/ayon_core/ui/components/table_filter.py
iLLiCiTiT: My point was to use single dictionary instead of 2 -> 2 `__getkey__` calls to get related information. I would use dedicated object too, but I'd say that should be used for the distint too, and that requires more then just suggestion ([here](https://github.com/ynput/ayon-core/pull/2054#discussion_r4005074056))...

BigRoy: Added https://github.com/ynput/ayon-core/pull/2054/commits/8f0606adac9bd34ce460784fd5cb092956d7bc19

The whole `val` versus `entry` juggling was a bit nasty too so I added a little `from_value` helper. Preferably that could've taken just the `entry` if it wasn't none, but that wouldn't have the fallback to the value as label 🤦‍♂️

client/ayon_core/ui/components/table_model.py
BigRoy: Moved with: https://github.com/ynput/ayon-core/pull/2054/commits/772601317d94ce782e2e27e526a2f307b2e88bd6

client/ayon_core/ui/components/table_model.py
iLLiCiTiT: Please at least change the argument description. The `Whether to offer "No {label}" and "Has {label}"` really does not say what it does.

`allow_empty (bool): Allow empty selection if 'single_select' it True. Result will be 'None'.`

The `single_select` does deserver better description too.

Like I wrote, the descriptions are use-case specific, but do not explain what it does technically. If you don't know (and understand) the use-cases the description will not tell you anything (I did not understand at all that it is related to selection).

BigRoy: 🤔 Well, it does say exactly what it _does_ as a result. When enabled, it gives you to adaditional filter options to filter by whether it's empty or not... haha. Anyway, I'll try to make a combination.

iLLiCiTiT: I probably still don't understand what it does. I don't know what `"No {label}"` and `"Has {label}"` is?

So, it is NOT related to `single_select`?

All it does is that it will add 2 specific action items like `"No task"` and `"Has task"`? If yes, then why it is called `allow_empty`? What empty?

I'm sorry to be stuck on this, but I would really like to understand what it actually does.

BigRoy: E.g. tags, can have values or not. (it can be empty multi-selection)

image

Same for "Description" (it can be empty string)

image

But e.g. Status, MUST be set - it always has a value so we don't offer "Has status" "No status" options.

image

BigRoy: Tweaked docstring with https://github.com/ynput/ayon-core/pull/2054/commits/af16bc45df0219126a52204ed403f88843cb3f39

iLLiCiTiT (approved):

PR #2059, opened 2026-09-15
Scene Inventory: Show lock state on version groups

### LINKED ISSUE #1575: Scene Inventory: Lock icon not visible on group item

Currently if any version been locked, user cant see it by default and need to dive into the version to see the actual product and its `locked` state icon.

Image

It would be great to see the `lock` icon straight away when in the `Scene Inventory`.

So its about incorporating a new item in the list for the lock icon or I can image just changing color of the `version` for locked status...

Simply to have visual cue that particular version been in a `locked` state might be super handy...

_No response_

### Additional context:

_No response_

BigRoy:
## Changelog Description

Add a lock state on the version column for the group as well, including showing partial locks.

This way, you don't need to expand the sections to see this state.

## Additional info

Fix #1575

**None are locked:**

Image

**Some are locked:**

Image

Image

**All are locked:**

Image

Image

## Testing notes:

1. Inventory lock state should also be visualized on the group


client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -129,6 +130,30 @@ def __init__(self, controller, parent=None):
...
+        """
+        locked_states = [
+            container_item.version_locked
+            for container_item in container_items
+        ]
+        if all(locked_states):
+            return True
+        if any(locked_states):
+            return CONTAINER_VERSION_PARTIALLY_LOCKED
+        return False
```
iLLiCiTiT: Another option might be this...
```suggestion
        locked_states = {
            container_item.version_locked
            for container_item in container_items
        }
        if True in locked_states:
            if False in locked_states:
                return CONTAINER_VERSION_PARTIALLY_LOCKED
            return True
        return False
```

iLLiCiTiT (approved):

client/ayon_core/tools/sceneinventory/model.py
```diff
@@ -375,11 +405,25 @@ def data(self, index, role):
...
+            return self._get_version_lock_tooltip(index)
+
         if col != 0:
             index = self.index(index.row(), 0, index.parent())
 
         return super().data(index, role)
 
+    def _get_version_lock_tooltip(self, index):
+        if index.column() != 0:
+            index = self.index(index.row(), 0, index.parent())
```
iLLiCiTiT:
```suggestion
            index = index.sibling(index.row(), 0)
```

BigRoy:
Still works for me with latest suggestions. :)

PR #2066, opened 2026-09-16
Support fuzzy search for folder name in Launcher

### LINKED ISSUE #322: YN-0973: Add fuzzy search on full entity path in the Launcher

The current search in the AYON Launcher only matches against individual entity names, making it difficult to isolate specific shots when sequence information is omitted from shot names. For example, searching for "0010" returns that shot across every sequence, and searching "0180" shows the sequence but hides nested shots like "0010" beneath it. The requested change would extend search to match against the full entity path using fuzzy logic, allowing users to combine terms such as "0180 0010" or descriptive strings like "Jairus Walk Cycle" to quickly narrow down and isolate the exact entity they are looking for. 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1531660747975688243/1531660753608900668)

_This issue was automatically created from AYON ticket [YN-0973]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=55843b506a4546c6a0e6b1451f1c6cb7)

moonyuet:
## Changelog Description
This PR is to support fuzzy search for folder name in Launcher. User can search `0100 0040` to find their destined folder type.

image

## Additional info
Marked it as draft as some discussions are needed for this. I am not sure if it would cause any potential danger as we do the filtering by folder_ids before this PR. 

## Testing notes:
1. AYON -> Launcher
2. Use filter search as the screenshot above

Resolves https://github.com/ynput/ayon-launcher/issues/322


client/ayon_core/tools/utils/folders_widget.py
```diff
@@ -421,6 +421,25 @@ def __init__(self):
...
+    def _match_name_filter(self, source_index) -> bool:
+        if not self._name_filter_terms:
+            return True
+
+        path_labels = []
+        while source_index.isValid():
+            display_role = source_index.data(QtCore.Qt.DisplayRole)
+            if display_role:
+                path_labels.append(str(display_role).casefold())
+            source_index = source_index.parent()
```
BigRoy: I believe if you enable `self.setRecursiveFilteringEnabled(True)` on the sort filter proxy model you wouldn't need to traverse into the parents as well, you just return True/False per index, right? 🤔 

Also, too bad we can't just rely on `FOLDER_PATH_ROLE` (which would be the full path to begin with) because that would match only names, and not labels.

However, we _could_ do this first:
```
if fuzzymatch(source_index.data(FOLDER_PATH_ROLE).casefold()) :
    return True
```
If we would assume that most matches would also match by the path. Plus this would also match then by name alongside just the label.

@iLLiCiTiT thoughts?

client/ayon_core/tools/utils/folders_widget.py
```diff
@@ -421,6 +421,25 @@ def __init__(self):
...
+
+    def set_name_filter(self, name: str) -> None:
+        self._name_filter_terms = name.casefold().split()
+        self.invalidateFilter()
+
+    def _match_name_filter(self, source_index) -> bool:
+        if not self._name_filter_terms:
+            return True
+
+        path_labels = []
```
iLLiCiTiT: Few questions:
1. Shouldn't we prepare the path when the item is created? Don't know the performance of this on bigger projects (@BigRoy could you try?).
2. Should we also look into the folder path (not just labels)?

client/ayon_core/tools/utils/folders_widget.py
```diff
@@ -421,6 +421,25 @@ def __init__(self):
...
+        while source_index.isValid():
+            display_role = source_index.data(QtCore.Qt.DisplayRole)
+            if display_role:
+                path_labels.append(str(display_role).casefold())
+            source_index = source_index.parent()
+
+        path = " ".join(reversed(path_labels))
+        return all(term in path for term in self._name_filter_terms)
```
iLLiCiTiT: Use `/` instead of space to allow path searching too.
```suggestion
        label_path = "/".join(reversed(path_labels))
        return all(term in label_path for term in self._name_filter_terms)
```

client/ayon_core/tools/utils/folders_widget.py
```diff
@@ -429,13 +448,17 @@ def set_folder_ids_filter(self, folder_ids: Optional[list[str]]):
...
         if self._folder_ids_filter is not None:
             if not self._folder_ids_filter:
                 return False
-            source_index = self.sourceModel().index(row, 0, parent_index)
             folder_id = source_index.data(FOLDER_ID_ROLE)
             if folder_id not in self._folder_ids_filter:
                 return False
+
+        if self._name_filter_terms:
+            return self._match_name_filter(source_index)
```
iLLiCiTiT: End filtering only if the name is not matching.
```suggestion
        if not self._match_name_filter(source_index):
            return False
```

client/ayon_core/tools/utils/folders_widget.py
iLLiCiTiT: > I believe if you enable self.setRecursiveFilteringEnabled(True) on the sort filter proxy model you wouldn't need to traverse into the parents as well, you just return True/False per index, right?
> Also, too bad we can't just rely on FOLDER_PATH_ROLE (which would be the full path to begin with) because that would match only names, and not labels.

You have to get the labels path (question [here](https://github.com/ynput/ayon-core/pull/2066#discussion_r4023840156)). Recursive is going down, if children are matching. This is the other way around, this is looking if parent is matching. `FOLDER_PATH_ROLE` contains names not labels (same [question](https://github.com/ynput/ayon-core/pull/2066#discussion_r4023840156)).

We can prepare both (when the items are created), and we can actually use only one dedicated role for it:
`FOLDER_PATH_FILTER` role that will contain `{folder_path} {folder_label_path}`. Something like `"/assets/gummy_bear /Assets/Gummy bear".casefold()`.

client/ayon_core/tools/utils/folders_widget.py
BigRoy: Do note that `path` searching would technically fail because we'd match against the labels. So we should perhaps include a match against `index.data(FOLDER_PATH_ROLE)` too.

client/ayon_core/tools/utils/folders_widget.py
BigRoy: > Should we also look into the folder path (not just labels)?

Yes please - I had a long comment prepared, but I think I lost it 🤔 but gist of it was, let's do it.

> Shouldn't we prepare the path when the item is created? Don't know the performance of this on bigger projects (@BigRoy could you try?).

Let's first see how it performs with it just here. If not acceptable, we'll see what happens if we move it.

client/ayon_core/tools/utils/folders_widget.py
```diff
@@ -421,6 +426,19 @@ def __init__(self):
...
+        self._name_filter_terms = []
+
+    def set_name_filter(self, name: str) -> None:
+        self._name_filter_terms = name.casefold().split()
+        self.invalidateFilter()
+
+    def _match_name_filter(self, source_index) -> bool:
+        if not self._name_filter_terms:
+            return True
+        folder_path_filter = source_index.data(FOLDER_PATH_FILTER_ROLE) or ""
```
iLLiCiTiT:
```suggestion
        folder_path_filter = source_index.data(FOLDER_PATH_FILTER_ROLE)
        if not folder_path_filter:
            return False
```

iLLiCiTiT (approved):

BigRoy (approved):
Works really well, even on a big project.

PR #2067, opened 2026-09-16
Tools: Simplify project sorting

### LINKED ISSUE #1474: Projects are not in alphabetical order

### Current Behavior:

Projects in Launcher are not sorted in alphabetical order.

Image

### Expected Behavior:

_No response_

### Version

1.6.3

Windows

### Steps To Reproduce:

Create a new project and check if projects in Launcher are in the right alphabetical order.

### Relevant log output:

```shell

```

### Additional context:

_No response_

iLLiCiTiT:
## Changelog Description
Simplified project sorting to Pinned, Unpinned and Archived.

## Additional info
Before this PR we did sort projects also by being a library project which also did allow us to add separator. By removing the library sorting the separator is not needed and the overall sorting mechanism is lot easier.

## Testing notes:
1. Projects are sorted by being pinned, unpinned and archived otherwise are sorted alphabetically.

Resolves https://github.com/ynput/ayon-core/issues/1474


BigRoy (approved):
LGTM

With **test pinned**

Image

Image

With **no pins**

Image

Image

Also, note that frontend does list them differently:

Image

It keeps the pinned project at the top, but ALSO in the list below.
And on Home > Projects it doesn't apply the pinning on frontend:

Image

BigRoy:
@LiborBatek can you confirm this is indeed ok for you too?

I wonder if there may actually be a preference for library projects to remain at the bottom. 🤔

iLLiCiTiT:
> Also, note that frontend does list them differently:

I semi-like this. I like the current project is at the top, and also in the projects list. Also like the search bar and the project code.

Something to look at that in future.

LiborBatek (approved):
The projects been sorted alphabetically and by pinning...

LGTM

@BigRoy  if I may say something, I dont mind the order (speaking of Library projects) but it would be nice to have that `pin` icon next to the project name and not on the very right side far away. Also separators would make it more readible...but those got removed already I guess...

iLLiCiTiT:
> it would be nice to have that pin icon next to the project name and not on the very right side far away

That's matching frontend.

> Also separators would make it more readible...

What separators?

BigRoy:
> What separators?

Instead of separating the libraries with separator.
Can we put a separator between the pinned and non-pinned?

iLLiCiTiT:
> Can we put a separator between the pinned and non-pinned?

Yes.

iLLiCiTiT:
Added separator between pinned projects.
image

Should be tested on all platforms.

BigRoy:
This does work.. but I'm not sure it reads very nice:

image

Can we dim the separator more? 

e.g. if we dim it by 50% opacity, it's already better:

image

Or 25%

image

I personally like the subtle one.

PR #2068, opened 2026-09-16
Browser: In Scene column via Column provider

BigRoy:
## Changelog Description

Move the logic for the _only in host integrations_ available **In Scene** column to a column provider.

## Additional info

Also:
- show correct in scene state for group by Product
- show no state on other group bys (instead of just showing "no")
- show in scene state even when browsing another project (e.g. library project)
- do single fast query for version id + product id if it's new server that supports it.
- do not reload full table on version load + do not reload all data on column provider refresh
- filtering by it now works correctly too, and sends no server requests via regular filter query logic.

## Testing notes:

1. In Scene column should still work.
2. In Scene column filters should still work.


iLLiCiTiT (approved):
In case we'd like to (at some point) be able to use browser with IPC we'd have to find a way how to do that without importing and using the host.

PR #2072, opened 2026-09-17
Browser: Folder list perform bulk query + fuzzy folder search

BigRoy:
## Changelog Description

Implement folder hierarchy bulk fetching and enhance folder search with fuzzy search capabilities like https://github.com/ynput/ayon-core/pull/2066

## Additional info

The bulk fetching makes it so that we don't always query the server per folder expansion, but also async do a single request for the full folder hierarchy and allow that to populate the children folders after the root folders are populated with the lighter queries. This allows a very big project to still show root folders quickly without having to wait for the bulk request.

And yes, I hate the complexity - but.... I also, like it. 🗡️ 
Anyway, it may very well be that the lazy tree model just isn't worth it here because getting the full folder list is usually fast enough.

## Testing notes:

1. Projects with large hierarchies should still load
2. After the bulk request no more folder queries should be done


BigRoy:
Better to just do the single bigger call - the complexity of also doing the smaller call hoping it's faster doesn't solve much.

PR #2074, opened 2026-09-17
Browser folders do not query per folder only use a single bulk request

BigRoy:
## Changelog Description

In browser folder widget - do not query per folder only use the bulk request.

## Additional info

Putting the draft up just so we have the intermediate state already, to get the discussion going.

## Testing notes:

1. Browser folder's list should work.


client/ayon_core/ui/components/tree_model.py
```diff
@@ -480,3 +492,257 @@ def reset(self) -> None:
         # connected to loading_changed observes a consistent model state.
         self._update_loading_state()
         self._fetch_children_async(self._root)
+
+
+class BulkTreeModel(QAbstractItemModel):
```
iLLiCiTiT, line 497: TBH I'm not sure about reusability of this model. This will work only if you have 1 column, as soon as you add more columns and different filters based on the column this will become unusable.

Also at the end, the class is so big because it has to implement everything what `QStandardItemModel` does implement for you. But `QStandardItemModel` allows more flexibility. It would also support using the async task.

For case of folders I'd say you can copy and modify what we have in `tools/utils/folders_widget.py` and use the async task to reset the items.

BigRoy:
Closing in favor of work being done on https://github.com/ynput/ayon-core/pull/2106

